Issue #13, opened 2018-03-02
Marko 4 unrecognized tag thrown for nested tag transformers


mlrawlings:
When using a component that has nested tags that are created using the nested tag transformer (like `<ebay-menu-item>` rather than `<ebay-menu:item>`, Marko 4 will throw an error during the parse stage (which occurs before the transform stage) saying it found an unrecognized tag.

**Solution:**
Add an empty tag definition for all valid nested tags:
```json
{
    "<ebay-menu-item>": {}
}
```

seangates:
Would that be alongside the other tag definition? e.g.:

```
{
    "<ebay-menu>": {
        "renderer": "./index.js",
        "transformer": "../../common/transformer/index.js"
    },
    "<ebay-menu-item>": {}
}
```

ianmcburnie:
Closing as fixed in 0.1.0 branch. Will merge to master branch at release time.

Issue #11, opened 2018-03-02
Components with nested transform tags don't work outside demos


mlrawlings:
The [`marko.json`](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu/marko.json) that specifies the nested tag transformer is only scoped to the demo and test components.

Any outside user of the components will not have their nested tags transformed:
```js
ebay_listbox_tag({
        name: "name"
      }, out, 0, function renderBody(out, ebay_listbox0) {
      out.w("<ebay-listbox-option value=\"1\" label=\"Option 1\"></ebay-listbox-option><ebay-listbox-option value=\"2\" label=\"Option 2\" selected></ebay-listbox-option><ebay-listbox-option value=\"3\" label=\"Option 3\"></ebay-listbox-option>");
    });
```

ianmcburnie:
Closing as fixed in 0.1.0 branch. Will merge to master branch at release time.

PR #15, opened 2018-03-03
Fixes unrecognized tag issue in Marko v4 and dash transform issue.

DylanPiercey:
## Description
The `marko.json` inside of each component (which contained the transform information) was only scoped relative to that folder which meant that the examples were loaded correctly but if you tried to actually use the component from outside the examples it was not working. This was fixed by moving all relevant `marko.json` config to the top level.

Also when using with Marko v4 the transform was being applied too late which caused an `unrecognized tag` error to be thrown. The fix for this was to include the child option tag names in the `marko.json` as well (the config for them is empty though) so that when Marko v4 is parsing it knows that they are in fact custom tags.

## Related Issue
issue https://github.com/eBay/ebayui-core/issues/11
issue https://github.com/eBay/ebayui-core/issues/13


seangates:
So ... the components themselves won't carry the tag definition?

DylanPiercey:
@seangates I'm trying out another solution which may allow for the tag definitions to be relative to the component. Will update the PR if this works...

seangates:
@DylanPiercey So, it's still required to have this at the root, all in one large file?

DylanPiercey:
@seangates currently pushing a fix and rebase (this was a couple typo's which @yomed and I caught earlier today). I tried to set this up with relative tag definitions however I was running into some errors that I haven't been able to resolve yet.

I think for now it would be best to just have all of this defined top level.

PR #32, opened 2018-03-08
Adding .npmrc and publishConfig

### LINKED ISSUE #31: Ensure registry is enforced correctly
Because we have an internal corp NPM registry, we need to ensure that we're actually publishing the core components to the public registry, and enforce this setting when publishing.

Resources:
https://docs.npmjs.com/files/npmrc
https://docs.npmjs.com/files/package.json#publishconfig

seangates:
## Description
We need to enforce the publishing of the registry.

## Context
Closes #31 

## References
https://docs.npmjs.com/files/npmrc
https://docs.npmjs.com/files/package.json#publishconfig
Adding it to publishConfig cannot be overridden: https://github.com/npm/npm/issues/10117#issuecomment-221056942

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+registry=https://registry.npmjs.org/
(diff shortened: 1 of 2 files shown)


seangates:
... @ianmcburnie really approves of this. Nice.

PR #36, opened 2018-03-08
README formatting and text fixes

### LINKED ISSUE #34: Docs: listbox/select README consistency.
There are currently some inconsistencies with the `ebay-listbox` components README.md vs the other components:

1. Headings contain code blocks.
2. "Examples" used instead of "Usage".
3. Headings prefixed with component name.

I'm not sure which format works best but I do believe we should make these consistent.

---

*ebay-listbox readme*
screen shot 2018-03-08 at 9 51 28 am

*ebay-button readme*
screen shot 2018-03-08 at 9 52 22 am

seangates:
## Description
Brings all the README documents in sync with one another (re: headings, nomenclature, etc.)

## References
Fixes #34

### DIFF
--- a/src/components/ebay-button/README.md
+++ b/src/components/ebay-button/README.md
@@ -1,14 +1,12 @@
 # ebay-button
 
-## ebay-button Tag
-
-### ebay-button Usage
+## ebay-button Usage
 
 ```marko
 <ebay-button>label</ebay-button>
 ```
 
-### ebay-button Attributes
+## ebay-button Attributes
 
 Name | Type | Stateful | Description
 --- | --- | --- | ---
@@ -20,7 +18,7 @@ Name | Type | Stateful | Description
 `disabled` | Boolean | Yes |
 `partially-disabled` | Boolean | No
 
-### ebay-button Events
+## ebay-button Events
 
 Event | Description
 --- | ---
--- a/src/components/ebay-button/README.md
+++ b/src/components/ebay-button/README.md
@@ -1,12 +1,15 @@
-# ebay-button
+# `ebay-button`
 
-## Usage
+## `<ebay-button>` Tag
+
+### `ebay-button` Usage
 
 ```marko
 <ebay-button>label</ebay-button>
 ```
 
-## Attributes
+### `ebay-button` Attributes
+
 Name | Type | Stateful | Description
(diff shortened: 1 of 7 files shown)


src/components/ebay-button/README.md
```diff
@@ -1,12 +1,15 @@
 # ebay-button
 
-## Usage
+## ebay-button Tag
```
yomed: Is the repetition needed for the nested tag case, and therefore applied to all cases (even components that are singular)?

src/components/ebay-button/README.md
```diff
@@ -1,12 +1,15 @@
 # ebay-button
```
yomed: I'm not a huge of the repetition below, but if we keep it, we could probably remove top level headings like this, and move all the other headings up a notch.

src/components/ebay-button/README.md
senthilp: Yes, I agree. The repetition of `ebay-button` seems redundant

seangates: Because headings get an `id` automatically generated for them (to create and easy table of contents with anchors) each one needs to be unique.

yomed: Is it certain we would actually use a TOC like https://github.com/thlorenz/doctoc? And if we might, shouldn't we defer this until that decision is made? The repetition and verbosity does seem a bit much, especially for cases like the `button`, which don't need it.

seangates: We are planning for the future right now. And, it’s not entirely about a TOC. I use IDs all the time to send people like to specific parts of a document. And, if the component docs were ever in a single page (not advocating for this) they would all be navigable.

Also, the _pattern_ is what we want to establish for everyone as we contribute, and accept outside contribution. Some components have more than one tag, where some don’t. I guess I’m trying to accommodate both scenarios.

Lastly, I have been using a Markdown linter in VS code, which has helped identify possible problem areas for the broad swathe of MD parsers which are out there.

All that said, I will look at how to make it less verbose and repetitive.

seangates:
@yomed Changes per your comments.

/cc @senthilp

seangates:
@yomed For sure. Simply trying to make things consistent and (somewhat) future-proof.

Issue #33, opened 2018-03-08
marko-dynamic-tag should be in devDependencies


yomed:
(It's currently in `dependencies`)

PR #37, opened 2018-03-09
Move marko-dynamic-tag to devDependencies

yomed:
## Description

`marko-dynamic-tag` is only used in the demo, so I'm moving it from `dependencies` to `devDependencies`. There's no `yarn.lock` change.

## References

Closes https://github.com/eBay/ebayui-core/issues/33


seangates:
Ah. I see your comment now.

Issue #30, opened 2018-03-07
Tests: run all server and browser tests against Marko v4


yomed:
All tests should ideally run against both v3 and v4. We may want to limit the v4 testing to CI (like with the integration suite), since it modifies dependencies.

PR #43, opened 2018-03-14
Run all tests with Marko v3 and v4

yomed:
## Description

- set up mechanism to run all server and browsers tests in v3 and v4
- get tests working in v4 by delaying an extra tick in some cases
- create separate task that is meant to only run in CI
- refactor some of the other scripts accordingly

## Context

Now that Marko v4 has improvements to the compat layer, we can run our entire test suite against both versions of Marko. Marko has some assumptions/globals which assumes that it is the only version installed. Therefore, we do some hacking around installing a single version of Marko at a time and running tests against that. Because of these installation steps, this code is really only meant for running in CI (it can still run locally, but most of the time there's no need). This is the case for the existing `integration` tests as well, so I repurposed the `build` task to be meant for CI.

## References

https://github.com/eBay/ebayui-core/issues/30


src/common/transformer/test/test.server.js
```diff
@@ -1,9 +1,20 @@
...
-const markoCompiler = require('marko/compiler');
 const prettyPrint = require('marko-prettyprint').prettyPrintAST;
-const CompileContext = require('marko/compiler/CompileContext');
-const Builder = require('marko/compiler/Builder');
+const markoCompiler = require('marko/compiler');
 const transformer = require('../');
+let CompileContext;
+let Builder;
+
+
```
seangates: My preference for no double lines. 🤷🏻‍♂️

seangates:
LGTM, otherwise!

src/common/transformer/test/test.server.js
yomed: FYI I opened an issue here to see if we want to apply that style rule globally   
https://github.com/eBay/eslint-config-ebay/issues/10

package.json
```diff
@@ -2,21 +2,24 @@
...
     "test:fps": "karma start",
-    "test:integration": "mocha integration/index.js",
-    "build": "yarn clean; yarn lint && yarn test",
+    "test:integration": "yarn installMarkoV3 && mocha integration && yarn installMarkoV4 && mocha integration && yarn installMarkoV3 && yarn clean:integration",
     "clean": "rimraf .coverage .karma .nyc_output demo/static",
+    "clean:integration": "rimraf integration/*.marko.js integration/static",
     "lint": "yarn lint:less && yarn lint:js",
     "lint:less": "stylelint '**/**.less'",
     "lint:js": "eslint .",
+    "build": "yarn lint && yarn test && mocha integration && yarn clean:integration && yarn installMarkoV4 && nyc marko test && mocha integration",
```
ianmcburnie: I'm having a little trouble following this line. I don't see a reference to installMarkoV3, which I was expecting. Is it not possible to replace a bunch of this stuff with just `yarn test:integration`?

yomed: The project uses v3 internally by default, so this is assuming that v3 is installed (which is the case most of the time, but especially so in the CI build). We can run `installMarkoV3` here, but it takes additional time and is basically a no-op. I tried to use `build` as an optimized task that's specifically meant for CI.

Alternatively, I set up `test:integration` as something that can run locally, regardless of current installation. That's why that case follows this:
- install v3
- run integration
- install v4
- run integration
- restore v3

The setup is somewhat convoluted, but it's because the underlying scenarios are complicated.

src/components/ebay-menu/test/test.browser.js
```diff
@@ -269,10 +269,11 @@ describe('given the menu is in the expanded state with radio items', () => {
 
     describe('when an item\'s inner span is clicked', () => {
         let spy;
-        beforeEach(() => {
+        beforeEach((done) => {
```
DylanPiercey: Why did we have to make the init async? What changed?

yomed: Not quite sure. These are all cases that were fine in v3, but failed in v4 until I made them async. I'm guessing that for some reason v4 needs the one extra tick from `setTimeout(done)` to fully set up the component.

package.json
ianmcburnie: Ok, that makes sense. Thanks. I think it would be a good idea to briefly document this on the CONTRIBUTING.md, under the CI section perhaps.

package.json
```diff
@@ -2,21 +2,24 @@
   "name": "@ebay/ebayui-core",
   "version": "0.1.0-1",
   "scripts": {
+    "installMarkoV3": "yarn add marko@^3 marko-widgets@^6 -D",
```
DylanPiercey: @yomed what are your thoughts on just having multiple versions of Marko as devDependencies?

IE:
```
"devDependencies": {
  ...
    "marko3": "npm:marko@3",
    "marko4": "npm:marko@4"
  ...
}
```

Which enables `require('marko3')` and `require('marko4'). We could then setup the tests to create a symlink before running which feels a bit cleaner to me than installing before running the tests, especially outside of the CI.

Thoughts?

yomed: I agree that that would be better -- that was actually my first attempt when I introduced the integration tests. It sort of worked (to begin with), but I ended up running into a lot of strange errors which I believe were related to Marko implicitly depending on there only being a single version installed. I don't remember the exact details, but I think some of them were around the Marko compiler relying on certain defined globals, which would get messed up with multiple versions. Maybe we can look to improve this in Marko in the future?

DylanPiercey: Alright. For now I don't think it's worth holding back this PR but this could be an enhancement we look into in the future.

Otherwise LGTM.

yomed:
I've filed some backlog improvements here https://github.com/eBay/ebayui-core/issues/47

Issue #51, opened 2018-03-16
ebay-notice: add dismiss functionality to notice


sharma46bhawana:

ianmcburnie:
Available in 0.2.0

PR #65, opened 2018-03-21
Notice: add dismiss functionality

sharma46bhawana:
## Description
The changes are only applicable in case of a `page-notice`.
- Adding the property `hidden`, so a parent widget can dismiss the widget by setting it to `true`.

- Adding 2 new inputs `dismissible` and `aria-close` to specify if the dismiss button should show up in the UI and for the label of the dismiss button, respectively.

- The dismissible of the notice emits a `notice-dismissed` event to notify the completion of the action.

## Context
This adds the dismiss functionality to the component

## References
Fixes issue #51 (Needs the button styles for notice from skin v3.5.0)

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -68,7 +68,7 @@
         "@class": "string",
         "@hidden": "boolean",
         "@aria-text": "string",
-        "@aria-close": "string",
+        "@aria-label-close": "string",
         "@dismissible":"boolean",
         "@*": "expression"
     }
--- a/marko.json
+++ b/marko.json
@@ -67,6 +67,7 @@
         "@status": "string",
         "@class": "string",
         "@aria-text": "string",
+        "@hidden":"boolean",
         "@*": "expression"
     }
 }
--- a/marko.json
+++ b/marko.json
@@ -66,8 +66,10 @@
         "@heading-level": "string",
         "@status": "string",
         "@class": "string",
+        "@hidden": "boolean",
         "@aria-text": "string",
-        "@hidden":"boolean",
+        "@aria-close": "string",
+        "@dismissible":"boolean",
         "@*": "expression"
     }
 }
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -25,6 +25,7 @@ Name | Type | Stateful | Description
 `heading-level` | String | No| used in case of "page" level notices to specify the heading tag according to the notice's placement
 `dismissible` | Boolean | No | used to specify the dismiss button in page notice
 `hidden` | Boolean | Yes | whether the widget is hidden or not (Note: not supported as initial attribute)
+`aria-label-close` | String | No | adding description for the button for a11y users
 
 ### ebay-notice Events
 
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -2,9 +2,9 @@
 
 The `<ebay-notice>` is a tag used to create a custom-designed notice element. The notice can be single or multi-line but each line should be wrapped inside a `<p>` tag.
 
-In case of dismissible notice if the parent of `<ebay-notice>` wants to close it, the property `hidden` can be used to change the state of the widget (other than the dismiss button provided).
+For the dismissible use case, if the parent of `<ebay-notice>` wants to close it, the property `hidden` can be used to change the state of the widget (other than the dismiss button provided).
 
-(Note: To use dismiss button in ebay-notice please include `@ebay/skin/icon`)
+(Note:  To use the dismiss button in `<ebay-notice>` please include `@ebay/skin/icon`)
 
 ## ebay-notice Usage
 
@@ -30,4 +30,4 @@ Name | Type | Stateful | Description
 
 Event | Description
 --- | ---
-`notice-dismissed` | the page notice was dismissed/closed
+`notice-change` | the page notice was dismissed/undismissed
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -4,6 +4,8 @@ The `<ebay-notice>` is a tag used to create a custom-designed notice element. Th
 
 In case of dismissible notice if the parent of `<ebay-notice>` wants to close it, the property `hidden` can be used to change the state of the widget (other than the dismiss button provided).
 
+(Note: To use dismiss button in ebay-notice please include `@ebay/skin/icon`)
+
 ## ebay-notice Usage
 
 ```marko
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -21,7 +21,7 @@ Name | Type | Stateful | Description
 `status`  | String | No | "priority" (default), "confirmation" or "information"
 `aria-text` | String | No | adding description for the notice for a11y users
 `heading-level` | String | No| used in case of "page" level notices to specify the heading tag according to the notice's placement
-`dismissible` | Boolean | Yes | used to specify the dismiss button in page notice
+`dismissible` | Boolean | No | used to specify the dismiss button in page notice
 `hidden` | Boolean | Yes | whether the widget is hidden or not (Note: not supported as initial attribute)
 
 ### ebay-notice Events
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -30,4 +30,4 @@ Name | Type | Stateful | Description
 
 Event | Data | Description
 --- | --- | ---
-`notice-change` | `{hidden}` | the page notice was dismissed/undismissed
+`notice-change` | `{ hidden }` | the page notice was dismissed/undismissed
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -28,6 +28,6 @@ Name | Type | Stateful | Description
 
 ### ebay-notice Events
 
-Event | Description
---- | ---
-`notice-change` | the page notice was dismissed/undismissed
+Event | Data | Description
+--- | --- | ---
+`notice-change` | `{hidden}` | the page notice was dismissed/undismissed
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -19,3 +19,10 @@ Name | Type | Stateful | Description
 `status`  | String | No | "priority" (default), "confirmation" or "information"
 `aria-text` | String | No | adding description for the notice for a11y users
 `heading-level` | String | No| used in case of "page" level notices to specify the heading tag according to the notice's placement
+`dismissible` | Boolean | Yes | used to specify the dismiss button in page notice
+
+### ebay-notice Events
+
+Event | Description
+--- | ---
+`notice-dismissed` | the page notice was dismissed/closed
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -31,4 +31,5 @@ Name | Type | Stateful | Description
 
 Event | Data | Description
 --- | --- | ---
-`notice-change` | `{ hidden }` | the page notice was dismissed/undismissed
+`notice-show` | `{ hidden }` | the page notice was shown
+`notice-hide` | `{ hidden }` | the page notice was hidden
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -2,6 +2,8 @@
 
 The `<ebay-notice>` is a tag used to create a custom-designed notice element. The notice can be single or multi-line but each line should be wrapped inside a `<p>` tag.
 
+In case of dismissible notice if the parent of `<ebay-notice>` wants to close it, the property `hidden` can be used to change the state of the widget (other than the dismiss button provided).
+
 ## ebay-notice Usage
 
 ```marko
@@ -20,6 +22,7 @@ Name | Type | Stateful | Description
 `aria-text` | String | No | adding description for the notice for a11y users
 `heading-level` | String | No| used in case of "page" level notices to specify the heading tag according to the notice's placement
 `dismissible` | Boolean | Yes | used to specify the dismiss button in page notice
+`hidden` | Boolean | Yes | whether the widget is hidden or not (Note: not supported as initial attribute)
 
 ### ebay-notice Events
 
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -29,7 +29,7 @@ Name | Type | Stateful | Description
 
 ### ebay-notice Events
 
-Event | Data | Description
---- | --- | ---
-`notice-show` | `{ hidden }` | the page notice was shown
-`notice-hide` | `{ hidden }` | the page notice was hidden
+Event | Description
+--- | ---
+`notice-show` | the page notice was shown
+`notice-hide` | the page notice was hidden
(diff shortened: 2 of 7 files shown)


src/components/ebay-notice/README.md
```diff
@@ -2,6 +2,10 @@
 
 The `<ebay-notice>` is a tag used to create a custom-designed notice element. The notice can be single or multi-line but each line should be wrapped inside a `<p>` tag.
 
+In case of dismissible notice if the parent of `<ebay-notice>` wants to close it, the property `hidden` can be used to change the state of the widget (other than the dismiss button provided).
```
seangates: Improvement:
>For the dismissible use case, if the parent ...

src/components/ebay-notice/README.md
```diff
@@ -19,3 +23,11 @@ Name | Type | Stateful | Description
...
 `aria-text` | String | No | adding description for the notice for a11y users
 `heading-level` | String | No| used in case of "page" level notices to specify the heading tag according to the notice's placement
+`dismissible` | Boolean | No | used to specify the dismiss button in page notice
+`hidden` | Boolean | Yes | whether the widget is hidden or not (Note: not supported as initial attribute)
+
+### ebay-notice Events
+
+Event | Description
+--- | ---
+`notice-dismissed` | the page notice was dismissed/closed
```
yomed: I think this should probably be `notice-dismiss` to be consistent with other command-form verbs in events.

src/components/ebay-notice/README.md
```diff
@@ -2,6 +2,10 @@
 
 The `<ebay-notice>` is a tag used to create a custom-designed notice element. The notice can be single or multi-line but each line should be wrapped inside a `<p>` tag.
 
+In case of dismissible notice if the parent of `<ebay-notice>` wants to close it, the property `hidden` can be used to change the state of the widget (other than the dismiss button provided).
+
+(Note: To use dismiss button in ebay-notice please include `@ebay/skin/icon`)
```
seangates: Improvement:
>... To use the dismiss button in `<ebay-notice>` please ...

src/components/ebay-notice/index.js
```diff
@@ -16,30 +18,53 @@ const constants = {
...
         headingTag: headingTag,
         contentTag: constants[type].contentTag,
+        dismissible,
         type,
         status,
+        hidden,
         ariaText: input.ariaText || '',
+        ariaClose: input.ariaClose || '',
         htmlAttributes: processHtmlAttributes(input),
         renderBody: input.renderBody,
```
yomed: I accidentally left this line in during https://github.com/eBay/ebayui-core/pull/58 -- maybe we can remove it as part of this PR.

src/components/ebay-notice/test/test.browser.js
```diff
@@ -0,0 +1,28 @@
...
+    let widget;
+    let button;
+
+    beforeEach(() => {
+        widget = renderer.renderSync({ type: 'page', dismissible: true }).appendTo(document.body).getWidget();
+        button = document.querySelector('.page-notice__close');
+    });
+    afterEach(() => widget.destroy());
+
+    describe('when the dismiss button is clicked', () => {
```
yomed: We should also test the programmatic access. You can see existing examples by searching for `programmatically`.

src/components/ebay-notice/test/test.browser.js
```diff
@@ -0,0 +1,28 @@
...
+        let spy;
+        beforeEach(() => {
+            spy = sinon.spy();
+            widget.on('notice-dismissed', spy);
+            testUtils.triggerEvent(button, 'click');
+        });
+
+        test('then it emits the marko event from notice-dismissed', () => {
```
yomed: We should test that the notice was actually dismissed too.

src/components/ebay-notice/index.js
```diff
@@ -16,30 +18,53 @@ const constants = {
...
     type: 'page',
-    status: 'confirmation'
+    status: 'priority',
+    hidden: false,
+    dismissible: false
 };
 
-function getTemplateData(state, input) {
+function getInitialState(input) {
+    const hidden = input.hidden || defaults.hidden;
```
yomed: In the docs, it says this is not supported as an initial attribute, in which case it should not use `input.hidden`.

src/components/ebay-notice/index.js
```diff
@@ -16,30 +18,53 @@ const constants = {
 
 const defaults = {
     type: 'page',
-    status: 'confirmation'
+    status: 'priority',
```
seangates: Why is the default status changing to `priority`? Is it because we're defaulting `dismissible` to false? Was this a conscious decision to have an "error" be the default notice?

In my view, the default notice should always be a `confirmation`, but I could be misunderstanding the component's requirements.

src/components/ebay-notice/index.js
```diff
@@ -16,30 +18,53 @@ const constants = {
...
+        hidden,
         ariaText: input.ariaText || '',
+        ariaClose: input.ariaClose || '',
         htmlAttributes: processHtmlAttributes(input),
         renderBody: input.renderBody,
         mainClass: [`${type}-notice`, `${type}-notice--${status}`, input.class],
         headingClass: `${type}-notice__status`,
-        contentClass: `${type}-notice__content`
+        contentClass: `${type}-notice__content`,
+        dismissedClass: dismissible ? `page-notice__close` : ``
```
yomed: You can use regular `'` here instead of template literals.

src/components/ebay-notice/template.marko
```diff
@@ -2,11 +2,15 @@
...
-    ${data.htmlAttributes}>
+    ${data.htmlAttributes}
+    if(!data.hidden)>
         <${data.headingTag}
             class=data.headingClass
             id="${data.status}-status-${widget.elId()}">
                 <span aria-label="${data.ariaText}" role="img"></span>
         </${data.headingTag}>
         <${data.contentTag} class=data.contentClass w-body/>
+         <button aria-label="${data.ariaClose}" class=data.dismissedClass w-onclick="onDismiss" if(data.dismissible)>
```
yomed: You should test the presence of this button for both dismissible and non-dismissible in the server side tests.

src/components/ebay-notice/index.js
CestDiego: How is it that one is supposed to change the hidden state from outside this component? should we just get the component manually in code and change it?

If this is the case. We should consider that a more declarative way would be to hook to `onInput` so that whenever the parent re-renders this component with different attributes, it changes.

yomed: That's a good point, that `input` can be for both initial attributes and stateful properties. We'll need to figure out a general solution for cases like this.

src/components/ebay-notice/README.md
sharma46bhawana: okay will change that, I was just thinking that as the event fired describes what already occurred.

src/components/ebay-notice/index.js
sharma46bhawana: okay I will remove it too.

src/components/ebay-notice/template.marko
sharma46bhawana: okay makes sense I will add it

sharma46bhawana:
@yomed We need to update our skin version but as skin 3.5.0 hasn't been released yet so I didn't upgrade it

src/components/ebay-notice/index.js
sharma46bhawana: done

src/components/ebay-notice/index.js
sharma46bhawana: I changed it to match the docs. I think that by default "error" would be most widely used hence I made `priority` as default

src/components/ebay-notice/README.md
```diff
@@ -19,3 +23,11 @@ Name | Type | Stateful | Description
...
 `aria-text` | String | No | adding description for the notice for a11y users
 `heading-level` | String | No| used in case of "page" level notices to specify the heading tag according to the notice's placement
+`dismissible` | Boolean | No | used to specify the dismiss button in page notice
+`hidden` | Boolean | Yes | whether the widget is hidden or not (Note: not supported as initial attribute)
+
+### ebay-notice Events
+
+Event | Description
+--- | ---
+`notice-change` | the page notice was dismissed/undismissed
```
yomed: A `change` event would need to pass data along too, or alternatively maybe we should have 2 events `notice-hide` and `notice-show` similar to the Menu.

sharma46bhawana: done

src/components/ebay-notice/index.js
```diff
@@ -16,30 +18,54 @@ const constants = {
...
+        hidden,
         ariaText: input.ariaText || '',
+        ariaClose: input.ariaClose || '',
         htmlAttributes: processHtmlAttributes(input),
-        renderBody: input.renderBody,
         mainClass: [`${type}-notice`, `${type}-notice--${status}`, input.class],
         headingClass: `${type}-notice__status`,
-        contentClass: `${type}-notice__content`
+        contentClass: `${type}-notice__content`,
+        dismissedClass: dismissible ? 'page-notice__close' : ''
```
yomed: (minor) `dismissibleClass`?

yomed: Actually, it looks like the button itself only shows up under `dismissible: true`, so maybe just put this class straight in the HTML.

sharma46bhawana: Okay I just that added it in `index` because we wanted `.marko` clear of any classes.

yomed: We're trying to keep the `.marko` files  light on logic, so that would normally apply to dynamic classes.

marko.json
```diff
@@ -66,7 +66,10 @@
         "@heading-level": "string",
         "@status": "string",
         "@class": "string",
+        "@hidden": "boolean",
         "@aria-text": "string",
+        "@aria-close": "string",
```
ianmcburnie: I'm not a fan of `aria-close`, as `aria` is a bit ambiguous (i.e. aria is not just used for labels). Maybe `close-icon-label` or `aria-label-close` instead.

Let's try and be consistent across the components and all decide on a convention. We can use slack if necessary.

In hindsight `aria-text` is a bit ambiguous too (but we all reviewed the PR, so not your fault!). I might rename that at some point.

src/components/ebay-notice/index.js
sharma46bhawana: okay, I will add it to the button then

marko.json
sharma46bhawana: I thought `aria-label-close` would become a too long name so I adopted for aria-close but I can change it as its more descriptive

src/components/ebay-notice/examples/4-dismissible-notice/template.marko
```diff
@@ -0,0 +1,7 @@
+<ebay-notice
+    status="information"
+    type="page"
+    aria-text="information" dismissible=true aria-close="Dismiss notice">
```
ianmcburnie: Minor, but this can just be `dismissable` alone, instead of `dismissable=true`, right?

src/components/ebay-notice/index.js
```diff
@@ -16,30 +18,53 @@ const constants = {
...
+
+function init() {
+    observer.observeRoot(this, ['hidden'], (hidden) => {
+        emitAndFire(this, 'notice-change', { hidden });
+    });
+}
+
+function onDismiss() {
+    this.setState('hidden', true);
+    emitAndFire(this, 'notice-change', { hidden: true });
```
ianmcburnie: If there is already an observer on this property (line 54 above), then why do we need to run emitAndFire again?

sharma46bhawana: This in case the dismiss button is explicitly clicked, and not changed through the hidden property

src/components/ebay-notice/examples/4-dismissible-notice/template.marko
sharma46bhawana: Yes I missed this, will change it

src/components/ebay-notice/index.js
ianmcburnie: But the click handler does change the hidden property, doesn't it?

sharma46bhawana: Yes it does, but the native setstate doesn't a callback so I am setting it explicitly.

ianmcburnie: @yomed Is this correct? I thought the propertyObserver created in the init() method would catch this.

yomed: Yes, this is right. The propertyObserver just observes the property on the DOM element, not the state value.  Using UI and API take different code paths initially, though they can share common processing afterwards. This particular example is super simple, so I'm not too concerned with the current setup.

Some examples of other patterns:
1) Marko has hooks to call a function whenever a certain value is updated. So we could create an `update_hidden()` function (the Marko hook) and then use that to call `emitAndFire()`.
2) We can create our own "common processing" function, which is roughly what we're doing now in Menu.

src/components/ebay-notice/README.md
```diff
@@ -19,3 +23,13 @@ Name | Type | Stateful | Description
...
+`dismissible` | Boolean | No | used to specify the dismiss button in page notice
+`hidden` | Boolean | Yes | whether the widget is hidden or not (Note: not supported as initial attribute)
+`aria-label-close` | String | No | adding description for the button for a11y users
+
+### ebay-notice Events
+
+Event | Data | Description
+--- | --- | ---
+`notice-show` | `{ hidden }` | the page notice was shown
+`notice-hide` | `{ hidden }` | the page notice was hidden
```
yomed: I don't think the events need data anymore.

src/components/ebay-notice/index.js
```diff
@@ -16,30 +18,65 @@ const constants = {
...
+function init() {
+    observer.observeRoot(this, ['hidden'], (hidden) => {
+        this.processAfterStateChange(hidden);
+    });
+}
+/**
+ * Common processing after data change via both UI and API
+ * @param {Boolean} hidden
+ */
+function processAfterStateChange(hidden) {
```
yomed: I think you can leave out the param and use `this.state.hidden` instead. Make sure to update the jsdoc too.

src/components/ebay-notice/test/test.browser.js
```diff
@@ -0,0 +1,87 @@
...
+            spy = sinon.spy();
+            widget.on('notice-hide', spy);
+            testUtils.triggerEvent(button, 'click');
+            setTimeout(done);
+        });
+        test('then root not present in the DOM', () => {
+            expect(document.querySelector('section.page-notice')).to.equal(null);
+        });
+
+        test('then it emits the marko event from notice-change', () => {
```
yomed: Search for `notice-change` to find the old leftover references.

src/components/ebay-notice/index.js
ianmcburnie: Thanks for clarifying. Pattern 1 sounds good to me, but I guess we need to be consistent with menu.

ianmcburnie: Also, this isn't a PR blocker, as Yoni says this case is super simple right now, but we should be very wary if we start doing this in multiple places or in more complex components than this.

src/components/ebay-notice/README.md
sharma46bhawana: done

src/components/ebay-notice/index.js
sharma46bhawana: done, thanks for the catch

src/components/ebay-notice/test/test.browser.js
sharma46bhawana: done, sorry about this.

Issue #21, opened 2018-03-05
New Component: Breadcrumbs


ianmcburnie:
Details to follow...

Related issue: https://github.com/eBay/skin/issues/82

ianmcburnie:
Available in 0.2.0.

PR #66, opened 2018-03-23
Breadcrumb

RajaRamu:
## Description
ebay-breadcrumb component

## References
Issue - https://github.com/eBay/ebayui-core/issues/21
https://dsl.ebay.com/ds6/breadcrumb
https://app.zeplin.io/project/5aaab921ced0cc2619637dae/dashboard
## Dependencies
Release of `@ebay/skin/breadcrumb` through https://github.com/eBay/skin/pull/102

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ For more information, please read [Building a UI Component in 2017 and Beyond](h
 
 ## Getting Started
 
-The eBayUI core components will shortly be available as the `@ebay/ebayui-core` package on [NPM](https://www.npmjs.com/).
+The eBayUI core components are available as the `@ebay/ebayui-core` package on [NPM](https://www.npmjs.com/package/@ebay/ebayui-core).
 
 Use npm or yarn to add the package dependency to your project:
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ For more information, please read [Building a UI Component in 2017 and Beyond](h
 *NOTE: Marko v3 also requires [Marko Widgets](https://github.com/marko-js/marko-widgets).*
 
 ### Components
-
+* [`ebay-breadcrumb`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-breadcrumb)
 * [`ebay-button`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-button)
 * [`ebay-menu`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-menu)
 * [`ebay-select`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-select)
--- a/src/components/ebay-breadcrumb/README.md
+++ b/src/components/ebay-breadcrumb/README.md
@@ -4,17 +4,25 @@
 ### Last item as the parent
 ```marko
 <ebay-breadcrumb aria-label='You are here'>
-    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>eBay</ebay-breadcrumb-item>
-    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>Auto Parts and Vehicles</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>eBay</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>Auto Parts and Vehicles</ebay-breadcrumb-item>
     <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>Motors Parts and Accessories</ebay-breadcrumb-item>
 </ebay-breadcrumb>
 ```
 ### Last item as the current
 ```marko
 <ebay-breadcrumb aria-label='You are here'>
-    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>eBay</ebay-breadcrumb-item>
-    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>Auto Parts and Vehicles</ebay-breadcrumb-item>
-    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}' current='true'>Motors Parts and Accessories</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>eBay</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>Auto Parts and Vehicles</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item>Motors Parts and Accessories</ebay-breadcrumb-item>
+</ebay-breadcrumb>
+```
+### Prevent-default behaviour and emit custom event (`breadcrumb-click`) onclick of items
+```marko
+<ebay-breadcrumb aria-label='You are here' prevent-default=true>
+    <ebay-breadcrumb-item href='https://...'>eBay</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>Auto Parts and Vehicles</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>Motors Parts and Accessories</ebay-breadcrumb-item>
 </ebay-breadcrumb>
 ```
 ## ebay-breadcrumb Attributes
@@ -22,12 +30,11 @@
 Name | Type | Stateful | Description
 --- | --- | --- | ---
 `aria-label` | String | No | aria-label for nav tag
+`prevent-default` | Boolean | No | this will be used for single page app. Custom event `breadcrumb-click` will be emitted on click of breadcrumb items
+
 
 ## ebay-breadcrumb-item Attributes
 
 Name | Type | Stateful | Description
 --- | --- | --- | ---
 `href` | String | No | anchor href
-`_sp` | String | No | sid tracking
-`navsrc` | String | No | this is for all the click tracking with SEO friendly approach. this is mainly used for answers platform's powered pages
-`current` | Boolean | No | this mark the item as current or parent
--- a/src/components/ebay-breadcrumb/README.md
+++ b/src/components/ebay-breadcrumb/README.md
@@ -20,7 +20,7 @@ Name | Type | Stateful | Description
 
 Event | Description | Data
 --- | ---
-`breadcrumb-click` | click breadcrumb items | `{ event, currentTarget: el }`
+`breadcrumb-click` | click breadcrumb items | `{ event, target:event.target, currentTarget: el }`
 
 ## ebay-breadcrumb-item Attributes
 
--- a/src/components/ebay-breadcrumb/README.md
+++ b/src/components/ebay-breadcrumb/README.md
@@ -1,0 +1,33 @@
+# ebay-breadcrumb
+
+## ebay-breadcrumb Usage
+### Last item as the parent
+```marko
+<ebay-breadcrumb aria-label='You are here'>
+    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>eBay</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>Auto Parts and Vehicles</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>Motors Parts and Accessories</ebay-breadcrumb-item>
+</ebay-breadcrumb>
+```
+### Last item as the current
+```marko
+<ebay-breadcrumb aria-label='You are here'>
+    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>eBay</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>Auto Parts and Vehicles</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}' current='true'>Motors Parts and Accessories</ebay-breadcrumb-item>
+</ebay-breadcrumb>
+```
+## ebay-breadcrumb Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`aria-label` | String | No | aria-label for nav tag
+
+## ebay-breadcrumb-item Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`href` | String | No | anchor href
+`_sp` | String | No | sid tracking
+`navsrc` | String | No | this is for all the click tracking with SEO friendly approach. this is mainly used for answers platform's powered pages
+`current` | Boolean | No | this mark the item as current or parent
--- a/src/components/ebay-breadcrumb/README.md
+++ b/src/components/ebay-breadcrumb/README.md
@@ -20,7 +20,7 @@ Name | Type | Stateful | Description
 
 Event | Description | Data
 --- | ---
-`breadcrumb-click` | click breadcrumb items | `{ event, target:event.target, currentTarget: el }`
+`breadcrumb-select` | click breadcrumb items | `{ event, target:event.target, currentTarget: el }`
 
 ## ebay-breadcrumb-item Attributes
 
--- a/src/components/ebay-breadcrumb/README.md
+++ b/src/components/ebay-breadcrumb/README.md
@@ -16,17 +16,17 @@ Name | Type | Stateful | Description
 `heading` | String | No | heading for breadcrumb which will be clipped
 `prevent-default` | Boolean | No | this will be used for single page app. when item is clicked, `breadcrumb-click` will be emitted.
 
-## ebay-breadcrumb-item Attributes
-
-Name | Type | Stateful | Description
---- | --- | --- | ---
-`href` | String | No | anchor href
-
 ## ebay-breadcrumb Events
 
 Event | Description
 --- | ---
 `breadcrumb-click` | click breadcrumb items
+
+## ebay-breadcrumb-item Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`href` | String | No | anchor href
 
 ### Examples
 More examples can be found `ebay-breadcrumb/examples`
--- a/src/components/ebay-breadcrumb/README.md
+++ b/src/components/ebay-breadcrumb/README.md
@@ -3,23 +3,7 @@
 ## ebay-breadcrumb Usage
 ### Last item as the parent
 ```marko
-<ebay-breadcrumb aria-label='You are here'>
-    <ebay-breadcrumb-item href='https://...'>eBay</ebay-breadcrumb-item>
-    <ebay-breadcrumb-item href='https://...'>Auto Parts and Vehicles</ebay-breadcrumb-item>
-    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>Motors Parts and Accessories</ebay-breadcrumb-item>
-</ebay-breadcrumb>
-```
-### Last item as the current
-```marko
-<ebay-breadcrumb aria-label='You are here'>
-    <ebay-breadcrumb-item href='https://...'>eBay</ebay-breadcrumb-item>
-    <ebay-breadcrumb-item href='https://...'>Auto Parts and Vehicles</ebay-breadcrumb-item>
-    <ebay-breadcrumb-item>Motors Parts and Accessories</ebay-breadcrumb-item>
-</ebay-breadcrumb>
-```
-### Prevent-default behaviour and emit custom event (`breadcrumb-click`) onclick of items
-```marko
-<ebay-breadcrumb aria-label='You are here' prevent-default=true>
+<ebay-breadcrumb heading='Page navigation'>
     <ebay-breadcrumb-item href='https://...'>eBay</ebay-breadcrumb-item>
     <ebay-breadcrumb-item href='https://...'>Auto Parts and Vehicles</ebay-breadcrumb-item>
     <ebay-breadcrumb-item href='https://...'>Motors Parts and Accessories</ebay-breadcrumb-item>
@@ -29,12 +13,20 @@
 
 Name | Type | Stateful | Description
 --- | --- | --- | ---
-`aria-label` | String | No | aria-label for nav tag
-`prevent-default` | Boolean | No | this will be used for single page app. Custom event `breadcrumb-click` will be emitted on click of breadcrumb items
-
+`heading` | String | No | heading for breadcrumb which will be clipped
+`prevent-default` | Boolean | No | this will be used for single page app. when item is clicked, `breadcrumb-click` will be emitted.
 
 ## ebay-breadcrumb-item Attributes
 
 Name | Type | Stateful | Description
 --- | --- | --- | ---
 `href` | String | No | anchor href
+
+## ebay-breadcrumb Events
+
+Event | Description
+--- | ---
+`breadcrumb-click` | click breadcrumb items
+
+### Examples
+More examples can be found `ebay-breadcrumb/examples`
--- a/src/components/ebay-breadcrumb/README.md
+++ b/src/components/ebay-breadcrumb/README.md
@@ -18,12 +18,12 @@ Name | Type | Stateful | Description
 
 ## ebay-breadcrumb Events
 
-Event | Description
+Event | Description | Data
 --- | ---
-`breadcrumb-click` | click breadcrumb items
+`breadcrumb-click` | click breadcrumb items | `{ event, currentTarget: el }`
 
 ## ebay-breadcrumb-item Attributes
 
-Name | Type | Stateful | Data | Description
+Name | Type | Stateful | Description
 --- | --- | --- | ---
-`href` | String | No | `{ event, currentTarget: el }` | anchor href
+`href` | String | No | anchor href
--- a/src/components/ebay-breadcrumb/README.md
+++ b/src/components/ebay-breadcrumb/README.md
@@ -13,8 +13,8 @@
 
 Name | Type | Stateful | Description
 --- | --- | --- | ---
-`heading` | String | No | heading for breadcrumb which will be clipped
-`prevent-default` | Boolean | No | this will be used for single page app. when item is clicked, `breadcrumb-click` will be emitted.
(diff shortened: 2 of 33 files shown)


seangates:
Broken build?

README.md
```diff
@@ -27,7 +27,7 @@ For more information, please read [Building a UI Component in 2017 and Beyond](h
 
 ## Getting Started
 
-The eBayUI core components will shortly be available as the `@ebay/ebayui-core` package on [NPM](https://www.npmjs.com/).
+The eBayUI core components are available as the `@ebay/ebayui-core` package on [NPM](https://www.npmjs.com/package/@ebay/ebayui-core).
```
seangates: You need to rebase.

RajaRamu:
retest

RajaRamu:
test please

src/components/ebay-breadcrumb/browser.json
```diff
@@ -0,0 +1,5 @@
+{
+    "dependencies": [
+        "require: ./index.js"
```
RajaRamu: I will add `@ebay/skin/breadcrumb` once its available else build is failing

src/components/ebay-breadcrumb/test/test.server.js
```diff
@@ -0,0 +1,65 @@
...
+    const $ = testUtils.getCheerio(context.render(mock.items));
+    const li = $('nav li');
+    expect(li.length).to.equals(4);
+    const anchor = $('nav li a');
+    expect(anchor.length).to.equals(4);
+});
+test('should return items html attributes', context => {
+    const $ = testUtils.getCheerio(context.render(mock.items));
+    const anchor = $('nav li a');
+    const _sp = [];
```
seangates: As with my comments before: this test should be removed as it's specific to eBay internal applications and processes, and does not apply in the public sphere.

src/components/ebay-breadcrumb/test/test.server.js
```diff
@@ -0,0 +1,65 @@
...
+});
+test('should return h2 tag', context => {
+    const $ = testUtils.getCheerio(context.render(mock.items));
+    const h2Tag = $('h2');
+    expect(h2Tag.length).to.equals(1);
+    expect(h2Tag.attr('id')).to.equals('breadcrumbs_heading');
+    expect(h2Tag.attr('class')).to.equals('clipped');
+    expect(h2Tag.html()).to.equals('You are here');
+});
+test('should return 4 li with anchor', context => {
```
seangates: These first four tests should be combined, or refactored to be more DRY. There's a ton of duplication of the rendering of `mock.items`.

src/components/ebay-breadcrumb/test/test.server.js
```diff
@@ -0,0 +1,65 @@
...
+});
+test('should create span tag if href is null ', context => {
+    const $ = testUtils.getCheerio(context.render(mock.itemsWithNoHref));
+    const li = $('nav li');
+    expect(li.length).to.equals(4);
+    const currentElement = $('span', li[li.length - 1]);
+    expect(currentElement.attr('aria-current')).to.equals('page');
+    expect(currentElement.attr('class')).to.equals('current');
+});
+test('should create <a> tag if href is NOT null for last item', context => {
```
seangates: This test can be combined with the first four.

src/components/ebay-breadcrumb/README.md
```diff
@@ -0,0 +1,40 @@
...
+## ebay-breadcrumb Usage
+### Last item as the parent
+```marko
+<ebay-breadcrumb aria-label='You are here'>
+    <ebay-breadcrumb-item href='https://...'>eBay</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>Auto Parts and Vehicles</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>Motors Parts and Accessories</ebay-breadcrumb-item>
+</ebay-breadcrumb>
+```
+### Last item as the current
```
yomed: I think the `README.md` should just have the simplest usage example (maybe this one), and leave example code in general for the `examples/` section.

src/components/ebay-breadcrumb/README.md
```diff
@@ -0,0 +1,40 @@
...
+    <ebay-breadcrumb-item href='https://...'>eBay</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>Auto Parts and Vehicles</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>Motors Parts and Accessories</ebay-breadcrumb-item>
+</ebay-breadcrumb>
+```
+## ebay-breadcrumb Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`aria-label` | String | No | aria-label for nav tag
```
yomed: We may want to rename this since it isn't actually an aria-label, but `clipped` text.  cc @ianmcburnie

src/components/ebay-breadcrumb/README.md
```diff
@@ -0,0 +1,40 @@
...
+`prevent-default` | Boolean | No | this will be used for single page app. Custom event `breadcrumb-click` will be emitted on click of breadcrumb items
```
yomed: I think the description here should be clarified to something like
> When an item is clicked, a custom event will be emitted instead of following the link

src/components/ebay-breadcrumb/examples/1-breadcrumb-prevent-default/template.marko
```diff
@@ -0,0 +1,7 @@
+
+<ebay-breadcrumb aria-label="You are here" prevent-default=true>
```
yomed: For boolean attributes, you can emit the `=true` part, since we're just looking for presence.

src/components/ebay-breadcrumb/examples/2-lastpage-as-current/template.marko
```diff
@@ -0,0 +1,6 @@
+<ebay-breadcrumb aria-label="You are here">
```
yomed: (minor) For `lastpage` examples, please separate it into `last-page` in the directory name so it reads correctly in the marketplace

src/components/ebay-breadcrumb/index.js
```diff
@@ -0,0 +1,60 @@
+const markoWidgets = require('marko-widgets');
+const emitAndFire = require('../../common/emit-and-fire');
+const processHtmlAttributes = require('../../common/html-attributes');
+const template = require('./template.marko');
+
+const constants = {
+    'currentTag': 'span',
```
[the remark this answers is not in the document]
yomed: @seangates I think this should either be enforced through our eslint rules, or not at all.

src/components/ebay-breadcrumb/index.js
```diff
@@ -0,0 +1,60 @@
+const markoWidgets = require('marko-widgets');
+const emitAndFire = require('../../common/emit-and-fire');
+const processHtmlAttributes = require('../../common/html-attributes');
+const template = require('./template.marko');
+
+const constants = {
```
yomed: Each of these constants is only used a single time, and doesn't really improve clarity, so I would skip this `constants` object altogether.

src/components/ebay-breadcrumb/index.js
```diff
@@ -0,0 +1,60 @@
...
+        ariaLabel: input.ariaLabel || '',
+        htmlAttributes: htmlAttributes,
+        preventDefault: input.preventDefault
+    };
+}
+function handleClick(event, el) {
+    const arg = {};
+    arg.event = event;
+    arg.currentTarget = el;
+    emitAndFire(this, constants.clickEvent, arg);
```
yomed:
```js
emitAndFire(this, 'breadcrumb-click', { event, currentTarget: el });
```

src/components/ebay-breadcrumb/README.md
```diff
@@ -0,0 +1,40 @@
...
+```marko
+<ebay-breadcrumb aria-label='You are here'>
+    <ebay-breadcrumb-item href='https://...'>eBay</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...'>Auto Parts and Vehicles</ebay-breadcrumb-item>
+    <ebay-breadcrumb-item href='https://...' _sp='' navsrc='{}' navsrc='{}'>Motors Parts and Accessories</ebay-breadcrumb-item>
+</ebay-breadcrumb>
+```
+### Last item as the current
+```marko
+<ebay-breadcrumb aria-label='You are here'>
```
yomed: Is `You are here` a typical example? Let's pick text that's realistic for the use case (across all examples).

src/components/ebay-breadcrumb/mock/index.js
```diff
@@ -0,0 +1,42 @@
...
+    getItem('Motors Parts & Accessories', 'https://www.ebay.com/rpp/motors-parts-accessories'),
+    getItem('Motorcycle Parts', 'https://www.ebay.com/b/Motorcycle-Parts/10063/bn_557636')];
+
+const itemsWithNoHref = [getItem('eBay', 'https://www.ebay.com/'),
+    getItem('Auto Parts and Vehicles', 'https://www.ebay.com/b/Auto-Parts-and-Vehicles/6000/bn_1865334'),
+    getItem('Motors Parts & Accessories', 'https://www.ebay.com/rpp/motors-parts-accessories'),
+    getItem('Motorcycle Parts', null)];
+
+module.exports = {
+    items: { ariaLabel: 'You are here',
```
yomed: (minor) For these objects, maybe put `ariaLabel` on the next line for better legibility.
```js
items: {
    ariaLabel: '',
    items,
    '*' : {}
}
```

src/components/ebay-breadcrumb/mock/index.js
```diff
@@ -0,0 +1,42 @@
+const item = { href: 'https://www.ebay.com/', navSrc: '{"actionKind":"NAVSRC","operationId":"2489527"}', _sp: 'p2489527.m4340.l9751.c1' };
+
+const getItem = (text, href) => {
+    const clonedItem = JSON.parse(JSON.stringify(item));
+    clonedItem.renderBody = (stream) => stream.write(text);
+    clonedItem.href = href;
+    return clonedItem;
+};
+const items = [getItem('eBay', 'https://www.ebay.com'),
```
yomed: You may want to call this something like `basicItems` instead of `items`, since `items` is already used for the nested `items` array.

src/components/ebay-breadcrumb/template.marko
```diff
@@ -0,0 +1,12 @@
+<if(data.items)>
+    <nav aria-labelledby="breadcrumbs_heading" class="breadcrumb" role="navigation" ${data.htmlAttributes} w-bind w-onclick=(data.preventDefault && 'handleClick') ${data.preventDefault}>
+        <h2 id="breadcrumbs_heading" class="clipped">${data.ariaLabel}</h2>
+        <ul>
+            <for(item in data.items)>
+                <li if(item)>
+                    <${item.tag} ${item.htmlAttributes} w-body=item.renderBody></${item.tag}>
```
yomed: (minor) You can use `/>` instead of the closing tag (since it has no inner content).

src/components/ebay-breadcrumb/test/test.browser.js
```diff
@@ -0,0 +1,47 @@
...
+const mock = require('../mock');
+const renderer = require('../');
+
+describe('breadcrumb with prevent-default state', () => {
+    let widget;
+    let list;
+    describe('when breadcrumb item is clicked', () => {
+        let clickSpy;
+        beforeEach((done) => {
+            widget = renderer.renderSync(mock.itemsWithPreventDefault).appendTo(document.body).getWidget();
```
yomed: Rendering should be done in the outermost `describe()` -- check some of the existing tests for this setup.

src/components/ebay-breadcrumb/test/test.server.js
```diff
@@ -0,0 +1,65 @@
+const expect = require('chai').expect;
+const testUtils = require('../../../common/test-utils/server');
+const mock = require('../mock/index');
+
+test('shoudl return zero nav element in the page', context => {
+    const input = {};
+    const $ = testUtils.getCheerio(context.render(input));
+    expect($('nav').length).to.equals(0);
```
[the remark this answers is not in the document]
yomed: `.to.equal()` please

src/components/ebay-breadcrumb/test/test.server.js
```diff
@@ -0,0 +1,65 @@
...
+    const anchor = $('nav li a');
+    const _sp = [];
+    const navsrc = [];
+    anchor.each((i, v) => {
+        _sp.push($(v).attr('_sp'));
+        navsrc.push($(v).attr('navsrc'));
+    });
+    expect(_sp.length).to.equals(4);
+    expect(navsrc.length).to.equals(4);
+});
```
yomed: Make sure to include these server tests:
https://github.com/eBay/ebayui-core/blob/31b96b8f4484092780bfc8acb71c47b6a6353b9f/src/components/ebay-menu/test/test.server.js#L84-L90

src/components/ebay-breadcrumb/examples/1-breadcrumb-prevent-default/template.marko
seangates: I think you mean "omit" the `=true` part, right @yomed?

src/components/ebay-breadcrumb/index.js
seangates: I agree with this.

src/components/ebay-breadcrumb/mock/index.js
seangates: Agree.

src/components/ebay-breadcrumb/index.js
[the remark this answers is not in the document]
seangates: I thought it was enforced. Hmmm.

src/components/ebay-breadcrumb/test/test.server.js
```diff
@@ -2,64 +2,59 @@ const expect = require('chai').expect;
 const testUtils = require('../../../common/test-utils/server');
 const mock = require('../mock/index');
 
-test('shoudl return zero nav element in the page', context => {
+describe('should render with basic breadcrumb and ', () => {
+    test('return nav element with input', context => {
+        const $ = testUtils.getCheerio(context.render(mock.basicItems));
```
seangates: You're still using the same render, so you can add a `beforeEach` to make the code simpler, and more DRY.

```
describe('should render with basic breadcrumb and ', () => {
    let $;

    beforeEach(context => {
        $ = testUtils.getCheerio(context.render(mock.basicItems));
    });

    test('return nav element with input', () => {
        ...
    });

    ...
});
````

File: src/components/ebay-breadcrumb/template.marko [hunk 22]
```diff
@@ -0,0 +1,12 @@
+<if(data.items)>
+    <nav aria-labelledby="breadcrumbs_heading" class="breadcrumb" role="navigation" ${data.htmlAttributes} w-bind w-onclick=(data.preventDefault && 'handleClick') ${data.preventDefault}>
```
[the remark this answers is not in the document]
RajaRamu: this is just a event delegation. Bound to parent to avoid too many event bindings with anchor

src/components/ebay-breadcrumb/README.md
```diff
@@ -0,0 +1,32 @@
...
+`heading` | String | No | heading for breadcrumb which will be clipped
+`prevent-default` | Boolean | No | this will be used for single page app. when item is clicked, `breadcrumb-click` will be emitted.
+
+## ebay-breadcrumb-item Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`href` | String | No | anchor href
+
+## ebay-breadcrumb Events
```
seangates: This section should be before any nested tags (e.g., before line 19).

src/components/ebay-breadcrumb/test/test.server.js
```diff
@@ -0,0 +1,59 @@
+const expect = require('chai').expect;
+const testUtils = require('../../../common/test-utils/server');
+const mock = require('../mock/index');
+
+describe('should render with basic breadcrumb and ', () => {
+    test('return nav element with input', context => {
+        const $ = testUtils.getCheerio(context.render(mock.basicItems));
```
seangates: You're still using the same render for the same mock, so you can add a beforeEach to make the code simpler, and more DRY.

```
describe('should render with basic breadcrumb and ', () => {
    let $;

    beforeEach(context => {
        $ = testUtils.getCheerio(context.render(mock.basicItems));
    });

    test('return nav element with input', () => {
        ...
    });

    ...
});
```

src/components/ebay-breadcrumb/test/test.browser.js
RajaRamu: it has two set of inputs `mock.itemsWithPreventDefault` and `mock.basicItems` to render. i do not see value of moving out to describe.

src/components/ebay-breadcrumb/test/test.server.js
RajaRamu: `beforeEach` will be called for every test in this block. so its a same right?

seangates: Yes, but think of the ~children~ developer who come after you. 😉

src/components/ebay-breadcrumb/template.marko
```diff
@@ -0,0 +1,12 @@
+<if(data.items)>
+    <nav aria-labelledby="breadcrumbs_heading" class="breadcrumb" role="navigation" ${data.htmlAttributes} w-bind w-onclick=(data.preventDefault && 'handleClick')>
+        <h2 id="breadcrumbs_heading" class="clipped">${data.heading}</h2>
+        <ul>
+            <for(item in data.items)>
+                <li if(item)>
```
seangates: Why the extra check? We're in a `<for>` loop so I would assume `item` is implicitly present.

ianmcburnie:
@RajaRamu please rebase your branch with 0.2.0 in order to get the updates to package.json and yarn.lock for latest prerelease of skin.

From within `0.2.0`, please run `git pull`
From within `21-breadcrumb`, please run `git rebase 0.2.0`
If no conflicts and your local git log has the two latest commits from 0.2.0, please push changes to your branch.

src/components/ebay-breadcrumb/test/test.browser.js
yomed: You're right that it doesn't make a practical difference for very simple cases like this, but this is part of the structure we've been following in browser tests. The outermost `describe()` (which should start with "given ...") describes a given initial setup, so rendering fits naturally in that stage.

src/components/ebay-breadcrumb/test/test.server.js
seangates: Go ahead and leave as-is. We can look into this later.

src/components/ebay-breadcrumb/test/test.server.js
```diff
@@ -0,0 +1,68 @@
...
+    const currentElement = $('span', li[li.length - 1]);
+    expect(currentElement.attr('aria-current')).to.equal('page');
+});
+test('return headingLevel', context => {
+    const $ = testUtils.getCheerio(context.render(mock.itemsWithHeadingLevel));
+    const h2Tag = $('h2');
+    expect(h2Tag.length).to.equal(0);
+    const h3Tag = $('h3');
+    expect(h3Tag.length).to.equal(1);
+});
```
yomed: Can you use the common utils to test pass-through on both the root and the `li`s?
https://github.com/eBay/ebayui-core/blob/fa69045f03d59b5c5f9b749645bb8af32f6dde37/src/components/ebay-menu/test/test.server.js#L47-L53

src/components/ebay-breadcrumb/index.js
```diff
@@ -0,0 +1,42 @@
+const markoWidgets = require('marko-widgets');
+const emitAndFire = require('../../common/emit-and-fire');
+const processHtmlAttributes = require('../../common/html-attributes');
+const template = require('./template.marko');
+
+function getTemplateData(state, input) {
+    const htmlAttributes = processHtmlAttributes(input);
+    const items = input.items || [];
+    let transformedItems = (items).map((item, index) => {
```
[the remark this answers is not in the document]
yomed: Ah ok. Maybe call the original var `inputItems`, and use `items` for the transformed ones.

src/components/ebay-breadcrumb/index.js
```diff
@@ -0,0 +1,42 @@
...
+        } else if (href) {
+            itemHtmlAttributes.href = href;
+        }
+        return {
+            tag: tag,
+            htmlAttributes: itemHtmlAttributes,
+            renderBody: item.renderBody
+        };
+    });
+    transformedItems = Object.keys(transformedItems).length > 0 ? transformedItems : null;
```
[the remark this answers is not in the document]
RajaRamu: in case of transformedItems is empty, then it will create <nav> tag unnecessarily. thats why we have this check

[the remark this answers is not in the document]
yomed: Maybe check `if (data.items.length)` instead, since it will always be an array.

src/components/ebay-breadcrumb/test/test.browser.js
```diff
@@ -0,0 +1,26 @@
...
+            list = document.querySelectorAll('nav li a');
+            clickSpy = sinon.spy();
+            widget.on('breadcrumb-click', clickSpy);
+            testUtils.triggerEvent(list[0], 'click');
+            setTimeout(done);
+        });
+        afterEach(() => widget.destroy());
+
+        it('then it emits the breadcrumb-click event', () => {
+            expect(clickSpy.calledOnce).to.equal(true);
```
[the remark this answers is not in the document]
RajaRamu: done

src/components/ebay-breadcrumb/test/test.server.js
RajaRamu: Tried with testUtils.testHtmlAttributes(context, 'nav li a') -- looks like its rendering with * attr input. it does not work as for me. can it be generalized?

src/components/ebay-breadcrumb/index.js
```diff
@@ -0,0 +1,43 @@
...
+    });
+    items = Object.keys(items).length > 0 ? items : null;
+    return {
+        items,
+        headingText: input.headingText || '',
+        headingLevel: input.headingLevel || 'h2',
+        htmlAttributes: htmlAttributes
+    };
+}
+function handleClick(event, el) {
```
senthilp: @mlrawlings @DylanPiercey What does Marko pass in to the second parameter `el`? Is it `event.target` or `event.currentTarget`?

src/components/ebay-breadcrumb/index.js
```diff
@@ -0,0 +1,43 @@
...
+    emitAndFire(this, 'breadcrumb-click', { event, currentTarget: el });
```
senthilp: `currentTarget` should ideally be `target`. The consumers probably need the anchor or span tag where the event originated and not the container `nav` element. They can still get it from the `event` object, but this will be easy.

RajaRamu: Emitting 3 values as `{ event, target: event.target, currentTarget: el }`

src/components/ebay-breadcrumb/index.js
RajaRamu: Emitting 3 values now - { event, target: event.target, currentTarget: el } since `event.currentTarget` is pointing to body

ianmcburnie: This doesn't feel consistent with the event fired by menu component. Let's meet to discuss.

src/components/ebay-breadcrumb/test/test.server.js
yomed: The util can be changed if needed, but I think it should work as-is for your case. Take a look at the 2 examples in the menu case, where it calls the util slightly differently for the cases of root vs. nested tag.

src/components/ebay-breadcrumb/browser.json
```diff
@@ -0,0 +1,6 @@
+{
+    "dependencies": [
+        "@ebay/skin/breadcrumb",
```
yomed: I think you'll need to depend on the `utility` to use the `.clipped` class.
https://github.com/eBay/ebayui-core/blob/b9f3790c3811ed85629b64126196074d02cc7c4a/src/components/ebay-carousel/browser.json#L3

Please also add a `breadcrumb` link to the main `README.md` section:
https://github.com/eBay/ebayui-core/blob/0.2.0/README.md#components

src/components/ebay-breadcrumb/index.js
senthilp: Maybe we just need the `event` parameter and mention in docs that this event is fired from the root element. Developers can then use `event.detail.event.target` or `event.detail.event. currentTarget` based on their needs. Also renaming `event` to maybe `origEvent` will help to avoid confusion. But we need to check how other parts of the codebase is doing it.

src/components/ebay-breadcrumb/browser.json
RajaRamu: done

src/components/ebay-breadcrumb/test/test.server.js
RajaRamu: this test is added at L30-L36

src/components/ebay-breadcrumb/README.md
```diff
@@ -20,7 +20,7 @@ Name | Type | Stateful | Description
 
 Event | Description | Data
 --- | ---
-`breadcrumb-click` | click breadcrumb items | `{ event, target:event.target, currentTarget: el }`
+`breadcrumb-select` | click breadcrumb items | `{ event, target:event.target, currentTarget: el }`
```
ianmcburnie: For consistency, I think we should make the event signature match `menu-select`:

`emitAndFire(this, 'menu-select', { index: itemIndex, checked: [itemIndex], el: itemEl });`

(replace `checked` with `current`)

Pagination would also be similar:

`emitAndFire(this, 'pagination-select', { index: itemIndex, current: [itemIndex], el: itemEl });`

In next release we can discuss how to make it easier to prevent the default click action.

RajaRamu: can we just pass `el` alone like below. i think passing index does not make sense for links on breadcrumb. thought?

```emitAndFire(this, 'breadcrumb-select', {el: event.target});```

RajaRamu: done as we discussed `emitAndFire(this, 'breadcrumb-select', {el: event.target});`

ianmcburnie: okay, I guess we can add the index and currentItem in future if necessary.

PR #69, opened 2018-03-27
Fixing to walk the whole tree when transforming;

### LINKED ISSUE #64: BUG: transform does not walk the tree of all nested tags
Currently the transform expects direct nested tags, and does not allow any other Marko tags.

For example, the `<ebay-select>` expects a direct nested tag `<ebay-select-option>`, and if you add any control tags, like `<if()>` or `<for()>`, you will break rendering.

The fix is to walk the entire tree of children tags and apply the transform to _any_ tag that matches, regardless of whether the nested tag is an immediate child of the tag.

seangates:
## Description
The previous transform was only one level deep, to the immediate children of the given component tag. This transformer now walks the entire tree looking for children that match the pattern of `{parentTag}-` (looking for the hypen, specifically).

Additionally, the tests now cover this nested tag, along with updating the paths for essential Marko APIs used during the transform.

## Context
This was reported from our first user, who couldn't get the nested tags to show the options, even though everything was set up as we would expect.

## References
Closes #64

### DIFF
--- a/src/common/transformer/test/test.server.js
+++ b/src/common/transformer/test/test.server.js
@@ -11,8 +11,9 @@ try {
     Builder = require('marko/compiler/Builder');
 } catch (e) {
     // v4 paths
-    CompileContext = require('marko/dist/compiler/CompileContext');
-    Builder = require('marko/dist/compiler/Builder');
+    const target = require('marko/env').isDebug ? 'src' : 'dist';
+    CompileContext = require(`marko/${target}/compiler/CompileContext`);
+    Builder = require(`marko/${target}/compiler/Builder`);
 }
 
 function getTransformedTemplate(srcString, templatePath) {
@@ -27,6 +28,7 @@ function getTransformedTemplate(srcString, templatePath) {
     );
 
     transformer(templateAST.body.array[0], context);
+
     return prettyPrint(templateAST).replace(/\>\s*</g, '><').trim();
 }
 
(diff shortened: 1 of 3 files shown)


src/common/transformer/test/test.server.js
```diff
@@ -11,8 +11,9 @@ try {
     Builder = require('marko/compiler/Builder');
 } catch (e) {
     // v4 paths
-    CompileContext = require('marko/dist/compiler/CompileContext');
-    Builder = require('marko/dist/compiler/Builder');
+    const target = require('marko/env').isDebug ? 'src' : 'dist';
+    CompileContext = require(`marko/${target}/compiler/CompileContext`);
+    Builder = require(`marko/${target}/compiler/Builder`);
```
DylanPiercey: Are we testing that this compiles properly in v3?

seangates: @mlrawlings ☝️ ?

seangates: I am assuming it does compile properly, given that it passes the v3 and v4 integration tests.

yomed:
Did you mean to include the empty example template? 
`3-nested-options/template.marko`

seangates:
@yomed I did not! That should have an example in it. Let me fix that.

ianmcburnie:
Is this meant to go into master?

seangates:
@ianmcburnie Yes, this is a bug that needs to be fixed in the 0.1.x minor version, and therefore would become 0.1.2.

PR #70, opened 2018-03-27
Changed semver range for makeup dependencies

ianmcburnie:
## Description

Changed semver range for makeup dependencies in 0.0.x version and regenerated yarn.lock.

## Context

The caret is not correct for 0.0.x. It meant patch updates would not be taken.

```
^0.2.3 := >=0.2.3 <0.3.0
^0.0.3 := >=0.0.3 <0.0.4
```

The regenerated yarn.lock now has latest patch releases in makeup which fix the .babelrc issue (https://github.com/makeup-js/makeup-expander/issues/5).

If approved these changes will be immediately published as v0.1.1 patch release.

## References

https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,10 +93,10 @@
     "lodash.get": "^4.4.2",
     "lodash.set": "^4.3.2",
     "lodash.throttle": "^4.1.1",
-    "makeup-expander": "^0.3.0",
-    "makeup-key-emitter": "~0.0.2",
-    "makeup-prevent-scroll-keys": "~0.0.1",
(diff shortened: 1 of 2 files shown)


yomed:
You might want to change the actual version numbers in the package.json since that would more accurately reflect what we actually depend on, but it doesn't matter much either way since the `yarn.lock` has the final version resolution anyway.

ianmcburnie:
Ah yes. Good idea. Will fix.

Issue #52, opened 2018-03-16
Build: allow local demo to optionally run on transpiled dist code


ianmcburnie:
So that we can catch and fix bugs in all browsers, we need to be running the transpiled dist code in our local demo.

Running `yarn start` should add a transpile step to it's pipeline, and local demo should have capability to be running on the transpiled code in dist.

Related to: #59

yomed:
As discussed, the default case can be against `src/`, but there should be an easy way to run against `dist/` for debugging edge cases. The `prepublish` and `postpublish` scripts take care of most of the `dist` setup, but we would also need to have a script that points demo to `dist/`. We could potentially make the publishing scripts more generic to achieve this.

DylanPiercey:
After talking with @yomed I think an easier solution here would just be to conditionally add the `lasso-babel-transform` to the demo based on `NODE_ENV`.

We could then trigger the demo to start in dist/production mode by using `NODE_ENV=production npm start` and also probably have an alias like `npm start:prod` or similar.

PR #73, opened 2018-04-02
Build: add command for demo with transpilation

yomed:
## Description

- add `yarn start:prod` command

## Context

This allows us to easily run the demo on transpiled source code (like for debugging older browsers). 

## References

https://github.com/eBay/ebayui-core/issues/52


marko.json
```diff
@@ -69,7 +69,7 @@
         "@hidden": "boolean",
         "@aria-text": "string",
         "@aria-label-close": "string",
-        "@dismissible":"boolean",
+        "@dismissible": "boolean",
```
yomed: This change will happen from prepublish/postpublish processing, so might as well sneak it in.

ianmcburnie:
This is great!

Issue #54, opened 2018-03-16
Build: integrate eslint-plugin-compat


ianmcburnie:
https://github.com/amilajack/eslint-plugin-compat

senthilp:
Pranav Jha has already done something similar in C2C. We should try to leverage that.

Issue #53, opened 2018-03-16
Documentation: add browser policy to README


ianmcburnie:
Should be identical to the eBay policy.

PR #74, opened 2018-04-02
Build: Add browser policy and eslint plugin

yomed:
## Description

- condense browser policy into a `browserslist` definition
- use this list to drive `eslint-plugin-compat`

## Context

This helps define our browser support both internally and externally. The eBay policy has 3 tiers, so effectively we need to support tier 1 and 2 (as a component library). We could delineate the exact tiers and browsers, but I don't really want to repeat either the actual eBay Browser Policy or our `browserslist`.

## References

https://github.com/amilajack/eslint-plugin-compat
#53 
#54 

## Screenshots
Failure looks like this:
```sh
  174:5  error  fetch is not supported in Safari 6, iOS Safari 7.0-7.1, IE Mobile 10, IE 9, Edge 12, Android Browser 3  compat/compat

✖ 1 problem (1 error, 0 warnings)

error An unexpected error occurred: "Command failed.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,19 @@
     "lint",
     "test"
   ],
+  "browserslist": [
+    "Chrome >= 49",
+    "Firefox >= 48",
+    "Safari >= 6",
+    "Edge >= 12",
+    "IE >= 9",
+    "Opera 47-48",
+
+    "ChromeAndroid >= 38",
+    "ios_saf >= 7",
+    "ie_mob >= 10",
+    "Android >= 3"
+  ],
   "files": [
     "dist/",
     "marko.json",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -34,6 +34,19 @@
...
+    "Safari >= 6",
+    "Edge >= 12",
+    "IE >= 9",
+    "Opera 47-48",
+
+    "ChromeAndroid >= 38",
+    "ios_saf >= 7",
+    "ie_mob >= 10",
+    "Android >= 3"
+  ],
```
seangates: Eventually this should come from a global require, correct? We store it in the @ebay org? https://github.com/browserslist/browserslist#shareable-configs

/cc @senthilp

senthilp: Yes, that is the plan. For now, this should be ok.

seangates: I've created a task to track and prioritize it: https://github.com/eBay/ebayui-core/issues/76.

PR #81, opened 2018-04-05
Carousel: initial version

yomed:
## Description

- major refactoring to carousel for initial version

Plus a few somewhat related features:
- pass lasso flags to demo page
- add `.only` support for demo debugging
- increase default mocha timeout

## Context

Massive improvements over the existing beta carousel version. Includes heavy browser testing too. The largest caveat at the moment is that when we use native scrolling (`touch` + `type="continuous"`) the typical API and events don't work in the same way as when we handle scrolling ourselves.

The code does have a handful of FIXMEs and TODOs, but nothing that would block merge IMO.

## Screenshots

(Lots more in the demo `examples/`)
screen shot 2018-04-05 at 2 41 35 pm


senthilp:
@yomed Just curious, where you able to test if the CSS transitions where rendered smoothly on a real device?

demo/index.js
```diff
@@ -41,8 +42,25 @@ app.get('/:component?', (req, res) => {
...
+    });
+
     req.model = model;
 
+    const md = new MobileDetect(req.headers['user-agent']);
+    const lassoFlags = ['skin-ds6'];
+    if (md.mobile() || md.tablet()) {
+        lassoFlags.push('touch');
+    } else {
+        lassoFlags.push('no-touch');
```
seangates: Is there a need for `no-touch`? It seems to me you could simply do wha you need with testing for the presence of `touch`. Lasso flags are meant to be boolean in nature.

Or, in the inverse, you could just test for whether it's _not_ mobile or tablet: `!md.mobile() && !md.tablet()`, then add a `no-touch` flag, since we're defaulting to mobile. Your Lasso would then just check for whether the `no-touch` flag is present or not (just like our attribute booleans).

yomed: Just to note, I'm currently defaulting to `no-touch` because of this: https://github.com/eBay/ebayui-core/pull/81/files/5d866c53d1e5c6339fefe54be94580fe7311e356#diff-9142d1141b49ae232a58033f834bf93cR17

If we want to keep this slightly future-proof, it could help to have both flags available, and have each component be able to use neither/either/both. Similar to how applications use both `isSmall` and `isLarge`. That said, you're right that at the moment, using only a `touch` flag would work.

yomed:
@senthilp Thanks for the reminder -- I have done some at various stages in development, but I'll do more testing on real devices before merge.

src/components/ebay-carousel/browser.json
```diff
@@ -1,9 +1,23 @@
 {
     "dependencies": [
+        "@ebay/skin/utility",
         "@ebay/skin/less",
         "@ebay/skin/iconfont",
         "@ebay/skin/dist/icon/ds4/icon.css",
```
DylanPiercey: Since this component is explicitly referencing the ds4 icons it can break the ds6 icon styles (in my case the close button). We should update this to use ds6 icons.

ianmcburnie:
Judging by the backlog of PRs, I will probably target carousel for v0.4.0. I will create the branch when 0.3.0 is out

src/components/ebay-carousel/style.less
```diff
@@ -1,115 +1,135 @@
-.carousel {
-    @transition-time: 0.25s;
+@ebay-carousel-transition-time: 0.25s;
+@ebay-carousel-transition-function: ease-out;
+@ebay-carousel-control-enabled-opacity: 0.92;
+@ebay-carousel-control-disabled-opacity: 0.3;
```
seangates: Does it need the `ebay-` prefix? I doubt there would be collisions by only having `carousel-`.

seangates: Just a thought. No request to change it.

src/components/ebay-carousel/style.less
```diff
@@ -1,115 +1,135 @@
...
-    &--discrete &__list {
-        display: flex;
+        &--next {
+            right: 0;
+        }
 
-        > li {
-            flex: 0 0 100%;
+        &.icon:before {
+            color: @color-core-gray-jet;
```
seangates: This appears to be a DS4 color? I think it needs to change to a DS6 color, possibly `@ds6-color-g206-gray`.

http://ebay.github.io/skin/ds6/#less

src/components/ebay-carousel/style.less
```diff
@@ -1,115 +1,135 @@
...
-    }
+        &.icon--chevron-light-right:before {
+            margin-left: 1px;
+        }
 
-    .icon--chevron-light-right {
-        &:before {
-            margin-left: -4px;
+        &.icon--chevron-light-left:before {
+            margin-left: -1;
```
seangates: This needs a unit. `px` I believe?

src/components/ebay-carousel/style.less
```diff
@@ -1,115 +1,135 @@
...
+
+        > li {
+            display: inline-block;
 
-        &__control--prev,
-        &__control--next {
-            display: none;
+            > button {
+                transition: background-color @ebay-carousel-transition-time @ebay-carousel-transition-function;
+                background-color: @color-core-gray-silver;
```
seangates: Same color comment as above. Please change to `@ds6-color-g206-gray`.

src/components/ebay-carousel/style.less
```diff
@@ -1,115 +1,135 @@
...
+                border: 0;
+                border-radius: 50%;
+                display: inline-block;
+                margin: 0 6px;
+                padding: 0;
+                height: 8px;
+                width: 8px;
+
+                &.carousel__dot--active {
+                    background-color: @color-core-gray-jet;
```
seangates: Same color comment as above. Please change to `@ds6-color-g206-gray`.

src/components/ebay-carousel/style-no-touch.less
```diff
@@ -0,0 +1,35 @@
...
+            pointer-events: auto;
+
+            &[aria-disabled="true"] {
+                opacity: @ebay-carousel-control-disabled-opacity;
+                cursor: default;
+            }
+        }
+
+        &:hover:not([aria-disabled="true"]) {
+            background-color: @color-core-gray-dim;
```
seangates: Need a DS6 color here (not sure which one).

src/components/ebay-carousel/style-no-touch.less
```diff
@@ -0,0 +1,35 @@
...
+
+            &.icon:before {
+                color: @color-core-white;
```
seangates: Need a DS6 color here: `@ds6-color-g201-gray`.

src/components/ebay-carousel/style-no-touch.less
```diff
@@ -0,0 +1,35 @@
...
+            }
+        }
+
+        &:hover:active:not([aria-disabled="true"]) {
+            background-color: @color-core-gray-davys;
```
seangates: Need a DS6 color here (not sure which one).

src/components/ebay-carousel/style-touch.less
```diff
@@ -0,0 +1,30 @@
+.carousel {
+    &__control {
+        &:active:not([aria-disabled="true"]) {
+            background-color: @color-core-gray-davys;
```
seangates: Need a DS6 color here (not sure which one).

src/components/ebay-carousel/style-touch.less
```diff
@@ -0,0 +1,30 @@
...
+
+            &.icon:before {
+                color: @color-core-white;
```
seangates: Need a DS6 color here: `@ds6-color-g201-gray`

src/components/ebay-carousel/style.less
```diff
@@ -1,115 +1,135 @@
...
-                color: @color-core-white;
-            }
+        &-1 .carousel__list > li {
+            flex: 0 0 100%;
         }
 
-        &:hover:active:not([aria-disabled=true]) {
-            background-color: @color-core-gray-davys;
+        &-2 .carousel__list > li {
+            flex: 0 0 calc(~"50% - 8px");
```
seangates: @DylanPiercey did this cool thing:

```flex: 0 0 calc(50% ~"-" 8px);```

File: src/components/ebay-carousel/style.less (hunk 5 above)
Comment (seangates) [reply src/components/ebay-carousel/style.less]: @yomed And this one?

Comment (yomed) [reply src/components/ebay-carousel/style.less]: Arrow icons are broken anyway right now, so all of the icon stuff has to be redone.

Comment (seangates) [reply src/components/ebay-carousel/style.less]: So, I assume you'll be touching this code when you fix them?

Comment (yomed) [reply src/components/ebay-carousel/style.less]: Right

**Comment by yomed** [comment]
@ianmcburnie FYI coveralls is showing the correct coverage in the top comment which it updated (+5%), but it looks like the actual Github check does not get updated correctly (-0.2% -- I think it shows coverage diff between last 2 commits). Not sure if this can be fixed in our coveralls configuration.

File: src/components/ebay-carousel/README.md [hunk 14]
```diff
@@ -4,25 +4,50 @@
...
+--- | --- | ---
+`carousel-update` | { [visibleIndexes] } | called whenever item visibility changes, including initialization
+`carousel-translate` | | start of carousel content translation
+`carousel-next` | | click next control
+`carousel-prev` | | click previous control
+
+### Additional Events for type="discrete"
+Event | Data | Description
+--- | --- | ---
+`carousel-slide` | `{ slide }` | new slide is navigated to (by controls, dots, or API)
```
ianmcburnie: Do we need this "slide" event? I think `carousel-prev` and `carousel-next` events give us what we need for tracking clicks on those controls. For the dots, perhaps we have a separate `carousel-pagination-click` event that could be used for tracking...? As far as tracking the API call, I don't think we need to do that, the app could handle that when it makes the call. Maybe I'm missing some other purpose of this event though.

Also wondering if "translate" is a bit of a technical term, and something more like "animate" or "move" might be less ambiguous...? I know translate is used in CSS, but trying to find a word that fit's in more with the carousel metaphor.

yomed: The main idea here was to have controls-based events (used for tracking) which are distinct from state-based events like movement. That was the main purpose of `translate` -- since the actual movement is closer to how teams would use it, like in lazyloading. I'm ok  with renaming this to `move`.

Having a separate `pagination-click` event would be ideal, but unfortunately that would need a bunch of refactoring. I did put a note in the code: `// TODO: add carousel-dot event and remove simulated dot clicks`. Currently the `slide` event kind of makes up for the lack of a `pagination-click`.

ianmcburnie: Ok thanks. That's clearer now. I'm fine with leaving 'slide' in for now so long as we deprecate it when pagination-click comes along. I also find the word 'slide' a bit confusing, as we are using it as both a noun and verb in our lexicon, I can live with it though (the documentation does clear up any confusion quickly), but if anybody can think of a better name before Monday, let us know!

ianmcburnie:
@seangates Are your requested changes in?

PR #84, opened 2018-04-06
Add no-skin lasso flag

CestDiego:
## Description
Add condition to now load skin with each component if there is a flag present.

## Context
Global Header ships their own scoped skin version and using ebayui-core components ships with their own unscoped skin code, this defeats the purpose of having a scoped skin version. This PR fixes that problem by letting Global Header import components without skin, and injecting their own scoped skin into then.

## References
https://github.com/eBay/ebayui-core/issues/75

### DIFF
--- a/src/components/ebay-breadcrumb/browser.json
+++ b/src/components/ebay-breadcrumb/browser.json
@@ -1,7 +1,12 @@
 {
     "dependencies": [
-        "@ebay/skin/breadcrumb",
-        "@ebay/skin/utility",
+        {
+            "if-not-flag": "ebayui-no-skin",
+            "dependencies": [
+                "@ebay/skin/breadcrumb",
+                "@ebay/skin/utility"
+            ]
+        },
         "require: ./index.js"
     ]
 }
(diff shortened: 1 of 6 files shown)


senthilp:
Thanks @CestDiego and @seangates

seangates:
I expect we leave this out of the documentation. It is only pertinent in very specific cases. In fact, only one we know of.

ianmcburnie:
Let's go with `ebayui-no-skin`, undocumented. Nobody will know about this feature except us and the people who are lurking in this thread (I'll create an issue to have their memories erased).

CestDiego:
@ianmcburnie sorry for the delay. I have changed the skinless flag to the `no-skin` alternative.

seangates:
Build failed, this time on something in the carousel. @YoniMedoff, can you look at it on Monday?

seangates:
Never mind. Reran the build. It was simply a timeout.

yomed:
@CestDiego Would `ebay-breadcrumb` need this too?
https://github.com/eBay/ebayui-core/blob/0d5453409f8d9ab2300f6a663c3723302595b5fb/src/components/ebay-breadcrumb/browser.json#L3-L4

CestDiego:
@yomed We definitely need that. I branched off of master when making changes then cherry picked on top of 0.2.0 and I missed this. Latest commit https://github.com/eBay/ebayui-core/commit/799038d fixes this

PR #86, opened 2018-04-06
convert to string to compare the option array & selected option

PrashantAshok:
## Description
The 3rd example on marketplace doesn't work

## Context
Select UI doesn't change on click and on keyboard when the options are rendered dynamically.

### DIFF
--- a/src/components/ebay-select/index.js
+++ b/src/components/ebay-select/index.js
@@ -185,7 +185,7 @@ function processAfterStateChange(el) {
  * @param {String} optionValue
  */
 function setSelectedOption(optionValue) {
-    const newOptionSelected = this.state.options.filter(option => option.value === optionValue)[0];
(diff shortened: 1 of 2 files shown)


ianmcburnie:
@PrashantAshok CI Job failed at following unit test:

```
1 failing
  1) src/components/ebay-select
       given the listbox is in the default state
         when the down arrow key is pressed
           then it emits the listbox-change event with the correct data:
      AssertionError: expected '1' to equal '2'
      + expected - actual
```

seangates:
Yup, we're looking at it together, @ianmcburnie. Will have an update soon.

ianmcburnie:
What's the latest on this? Is fix going into 0.2.0?

seangates:
The latest is that I haven't been able to get past the failed test, and it only happens with this branch, so it's clearly something with how this logical change affects the test. Let me work with @PrashantAshok to close the loop today.

PR #90, opened 2018-04-09
Select: Fixes clicked element returned;

### LINKED ISSUE #89: Select: differing elements returned when clicking on an option
When clicking on one of the `<ebay-select-options>` you could possibly have the option `<div>` or the label `<span>` returned as the clicked element. We should default to the option `<div>`.

### Expected

The `ebay-select-option` div with the `role="option"` is returned.

### Actual

Depending on where you click, either the `ebay-select-option` div with the `role="option"` is returned, or the `ebay-select-option`'s label is returned, which is not the option, but simply the text inside the option.

seangates:
## Description
When a user clicks on the select's option, you could get different elements returned in the event. This PR will look for the target with the `optionValue`, which is expected to be available.

Additionally, removing the `optionValue` data attribute from the nested elements, since they should not have them.

## References
Fixes #89

### DIFF
--- a/src/components/ebay-select/index.js
+++ b/src/components/ebay-select/index.js
@@ -104,7 +104,18 @@ function handleCollapse() {
  * @param {MouseEvent} event
  */
 function handleOptionClick(event) {
-    this.processAfterStateChange(event.target);
+    let el;
+
+    // find the element with the data
+    // start with the target element
+    if (event.target.dataset.optionValue) {
+        el = event.target;
+    // check up the tree one level (in case option-label or status was clicked)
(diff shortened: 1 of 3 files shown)


ianmcburnie:
@seangates should this go into 0.2.1 patch release instead?

yomed:
This fix could use some accompanying tests. Would you be able to test it by mocking up different types of options?

seangates:
@yomed and @ianmcburnie, I added a test (and modified one) and things look better. Can you both take a quick look, and approve if all looks good?

seangates:
Will release this as 0.2.1 after it's merged.

PR #96, opened 2018-04-13
Added the resize util

### LINKED ISSUE #95: Create a resize event util

sharma46bhawana:
## Description
Added a resize util that reduces the callbacks on window resize.

## Context
Added a resize util that reduces the callbacks on window resize.

## References
Fixes #95


sharma46bhawana:
A special thanks to @DylanPiercey for helping me! :)

src/common/resize-event-util/index.js
```diff
@@ -0,0 +1,26 @@
+const handlers = [];
```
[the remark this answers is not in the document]
DylanPiercey: @yomed I think that would make sense as well.

Also regarding the rational:
` this.subscribeTo(window).on('resize', debounce(() => this.refresh(), 50)); `
Can cause things to jump around as you are resizing the page (which isn't the worst thing but looks kinda ugly). This approach ensures that we are running these resize based changes in line with when when browser renders.

The other benefit to this approach is that the resize event handler is only invoked at most once per animation frame where as the debounced resize function (although it does nothing but defer execution) is actually called on every resize event and this small cost increases as more component instances setup their own handler. This is done by adding the resize handler and immediately removing it once it has been invoked and only adding it again for the next animation frame.

Since a few components need to react to the resize event it seemed to make sense to extract this functionality out into a utility and also address the issue of delayed/debounced updates.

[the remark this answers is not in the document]
sharma46bhawana: Okay I will make the changes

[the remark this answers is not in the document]
aselvaraj: @DylanPiercey @sharma46bhawana this is fantastic!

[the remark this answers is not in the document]
senthilp: @DylanPiercey Smart approach. But one thing to note is that RAF will also fire at the same rate as resize event because it happens before recalc/layout/paint ([explanation](https://bugs.webkit.org/show_bug.cgi?id=177484)). A simple demo that I created [https://resize.glitch.me/](https://resize.glitch.me/). Event count is same in Chrome, whereas Safari RAF count is less because of the bug. 

Having said that your approach still saves on the number of events getting attahced to `resize` itself. At max we will have only one event listner.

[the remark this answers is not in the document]
DylanPiercey: @senthilp thanks for the link. I did realize that the resize events emitted were in tandem with the   RAF but I had (incorrectly) thought the issue was the other way around and that chrome had the unexpected behaviour and was doing something like it does for `passive` scroll handlers.

Anyway as you say there is still a benefit to this approach, it's just much smaller than I had imagined.
Ultimately though if we decide that this is truly bottlenecking the pages on resize (which doesn't happen too often 😆) we can always update this little utility to revert to the debouncing behaviour while retaining the other benefits.

src/common/event-utils/index.js
```diff
@@ -24,4 +24,35 @@ function handleUpDownArrowsKeydown(e, callback) {
     handleKeydown([38, 40], e, callback);
 }
 
-module.exports = { handleActionKeydown, handleEscapeKeydown, handleUpDownArrowsKeydown };
+const handlers = [];
+function addEventListener(_, handler) {
```
ianmcburnie: What is the underscore for?

src/common/event-utils/index.js
```diff
@@ -24,4 +24,35 @@ function handleUpDownArrowsKeydown(e, callback) {
...
+            window.addEventListener('resize', handleResize);
+        }
+    }, 16);
+}
+
+module.exports = {
+    handleActionKeydown,
+    handleEscapeKeydown,
+    handleUpDownArrowsKeydown,
+    'resizeObserver': {
```
ianmcburnie: Does `resizeObserver` object need single quotes?

src/common/event-utils/index.js
```diff
@@ -24,4 +24,35 @@ function handleUpDownArrowsKeydown(e, callback) {
...
 
-module.exports = { handleActionKeydown, handleEscapeKeydown, handleUpDownArrowsKeydown };
+const handlers = [];
+function addEventListener(_, handler) {
+    if (handlers.length === 0) {
+        window.addEventListener('resize', handleResize);
+    }
+    handlers.push(handler);
+}
+function removeEventListener(_, handler) {
```
ianmcburnie: Are `addEventListener` and `removeEventListener` methods only intended for window resize event? If so, these methods should be renamed accordingly. Same applies to the `handlers` variable.

DylanPiercey: @ianmcburnie they have be named (or at least exported as) `addEventListener` and `removeEventListener` to play nice with the `subscribeTo` method in Marko components.

ianmcburnie: Thanks. I expect it will become clearer when I see the integration with pagination component.

yomed: Right, maybe this should be part of the pagination PR, since it's a little strange out of context.

src/common/event-utils/index.js
sharma46bhawana: I don't think it does. I will remove it.

src/common/event-utils/index.js
sharma46bhawana: We don't need it but the `subscribeTo` takes in 2 arguments hence the ` _`

sharma46bhawana:
I don't have the permission to merge can someone else merge it?

Issue #91, opened 2018-04-12
Button: consider adding label attribute


yomed:
The button component currently allows stateful updating of its content through `w-body`, so users would have to update the state of the parent to change it. We may want to add a `label` attribute which can be used instead of `w-body` for use cases where there doesn't need to be complex HTML inside the button tag, and where users might want more easy access to stateful updating. 
cc @RajaRamu

PR #99, opened 2018-04-17
Button: add label attribute

yomed:
I'm not sure if we should do this change or not, so I'm really looking for feedback this far. If we do decide to go with this, I'll flesh out the tests for it.

## Description

- adds `label` attribute for `<ebay-button>` as an alternative to HTML body
- order demo examples correctly (first time we have more than 9 examples)

## Context

The button currently accepts its inner content with an HTML-like body. This allows complex HTML inside the button, but means that it can be more complicated to statefully update the content. A common use case with the button is to use a simple string label. For this case, it makes sense to expose that attribute as a stateful property. I was able to support both methods at the same time.

Note that this does add clientside rendering to the button. I'm also not sure if this pattern would apply all the time, or just for very simple components like this one.

## References

https://github.com/eBay/ebayui-core/issues/91

### DIFF
--- a/src/components/ebay-button/template.marko
+++ b/src/components/ebay-button/template.marko
@@ -1,6 +1,5 @@
 <${data.tag}
     w-bind
-    w-body
     w-onclick="handleClick"
     w-onkeydown="handleKeydown"
     type="button"
@@ -9,4 +8,10 @@
     disabled=data.disabled
     aria-disabled=data.partiallyDisabled
     ${data.htmlAttributes}>
+    <if(data.label)>
+        ${data.label}
+    </if>
+    <else>
+        <span w-body body-only-if(true)/>
+    </else>
 </>
--- a/demo/index.js
+++ b/demo/index.js
@@ -33,11 +33,12 @@ app.get('/:component?', (req, res) => {
     const model = {
         name: req.params.component,
         examples: fs.readdirSync(examplesPath).map(example => ({
+            num: parseInt(example.split('-')[0]),
             name: example.split('-').slice(1, example.length).join(' '),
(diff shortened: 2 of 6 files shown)


src/components/ebay-button/template.marko
```diff
@@ -9,4 +8,10 @@
     disabled=data.disabled
     aria-disabled=data.partiallyDisabled
     ${data.htmlAttributes}>
+    <if(data.label)>
+        ${data.label}
+    </if>
+    <else>
+        <span w-body body-only-if(true)/>
```
seangates: So, it would always add an inner `<span>`?

yomed: Exactly. This way we render `label` if it exists, and otherwise fall back to `w-body` (span is not rendered).

seangates:
You should probably write at least a couple of tests for this change.

yomed:
@seangates For sure, I just deferred it for now to get feedback:
> I'm not sure if we should do this change or not, so I'm really looking for feedback this far. If we do decide to go with this, I'll flesh out the tests for it.

seangates:
Ah, thanks. I think I just didn't read that part (or did not remember reading it).

ianmcburnie:
Which one takes precedence, label or body?

Also, in order to avoid confusion with `aria-label`, something like `text` or `value` (similar to input type="button") might be a better name.

yomed:
@ianmcburnie Currently `label` takes precedence, since I think I assumed that you would only include that attribute if you actually wanted it to be used. I used the name `label` since Menu and Select use that. We could potentially change them all?

Right now I'm leaning against merging this PR (and fixing the associated issue), because it does complicate things a bit, and I'm not sure what the criteria is for putting this feature into any given component. But, it does have some value. So it could be that we only need to add this if/when the need arises from teams that use it.

PR #100, opened 2018-04-17
New component: ebay-pagination

### LINKED ISSUE #20: New Component: Pagination
Details to follow...

Related issue: https://github.com/eBay/skin/issues/81

sharma46bhawana:
## Description
Contains a new component `ebay-pagination` which constructs page navigation.

## Context
- The pagination is constructed with a parent element `<ebay-pagination>` which takes as input a boolean value `fake-link` (for the ajax links), the `aria-*` (for `next`, `previous` and `heading` of the component) and its child element `<ebay-pagination-item>` which takes as input boolean type `previous`, `next` and `current` to provide context for the item and `disabled` to provide information about the navigating links.
- There are 3 events that will be fired: `pagination-previous`, `pagination-next` and `pagination-select` each returns the `target` with `pagination-select` emitting an additional `value` of the page clicked/pressed.
- When the `previous` or `next` is not provided in the input for `<ebay-pagination-item>` the navigation is taken as `disabled`. 

`(Note: To pass the tests I have added the resize-event-utils changes here, once that gets merged I will rebase it from 0.3.0)`

## References
Closes #20 

## Screenshots
screen shot 2018-04-17 at 12 06 40 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ For more information, please read [Building a UI Component in 2017 and Beyond](h
 * [`ebay-dialog`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-dialog)
 * [`ebay-menu`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-menu)
 * [`ebay-notice`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-notice)
+* [`ebay-pagination`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-pagination)
 * [`ebay-select`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-select)
 
 ## Getting Started
--- a/demo/browser.json
+++ b/demo/browser.json
@@ -6,12 +6,13 @@
         "@ebay/skin/marketsans",
         "./atom-light-syntax.css",
         "./style.less",
+        "../src/components/ebay-breadcrumb",
         "../src/components/ebay-button",
         "../src/components/ebay-carousel",
         "../src/components/ebay-dialog",
         "../src/components/ebay-menu",
-        "../src/components/ebay-select",
         "../src/components/ebay-notice",
-        "../src/components/ebay-breadcrumb"
+        "../src/components/ebay-pagination",
+        "../src/components/ebay-select"
     ]
 }
--- a/marko.json
+++ b/marko.json
@@ -93,5 +93,25 @@
             "@*": "expression"
         }
     },
-    "<ebay-breadcrumb-item>": {}
+    "<ebay-breadcrumb-item>": {},
+    "<ebay-pagination>": {
+        "renderer": "./src/components/ebay-pagination/index.js",
+        "transformer": "./src/common/transformer/index.js",
+        "@class": "string",
+        "@accessibility-prev": "string",
+        "@accessibility-next": "string",
+        "@accessibility-current": "string",
+        "@hijax": "boolean",
+        "@*": "expression",
+        "@items <item>[]": {
+            "@class": "string",
+            "@current": "boolean",
+            "@disabled": "string",
+            "@href": "string",
+            "@type": "string",
+            "@role": "string",
+            "@*": "expression"
+        }
+    },
+    "<ebay-pagination-item>": {}
 }
--- a/src/components/ebay-pagination/README.md
+++ b/src/components/ebay-pagination/README.md
@@ -1,0 +1,53 @@
+# ebay-pagination
+
+The `<ebay-pagination>` is a tag used to create a pagination navigation.
+
+## ebay-pagination Tag
+
+### ebay-pagination Usage
+
+```marko
+<ebay-pagination aria-label-prev="Previous page" aria-label-next="Next page" curr-text="Results Pagination - Page 2">
+    <ebay-pagination-item href="#" previous disabled/>
+    <ebay-pagination-item href="#">item 1</ebay-pagination-item>
+    <ebay-pagination-item href="#" current>item 2</ebay-pagination-item>
+    <ebay-pagination-item href="#">item 3</ebay-pagination-item>
+    <ebay-pagination-item href="#" next/>
+</ebay-pagination>
+```
+
+### ebay-pagination Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`class` | String | No | custom class
+`accessibility-prev` | String | No | aria-label for previous arrow button
+`accessibility-next` | String | No | aria-label for next arrow button
+`accessibility-current` | String | No | Description for the current page (e.g. Results of Page 1)
+`hijax` | Boolean | No | Use pagination links for an ajax reload
+
+### ebay-pagination Events
+
+Event | Data | Description
+--- | --- | ---
+`pagination-previous` |  `{ event, el: event.target }`| clicked previous arrow button
+`pagination-next` | `{ event, el: event.target }` | clicked next arrow button
+`pagination-select` | `{ event, el: event.target , value: "Selected page" }` | page selected clicked
+
+## ebay-pagination-item Tag
+
+### ebay-pagination-item Usage
+
+```marko
+<ebay-pagination-item>1</ebay-pagination-item>
+```
+
+### ebay-pagination-item Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`disabled` | Boolean | No | Item is disabled or not (Note: should only be used in case of navigation)
+`href` | String | No | for link that looks like a menu-item
+`current` | Boolean | No | the current page
+`next` | Boolean | No | whether or not the input is the next page (when not provided taken as `disabled`)
+`previous` | Boolean | No | whether or not the input is the next page (when not provided taken as `disabled`)
--- a/src/components/ebay-pagination/README.md
+++ b/src/components/ebay-pagination/README.md
@@ -46,8 +46,8 @@ Event | Data | Description
 
 Name | Type | Stateful | Description
 --- | --- | --- | ---
-`disabled` | Boolean | No | Item is disabled or not (Note: should only be used in case of navigation)
+`disabled` | Boolean | No | Previous/next button is disabled or not
+`class` | String | No | custom class
 `href` | String | No | for link that looks like a menu-item
 `current` | Boolean | No | the current page
-`next` | Boolean | No | whether or not the input is the next page (when not provided taken as `disabled`)
-`previous` | Boolean | No | whether or not the input is the next page (when not provided taken as `disabled`)
+`type` | String | No | "previous", "next" or "page"(default). To specify if the information entered is for the previous or next arrrow button or a page. If the `type='previous|next'` isn't provided the previous/next arrow buttons will be taken as `disabled`
--- a/src/components/ebay-pagination/browser.json
+++ b/src/components/ebay-pagination/browser.json
@@ -1,0 +1,14 @@
+{
+    "dependencies": [
+        {
+            "if-not-flag": "ebayui-no-skin",
+            "dependencies": [
+                "@ebay/skin/pagination",
+                "@ebay/skin/utility",
+                "@ebay/skin/icon"
+            ]
+        },
+        "require: marko-widgets",
+        "require: ./index.js"
+    ]
+}
--- a/src/components/ebay-pagination/index.js
+++ b/src/components/ebay-pagination/index.js
@@ -1,0 +1,159 @@
+const processHtmlAttributes = require('../../common/html-attributes');
+const eventUtils = require('../../common/event-utils');
+const emitAndFire = require('../../common/emit-and-fire');
+const template = require('./template.marko');
+
+const constants = {
+    indexForNavigation: 2,
+    minPagesRequired: 5,
+    maxPagesAllowed: 9,
+    margin: 8
+};
+
+function getInitialState(input) {
+    let prevItem;
+    let nextItem;
+    const items = [];
+    const inputItems = input.items || [];
+    const isHijax = input.hijax || false;
+    let role;
+
+    if (isHijax) {
+        role = 'button';
+    }
+
+    for (let i = 0; i < inputItems.length; ++i) {
+        const item = inputItems[i];
+        const href = item.href;
+        const current = item.current;
+
+        const tempItem = {
+            role,
+            href,
+            current: Boolean(current) || false,
+            htmlAttributes: processHtmlAttributes(item),
+            renderBody: item.renderBody
+        };
+        if (item.type === 'previous') {
+            prevItem = tempItem;
+            prevItem.class = 'pagination__previous';
+            prevItem.disabled = Boolean(item.disabled) || !href;
+            continue;
+        } else if (item.type === 'next') {
+            nextItem = tempItem;
+            nextItem.class = 'pagination__next';
+            nextItem.disabled = Boolean(item.disabled) || !href;
+            continue;
+        } else {
+            tempItem.class = ['pagination__item', item.class];
+        }
+        items.push(tempItem);
+    }
+
+    return {
+        isHijax,
+        nextItem: nextItem || { class: 'pagination__next', disabled: true },
+        prevItem: prevItem || { class: 'pagination__previous', disabled: true },
+        items,
+        accessibilityPrev: input.accessibilityPrev || 'Previous page',
+        accessibilityNext: input.accessibilityNext || 'Next page',
+        accessibilityCurrent: input.accessibilityCurrent || 'Results Pagination - Page 1',
+        classes: ['pagination', input.class],
+        htmlAttributes: processHtmlAttributes(input)
+    };
+}
+
+function getTemplateData(state) {
+    return state;
+}
+
+function init() {
+    this.pageContainerEl = this.el.querySelector('.pagination__items');
+    this.pageEls = this.pageContainerEl.children;
+    this.containerEl = this.el;
+    this.subscribeTo(eventUtils.resizeUtil).on('resize', refresh.bind(this));
+    this.refresh();
+}
+
+function refresh() {
+    const containerWidth = this.containerEl.offsetWidth;
+    const pageNumWidth = this.pageEls[0].offsetWidth + constants.margin;
+    const numPagesAllowed = Math.floor(containerWidth / pageNumWidth) - constants.indexForNavigation;
+    const adjustedNumPages = Math.min(constants.maxPagesAllowed,
+        Math.max(numPagesAllowed, constants.minPagesRequired));
+    const totalPages = this.pageEls.length;
+
+    // Let's show all the pages that we can.
+    for (let i = 5; i < adjustedNumPages && i < totalPages; ++i) {
+        if (this.pageEls[i].hasAttribute('hidden')) {
+            this.pageEls[i].removeAttribute('hidden');
+        }
+    }
+
+    // Now that we are showing all the pages that we can, lets hide remaining pages.
+    for (let i = adjustedNumPages; i < totalPages; ++i) {
+        this.pageEls[i].setAttribute('hidden', true);
+    }
+}
+
+function handleHijax(e) {
+    if (this.state.isHijax) {
+        e.preventDefault();
+    }
+}
+
+/**
+ * Handle normal mouse click for item, next page and previous page respectively.
+ * @param {MouseEvent} event
+ */
+function handlePageClick(event) {
+    this.handleHijax(event);
+    emitAndFire(this, 'pagination-select', { el: event.target, value: event.target.innerText });
+}
+
+function handleNextPage(event) {
+    this.handleHijax(event);
+    emitAndFire(this, 'pagination-next', { el: event.target });
+}
+
+function handlePreviousPage(event) {
+    this.handleHijax(event);
+    emitAndFire(this, 'pagination-previous', { el: event.target });
+}
+
+/**
+ * Handle accessibility for item, next page and previous page respectively.
+ * @param {KeyboardEvent} event
+ */
+function handlePageKeyDown(event) {
+    eventUtils.handleActionKeydown(event, () => {
+        this.handlePageClick(event);
+    });
+}
+
+function handleNextPageKeyDown(event) {
+    eventUtils.handleActionKeydown(event, () => {
+        this.handleNextPage(event);
+    });
+}
+
+function handlePreviousPageKeyDown(event) {
+    eventUtils.handleActionKeydown(event, () => {
+        this.handlePreviousPage(event);
+    });
+}
+
+module.exports = require('marko-widgets').defineComponent({
+    template,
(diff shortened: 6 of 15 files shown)


sharma46bhawana:
I don't have write access to this repository so I can't re-run the Travis-ci job. I see it failing due to timeout. Can someone rerun it?

seangates:
@sharma46bhawana Restarted it.

marko.json
```diff
@@ -93,5 +93,25 @@
...
     },
-    "<ebay-breadcrumb-item>": {}
+    "<ebay-breadcrumb-item>": {},
+    "<ebay-pagination>": {
+        "renderer": "./src/components/ebay-pagination/index.js",
+        "transformer": "./src/common/transformer/index.js",
+        "@class": "string",
+        "@aria-label-prev": "string",
+        "@aria-label-next": "string",
+        "@curr-text": "string",
```
yomed: Just to note, I used different names for approximately the same attributes, so we should try to keep these consistent (I'm fine changing mine too). @ianmcburnie Do you prefer certain names for these a11y attributes?

src/components/ebay-pagination/README.md
```diff
@@ -0,0 +1,56 @@
...
+</ebay-pagination>
+```
+
+### ebay-pagination Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`class` | String | No | custom class
+`aria-label-prev` | String | No | aria-label for previous button
+`aria-label-next` | String | No | aria-label for previous button
```
yomed: `for next button`

src/components/ebay-pagination/examples/1-basic/template.marko
```diff
@@ -0,0 +1,13 @@
+<ebay-pagination aria-label-prev="Previous page" aria-label-next="Next page" curr-text="Results Pagination - Page 1">
```
yomed: In the carousel, I omitted a11y attributes on the examples, except for one example that holds all of them: https://github.com/eBay/ebayui-core/pull/81/files#diff-b6c1a03d5e16301d2315f34a8b093c47R13

This way we can showcase the simplest usage (since all the a11y strings have defaults).

src/components/ebay-pagination/examples/1-basic/template.marko
```diff
@@ -0,0 +1,13 @@
...
+    <ebay-pagination-item href="#" previous></ebay-pagination-item>
+    <ebay-pagination-item href="#">1</ebay-pagination-item>
+    <ebay-pagination-item href="#" current>2</ebay-pagination-item>
+    <ebay-pagination-item href="#hh">3</ebay-pagination-item>
```
yomed: If `hh` doesn't mean anything, remove it so we can keep this simple (here and in the other examples).

src/components/ebay-pagination/README.md
```diff
@@ -0,0 +1,56 @@
...
+`curr-text` | String | No | Description for the current page (e.g. Results of Page 1)
+`fake-link` | Boolean | No | Use pagination links for an ajax reload
+
+### ebay-pagination Events
+
+Event | Data | Description
+--- | --- | ---
+`pagination-previous` |  `{ event, el: event.target }`| clicked previous arrow
+`pagination-next` | `{ event, el: event.target }` | clicked next arrow
```
yomed: (minor) `button` instead of `arrow` for consistency

src/components/ebay-pagination/examples/1-basic/template.marko
```diff
@@ -0,0 +1,13 @@
...
+    <ebay-pagination-item href="#">4</ebay-pagination-item>
+    <ebay-pagination-item href="#">5</ebay-pagination-item>
+    <ebay-pagination-item href="#">6</ebay-pagination-item>
+    <ebay-pagination-item href="#">7</ebay-pagination-item>
+    <ebay-pagination-item href="#">8</ebay-pagination-item>
+    <ebay-pagination-item href="#">9</ebay-pagination-item>
+    <ebay-pagination-item href="#" next disabled></ebay-pagination-item>
```
yomed: Just curious about the decision to make prev/next the same as `item`s, rather than using attributes like `prev-url` and `next-url` on the root.

src/components/ebay-pagination/index.js
```diff
@@ -0,0 +1,173 @@
...
+    if (isRoleButton) {
+        role = 'button';
+    }
+
+    for (let i = 0; i < inputItems.length; ++i) {
+        const item = inputItems[i];
+        const href = item.href;
+        const current = item.current;
+
+        tempObj = {
```
yomed: Looks like you can declare `const tempObj` here, instead of above.

src/components/ebay-pagination/index.js
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    // Now that we are showing all the pages that we can, lets hide remaining pages.
+    for (let i = adjustedNumPages; i < totalPages; ++i) {
+        this.pageEls[i].setAttribute('hidden', true);
+    }
+}
+
+function isPreventDefault(e) {
+    if (this.state.isRoleButton) {
```
yomed: `isRoleButton` makes sense above, but not here. Maybe rename it to `hasFakeLinks`

src/components/ebay-pagination/index.js
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    // Now that we are showing all the pages that we can, lets hide remaining pages.
+    for (let i = adjustedNumPages; i < totalPages; ++i) {
+        this.pageEls[i].setAttribute('hidden', true);
+    }
+}
+
+function isPreventDefault(e) {
```
yomed: I think this needs better naming, since it currently implies a boolean return.

src/common/event-utils/index.js
```diff
@@ -24,4 +24,35 @@ function handleUpDownArrowsKeydown(e, callback) {
     handleKeydown([38, 40], e, callback);
 }
 
-module.exports = { handleActionKeydown, handleEscapeKeydown, handleUpDownArrowsKeydown };
+const handlers = [];
```
yomed: I think you should nest all of this code inside of a `resizeEvent` object, and export that. The `observer` name conflicts a bit with how we use it in `propertyObserver`. It's also a bit strange having these generic names in the top scope of this module which holds lots of utils.

src/components/ebay-pagination/test/test.browser.js
```diff
@@ -0,0 +1,108 @@
...
+        let spy;
+        beforeEach(() => {
+            spy = sinon.spy();
+            widget.on('pagination-select', spy);
+            testUtils.triggerEvent(pageItem, 'click');
+        });
+
+        test('then it emits the marko event called pagination-select', () => {
+            expect(spy.calledOnce).to.equal(true);
+            expect(Object.keys(spy.args[0][0])).to.deep.equal(['el', 'value']);
```
yomed: Can you test the actual values rather than just the keys? I.e. `expect(eventData.value).to.equal(x)`

src/components/ebay-pagination/test/test.browser.js
```diff
@@ -0,0 +1,108 @@
...
+            widget.on('pagination-select', spy);
+            testUtils.triggerEvent(pageItem, 'keydown', 32);
+        });
+
+        test('then it emits the marko event called pagination-select', () => {
+            expect(spy.calledOnce).to.equal(true);
+            expect(Object.keys(spy.args[0][0])).to.deep.equal(['el', 'value']);
+        });
+    });
+});
```
yomed: Would be good to test the resize behavior too, something like this:
https://github.com/eBay/ebayui-core/pull/81/files#diff-312b1af3550afda88d75e63a6c6bfcc8R31

src/components/ebay-pagination/test/test.server.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+
+    test('handles pass-through html attributes', context => {
+        testUtils.testHtmlAttributes(context, '.pagination');
+    });
+
+    test('handles custom class', context => {
+        testUtils.testCustomClass(context, '.pagination');
+    });
```
yomed: Can you do `testHtmlAttributes` and `testCustomClass` on `pagination-item` too?

README.md
```diff
@@ -22,8 +22,9 @@ For more information, please read [Building a UI Component in 2017 and Beyond](h
 * [`ebay-breadcrumb`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-breadcrumb)
 * [`ebay-button`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-button)
 * [`ebay-menu`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-menu)
-* [`ebay-select`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-select)
 * [`ebay-notice`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-notice)
+* [`ebay-pagination`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-pagination)
+* [`ebay-select`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-select)
```
seangates: @ianmcburnie and I love you for making this alphabetical.

demo/browser.json
```diff
@@ -12,6 +12,7 @@
         "../src/components/ebay-menu",
         "../src/components/ebay-select",
         "../src/components/ebay-notice",
-        "../src/components/ebay-breadcrumb"
+        "../src/components/ebay-breadcrumb",
+        "../src/components/ebay-pagination"
```
seangates: Maybe make this alphabetical?

src/components/ebay-pagination/README.md
sharma46bhawana: Oops!

src/components/ebay-pagination/examples/1-basic/template.marko
sharma46bhawana: Yeah I was just checking something and forgot to remove it.

README.md
sharma46bhawana: I learned from the best 😉

demo/browser.json
sharma46bhawana: will do

src/components/ebay-pagination/examples/1-basic/template.marko
sharma46bhawana: There might be other html-attributes like tracking that might need to be added so I made them an item. Otherwise the root would become too cluttered

marko.json
sharma46bhawana: I referred to the old carousal while naming I think. I will take a look at your names and change them.

src/components/ebay-pagination/examples/1-basic/template.marko
sharma46bhawana: Okay will update these.

src/components/ebay-pagination/README.md
sharma46bhawana: I didn't want to say button as they aren't technically a `button` tag, so it might cause confusion

src/common/event-utils/index.js
sharma46bhawana: Should I make the changes in the other PR?

src/components/ebay-pagination/test/test.browser.js
sharma46bhawana: Okay will update these

src/components/ebay-pagination/test/test.browser.js
sharma46bhawana: Yes! I will add the test cases.

marko.json
seangates: If anything, it should be `current-text`.

src/components/ebay-pagination/examples/1-basic/template.marko
seangates: I can see it either way. But I can definitely see it being cluttered, especially as teams will be adding translated strings for all the accessibility attributes.

@yomed Do you feel using an attribute fits the style of what we're doing? More HTML-like?

src/components/ebay-pagination/index.js
seangates: Brainstorm:
`applyPreventDefault`
`doPreventDefault`
`fakeLinkPreventDefault`

seangates: `handleFakeLink`

src/components/ebay-pagination/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### ebay-pagination Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`class` | String | No | custom class
+`aria-label-prev` | String | No | aria-label for previous button
+`aria-label-next` | String | No | aria-label for previous button
+`curr-text` | String | No | Description for the current page (e.g. Results of Page 1)
+`fake-link` | Boolean | No | Use pagination links for an ajax reload
```
ianmcburnie: I like the purpose of this attribute (and breadcrumbs should be updated to match, in a separate PR), but we might need to choose another name for this. Maybe `type="spa"` or `type="hijax"` or something?

src/common/event-utils/index.js
yomed: Personally I think the event util changes should be part of this PR since the need arose from pagination, but I'm not sure if the others wanted you to keep it separate.

src/components/ebay-pagination/README.md
yomed: Ok, I think just keep consistency with whatever terminology you use. This stood out because for the attributes it has "previous button" and here it has "previous arrow."

src/components/ebay-pagination/examples/1-basic/template.marko
yomed: Ok, makes sense to keep it off the root if you anticipate more usage there. Did you also consider splitting these out into something like `<ebay-pagination-control>`? For me, it seems like the current pattern is a bit strange since the controls don't share a lot with the actual items, but I also haven't been exposed to all the use cases.

src/components/ebay-pagination/README.md
sharma46bhawana: I can have a boolean value as `hijax`. Having a `type="hijax"` isn't needed here as we don't have a 3rd value and we don't necessarily need a type for the default case.

ianmcburnie:
As mentioned in standup today, I'm going to hold this back for the 0.4.0 release, in order to lighten load on QE for 0.3.0 testing.

src/components/ebay-pagination/examples/2-fake-link/template.marko
```diff
@@ -0,0 +1,13 @@
+<ebay-pagination hijax>
```
yomed: Rename to `2-hijax`

src/components/ebay-pagination/index.js
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    // Now that we are showing all the pages that we can, lets hide remaining pages.
+    for (let i = adjustedNumPages; i < totalPages; ++i) {
+        this.pageEls[i].setAttribute('hidden', true);
+    }
+}
+
+function handleFakeLink(e) {
+    if (this.state.hasFakeLink) {
```
yomed: Rename to `handleHijax` and `state.isHijax`

src/components/ebay-pagination/index.js
```diff
@@ -0,0 +1,172 @@
...
+    if (hasFakeLink) {
+        role = 'button';
+    }
+
+    for (let i = 0; i < inputItems.length; ++i) {
+        const item = inputItems[i];
+        const href = item.href;
+        const current = item.current;
+
+        const tempObj = {
```
yomed: Rename to `tempItem` or similar

src/components/ebay-pagination/index.js
```diff
@@ -0,0 +1,172 @@
...
+        nextItem = {
+            class: 'pagination__next',
+            disabled: true
+        };
+    }
+
+    return {
+        hasFakeLink,
+        nextItem,
+        prevItem,
```
yomed: You can tighten up the above checks by doing:
```js
prevItem: prevItem || { class: 'pagination__previous', disabled: true };
nextItem: nextItem || { class: 'pagination__next', disabled: true };
```

src/components/ebay-pagination/browser.json
```diff
@@ -0,0 +1,10 @@
+{
+    "dependencies": [
+        {
+            "if-not-flag": "ebayui-no-skin",
+            "path": "@ebay/skin/pagination"
```
yomed: This would also need `"@ebay/skin/utility"` for using the `.clipped` class

src/components/ebay-pagination/template.marko
```diff
@@ -0,0 +1,43 @@
...
+        class=data.prevItem.class
+        href=data.prevItem.href
+        role=data.role
+        w-onkeydown="handlePreviousPageKeyDown"
+        w-onclick="handlePreviousPage"
+        if(data.prevItem)>
+       <span></span>
+    </a>
+    <ol class="pagination__items">
+        <for(item in data.items | status-var=loop)>
```
yomed: `status-var` is not used, and can be removed

src/components/ebay-pagination/test/test.server.js
```diff
@@ -0,0 +1,57 @@
...
+    test('renders basic version', context => {
+        const $ = testUtils.getCheerio(context.render(mock.basicLinks));
+        expect($('nav.pagination').length).to.equal(1);
+        expect($('.pagination__previous').length).to.equal(1);
+        expect($('.pagination__next').length).to.equal(1);
+        expect($('h2.clipped').length).to.equal(1);
+    });
+
+    test('renders fake version', context => {
+        const $ = testUtils.getCheerio(context.render(mock.fakeLinks));
```
yomed: `renders hijax version`
`mock.hijax`

src/components/ebay-pagination/test/test.browser.js
```diff
@@ -0,0 +1,110 @@
...
+        });
+
+        test('then it emits the marko event called pagination-select', () => {
+            expect(spy.calledOnce).to.equal(true);
+            const eventData = spy.getCall(0).args[0];
+            expect(Object.keys(spy.args[0][0])).to.deep.equal(['el', 'value']);
+            expect(eventData.value).to.be.equal('1');
+        });
+    });
+});
```
yomed: Run `yarn tr` locally to see the coverage report, and fill in the missing gaps where possible.

src/components/ebay-pagination/examples/2-fake-link/template.marko
seangates: Rename the directory from `2-fake-link` to `2-hijax`, right?

src/components/ebay-menu/index.js
```diff
@@ -75,8 +75,8 @@ function getInitialState(input) {
         isFake,
         label: input.label,
         class: input.class,
-        grow: Boolean(input.grow),
```
ianmcburnie: I'm seeing some non pagination related diffs. I think you need to rebase/merge to get latest from 0.3.0.

src/components/ebay-pagination/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The `<ebay-pagination>` is a tag used to create a pagination navigation.
+
+## ebay-pagination Tag
+
+### ebay-pagination Usage
+
+```marko
+<ebay-pagination aria-label-prev="Previous page" aria-label-next="Next page" curr-text="Results Pagination - Page 2">
+    <ebay-pagination-item href="#" previous disabled/>
```
ianmcburnie: I'm not sure if there is any benefit in letting the developer declaratively add the prev and next items? Can't we abstract away that detail for them?

marko.json
```diff
@@ -93,5 +93,26 @@
             "@*": "expression"
         }
     },
-    "<ebay-breadcrumb-item>": {}
+    "<ebay-breadcrumb-item>": {},
+    "<ebay-pagination>": {
+        "renderer": "./src/components/ebay-pagination/index.js",
+        "transformer": "./src/common/transformer/index.js",
+        "@class": "string",
+        "@accessibility-prev": "string",
```
ianmcburnie: We could abbreviate `accessibility` to `a11y` if everyone is in agreement. But I'm fine keeping as is.

src/components/ebay-pagination/README.md
yomed: @ianmcburnie FYI there was some discussion of this here: https://github.com/eBay/ebayui-core/pull/100#discussion_r182160536

marko.json
DylanPiercey: The notice and the dialog have `aria-label-x` attributes. Wouldn't it make sense to be consistent with those? IE: `aria-label-prev` `aria-label-next` and `aria-label-current`.

yomed: We might want to move towards more generic naming like `accessibility` or `a11y` so we can be consistent with cases where the strings don't directly map to `aria-label`s. https://github.com/eBay/ebayui-core/pull/81/files#diff-e56f5fdf7a730e2c503dd8217ef07299R21

sharma46bhawana: Should I send a separate PR to change the naming in notice component?

src/components/ebay-pagination/README.md
ianmcburnie: Thanks. Ok, let's go with the tag for now if we need to add attributes to it, but I'm not a fan of them being booleans, because it would never make sense to have them both set to true. Maybe `type="item(default)|prev|next"`?

ianmcburnie:
Not sure if I mentioned this already, but we should have an example of the hijax use case on the examples page. Just trigger a javascript alert or something in the handler.

ianmcburnie:
Can you clarify this comment in the description please:

`When the previous or next is not provided in the input for <ebay-pagination-item> the navigation is taken as disabled`.

Does this only apply to the first and last items?

src/components/ebay-pagination/README.md
sharma46bhawana: Okay I will change it to type instead of a boolean.

marko.json
sharma46bhawana: Okay I will keep the naming here as is and send a separate PR for notice

sharma46bhawana:
@ianmcburnie I was thinking of a scenario when service doesn't return any information for `previous/next` buttons. The developer shouldn't need to write an extra `<ebay-pagination-item type="previous/next"/>` just because the markup needs it. So in case service doesn't send any extra information, if the developer doesn't write the markup the pagination arrows would be disabled.

yomed:
We could also ask the Design team, "If there is a small fixed number of pages, and next/prev functionality is not desired, should the controls be disabled or omitted?"

To me it seems that it's better to omit the controls rather than showing both as disabled. In the carousel, I hide them via CSS if they are both disabled (but in that case, being enabled/disabled depends on screen size which can change).

ianmcburnie:
Issue created for enhancements to root node API: https://github.com/eBay/ebayui-core/issues/122

PR #102, opened 2018-04-18
Allow style to pass through the button

CestDiego:
# Description
Add style as a pass through to the 

## Context

Given that marko treats `class` and `style` differently. We should allow the style to also be explicitly defined in the template, this way we don't run into trouble when using the ebay-button inside a nested component.

```
// this is button wrapper file
<ebay-button ...input />
```

```
// this is the user of the  button-wrapper
$ const style = { color: red }
<button-wrapper fluid style=style > some text <button-wrapper />
```

This doesn't add `style` to the `input['*']` object. And if we otherwise did:
```
// this is button wrapper file
<ebay-button style=input.style />
```

What we get is:

image

### DIFF
--- a/src/components/ebay-button/index.js
+++ b/src/components/ebay-button/index.js
@@ -46,6 +46,7 @@ function getTemplateData(state, input) {
 
     model.tag = tag;
     model.classes = classes;
+    model.style = input.style;
     model.disabled = state.disabled;
     model.partiallyDisabled = input.partiallyDisabled ? 'true' : null; // for aria-disabled
     model.htmlAttributes = processHtmlAttributes(input);
--- a/src/components/ebay-button/template.marko
+++ b/src/components/ebay-button/template.marko
@@ -5,6 +5,7 @@
     w-onkeydown="handleKeydown"
     type="button"
     class=data.classes
+    style=data.style
     href=data.href
     disabled=data.disabled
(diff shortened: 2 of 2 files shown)


yomed:
If we did this, we would have to do it to all components. So it would be ideal if we could avoid this. It also seems like there's an easy workaround here by using `const style = 'color:red'` and `style=input.style`.

Aside from that, it seems that the larger fix is just to support cta. Either as its own `<ebay-cta>` component, or through `<ebay-button>`.

ianmcburnie:
If this is risky or still needs more investigation, let's leave this till next release (0.4.0) . We already have QA trying to rush through 0.3.0 before end of week, so I'd rather not add to their burden.

yomed:
@ianmcburnie Do you think we should work on `<ebay-cta>` as a separate component from button, given the functionality differences? Would we need to add custom color support as part of Skin?

ianmcburnie:
Let's sync up on this next week if we can. There is a side discussion on whether CTA is always a hyperlink or not, i.e. the validity of this statement from DSL:

`“The behavior of an action button (CTA) is to always take users to another destination.”`

ianmcburnie:
Sounds like passing of style is an issue for all components, not just button, so I'm going to close this PR and create an issue for us so that we can spend more time thinking about it.

Issue #132

PR #104, opened 2018-04-19
Updating styles for the skin change to combobox;

seangates:
## Description
- updated class names to use the Skin's new convention for `combobox`
- fix for a whitespace lint issue in one of the util files 🤷‍♂️

### DIFF
--- a/src/components/ebay-select/index.js
+++ b/src/components/ebay-select/index.js
@@ -7,11 +7,11 @@ const processHtmlAttributes = require('../../common/html-attributes');
 const observer = require('../../common/property-observer');
 const template = require('./template.marko');
 
-const listboxOptionsClass = 'listbox__options';
-const listboxExpanderClass = 'listbox__control';
+const listboxOptionsClass = 'combobox__options';
+const listboxExpanderClass = 'combobox__control';
 const listboxHostSelector = `.${listboxExpanderClass} > input`;
-const listboxBtnClass = 'listbox__control';
-const listboxOptionSelector = '.listbox__option[role=option]';
+const listboxBtnClass = 'combobox__control';
+const listboxOptionSelector = '.combobox__option[role=option]';
 
 function getInitialState(input) {
     const options = (input.options || []).map(option => {
@@ -45,7 +45,7 @@ function getInitialState(input) {
 }
 
 function getTemplateData(state) {
-    const listboxClass = ['listbox', state.class];
+    const listboxClass = ['combobox', state.class];
     const btnClass = [listboxBtnClass];
     const optionsClass = [listboxOptionsClass];
 
--- a/src/components/ebay-select/index.js
+++ b/src/components/ebay-select/index.js
@@ -92,11 +92,11 @@ function init() {
 }
 
 function handleExpand() {
-    emitAndFire(this, 'combobox-expand');
+    emitAndFire(this, 'listbox-expand');
 }
 
 function handleCollapse() {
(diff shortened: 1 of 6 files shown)


src/components/ebay-select/index.js
```diff
@@ -7,11 +7,11 @@ const processHtmlAttributes = require('../../common/html-attributes');
 const observer = require('../../common/property-observer');
 const template = require('./template.marko');
 
-const listboxOptionsClass = 'listbox__options';
-const listboxExpanderClass = 'listbox__control';
+const listboxOptionsClass = 'combobox__options';
```
yomed: Do you want to remove all the old references to `listbox`? These could easily be changed like `listboxOptionsClass` to `optionsClass`. I also see there are some other leftover ones in the event names and elsewhere, which could be updated.

seangates: This will all be done when we split this into `select` and `combobox`. It's all leftover from when it started out as `listbox`.

So, I feel we can leave it for now, as we will be working to get the native select version out later. I think that will be a v1 type of breaking change, too.

yomed: IMO this is a good opportunity to clean it up, given that this PR removes the last practical usage of `listbox`. Doesn't make much sense to have the remaining references, even pending later cleanup. We're still in major version 0, so breaking changes shouldn't be an issue.

ianmcburnie: Hmm, I feel we should rename as many of these internal variables now as we can to avoid confusion for others who might be working on this later.

seangates: Okay. Let me get on that today.

seangates:
@ianmcburnie @yomed Updated. Please review again. 👍

seangates:
@ianmcburnie @yomed Also ... I'm thinking we might need to rename the component altogether now, because the `combobox-change` event doesn't makes sense when the tag is `<ebay-select>`. We'd be really confusing our developers.

seangates:
@ianmcburnie So, the event names?

ianmcburnie:
Ohhhhh, my apologies. I was a little hasty with my response. They can stay the same. It is still a listbox that is being expanded and collapsed. The combobox is just the input. Sorry for the confusion!!!

We *could* maybe change them later to `select-*`, but that is a breaking change, and possibly what you were referring to earlier. Sorry!

seangates:
@ianmcburnie Cool. Let me revert those.

ianmcburnie:
If we *really* wanted to change them now we could continue firing the old events in addition to the new events. Then just remove them in 1.0.0.

seangates:
Let's do it later.

seangates:
@ianmcburnie @yomed Okay, we should be good to go here. Let me know.

ianmcburnie:
I think it's pretty close. To quickly try & recap:

Mobile will render a select element only. No ARIA readonly combobox or ARIA listbox. This is slated for 0.4.0 release.

Web will render an ARIA readonly combobox that opens an ARIA listbox. Under the hood a select element is also rendered that stores the form data.

For <ebay-select> there will never be an *editable* combobox version (because we are trying to stay as close to select tag functionality as possible, which is always readonly). So we are done with <ebay-select> after 0.4.0.

For an editable combobox (i.e, an input tag with a datalist), we'll create a new <ebay-combobox> component that will NOT be backed by a select element under the hood. This is because the input element will store the form data value.

Issue #106, opened 2018-04-19
Dialog: single line dialog looks bad (edge case)


hramakumar:
When the dialog content comes in single line without header. The line width makes it look smaller than the close button.
![image](https://user-images.githubusercontent.com/13157115/39020709-ac2d23c0-43e2-11e8-9327-2b31ed5dfd52.png)

hramakumar:
expected: We need to have a default one line spacing at the top to make it look neat.

Issue #105, opened 2018-04-19
Dialog: clicking or moving the browser scroll bar does not work


hramakumar:
Repo:
  create a dialog with long text content so that the dialog becomes scrollable.
when we scroll over the active region it works fine but when we scroll on the window scroll bar(on the right of the page)  it does not work.

The scroll bar should be either inside the dialog  or if its going to scroll along with the browser scroll bar then we need to make the browser scroll bar clickable. 

There is no quick way to scroll up and down

hramakumar:
![image](https://user-images.githubusercontent.com/13157115/39020012-352859e0-43e0-11e8-8715-154b49b410ea.png)

hramakumar:
in the above screenshot if look at the scrollbar on the side of the browser. I am not able to move that scroll bar when the mouse places at the edge of the browser.

ianmcburnie:
Fixed in 0.3.0.

PR #111, opened 2018-04-24
ebay-dialog: Update to match latest version of skin.

DylanPiercey:
## Description
This PR updates the `ebay-dialog` to no longer rely on a `dialog__mask` element (instead the mask is the background of the dialog element).

## References
#105 and #106


seangates:
@ianmcburnie I don't see any bad merge stuff. Only 6 files.

PR #117, opened 2018-04-25
Fix body scrolling on open dialogs for mweb.

### LINKED ISSUE #107: Dialog: main page content is still scrollable in MWEB devices
in devices the areas outside the dialog border moves when I try to scroll outside the dialog section.
This gives a bad user experience.
Repro:
IOS devices
open a scrollable dialog
try touch, hold and scroll in the area outside the dialog borders.
The parent page at the back moves.

DylanPiercey:
## Description
This changes the way scrolling is prevented on the `ebay-dialog` is open.
Previously `overflow:hidden` was set on the body which was not respected in mweb.
Now `position:fixed` is used and the scroll position is accounted for by using negative margins.

I extracted the `body scrolling` functionality into the common js just because it was getting slightly unruly in the component itself.

## References
Closes #107

### DIFF
--- a/src/common/body-scroll/index.js
+++ b/src/common/body-scroll/index.js
@@ -4,6 +4,15 @@ let previousStyles;
 module.exports = {
     /**
      * Prevents the `<body>` element from scrolling.
+     *
+     * This is done by forcing the body to be `fixed` and setting it's height/width inline.
+     * The body is then translated the reverse of the scroll position using negative margins.
+     *
+     * This approach was chosen over the more common just `overflow:hidden` to address some
+     * issues with iOS and Android browsers.
+     *
+     * Finally the scroll position is also stored so that it can be reapplied after restoring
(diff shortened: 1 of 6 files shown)


src/common/body-scroll/index.js
```diff
@@ -0,0 +1,40 @@
+let previousPosition;
+let previousStyles;
+
+module.exports = {
+    /**
+     * Prevents the `<body>` element from scrolling.
```
ianmcburnie: I think maybe just a little extra commenting, one or two lines, to explain how this is achieved.

DylanPiercey:
@ianmcburnie updated the docs. Let me know if you think anything else is missing 😄.

ianmcburnie:
Oops, I prematurely merged this before @seangates or @yomed also approved. Feel free to make comments before we pre-release later today.

Issue #114, opened 2018-04-25
Select: options should support custom atrributes


bhoomisatasiya:
```
<ebay-select-option 
        value=actionParams
        label=label
        selected=selected
        data-action=action
        data-action-params=actionParams
 />

PR #118, opened 2018-04-26
Select: fix pass-through on nested options

yomed:
## Description

For `ebay-select-option`:
- fix custom class 
- pass through `htmlAttributes`
- add tests (diff is only big because I added new `describe()`s)

## Context

Looks like these were basically typos from earlier. The tests should help keep this working.

## References

Fixes https://github.com/eBay/ebayui-core/issues/114


ianmcburnie:
@seangates You good with this? (and I hope you feel better soon!)

Issue #112, opened 2018-04-24
Breadcrumbs: add "hijax" boolean attribute


ianmcburnie:
Let's bring this component in line with the incoming pagination component, and add a `hijax` boolean attribute that operates in the same way.

PR #124, opened 2018-05-01
Breadcrumb: add hijax

yomed:
## Description

- add `hijax` functionality to breadcrumb component
- create `preventDefaultIfHijax` util
- standardize usage across breadcrumb and pagination
- improve hijax examples
- minor related cleanup
- update tests and docs

## Context

Standardizing `hijax` usage for similar cases in the two components.

Note: `breadcrumb-select` currently fires even on `span` elements, which is unchanged in the `hijax` case. This may need fixing/refactoring, but I avoided changing that behavior as part of this PR.

## References

#112

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -90,8 +90,10 @@
     "<ebay-breadcrumb>": {
         "renderer": "./src/components/ebay-breadcrumb/index.js",
         "transformer": "./src/common/transformer/index.js",
+        "@class": "string",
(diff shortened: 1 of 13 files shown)


ianmcburnie:
After we've reviewed, I think we can put this out in 0.4.0-1 later this week, after QA is done with 0.4.0-0.

Issue #127, opened 2018-05-01
Dialog: Transition utility not working when children have transitions.


DylanPiercey:
The transition utility currently works by trying to calculate how many transitions the dialog has and then waiting for them to complete before marking the entire transition as complete. This is currently causing issues since the children of the dialog can contain transitions that are unrelated to the dialog transitions and this trips up the utility.

PR #128, opened 2018-05-01
Transition utility fixes.

DylanPiercey:
## Description
Currently the transition utility tries to figure out how many transitions it should wait for before marking the animation as completed. In this PR the options for the transition utility are changed and it now requires passing in a list of elements to wait for their transitions to end.

## References
#127

### DIFF
--- a/src/components/ebay-dialog/index.js
+++ b/src/components/ebay-dialog/index.js
@@ -10,8 +10,10 @@ const template = require('./template.marko');
 
 function init() {
     this.dialogEl = this.getEl('dialog');
+    this.windowEl = this.getEl('window');
     this.closeEl = this.getEl('close');
     this.bodyEl = this.getEl('body');
+    this.transitionEls = [this.windowEl, this.dialogEl];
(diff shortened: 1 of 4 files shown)


src/components/ebay-dialog/index.js
```diff
@@ -10,8 +10,10 @@ const template = require('./template.marko');
 
 function init() {
     this.dialogEl = this.getEl('dialog');
+    this.windowEl = this.getEl('window');
```
senthilp: Does listening on `window`'s `transitionend` event wait for all the dialog's children transition to complete? Curios on why you are adding `window` to the wait list.

DylanPiercey: It’s the dialog window, not the actual window. The two elements that have transitions are the dialog itself and the `dialog__window`.

DylanPiercey:
This is a fully backwards compatible change.

ianmcburnie:
Squash and merge away! (I assume you have merge permissions)

DylanPiercey:
👍

PR #129, opened 2018-05-02
Upgraded to makeup-expander v0.4.0 (fixes #92)

### LINKED ISSUE #92: Menu: keep menu open on mouseout
Consensus seems to be that menu (and select) should not auto-collapse on mouseout, but only when clicking elsewhere on page.

It would appear that the current makeup-expander autoCollapse option would need slight modification to support this. Here it is currently:

`options.autoCollapse`: whether click-activated expander should collapse when keyboard or mouse leaves the content or widget (default: false)

We probably want to continue to with the auto collapse when keyboard leaves the widget, but not when mouse leaves the widget.

ianmcburnie:
## Description

Upgraded to makeup-expander v0.4.0.

The latest version of makeup-expander has some refinements based around autoCollapse. These refinements separate the keyboard and mouse collapse behaviour.

In short:

* menu will no longer collapse when moving mouse out of the menu overlay
* menu will now collapse when clicking outside of menu overlay
* menu will continue to collapse when keyboard focus leaves widget
* menu will continue to collapse when ESC key is pressed on a menu item

Fixes #92.

## References

https://github.com/makeup-js/makeup-expander/commit/3be210249f23bab7f6103993f127b6c81b2557d1

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
   "dependencies": {
     "lodash.get": "^4.4.2",
     "lodash.set": "^4.3.2",
-    "makeup-expander": "^0.3.2",
+    "makeup-expander": "~0.4.0",
     "makeup-focusables": "~0.0.3",
     "makeup-key-emitter": "~0.0.3",
     "makeup-keyboard-trap": "^0.0.9",
--- a/src/components/ebay-menu/index.js
+++ b/src/components/ebay-menu/index.js
@@ -156,7 +156,7 @@ function init() {
     const expander = new Expander(this.el, { // eslint-disable-line no-unused-vars
         hostSelector: buttonSelector,
         focusManagement: 'focusable',
-        click: true,
+        expandOnClick: true,
         autoCollapse: true
     });
 }
@@ -257,7 +257,8 @@ function handleItemKeydown(e) {
     });
 
     eventUtils.handleEscapeKeydown(e, () => {
-        this.buttonEl.focus(); // triggers collapse through makeup
+        this.buttonEl.focus();
+        this.setState('expanded', false);
     });
 }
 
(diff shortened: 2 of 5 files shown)


ianmcburnie:
For some reason, the following two tests are failing in Travis, but working in local env:

```
describe('when the escape key is pressed', () => {
        let spy;
        beforeEach(() => {
            spy = sinon.spy();
            widget.on('menu-collapse', spy);
            testUtils.triggerEvent(firstItem, 'keydown', 27);
        });

        test('then it collapses', () => {
            expect(button.getAttribute('aria-expanded')).to.equal('false');
        });

        test('then it emits the marko collapse event', () => {
            expect(spy.calledOnce).to.equal(true);
        });
});
```

yomed:
@ianmcburnie It looks like it's only failing under v4. You can debug locally by running `yarn installMarkoV4` and then running tests as usual. I can take a look too.

DylanPiercey:
My guess is that this is because the update is async in Marko 4. If you updated the `beforeEach` call to wait for Marko to finish updating then I think this should work.

```marko
        beforeEach((done) => {
            spy = sinon.spy();
            widget.on('menu-collapse', spy);
            widget.once('update', done);
            testUtils.triggerEvent(firstItem, 'keydown', 27);
        });
```

yomed:
@DylanPiercey I think that's part of it, but we shouldn't use `once`, because that will let the case of calling the event twice go through (which is a bug).

ianmcburnie:
@yomed Thanks. Yes, you are right, it's also failing under MarkoV4 locally. Which makes this easier to debug at least.

seangates:
Pulled and ran it. LGTM.

Issue #120, opened 2018-04-30
New Component: Radio


PrashantAshok:
Create a component that renders a radio button that matches the spec here. http://ebay.github.io/skin/ds6/#radio

PR #134, opened 2018-05-02
Radio Component

PrashantAshok:
## Description
ebay-radio component which renders a radio

## References
Issue - #120 
https://dsl.ebay.com/ds6/forms

### DIFF
--- a/src/components/ebay-radio/examples/3-grouped-radio/template.marko
+++ b/src/components/ebay-radio/examples/3-grouped-radio/template.marko
@@ -30,4 +30,3 @@
     var fieldSet = document.querySelector("fieldset");
     fieldSet.addEventListener("radio-change", e => console.log(e.detail), true);
 })()</script>
-
--- a/src/components/ebay-radio/examples/3-grouped-radio/template.marko
+++ b/src/components/ebay-radio/examples/3-grouped-radio/template.marko
@@ -28,5 +28,5 @@
 
 <script>(function() {
     var fieldSet = document.querySelector("fieldset");
-    fieldSet.addEventListener("radio-click", e => console.log(e.detail), true);
+    fieldSet.addEventListener("radio-select", e => console.log(e.detail), true);
 })()</script>
--- a/src/components/ebay-radio/examples/3-grouped-radio/template.marko
+++ b/src/components/ebay-radio/examples/3-grouped-radio/template.marko
@@ -6,7 +6,7 @@
             aria-label="Grouped radio example"
             id="group-radio-1"
             value="1"
-            name="radio-group"></ebay-radio>
+            name="radio-group"/>
         <label class="field__label field__label--end" for="group-radio-1">Option 1</label>
     </span>
     <span class="field">
@@ -15,7 +15,7 @@
                 aria-label="Grouped radio example"
                 id="group-radio-2"
                 value="2"
-                name="radio-group"></ebay-radio>
+                name="radio-group"/>
         <label class="field__label field__label--end" for="group-radio-2">Option 2</label>
     </span>
     <span class="field">
@@ -24,7 +24,7 @@
             aria-label="Grouped radio example"
             id="group-radio-3"
             value="3"
-            name="radio-group"></ebay-radio>
+            name="radio-group"/>
         <label class="field__label field__label--end" for="group-radio-3">Option 3</label>
     </span>
 </fieldset>
--- a/src/components/ebay-radio/examples/3-grouped-radio/template.marko
+++ b/src/components/ebay-radio/examples/3-grouped-radio/template.marko
@@ -1,0 +1,30 @@
+<fieldset>
+    <legend>Choose an Option</legend>
+    <span class="field">
+        <ebay-radio
+            class="field__control"
+            aria-label="Grouped radio example"
+            id="group-radio-1"
+            value="1"
+            name="radio-group"></ebay-radio>
+        <label class="field__label field__label--end" for="group-radio-1">Option 1</label>
+    </span>
+    <span class="field">
+            <ebay-radio
+                class="field__control"
+                aria-label="Grouped radio example"
+                id="group-radio-2"
+                value="2"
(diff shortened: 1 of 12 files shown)


src/components/ebay-radio/examples/3-grouped-radio/template.marko
```diff
@@ -0,0 +1,30 @@
+<fieldset>
+    <legend>Choose an Option</legend>
+    <span class="field">
+        <ebay-radio
+            class="field__control"
+            aria-label="Grouped radio example"
```
ianmcburnie: If we have a label tag, then we can drop the `aria-label`.

src/components/ebay-radio/examples/3-grouped-radio/template.marko
```diff
@@ -0,0 +1,30 @@
...
+            id="group-radio-3"
+            value="3"
+            name="radio-group"/>
+        <label class="field__label field__label--end" for="group-radio-3">Option 3</label>
+    </span>
+</fieldset>
```
ianmcburnie: I think it would be nice if we could add a simple bit of javascript to log the radio-change event to console. Similar I think to how @sharma46bhawana added some script for the hijax pagination example.

src/components/ebay-radio/index.js
```diff
@@ -0,0 +1,29 @@
...
+    };
+}
+
+function getTemplateData(state) {
+    return state;
+}
+
+function handleClick(originalEvent) {
+    if (!this.state.disabled) {
+        emitAndFire(this, 'radio-change', { originalEvent, selected: originalEvent.target.value });
```
ianmcburnie: Will event fire even if radio is currently checked? I don't think it should, right? If so we'll need a test case for this.

PrashantAshok: yep there is a [testcase](https://github.com/eBay/ebayui-core/pull/134/files#diff-d52de5f96198a93fd0388bde6d4395beR39) for it.

src/components/ebay-radio/examples/3-grouped-radio/template.marko
PrashantAshok: added for checkbox as well.

src/components/ebay-radio/index.js
ianmcburnie: That test says "given radio button is disabled", but I'm talking about *checked* state, not disabled state.

src/components/ebay-radio/examples/3-grouped-radio/template.marko
ianmcburnie: Changes look good. But in the event detail noticed that `selected` is an integer and not zero index based. This got me thinking that we should make this event consistent with menu & select which returns `index` and `checked`.

src/components/ebay-radio/index.js
PrashantAshok: ohk. Yes, the event will fire even if the radio is checked. It follows the same behavior as native [demo](http://jsbin.com/jaxifuvige/edit?html,js,console,output)

ianmcburnie: That native demo is firing a click event though. Not a change event.

src/components/ebay-radio/examples/3-grouped-radio/template.marko
PrashantAshok: discussed offline, since checkbox/radio is just single here, don't have access to `array` of inputs, hence can only emit the `value`

ianmcburnie: `selected` is a little ambiguous as a name to end user (i.e. is it selectedIndex or selectedValue or something else?).  And as this thing is the only thing that can be selected (i.e. it's not in a group) I say we rename it to `value`.

We could also just potentially drop it altogether, as it's available in the originalEvent. But let's maybe just rename it and leave it there for now.

ianmcburnie: This got me to thinking that we could, if we wanted, do some clever stuff and sniff out other radio components with the same name attribute. Then we could return the index.

PrashantAshok: updated the property to `value`.

PrashantAshok: reg your second point about group radio - that would be trying to access elements outside the current component - doesn't seem right. Maybe next iteration we could come up `ebay-radio-group`?

ianmcburnie:
@PrashantAshok Sorry to do this to you, I know first we had the event as `radio-change`, but that overlapped with the DOM change event, then we moved to `radio-click` as that was more explicit this was a pass through of the click event. But then I remember we have `breadcrumb-select` and maybe a few others using "select" that are pass throughs for the click event. So in the interest of consistency maybe we go with `radio-select` for now, and then review all these names later.

@yomed @seangates thoughts?

ianmcburnie:
Or maybe `select` makes sense in the context of a group of items (like breadcrumbs and menu items). Whereas `click` makes sense in the context of a single item (like radio)...?

ianmcburnie:
And for checkbox I guess we have an `unselect`?

PrashantAshok:
hmm, `radio-click` is better, since that is the event being handled & emitted? `w-onclick`

PrashantAshok:
for checkbox - we just have the `checkbox-click`. Did you meant sending `checked: true/false` as part of the event data?

ianmcburnie:
Breadcrumbs and pagination are also handling the click event, but using `breadcrumb-select` and `pagination-select` as the name. I say we keep it consistent for now and review again later.

I guess the `checkbox-select` event could just contain `{ originalEvent, value, checked }` like you say. Maybe no need for a separate event when unchecked.

ianmcburnie:
Okay, we are going ahead with `radio-select` and `checkbox-select` in order to move things forward. Checkbox event should contain the checked state too.

Issue #42, opened 2018-03-10
New Component: Icon


ianmcburnie:
Placeholder issue created. More details to follow....

(basically a carbon copy of ebayui-image I think)

RajaRamu:
it depends on skin-icon - https://github.com/eBay/skin/pull/124. please assign this to me

PR #136, opened 2018-05-03
Icon Component

yomed:
## Description
- adds new `<ebay-icon>` component
- edit demo slightly to avoid dynamically loaded icons
- minor refactoring to testUtils

## Context

General technical details/limitations are in the README.md.

I'm not currently warning about dynamic icon usage, because it could be legitimate. I clarified in the README that dynamic icons only break the stamper, so as long as you include the icon on the page once statically, you can do any dynamic inclusion.

## References

https://github.com/eBay/ebayui-core/issues/42

## Screenshots

(More in the `examples/`, but this a good representation of most of the work)
screen shot 2018-05-03 at 12 39 53 pm


demo/template.marko
```diff
@@ -17,7 +17,18 @@
             <div class="example">
                 <h4>${example.name}</h4>
                 <div class="example-code">$!{example.code}</div>
-                <div><dynamic-tag options={path: example.path}/></div>
+                <div>
+                    <!-- icons need to be statically imported -->
+                    <if(data.params.component === 'ebay-icon')>
```
DylanPiercey: Is this still needed?

src/components/ebay-icon/stamper/index.js
```diff
@@ -0,0 +1,87 @@
...
+            const foundIcons = currentNode.exports && currentNode.exports._ebay_icons;
+
+            if (foundIcons) {
+                icons.push(...foundIcons);
+            }
+        }
+    });
+
+    // remove duplicates
+    pageCache[pageTemplateId] = icons.filter((value, index) => icons.indexOf(value) === index);
```
DylanPiercey: Micro optimization would be to use a `Set` here for the `icons` and then we don't need to remove duplicates ourselves.

src/components/ebay-icon/transformer.js
```diff
@@ -0,0 +1,24 @@
+// Transforms inline icons to artificially require specific icon files
```
DylanPiercey: This comment is now off. Should be something like:

```js
// Exposes all icons used in a component as a list of icon names which we can discover statically through the require cache.
```

demo/template.marko
yomed: It looks like this is still needed, because the icons are required too late to be used by the stamper. Theoretically it would be ok after the first request I think, but those don't work either because we're caching the page's icons on the first request.

I'll have to do this static import for our internal components that use inline `<ebay-icon>` too.

src/components/ebay-icon/stamper/index.js
yomed: Maybe we should defer this until we set up the babel runtime transform https://github.com/eBay/ebayui-core/issues/156

yomed: @DylanPiercey I looked at using `Set`, but it seems like it might not be worth it. We're doing other array stuff like `icons.map()`, and using `Set` would mean conversions like `Array.from(icons.values()).map()`. So probably simplest to use array across the board.

DylanPiercey: Alrighty, lets hold off on this then.

src/components/ebay-icon/index.js
```diff
@@ -0,0 +1,39 @@
...
+    const isInline = type === 'inline';
+    const accessibilityText = input.accessibilityText;
+    const name = input.name;
+    let titleId;
+    let accessibilityAttributes;
+
+    if (isInline) {
+        if (accessibilityText) {
+            titleId = `icon-title-${Math.random().toString(36).substr(2, 9)}`;
+            accessibilityAttributes = { 'aria-labelled-by': titleId, focusable: 'true', role: 'img' };
```
ianmcburnie: I think we should leave focusable as false. If it's an interactive image it should be nested inside a button or link (which will take care of focus) by app team.

yomed:
@ianmcburnie Thanks, should be all set now.

PR #145, opened 2018-05-05
Carousel: fix initial client render in v4

### LINKED ISSUE #142: ebay-carousel: with no element hidden- arrows are not disabled
Repro:
go to https://go/ebayui/test2
search for 'All expanded - both arrows disabled' in the Carousel section
no items are hidden- still I see the arrows
expected behaviour:
The arrows should not show up if now of the cards are hidden

Resizing the browser fixes the problem

### LINKED ISSUE #143: ebay-carousel: Starting index specified- yet starts from beginning
Repro:
Go to https://go/ebayui/test2
look at "Starting index -start with card 3" in the Carousel section
It start from card 1
expected:
as the starting index is specified as 2 the carousal visible area is expected to start from card 3

(also in the page html, it shows hidden = false for all cards)

resizing the browser fixes the problem

yomed:
## Description

- add extra `this.refresh()` in `init()`

## Context

The carousel does an initial call to `this.refresh()` to do sizing calculations and set up initial state based on client rendering. Everything works fine in v3, but for some reason, the top `this.refresh()` call does not work correctly in v4. I started out fixing this by doing `setTimeout(refresh.bind(this))`, but that causes lots of problems in the tests, which I didn't make much progress on. I noticed that a second call magically fixes everything, and all the tests are still fine.

@DylanPiercey I'm guessing this is related to Marko internals, especially with it working fine in v3.

## References

Fixes #142 
Fixes #143

### DIFF
--- a/src/components/ebay-carousel/index.js
+++ b/src/components/ebay-carousel/index.js
@@ -85,6 +85,7 @@ function init() {
 
     this.subscribeTo(resizeUtil).on('resize', refresh.bind(this));
     this.refresh();
+    this.refresh(); // FIXME: currently needs a second call in v4
 }
 
 function refresh() {


senthilp:
I hope this doesn't impact Marko V3 performance. We should probably open another issue to follow-up and close once it is fixed on the Marko side.

yomed:
@senthilp There's no real impact to v3 performance. It's actually a good sign that all the tests pass without any other change, which shows that this is roughly idempotent (e.g. events don't double fire).

yomed:
@hramakumar This should unblock carousel testing (once we publish).

PR #147, opened 2018-05-06
Added a check to return the correct target

### LINKED ISSUE #146: Pagination: clicking on pagination arrows gives incorrect el
Clicking on the pagination arrow in the center gives the el as `span` instead of `a` tag.

This issue was initially found by @bsatasiya

sharma46bhawana:
## Description
The changes here fix the issue where if a mouse click is registered at the center of the pagination arrow the returned target is an empty `span` as the content of the pagination arrows is in a child `span` inside the actual `a` that has all the `data-*`

## Context
For all the pagination arrow clicks added a check for the `span` tag, if it was indeed a `span` it returns the `parentNode` else it returns the current target. 

## References
Fixes #146

### DIFF
--- a/src/components/ebay-pagination/index.js
+++ b/src/components/ebay-pagination/index.js
@@ -119,7 +119,7 @@ function handlePreviousPage(event) {
  * gets the correct target for pagination arrow button
  */
 function getCorrectElement(event) {
-    if (event.target.nodeName === 'SPAN') {
+    if (event.target.nodeName.toUpperCase() === 'SPAN') {
         return event.target.parentNode;
     }
     return event.target;
--- a/src/components/ebay-pagination/index.js
+++ b/src/components/ebay-pagination/index.js
@@ -107,12 +107,22 @@ function handlePageClick(event) {
 
 function handleNextPage(event) {
     eventUtils.preventDefaultIfHijax(event, this.state.hijax);
-    emitAndFire(this, 'pagination-next', { el: event.target });
+    emitAndFire(this, 'pagination-next', { el: getCorrectElement(event) });
 }
 
 function handlePreviousPage(event) {
     eventUtils.preventDefaultIfHijax(event, this.state.hijax);
-    emitAndFire(this, 'pagination-previous', { el: event.target });
+    emitAndFire(this, 'pagination-previous', { el: getCorrectElement(event) });
+}
+
+/**
+ * gets the correct target for pagination arrow button
+ */
+function getCorrectElement(event) {
+    if (event.target.nodeName === 'SPAN') {
+        return event.target.parentNode;
+    }
+    return event.target;
 }
 
 /**
@@ -144,6 +154,7 @@ module.exports = require('marko-widgets').defineComponent({
     handlePageClick,
     handleNextPage,
     handlePreviousPage,
+    getCorrectElement,
     handlePageKeyDown,
     handleNextPageKeyDown,
     handlePreviousPageKeyDown,
--- a/src/components/ebay-pagination/index.js
+++ b/src/components/ebay-pagination/index.js
@@ -71,6 +71,8 @@ function init() {
     this.pageContainerEl = this.el.querySelector('.pagination__items');
     this.pageEls = this.pageContainerEl.children;
     this.containerEl = this.el;
+    this.previousPageEl = this.el.querySelector('.pagination__previous');
+    this.nextPageEl = this.el.querySelector('.pagination__next');
     this.subscribeTo(eventUtils.resizeUtil).on('resize', refresh.bind(this));
     this.refresh();
 }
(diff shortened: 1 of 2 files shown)


src/components/ebay-pagination/index.js
```diff
@@ -107,12 +107,22 @@ function handlePageClick(event) {
...
     eventUtils.preventDefaultIfHijax(event, this.state.hijax);
-    emitAndFire(this, 'pagination-previous', { el: event.target });
+    emitAndFire(this, 'pagination-previous', { el: getCorrectElement(event) });
+}
+
+/**
+ * gets the correct target for pagination arrow button
+ */
+function getCorrectElement(event) {
+    if (event.target.nodeName === 'SPAN') {
```
senthilp: @sharma46bhawana It is better to explicitly upper case it `event.target.toUpperCase() === 'SPAN'`, to avoid weird [browser issues](https://johnresig.com/blog/nodename-case-sensitivity/).

src/components/ebay-pagination/index.js
```diff
@@ -107,12 +107,22 @@ function handlePageClick(event) {
...
 function handlePreviousPage(event) {
     eventUtils.preventDefaultIfHijax(event, this.state.hijax);
-    emitAndFire(this, 'pagination-previous', { el: event.target });
+    emitAndFire(this, 'pagination-previous', { el: getCorrectElement(event) });
+}
+
+/**
+ * gets the correct target for pagination arrow button
+ */
+function getCorrectElement(event) {
```
jamesrweb:
```javascript
/**
 * @function getPreferredElement
 * @param {Event} event
 * @param {String} undesiredNode
 * @param {String} desiredNode
 */
function getPreferredElement(event, undesiredNode, desiredNode) {
  if(!(event instanceof Event)) {
    throw new Error(`Function ${arguments.callee.name}: Parameter 1 must be of type Event`);
  }

  if(typeof undesiredNode != 'string' || typeof desiredNode != 'string') {
    throw new Error(`Function ${arguments.callee.name}: Parameter 2 and 3 must be of type String`);
  }

  if (event.target.nodeName.toLowerCase() === desiredNode) {
    return event.target;
  }

  if (event.target.nodeName.toLowerCase() === undesiredNode) {
    const desiredElement = event.target.closest(desiredNode);

    if (desiredElement) {
      return desiredElement;
    }

    throw new Error(`Function ${arguments.callee.name}: Undesired element found but the desired element was not found`);
  }

  return event.target;
}

//example usage
const paginationLink = document.getElementById('#next-arrow');

paginationLink.addEventListener('submit', event => {
  const link = getPreferredElement(event, 'span', 'a');
  // ...
});
```

jamesrweb: What about something like this, so it is a bit more re-useable? Let me know what you think.

```javascript
/**
 * @function getPreferredElement
 * @param {Event} event
 * @param {String} undesiredNode
 * @param {String} desiredNode
 */
function getPreferredElement(event, undesiredNode, desiredNode) {
  if(!(event instanceof Event)) {
    throw new Error(`Function ${arguments.callee.name}: Parameter 1 must be of type Event`);
  }

  if(typeof undesiredNode != 'string' || typeof desiredNode != 'string') {
    throw new Error(`Function ${arguments.callee.name}: Parameter 2 and 3 must be of type String`);
  }

  if (event.target.nodeName.toLowerCase() === desiredNode) {
    return event.target;
  }

  if (event.target.nodeName.toLowerCase() === undesiredNode) {
    const desiredElement = event.target.closest(desiredNode);

    if (desiredElement) {
      return desiredElement;
    }

    throw new Error(`Function ${arguments.callee.name}: Undesired element found but the desired element was not found`);
  }

  return event.target;
}

//example usage
const paginationLink = document.getElementById('#next-arrow');

paginationLink.addEventListener('submit', event => {
  const link = getPreferredElement(event, 'span', 'a');
  // ...
});
```

senthilp: Good one. Maybe we should add this to common utilities.

yomed: A generic utility like this would be nice, but IMO this change is ok for a quick fix. Though, you would need to test it either way. We have similar logic in the menu:
https://github.com/eBay/ebayui-core/blob/5b5b22d37f89620b1280299a9834819f07205361/src/components/ebay-menu/index.js#L214-L217

https://github.com/eBay/ebayui-core/blob/5b5b22d37f89620b1280299a9834819f07205361/src/components/ebay-menu/test/test.browser.js#L270-L289

ianmcburnie: I agree with Yoni. Let's keep this simple, as is, for now.

ianmcburnie:
@sharma46bhawana What is the status of this PR? Is it ready for re-review? Or are you still working on it?

sharma46bhawana:
@ianmcburnie  I have addressed the review comments and its ready to re-review

PR #153, opened 2018-05-08
Misc dialog fixes.

### LINKED ISSUE #113: Dialog: scroll position keeps old state when reopening on mweb
Repro:
open scrollable dialog
scroll to a place in the dialog
close the dialog
open it again
it opens up at the place where we left(mweb)
it goes to the top(dweb)

DylanPiercey:
## Description
The primary enhancement in this PR is that a dialog's scroll position is reset once it is closed.
This behavior was inconsistent between dweb and mweb (related: #113).

Additionally this PR contains the following small changes/fixes:
* Fix preventing body scroll multiple times.
* Fix typo in body sizing after preventing scrolling (previously height/width were backwards and caused issues).
* Only set margin-top/left if needed when preventing body scroll.

## References
Fixes #113

### DIFF
--- a/src/common/body-scroll/index.js
+++ b/src/common/body-scroll/index.js
@@ -1,5 +1,6 @@
 let previousPosition;
 let previousStyles;
+let isPrevented = false;
 
 module.exports = {
     /**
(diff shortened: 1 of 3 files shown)


DylanPiercey:
The CI is failing for the Marko 4 tests, I’ll update the PR in the morning 🙂.

DylanPiercey:
Everything should be good now. The issue was that I am restoring the scroll for the dialog to the top a frame after It closes to avoid jank, in Marko v4 the component was being destroyed before this happened (which is perfectly valid). I just updated the PR to handle canceling the scroll reset if the component is rerendered or destroyed.

Should be good to go now 😄.

PR #164, opened 2018-05-11
Pass originalEvent when custom event maps to interaction

### LINKED ISSUE #126: Pass originalEvent for all custom events with direct mapping to DOM event

yomed:
## Description
- adds `originalEvent` for breadcrumb, carousel, and pagination
- update docs
- includes minor cleanup and related refactoring 

## Context
I looked through each component for events that only map to user interaction, and then added `originalEvent` to the custom event data. There are some cases where the event can be triggered by both UI and API. For those cases, I am not passing the `originalEvent`, because I think it seems strange to sometimes pass it and sometimes not for the same event. We can see if teams actually need this functionality later on.

## References
Follow up to #141 
Fixes #126

### DIFF
--- a/src/components/ebay-breadcrumb/index.js
+++ b/src/components/ebay-breadcrumb/index.js
@@ -45,9 +45,9 @@ function getInitialState({ hijax }) {
     return { hijax };
 }
 
-function handleClick(event) {
-    eventUtils.preventDefaultIfHijax(event, this.state.hijax);
-    emitAndFire(this, 'breadcrumb-select', { el: event.target });
+function handleClick(originalEvent) {
+    eventUtils.preventDefaultIfHijax(originalEvent, this.state.hijax);
+    emitAndFire(this, 'breadcrumb-select', { originalEvent, el: event.target });
 }
 
 module.exports = markoWidgets.defineComponent({
--- a/src/components/ebay-breadcrumb/index.js
+++ b/src/components/ebay-breadcrumb/index.js
@@ -47,7 +47,7 @@ function getInitialState({ hijax }) {
 
 function handleClick(originalEvent) {
     eventUtils.preventDefaultIfHijax(originalEvent, this.state.hijax);
-    emitAndFire(this, 'breadcrumb-select', { originalEvent, el: event.target });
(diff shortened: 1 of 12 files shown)


src/components/ebay-breadcrumb/index.js
```diff
@@ -45,9 +45,9 @@ function getInitialState({ hijax }) {
     return { hijax };
 }
 
-function handleClick(event) {
-    eventUtils.preventDefaultIfHijax(event, this.state.hijax);
-    emitAndFire(this, 'breadcrumb-select', { el: event.target });
+function handleClick(originalEvent) {
+    eventUtils.preventDefaultIfHijax(originalEvent, this.state.hijax);
+    emitAndFire(this, 'breadcrumb-select', { originalEvent, el: event.target });
```
PrashantAshok: i think you meant `originalEvent.target`
```emitAndFire(this, 'breadcrumb-select', { originalEvent, el: originalEvent.target });```

seangates: Yup. I think it should be `originalEvent.target`, too.

ianmcburnie:
Looks like there are some conflicts that need resolving.

ianmcburnie:
`There are some cases where the event can be triggered by both UI and API`

This gets me back to thinking that we may need to more explicitly separate and name some of our events. For example, `pagination-next-click` which is clearly UI based, as opposed to `pagination-next` which is API based.

yomed:
Maybe, though I don't think there's any rush on this. A more apt example now might be `notice-hide`, which can be triggered from both UI and API. It's possible that users would need `originalEvent`, but depending on what the reasoning is, we might not send that along anyway. IMO it's hard to gauge this right now since we don't have a lot of cases for either API usage, or `originalEvent` usage.

senthilp:
Agree, for now, we should be ok. We just need to follow this guideline — if there is a native event (UI triggered), `originalEvent` has to be passed.

ianmcburnie:
Yep, no rush. I think we'll have it all figured out by v1.0.0 (i.e. the grand opening).

PR #173, opened 2018-05-14
Pagination: don't emit events from disabled controls

### LINKED ISSUE #171: Pagination: Disabled next or Prev button should not emit event.

yomed:
## Description
- don't emit events when control is disabled
- add tests

## Context
This is the same behavior we use for other components like the button.

## References
Fixes #171


ianmcburnie:
Changing this to target 0.4.0. Seems a safe code change to me.

PR #175, opened 2018-05-14
Pagination: fix pass-through on controls

### LINKED ISSUE #170: Pagination: Prev , next button should support custom attributes.

sharma46bhawana:
## Description
Adding html attributes pass through in navigation arrows just like the pagination items.

## Context
Adding it as it was missed in earlier PR

## References
Fixes #170

### DIFF
--- a/src/common/test-utils/server.js
+++ b/src/common/test-utils/server.js
@@ -10,33 +10,36 @@ function getCheerio(output) {
 }
 
 /**
- * Put input in single-entry array when specified (for marko nested tags)
- * @param {Object} input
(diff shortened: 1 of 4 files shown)


yomed:
Would `htmlAttributes` be needed for the default cases too?
https://github.com/eBay/ebayui-core/blob/0fb57ad3f0b55747eae92e8ab7580c6fc1369d9c/src/components/ebay-pagination/index.js#L55-L56

ianmcburnie:
Feels safe enough to add to 0.4.0.

PR #179, opened 2018-05-15
Carousel: misc improvements

### LINKED ISSUE #161: Carousel: needs optimization of state handling
Related PR: https://github.com/eBay/ebayui-core/pull/145

From @DylanPiercey, our resident Marko expert:
>When you have an `update_` property and you update the state during that call it actually ignores any state mutations because it assumes that none happen and that everything should be handled by your `update_` handlers. Technically we could make it so that Marko checks to see if the state has been updated after running these handlers and continuing with the update but this might get weird since you could fairly easily get yourself into infinite loops.
>
>Also I think it’s better to break up this logic and trigger your new state changes on a new frame. For the case of the carousel it’d probably be good to wrap everything inside `performSlide` in a `requestAnimationFrame` anyways since we are reading the dom.
>
>The reason the double update works is because on the first pass the index changes (which calls the handler and ignores any state updates triggered within the handler) and then the second call just manually applies a rerender since no state changed and it skips the `update_` path.

The current code is ok for now, but we should look to refactor a few things:
1) Remove `update_index` function, along with FIXMEs around `this.update()`;
2) Improve DOM performance of `performSlide()`

### LINKED ISSUE #177: Carousel: broken when updating index via parent state
I am setting the Index to animate the carousel to selected index. 
it works fine when I load the page from the server side. 
but when Click a pill by scrolling to that pill, I make an ajax call,
which updated all the components on the page including the Carousel, 
the selected index is not in the view port, and the right arrow is broken too.

giphy

yomed:
## Description
- avoid use of `update_` methods, so all state goes through marko
- clean up current `FIXME`s in state handling
- put DOM manipulation inside `requestAnimationFrame`
- normalize `index` values when out of bounds
- avoid processing for carousels that don't have room to move at all
- clarify docs on `touch` usage

## References
Fixes #161 
Fixes #177


ianmcburnie:
Also fixes this issue I found on demo page:

screen shot 2018-05-15 at 11 57 04 am

yomed:
FYI I'm working on a larger refactor to fix https://github.com/eBay/ebayui-core/issues/177, but that should probably be kept separate from this safer set of changes.

yomed:
@DylanPiercey I've updated this PR with the latest. The diff isn't too helpful, but basically I tried to cover all the fundamental state issues we talked about. I also added a new test to cover the parent state case.

src/components/ebay-carousel/index.js
```diff
@@ -85,120 +85,104 @@ function init() {
...
 
-function handlePrev(originalEvent) {
-    if (!this.state.prevControlDisabled) {
-        if (this.state.isDiscrete) {
-            this.simulateDotClick(this.state.slide - 1);
-        } else if (this.state.isContinuous) {
-            this.setState('index', this.calculatePrevIndex());
-        }
-        emitAndFire(this, 'carousel-prev', { originalEvent });
+    if (this.state.index > this.lastIndex) {
```
DylanPiercey: Shouldn't this be `this.state.index !== this.lastIndex`? I might be missing something.

src/components/ebay-carousel/index.js
```diff
@@ -85,120 +85,104 @@ function init() {
...
- * @param {Integer} index
+ * Handle movement to current state.index
  */
-function performSlide() {
-    // FIXME: API manipulation is disabled in native scroll case
-    if (this.state.index >= 0 && this.state.index <= this.lastIndex && !this.usesNativeScroll) {
-        const oldFirstVisibleIndex = this.state.firstVisibleIndex;
-        const oldLastVisibleIndex = this.state.lastVisibleIndex;
+function processMovement() {
+    this.processMovementFrame = requestAnimationFrame(() => {
```
DylanPiercey: We should probably `cancelAnimationFrame` before this line. Just incase they change the state twice in a frame we can avoid doing extra work.

src/components/ebay-carousel/index.js
yomed: `lastIndex` is used as in "A carousel of length 4 would have lastIndex=3." I'm using `activeIndex` below to keep track of when to handle the state updates.

PR #181, opened 2018-05-15
Dialog scrolling to top of page IE

### LINKED ISSUE #180: ebay-dialog: in IE when dialog opens up, The parent screen goes to the top
In IE  when the dialog open up the background screen goes to the top instead of staying where it is.

before dialog opens
image

after dialog opens
image
 **expected**

image

DylanPiercey:
## Description
This changes from using `scrollY` and `scrollY` to `pageYOffset` and `pageXOffset ` which have better browser support.

## Context
Using `scroll` properties do not work in IE.

## References
Fixes #180

### DIFF
--- a/src/common/body-scroll/index.js
+++ b/src/common/body-scroll/index.js
@@ -17,20 +17,20 @@ module.exports = {
      */
     prevent() {
         const { body } = document;
-        const { scrollX, scrollY } = window;
+        const { pageXOffset, pageYOffset } = window;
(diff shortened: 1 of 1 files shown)


ianmcburnie:
Going to keep this for 0.5.0, just to lessen chance of regression in other browsers

DylanPiercey:
Sounds good. I did test this against FF, Chrome, Safari, IE11 and Edge and it appears to be working in all of them now.

PR #183, opened 2018-05-16
README - Update for Marko v4 users

trg:
## Description

Updates README with steps for users of latest Marko version trying out the `<ebay-menu>` example in the README.

## Context

First time `ebayui` consumers who are on Marko v4 immediately run into runtime errors with the provided README example; and may not be aware that `ebayui` components require an additional compatibility layer given that the required version of Marko is "3+".

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ For more information, please read [Building a UI Component in 2017 and Beyond](h
 * [Marko](https://markojs.com) (v3+)
 * [eBay Skin](https://ebay.github.io/skin/) (v4+)
 
-> Note:\
-> Marko v3 requires [Marko Widgets v6](https://github.com/marko-js/marko-widgets)\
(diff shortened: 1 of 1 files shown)


seangates:
I think this note should go under "Getting Started", _before_  the "Custom Tags" header.

yomed:
@seangates Maybe we can just update the `NOTE` here:
https://github.com/eBay/ebayui-core/blob/master/README.md#requirements

> Marko v3 requires Marko Widgets v6
Marko v4 requires Marko Widgets v7

seangates:
That works, too!

ianmcburnie:
Let's go with Yoni's suggestion.

trg:
Copy updated: 8d46cfe

Should this be pulled into master or into `0.5.0`?

seangates:
Thanks, @trg!

Issue #172, opened 2018-05-14
Menu: use inline SVG for menu button icon


ianmcburnie:
When app team overrides the background color and foreground color of menu button, the icon remains fixed as black. We should change from using span with background svg to inline svg, so that the icon will inherit the foreground colour.

yomed:
I think this would need changes in Skin to make `svg.expand-btn__icon` a smaller size.

PR #189, opened 2018-05-17
Integrate icon into carousel and dialog

yomed:
## Description

- add `<ebay-icon>` to carousel and dialog templates
- add `/ebay-icon` to the browser.json files (with minor reordering cleanup)

## Context

I'm leaving the menu/select for later since I think those would work better with Skin changes (https://github.com/eBay/ebayui-core/issues/172).

@PrashantAshok You should be able to update checkbox/radio to use this pattern. If we want to get those in quickly, I can also just apply this to those components as part of this PR (after they're merged).

@DylanPiercey I'm leaving the dynamic demo setup alone for now, because most icons are being imported by the `<ebay-icon>` examples, which are getting cached since every component uses the same template.

## References

Follow up to https://github.com/eBay/ebayui-core/pull/136


PrashantAshok:
I might be wrong, but didn't we discuss not to add `ebay-icon` to other components, because of issues involved with SPA. @senthilp?

senthilp:
Yes, that seems right. We will sync-up on this as a group and make a call.

yomed:
There shouldn't be an issue using `<ebay-icon>` in SPAs (either on it's own, or in components). It's just that SPA teams will need to handle their own stamping, rather than relying on our magic stamper.

senthilp:
Will you be using inline or background for carousel and dialog?

yomed:
Sorry, the diff is clearer now after rebase. @senthilp Carousel and dialog use inline because both cases need to change the icon color during interaction.

PR #192, opened 2018-05-18
Prevent closing dialog when clicking on whitespace in the panel.

### LINKED ISSUE #188: ebay-dialog: Clicking within dialog whitespace closes it?
I'm not sure if this is part of the pattern, but tapping/clicking within the dialog that's not content causes it to close.

DylanPiercey:
## Description
This change fixes the dialog closing when clicking on whitespace at the bottom of the dialog panel.

## Context
Ideally this change would be implemented in @ebay/skin by making the `.dialog__body` fill the `.dialog__window`, or by finding an alternative to the `padding` on the `.dialog__window` however I wasn't able to find a solution that worked for all types of dialogs and was working in all browsers. Hopefully we can come back to this.

## References
Fixes #188

### DIFF
--- a/src/components/ebay-dialog/index.js
+++ b/src/components/ebay-dialog/index.js
@@ -162,12 +162,24 @@ function show() {
     this.setState('open', true);
 }
 
-function close(ev) {
-    if (ev && this.bodyEl.contains(ev.target)) {
-        return;
+function close() {
+    this.setState('open', false);
+}
+
+function handleDialogClick({ target, clientY }) {
+    const { closeEl, windowEl } = this;
+
+    // Checks if we clicked inside the white panel of the dialog.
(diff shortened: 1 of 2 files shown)


DylanPiercey:
@ianmcburnie I'm not necessarily against having two methods. To be clear though with the current implementation you can still call `widget.close()` which will always close the dialog since it first does a check to see if this was called with an event.

ianmcburnie:
@DylanPiercey Yeah, I did see that. Both ways achieve the same end result. But I personally feel that the template is a little more readable with two methods, and the close method is simplified. I think onDialogClick should figure out where the click happened, and call the close method if necessary.

DylanPiercey:
@ianmcburnie I updated the PR based on your suggestion. Let me know if that's what you were thinking.

DylanPiercey:
Fixed the conflict 😄.

PR #196, opened 2018-05-22
Docs: updated Pull Request section of contributing guide

ianmcburnie:
This PR is intended for `master` branch.

## Description

Updated the pull requests section of the contributing guide.

## Context

The old guidelines did not reflect our actual working model, and some personal preferences. Hopefully these updates are *better* and a little more flexible, but not saying they are perfect either ;-)

## References

https://github.com/eBay/ebayui-core/blob/contributing_updates/CONTRIBUTING.md#pull-requests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,7 +50,7 @@ To ensure that all components function correctly in both Marko v3 and v4, we run
 
 ## Pull Requests
 
-This section contains information and in-house etiquette regarding pull requests (or "PRs" as they are often called).
+This section contains information and etiquette regarding pull requests (or "PRs" as they are often called).
 
 ### Before You Start
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,21 +48,52 @@ Please ensure all scripts in `package.json` are Unix *and* Windows friendly.
 
 To ensure that all components function correctly in both Marko v3 and v4, we run our entire test suite against both versions. This requires automated installation of Marko versions, and is best executed in the CI (to avoid conflicting with local environments). The CI runs a specific `build:ci` task in the NPM scripts for this purpose.
 
-## Pull Requests
+## Pull Requests (PRs)
 
-Please ensure an issue exists in GitHub before writing any code. Every pull request must have an associated issue for tracking and triage purposes.
+This section contains information and in-house etiquette regarding pull requests.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -50,19 +50,50 @@ To ensure that all components function correctly in both Marko v3 and v4, we run
 
 ## Pull Requests
 
-Please ensure an issue exists in GitHub before writing any code. Every pull request must have an associated issue for tracking and triage purposes.
+This section contains information and in-house etiquette regarding pull requests (or "PRs" as they are often called).
```
seangates: Is this limited to "in-house" etiquette?

ianmcburnie: I guess I meant "house rules". But I can probably remove this word to avoid confusion.

ianmcburnie:
@yomed There are now two options of what to do with your commit history (squash or don't squash), compared to just one previously (must squash). This is based on our behaviour and patterns I've seen within the team. I would not want to add any more options. Based on our behaviour, option 2 has been the most common, because there's rarely been a case where we wanted to keep more than 1 commit from our feature branch history. But some people prefer option 1 regardless.

There are 3 new sections: "Moving Code", "Avoiding Conflicts" and "Merging Code" - all created to document the processes we have been following. So I would say there are some changes from previously.

Cleanup in the same commit is fine, to an extent, we all do it. What I'm saying is that it's not a good experience for the reviewer if you move a lot of code and files and make changes within those blocks in the same commit. Maybe I can make that clearer. Also, the language is "Please try", which doesn't feel strong to me. The sentiment is to "isolate as best as you can". 

I have no objection to 10+ commits, and neither does the document, so long as each commit is a useful atomic commit (i.e. a state we might wish to rollback to on master branch) that does not leave the app in a broken state. If they are not useful, get rid of them.

yomed:
Ok. My object to the strong language is mostly around how PRs will be "closed immediately" or we'll "reject based on ____." Sure, we can close/reject PRs whenever we feel like it, but to me, spelling it out like this kind of makes it sound like we _want_ to reject PRs.

ianmcburnie:
@yomed Ah, okay. I see. Fair point. Maybe I could remove all of those "The reviewer has the right to ..." statements, or just reword them slightly. What do others think?

ianmcburnie:
Going to merge, as I got two approvals. We can change the language based on feedback later if necessary.

PR #209, opened 2018-06-08
63 mweb select

### LINKED ISSUE #63: Select: support native mweb use case
Related ticket to be tackled first: eBay/skin#54.

seangates:
## Description
This set of changes completes the split between `combobox` and `select`, and allows us to allow teams to have a native `select` for their mWeb implementations of their pages/apps.

## References
Closes #63.


seangates:
The CI will fail until we publish Skin (either v5, or another pre-release). I'm hoping to release Skin today.

seangates:
Okay, Skin v5.0.2 was published yesterday, so this should be working. Let me investigate.

src/components/ebay-combobox/browser.json
```diff
@@ -0,0 +1,11 @@
+{
+    "dependencies": [
+        {
+            "if-not-flag": "ebayui-no-skin",
+            "path": "@ebay/skin/listbox"
```
ianmcburnie: Should this be `@ebay/skin/combobox` instead of `@ebay/skin/listbox `?

src/components/ebay-combobox/browser.json
```diff
@@ -0,0 +1,11 @@
...
+        },
+        "../../../src/components/ebay-button",
```
ianmcburnie: Do we need the dependency on button?

seangates:
@ianmcburnie Please have a look. Updated per comments.

Also, I removed the fixed-width `span` from each of the examples for the `ebay-select` as it's not needed.

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,244 @@
...
+
+    return {
+        class: comboboxClass,
+        btnClass,
+        optionsClass,
+        name: state.name,
+        selectedOption: state.selected,
+        options: state.options,
+        grow: state.grow,
+        disabled: state.disabled,
```
DylanPiercey: Disabled does not seem to do anything in this component.

seangates: I think it does here: https://github.com/eBay/ebayui-core/pull/209/files/3fb66418b28ce66c87a6246909619e326ac3844a#diff-854be3f0318a650eed139adb0e266c38R12

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,244 @@
...
+        htmlAttributes: state.htmlAttributes
+    };
+}
+
+function init() {
+    const optionEls = this.el.querySelectorAll(comboboxOptionSelector);
+
+    if (this.state.options && this.state.options.length > 0) {
+        this.expander = new Expander(this.el, {
```
DylanPiercey: Maybe we should change the expander to be initialized after Marko renders (onUpdate) instead of here.
I think that with this setup if you first render a select with no options, and then again with options it would not work properly.

DylanPiercey: Also if we were to do that we could change it so that the `makeup-expander` is not initialized if the the select is disabled (or empty).

seangates: Yes, we've found that this was an issue. This would clear that up.

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,256 @@
...
+            observer.observeInner(this, optionEl, 'selected', `options[${i}]`, 'options', selectedObserverCallback);
+        });
+
+        scrollKeyPreventer.add(this.el.querySelector(comboboxHostSelector));
+        scrollKeyPreventer.add(this.el.querySelector(`.${comboboxOptionsClass}`));
+    }
+}
+
+function onUpdate() {
+    if (!this.expander && this.state.options && this.state.options.length > 0) {
```
DylanPiercey: This will still break if someone is toggling `disabled` on the component. We need to (onUpdate) check to ensure the the expander is enabled or disabled depending on the current `disabled` state.

seangates: Yeah, I pushed this but am still working on it.

ianmcburnie:
Let's not try and fix pre-existing issues with ebay-select in this PR. The scope of this PR is only to rename/reorganise. Let's file them as separate issues for tracking and we will triage them appropriately. 

@seangates Can you please file the two issues we discussed offline.

seangates:
Yup. Will do.

PR #213, opened 2018-06-12
Convert icon to inline and dedupe svgs instead of using stamper.

### LINKED ISSUE #202: Icon: SPA support for ebay-stamper
# Feature Request

## Description
Currently the `ebay-stamper` component does not work properly when used in a SPA.

## Potential Solutions / Notes
* Expose the `ebay-stamper` and allow teams to manually specify which svg's to inline.
* Setup all symbols in an external svg on a cdn.
* Potentially expose the logic for getting the icons used in a page.

### LINKED ISSUE #206: Icon: requires a `body`. When serving content through AJAX we do not have a body.
# Bug Report

## Description
Create a route in ajax that just does this:
```
<lasso-page
        package-path="./browser.json"
    />
 <ebay-dialog > Hello </ebay-dialog>
```
in the controller, send down json to the browser so the user can inject it.
```
        template.render({}, (err, output) => {
            const renderedComponents = require('marko/components').getRenderedComponents(output.out);
            return resolve({
                renderedComponents,
                html: output.getOutput()
            });
        });
```

Observe how there is no ebay-icon transform done in the response.

DylanPiercey:
## Description
In this PR the `eBay-icon-stamper` component is removed and the implementation of `ebay-icon` is changed to automatically include the `symbol` for the icon (the first time its rendered) under a `<def>` tag in the stamped out svg. Subsequent icons will not output the symbol multiple times and instead piggy back off of the existing symbol.

Once the `ebay-icon` component mounts in the browser it automatically creates a single `svg` at the top of the body and all `symbol`s are moved there. The symbols are never deleted but can be added to as new components with unrendered icons are rendered.

## Context
This allows for better support for the SPA use case with `ebay-icon`. With the above approach the implementation is made isomorphic and works well enough for both the SSR and SPA use cases.

The only downside is that now an icon's symbol can be sent down twice (once in the initial SSR of the document if it was rendrered) and again in the JavaScript to support client side rendering. I believe the impact would be very minimal.

## Details

Lets take a look at the `ebay-dialog` as an example to illustrate the above.

When the server renders the first `ebay-dialog` with it's `close` button the following will be rendered:
image

*Note: The [defs](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/defs) are hidden and work similar to a hidden SVG. Only the first time that icon is rendered will it's stamp contain the `<def>`.*

The next rendered dialog in the page would look as it did before:
image

Again, after the components mount in the browser all Icons are moved to an svg at the top of the body. This is done to maintain any rendered symbols even though the components may be removed at any time via updates/rerenders.

## References
Fixes #206
Fixes #202 

This also fixes the icons rendering in the marketplace.


senthilp:
@DylanPiercey In this scenario, the icon symbols will be sent twice, even if I am not doing client-side rendering right? This is because we are always bundling the JS.
We probably need another review of it, similar to what Yoni did a month back.

DylanPiercey:
@senthilp yes the icons used in the page will be sent at least once (in the js bundle) and then once more if they are displayed on the initial server side render in an inline svg.

The worst case scenario would be that the page uses all icons and renders all icons on the page on the first render also. In this case the minified js adds about 500 bytes per symbol and with 38 icons currently that means we could be potentially sending down an extra 19kb.

This number would be cut down a fair amount once the JS is gzipped also and on top of that the JS here should be cached unlike the inline svg's which we are already sending.

I think the extra weight is worth it as it simplifies the component's internals and also allows us to fully support SPA's.

src/components/ebay-icon/index.js
```diff
@@ -1,17 +1,48 @@
...
+        rootSvg = document.createElement('svg');
+        rootSvg.hidden = true;
+        document.body.insertBefore(rootSvg, document.body.firstChild);
+    }
+
+    // If there were any symbols rendered then we move them to the svg above after rendering them.
+    const defs = this.getEl('defs');
+
+    if (defs) {
+        const symbol = defs.firstChild;
```
PrashantAshok: I know it's `symbol` and `Symbol`, maybe a "safer" variable name?

src/components/ebay-icon/template.marko
```diff
@@ -4,6 +4,7 @@
 </if>
 <else-if(data.isInline)>
     <svg w-bind ${data.htmlAttributes} ${data.accessibilityAttributes} focusable="false" class=data.classes>
+        <defs w-id="defs" w-body if(data.renderDefs)/>
```
aselvaraj: Can the symbol be directly included here? why do we need a transformer?

DylanPiercey: @aselvaraj the thing is we want to make it so that you only get whichever symbols you use in the javascript bundle.

The transform for this implementation of `ebay-icon` does this:

```marko
 <ebay-icon type="inline" name="close"/>

 <!-- Becomes -->

<ebay-icon type="inline" name="close"><include('$DIRNAME/symbols/close.marko')/></ebay-icon>
```

Which basically just makes it so that you automatically include only the specific symbol you are asking for.

Issue #215, opened 2018-06-14
Menu: support "size" attribute


PrashantAshok:
# Feature Request

## Description
Make `ebay-menu` support size attribute similar to how `ebay-button` supports `size="small"` attribute. 

## Context
To support cases where a `ebay-button` (`size="small"`) would be placed alongside `ebay-menu`

## Screenshots
<img width="319" alt="screen shot 2018-06-14 at 12 16 40 pm" src="https://user-images.githubusercontent.com/5200733/41435993-4b473c34-6fd5-11e8-9f95-b277a52e76f6.png">

PR #238, opened 2018-07-02
Menu: add size attribute for button

yomed:
## Description

- add menu size attribute
- pass size to `<ebay-button>`
- update docs and tests
- minor reordering of menu examples so that the button variations are together

## Context

Note: This only affects the size of the button, and does not change the actual menu content at all.

@ianmcburnie This doesn't actually apply the `expand-btn--{size}` class, and instead would result in `btn--{size}`. I'm taking advantage of the fact that both work for this case. Longer term, we need to decide if the `expand` button type should be supported separately within `<ebay-button>` itself. Right now, the `<ebay-menu>` is responsible for setting up the `expand-btn`, so `<ebay-button>` doesn't know to format the classes specifically for that.

Relatedly, `priority` does not currently work with `expand-btn`, so I'm deferring that until it is supported in Skin.

## References

#215 

## Screenshots

screen shot 2018-07-02 at 2 32 09 pm


PrashantAshok:
@yomed, thanks for this PR. Reg your comment on the priority, it looks like skin supports [it](https://github.com/eBay/skin/blob/484e2ba26f87b2085fa5882795ff62192538f7aa/src/less/button/ds6/button-base.less#L93)

seangates:
I just noticed that the borderless menu should have a transparent background.

yomed:
@PrashantAshok Yeah looks good in Skin -- I'll get to supporting priority once https://github.com/eBay/ebayui-core/pull/240 is in (needed for the icon color to work correctly with `priority="primary"`).

@seangates That's a Skin issue right? Could you create an issue there?

PR #239, opened 2018-07-02
Select: split pass-through attributes between root and select

### LINKED ISSUE #208: Select: attributes pass through
Hey guys,

I'm trying to use `ebay-select` component on my page and stumbled upon some limitations.

There are cases, when i need to pass `aria-required` or `aria-invalid`, etc. And i expect such attributes to be passed-through to the `select` element that component produces. But in reality it's being appended to the root element instead.

As far as i believe, the only attribute to be applied to the root element of produced component should be `class`. Everything else should be applied to the 'control' element.

yomed:
## Description

- move `htmlAttributes` (besides `class`) to `<select>` tag
- remove redundant `name` attribute
- update docs and tests

## Context

@seangates With the HTML attributes moved to the `<select>`, we no longer need an explicit `name` attribute. However, since it's a pretty important attribute in using this component in general, I left it as-is in all the example code.

## References

Fixes #208


ianmcburnie:
Just a heads up that this will go into 0.7.0 branch.

seangates:
Although, `name` was there to specifically help enforce developers giving the select a `name`. But, I realize it's not required. Even though we assume they know they need it, maybe we need to remind them like, "Don't forget to add the `name` attribute to the select!".

Issue #219, opened 2018-06-21
Menu: Support "priority" attribute


PrashantAshok:
# Feature Request

## Description
Make ebay-menu support `priority` attribute similar to how ebay-button has the priority attribute.

## Context
To support cases where a `ebay-button` (priority="primary") would be placed alongside `ebay-menu`

## Screenshots
<img width="312" alt="screen shot 2018-06-20 at 4 41 57 pm" src="https://user-images.githubusercontent.com/5200733/41749584-1ff34e0a-756c-11e8-9794-554dbb7bb979.png">

yomed:
Currently waiting on Skin 5.1.0-1

ianmcburnie:
Skin 5.1.0-1 is now ready.

PR #240, opened 2018-07-03
Menu: integrate ebay-icon for chevron

yomed:
## Description

- add `no-skin-classes` attribute to `<ebay-icon>`
- update icon tests
- integrate `<ebay-icon>` as inline svg in menu

## Context

The button icons in Skin (`expand` and `cta`) are used without the traditional `icon` classes. We still want to use `<ebay-icon>` for the markup and other magic, so I added a new `no-skin-classes` attribute. I considered keeping it undocumented, but I think it could potentially be useful for teams, so I noted it in the `README`.

With this change, the icon color matches the button.

## References

https://github.com/eBay/ebayui-core/issues/172
https://github.com/eBay/skin/issues/250

## Screenshots

Before:
screen shot 2018-07-03 at 11 20 55 am

After:
screen shot 2018-07-03 at 11 21 14 am


src/components/ebay-icon/index.js
```diff
@@ -61,7 +61,7 @@ function getTemplateData(state, input, out) {
         accessibilityAttributes,
         titleId,
         htmlAttributes: processHtmlAttributes(input),
-        classes: ['icon', `icon--${name}`, input.class]
+        classes: input.noSkinClasses ? [input.class] : ['icon', `icon--${name}`, input.class]
```
senthilp: @yomed Is this because we are using inline SVG icons in `<ebay-menu>`? Or the calling component is taking care of the classes?

yomed: There are a few special case button icons in Skin, where they follow all the rules of regular icons, except without applying the regular classes. In this case, the menu needs to use this for the `expand-btn__icon`.

yomed:
@seangates Can you be specific about what kinds of abuse? I haven't opened up custom menu icons itself -- the menu still only uses the `chevron-down-bold` icon.

seangates:
Ah, the name is the piece used to control which icon is used. Apologies. I had misunderstood that it was the classes passed through.

seangates:
So much for looking at this code while watching a WC match.

yomed:
Note that https://github.com/eBay/ebayui-core/issues/219 is dependent on this PR.

Issue #216, opened 2018-06-15
Infinitely Looping, Auto-Playing Carousels


trg:
# Feature Request

Autoplaying, Infinitely looping slideshow

## Description

- Auto-playing / Auto-pagination.  The carousel paginates itself without user interaction.
- The carousel loops around to the first slide/set of slides when the user reaches the end of the slides, however to the user the slides always appear to slide in from the correct direction.

## Context

Ebay.com's desktop homepage currently has a carousel with this functionality.

## Screenshots

![jun-15-2018 16-13-33](https://user-images.githubusercontent.com/151419/41487544-1b04351c-70b7-11e8-9a39-42d65b72de5d.gif)

Issue #149, opened 2018-05-07
Carousel: layout and js breaks without any items


hramakumar:
go to https://go/ebayui/test2
search for "empty carousal"

actual:
![image](https://user-images.githubusercontent.com/13157115/39726863-551140ae-5205-11e8-895e-25fc028e7e5e.png)

yomed:
In addition to the layout issue, the the js call stack can run out. Since there are no items, `lastIndex` becomes -1. Then the index continuously loops between 0 and -1, since it uses `lastIndex` as the upper bound.

PR #245, opened 2018-07-06
Carousel: Refactor + Bug fixes / Improvements

### LINKED ISSUE #176: Carousel: movement controls don't apply for native scrolling
Filing this ticket to keep track of the issue as we have closed the old ticket.

In the mWeb use case, whether or not a start index is defined, the continuous carousel starts with the first card. Currently, a touch-enabled carousel with `type="continuous"` will use native scrolling which has more limited functionality.

**actual**
image

**expected:**
image

### LINKED ISSUE #197: Carousel: starting index cuts off content in continuous with native scrolling
A little hard to tell in the picture, but basically it has a starting index of 2, which makes all the content inaccessible before the third card. 

Movement is currently handled differently for the native scrolling continuous case (per https://github.com/eBay/ebayui-core/issues/176), but we should fail more gracefully here.
screen shot 2018-05-22 at 6 39 37 pm

### LINKED ISSUE #220: Carousel: The rightmost image is cut off when element margin is overridden
# Bug Report

## Description
If carousel images have same width/height, when application overrides `.carousel__list li` margin, on desktop the rightmost image in the carousel is cut off.

## Workaround
Not override carousel element margin

## Screenshots
screen shot 2018-06-25 at 1 15 43 pm

## Reference
[Running app](http://xc.addonfe.stratus.qa.ebay.com/trends), the 2nd carousel shows the issue.

### LINKED ISSUE #241: ebay-carousel: carousal navigations are very slow
# Bug Report

## Description
Looks like carousal takes a while to move slides
go to https://go/ebayui/test2
click right arrow on any discrete carousal
it takes some time delay to move to the next slide

https://www.useloom.com/share/40e79ec813364fa992996e0cb5f2b23c

DylanPiercey:
Alright, here is part one of my carousel PR.
This refactor was done primarily to normalize various options on the carousel, simplify it's rendering logic and to fix some pre existing issues.

## Fixes
* fixed #220 by exposing a new `gap` attribute to override the margin.
* fixed #197
* fixed #176 
* Partially fixes, #149  and fixed #241
* Previously if items were too large in the carousel the slide controls would be off after navigating.

## Improvements
* Carousel now re-renders less often than previously.
* Re-renders are now shallow and don't actually cause a re-render of the carousel items unless needed.
* Dot controls animation was tweaked to avoid transitioning background-color for better performance.
* The carousel is now forced onto a new layer (`translate3d`) which improves performance in the native scrolling mode quite a bit.

## Features
* On mobile: native scrolling now automatically snaps to the closest items and emits events.
* On mobile: the controls now work and animate properly (#176).
* You can now have any number of `items-per-slide` (previously capped at 5).
* Auto looping index, working toward #216, now when you set the `index` above the bounds it goes to the first item, and 0 goes to the last.

## Breaking Changes
The changes below are not absolutely mandatory however during this PR I had some time to think about what we really need. This will probably warrant more discussion and if there are concerns with these changes I can update the PR.

* The `type` attribute was removed. Everything that could be gleaned from this attribute can be determined from just checking if `items-per-slide` is set and I feel that having two types of carousel is slightly confusing. In this implementation `items-per-slide` automatically sizes the items and adds to slide controls.

* The `slide` attribute was removed. In the previous implementation there were several branches that were accounting for both a `slide` and an `index` attribute. The slide can always be found from the index, and is relevant in (what was previously) `discrete` and `continuous` carousels. In this PR the slide is always calculated from the `index` and there is now one source of truth and a simpler API.

* Exposed methods on the widget have changed. This is not something we are advertising on our public API but I know at one point we recommended using the `refresh` method. I believe anyone that was using that method was just using it while waiting for a patch to the carousel. Users should not need be aware of these methods anyways, however to achieve the same effect now you can use `widget.update()` instead.

* `move` event has been removed. This was mostly because the hooks required to implement this clutter the codebase and as far as I can tell little value is added. Users should mostly want to use the `carousel-update` event which provides the current active indexes. If users really want to know when the carousel moves they can also listen for the `update` event which effectively gives the same result.

## Context
I planned on just working toward #216 but managing state of the component and keeping track of things was becoming a bit tricky.

### DIFF
--- a/src/components/ebay-carousel/README.md
+++ b/src/components/ebay-carousel/README.md
@@ -14,25 +14,18 @@ This component will bundle different resources depending on Lasso flags provided
 Use `touch` for touch devices, and `no-touch` for non-touch devices. Without flags, `no-touch` is assumed.
 [//]: # (TODO: `touch` should ideally be default, but currently the js behavior doesn't execute with native scrolling)
 
-Note: The `touch` carousel with type="continuous" will use native scrolling. This behavior is used instead of our controlled movement, so some functionality is limited. In this case, the carousel will not respond to `index` changes, and will not emit events.
-
 ## ebay-carousel Attributes
 Name | Type | Stateful | Description
 --- | --- | --- | ---
-`type` | String | No | "continuous" (default) or "discrete"
 `accessibility-prev` | String | No | aria-label for previous control (default: "Previous Slide")
 `accessibility-next` | String | No | aria-label for next control (default: "Next Slide")
-
-### Additional Attributes for type="continuous"
-Name | Type | Stateful | Description
---- | --- | --- | ---
 `index` | String | Yes | 0-based index position
+`items-per-slide` | String | No | automatically fit a number of items for each carousel slide and enable slide controls.
+`gap` | String | No | override the margin between carousel items (default: "16px")
 
-### Additional Attributes for type="discrete"
+### Additional Attributes for when items-per-slide is set.
 Name | Type | Stateful | Description
 --- | --- | --- | ---
-`slide` | String | Yes | 1-based slide position
-`items-per-slide` | String | No | 1 - 5 -- number of items for a single carousel slide (default: 1)
 `accessibility-status` | String | No | status text (default: "Showing Slide {currentSlide} of {totalSlides} - Carousel")
 `accessibility-current` | String | No | pagination current slide text (default: "Current Slide {currentSlide} - Carousel")
 `accessibility-other` | String | No | pagination other slide text (default: "Slide {slide} - Carousel")
@@ -41,11 +34,10 @@ Name | Type | Stateful | Description
 Event | Data | Description
 --- | --- | ---
 `carousel-update` | { [visibleIndexes] } | called whenever item visibility changes, including initialization
-`carousel-move` | | start of carousel content movement animation
 `carousel-next` | { originalEvent } | click next control
 `carousel-prev` | { originalEvent } | click previous control
 
-### Additional Events for type="discrete"
+### Additional Events for when items-per-slide is set.
 Event | Data | Description
 --- | --- | ---
 `carousel-slide` | `{ slide }` | new slide is navigated to (by controls, dots, or API)
--- a/src/components/ebay-carousel/style-touch.less
+++ b/src/components/ebay-carousel/style-touch.less
@@ -9,22 +9,24 @@
         }
     }
 
-    &--continuous &__container {
-        overflow-x: scroll; // also used in js to determine scroll behavior
+    &__list {
+        overflow-x: auto; // also used in js to determine scroll behavior
         -webkit-overflow-scrolling: touch;
+        -ms-overflow-style: none;
+
+        &::-webkit-scrollbar {
+            display: none;
+        }
     }
 
-    &--discrete {
-        // show controls on touch since there's no hover
-        // TODO: swipe support on mobile discrete?
-        .carousel__control {
-            opacity: @ebay-carousel-control-enabled-opacity;
-            pointer-events: auto;
+    // show controls on touch since there's no hover
+    .carousel__control {
+        opacity: @ebay-carousel-control-enabled-opacity;
+        pointer-events: auto;
 
-            &[aria-disabled="true"] {
-                opacity: @ebay-carousel-control-disabled-opacity;
-                cursor: default;
-            }
+        &[aria-disabled="true"] {
+            opacity: @ebay-carousel-control-disabled-opacity;
+            cursor: default;
         }
     }
 }
--- a/src/components/ebay-carousel/style.less
+++ b/src/components/ebay-carousel/style.less
@@ -15,20 +15,18 @@
 
     &__list {
         transition: transform @ebay-carousel-transition-time @ebay-carousel-transition-function;
-        transform: translateX(0);
+        will-change: transform; // Ensures that the list is on a new layer for better scrolling perf.
+        transform: translate3d(0,0,0); // Same as above, but for backward compatibility to (ie10).
+        display: flex;
         position: relative;
-        box-sizing: border-box;
         width: 100%;
         margin: 0;
         padding: 0;
 
         > li {
+            flex-grow: 0;
+            flex-shrink: 0;
             list-style: none;
-            margin-right: 16px;
-
-            &:last-of-type {
-                margin-right: 0;
-            }
         }
     }
 
@@ -69,48 +67,6 @@
         }
     }
 
-    &--continuous {
-        .carousel__list {
-            display: inline-block;
-
-            > li {
-                display: inline-block;
-            }
-        }
-    }
-
-    &--discrete {
-        .carousel__list {
-            display: flex;
-
-            // avoid flex shorthand due to ie11 bugs
-            > li {
-                flex-grow: 0;
-                flex-shrink: 0;
-            }
-        }
-
-        &-1 .carousel__list > li {
-            flex-basis: 100%;
-        }
-
-        &-2 .carousel__list > li {
-            flex-basis: calc(50% ~"-" 8px);
-        }
-
-        &-3 .carousel__list > li {
-            flex-basis: calc(33.333% ~"-" 10.666px);
-        }
-
-        &-4 .carousel__list > li {
-            flex-basis: calc(25% ~"-" 12px);
-        }
-
-        &-5 .carousel__list > li {
-            flex-basis: calc(20% ~"-" 16px);
-        }
-    }
-
     &__dots {
         margin: 8px 0;
         padding-left: 0;
@@ -120,8 +76,7 @@
             display: inline-block;
 
             > button {
-                transition: background-color @ebay-carousel-transition-time @ebay-carousel-transition-function;
-                background-color: @ds6-color-g204-gray;
+                background-color: @ds6-color-g206-gray;
                 cursor: pointer;
                 border: 0;
                 border-radius: 50%;
@@ -130,9 +85,23 @@
                 padding: 0;
                 height: 8px;
                 width: 8px;
+                overflow: hidden;
 
-                &.carousel__dot--active {
-                    background-color: @ds6-color-g206-gray;
+                // The active dot uses an after to color it in which transitions opacity.
+                // This was done to avoid transitioning the background color of the element
+                // while still allowing us to use exact ds colors.
+                &::after {
+                    transition: opacity @ebay-carousel-transition-time @ebay-carousel-transition-function;
+                    content: "";
+                    display: block;
+                    width: 100%;
+                    height: 100%;
+                    opacity: 1;
+                    background-color: @ds6-color-g204-gray;
+                }
+
+                &.carousel__dot--active::after {
+                    opacity: 0;
                 }
             }
         }
--- a/src/components/ebay-carousel/style.less
+++ b/src/components/ebay-carousel/style.less
@@ -24,6 +24,7 @@
         padding: 0;
 
         > li {
+            display: inline-block;
             flex-grow: 0;
             flex-shrink: 0;
             list-style: none;
--- a/src/components/ebay-carousel/template.marko
+++ b/src/components/ebay-carousel/template.marko
@@ -1,48 +1,56 @@
 <div class=data.classes w-bind ${data.htmlAttributes}>
+    <var config=data.config/>
     <var statusId=("carousel-status-" + widget.id)/>
-    <div class=("carousel__container" + (data.bothControlsDisabled ? " carousel__container--controls-disabled" : ""))>
+    <div w-id="container" class=["carousel__container", data.bothControlsDisabled && "carousel__container--controls-disabled"]>
         <span
-            if(data.isDiscrete)
+            if(data.totalSlides >= 1)
             id=statusId
             class="clipped"
             role="status"
             aria-live="polite">
-            ${data.statusText}
+            ${data.accessibilityStatus}
         </span>
-        <button
-            class="carousel__control carousel__control--prev"
-            w-onclick="handlePrev"
+        <button class="carousel__control carousel__control--prev"
+            w-onclick=(!data.prevControlDisabled && "handleMove")
+            data-direction=-1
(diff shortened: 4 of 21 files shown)


src/components/ebay-carousel/style.less
```diff
@@ -15,20 +15,17 @@
...
-        transform: translateX(0);
+        transform: translate3d(0,0,0); // Ensures that the list is on a new layer for better scrolling perf.
+        display: flex;
         position: relative;
-        box-sizing: border-box;
         width: 100%;
         margin: 0;
         padding: 0;
 
         > li {
```
PrashantAshok: can't we not define & use a `class` here? css performance ;-)

src/components/ebay-carousel/style.less
```diff
@@ -120,8 +75,7 @@
             display: inline-block;
 
             > button {
```
PrashantAshok: same comment as above, can't we not define & use a `class` here?

DylanPiercey: I think the css could be optimized further as well but I tried to keep it as close as possible to the existing css for this PR. Perhaps in a future PR we could optimize the selectors further.

src/components/ebay-carousel/README.md
```diff
@@ -14,25 +14,17 @@ This component will bundle different resources depending on Lasso flags provided
 Use `touch` for touch devices, and `no-touch` for non-touch devices. Without flags, `no-touch` is assumed.
 [//]: # (TODO: `touch` should ideally be default, but currently the js behavior doesn't execute with native scrolling)
 
-Note: The `touch` carousel with type="continuous" will use native scrolling. This behavior is used instead of our controlled movement, so some functionality is limited. In this case, the carousel will not respond to `index` changes, and will not emit events.
-
 ## ebay-carousel Attributes
 Name | Type | Stateful | Description
 --- | --- | --- | ---
-`type` | String | No | "continuous" (default) or "discrete"
```
senthilp: How will this work on mobile devise if there is only 1 item per slide and I want that to be continuous? Meaning I am using natural mobile horizontal scroll and can leave the item in an in-between state.

src/components/ebay-carousel/style.less
```diff
@@ -15,20 +15,17 @@
 
     &__list {
         transition: transform @ebay-carousel-transition-time @ebay-carousel-transition-function;
-        transform: translateX(0);
+        transform: translate3d(0,0,0); // Ensures that the list is on a new layer for better scrolling perf.
```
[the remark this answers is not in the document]
DylanPiercey: `will-change` would definitely better, just last I checked the browser support wasn’t there. Not sure if autoprefixer is going to automatically pick that one up either.

src/components/ebay-carousel/README.md
DylanPiercey: So if you want 1 item per slide you can use that option. In this PR on mobile discrete carousels will also work in native scroll mode in mobile. After the scroll has completed it will then animate to the closest slide.

Items never stay in an in between state. If that is a requirement then we can change that fairly easily but I think it works well currently.

src/components/ebay-carousel/style.less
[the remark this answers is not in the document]
senthilp: Looks like `will-change` works [everywhere](https://caniuse.com/#search=will-change) except Edge. But I am not sure if even the `translate3d` hack will trigger a new layer in Edge.

src/components/ebay-carousel/README.md
```diff
@@ -14,25 +14,17 @@ This component will bundle different resources depending on Lasso flags provided
...
 `accessibility-prev` | String | No | aria-label for previous control (default: "Previous Slide")
 `accessibility-next` | String | No | aria-label for next control (default: "Next Slide")
-
-### Additional Attributes for type="continuous"
-Name | Type | Stateful | Description
---- | --- | --- | ---
 `index` | String | Yes | 0-based index position
+`items-per-slide` | String | No | automatically fit a number of items for each carousel slide and enable slide controls.
```
yomed: Do we need a line for `gap` too here?

src/components/ebay-carousel/style-touch.less
```diff
@@ -9,22 +9,25 @@
...
     }
 
-    &--discrete {
-        // show controls on touch since there's no hover
-        // TODO: swipe support on mobile discrete?
-        .carousel__control {
-            opacity: @ebay-carousel-control-enabled-opacity;
-            pointer-events: auto;
+    // show controls on touch since there's no hover
+    // TODO: swipe support on mobile discrete?
```
yomed: Can probably remove this TODO now.

src/components/ebay-carousel/test/test.browser.js
```diff
@@ -3,10 +3,14 @@ const expect = require('chai').expect;
...
 const renderer = require('../');
-
-const privates = renderer.privates;
-const constants = privates.constants;
-const containerWidth = 800; // puppeteer default
+const styleOverrides = document.createElement('style');
+// No need to run the full length animation when testing the carousel.
+styleOverrides.innerHTML = `
+    .carousel__list {
+        transition: transform 0.1s linear;
```
yomed: Should we disable transition altogether here? Or if we do need to keep it, could we make it even shorter? By the way, maybe make this a one-liner since intent+comment is very clear.

src/components/ebay-carousel/test/test.browser.js
```diff
@@ -66,11 +70,11 @@ describe('given the carousel starts in the default state with items', () => {
             widget.on('carousel-move', moveSpy);
             widget.on('carousel-update', updateSpy);
             root.index = 1;
-            delay(done);
+            widget.subscribeTo(list).once('transitionend', done);
         });
 
         it('then it emits the marko move event', () => {
-            expect(moveSpy.calledOnce).to.equal(true);
+            // expect(moveSpy.calledOnce).to.equal(true);
```
yomed: Assuming we're cool with killing the `move` event in favor of update, should be ok to delete all the tests like this.

src/components/ebay-carousel/test/test.browser.js
```diff
@@ -80,13 +84,17 @@ describe('given the carousel starts in the default state with items', () => {
...
         });
 
         it('then it calculates item visibility correctly', () => {
-            expect(widget.firstVisibleIndex).to.equal(1);
-            expect(widget.lastVisibleIndex).to.equal(3);
+            const { state: { items } } = widget;
+            const visibleIndexes = [];
+            items.forEach((item, i) => {
+                if (item.fullyVisible) visibleIndexes.push(i);
+            });
```
yomed: Maybe bring this function to the top so it can be reused across the tests that need this.

src/components/ebay-carousel/README.md
DylanPiercey: Ah yes.

src/components/ebay-carousel/style-touch.less
DylanPiercey: 👍

src/components/ebay-carousel/test/test.browser.js
DylanPiercey: Not against making it a one liner, but then the linter would not be happy. Also the reason I don't have it set to be a 0s transition is because then the `transitionend` events won't fire.

yomed: Ah, right. Maybe we could shorten it to 0.05s or less? It's not super noticeable in this case, but does add up. (I guess it's more noticeable in the `dialog`).

src/components/ebay-carousel/test/test.browser.js
DylanPiercey: Right, will clean this up and also mention it in breaking changes.

src/components/ebay-carousel/test/test.browser.js
DylanPiercey: Yeah I can decrease it more. I plan on creating another PR that does the same thing for the dialog tests.

DylanPiercey:
Re testing the mobile functionality:

I could pull out the the `onScrollEnd` and `transition` scroll handlers and test those independently which would be easier. I do agree we need a better solution for this but I am not sure what it is at the moment.

yomed:
Sounds good, I think that testing setup should be fine for the interim.

src/components/ebay-carousel/utils/scroll-end/index.js
```diff
@@ -0,0 +1,77 @@
...
+ * (Only works in mobile and relies on touch events)
+ *
+ * @param {HTMLElement} el The element which scrolls.
+ * @param {(offset: number)=>{}} fn The function to call after scrolling completes.
+ * @return {function} A function to cancel the scroll listener.
+ */
+module.exports = function onScrollEnd(el, fn) {
+    let frame;
+    let stage = 0;
+    el.addEventListener('touchmove', handleTouchMove);
```
senthilp: Should we make these [passive](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners) events listeners, to ensure more smoothness?

src/components/ebay-carousel/template.marko
```diff
@@ -1,48 +1,55 @@
...
             aria-describedby=statusId
             aria-label=data.accessibilityNext
-            aria-disabled=String(data.nextControlDisabled)>
+            aria-disabled=(data.nextControlDisabled && 'true')>
             <ebay-icon type="inline" name="chevron-right"/>
         </button>
     </div>
-    <ul if(data.isDiscrete) class="carousel__dots" role="navigation">
-        <li for(i from 1 to data.totalSlides)>
+    <ul class="carousel__dots" role="navigation" if(data.totalSlides > 1)>
```
yomed: Is the pagination assumed to be present for the discrete case? I.e. will things break if `totalSlides === 1`? I wasn't sure if your use of `totalSlides >= 1` above versus `totalSlides > 1` was intentional.

src/components/ebay-carousel/test/test.browser.js
```diff
@@ -80,13 +85,14 @@ describe('given the carousel starts in the default state with items', () => {
         });
 
         it('then it applies a translation', () => {
-            const translation = mock.itemWidth + constants.margin;
-            expect(list.style.transform).to.equal(`translateX(-${translation}px)`);
+            const { offsetLeft } = list.children[1];
+            expect(list.style.transform).to.equal(`translate3d(-${offsetLeft}px, 0px, 0px)`);
```
yomed: The `translate3d` string is repeated a lot, maybe consider a helper?
```js
expect(getTranslation(list.style.transform)).to.equal(offsetLeft);
```

src/components/ebay-carousel/utils/scroll-end/index.js
```diff
@@ -0,0 +1,77 @@
...
+ * Calls a function every time scrolling completes for an element.
+ * Returns a function to stop listening for scroll ends events.
+ *
+ * (Only works in mobile and relies on touch events)
+ *
+ * @param {HTMLElement} el The element which scrolls.
+ * @param {(offset: number)=>{}} fn The function to call after scrolling completes.
+ * @return {function} A function to cancel the scroll listener.
+ */
+module.exports = function onScrollEnd(el, fn) {
```
yomed: (minor) You may want to name the file `on-scroll-end` also to clarify the usage, especially as opposed to `scroll-transition` which actually executes scrolling.

yarn.lock
```diff
@@ -108,48 +108,48 @@
     lasso-package-root "^1.0.1"
     resolve-from "^4.0.0"
 
-"@marko/create@^3.1.0":
-  version "3.1.0"
-  resolved "https://registry.npmjs.org/@marko/create/-/create-3.1.0.tgz#c621f9fa1cb8c07fd139eca7b6f322121d6613c4"
+"@marko/create@^3.1.6":
```
yomed: Was the yarn re-wrap intentional?

src/components/ebay-carousel/utils/scroll-end/index.js
DylanPiercey: 👍

src/components/ebay-carousel/utils/scroll-end/index.js
DylanPiercey: Should not be needed in this case as the event handler is removed immediately after the first touch move [here](https://github.com/eBay/ebayui-core/blob/6fcd95b6ea92c540f19e0eef8b0b9863bbf22f19/src/components/ebay-carousel/utils/scroll-end/index.js#L21). Then It waits for a touch end.

yarn.lock
DylanPiercey: Nope, thanks 😄

src/components/ebay-carousel/test/test.browser.js
DylanPiercey: 👍

DylanPiercey:
@yomed I think I've done everything you mentioned. Regarding the lock file, it actually needed to be updated for Marko-cli to pull in the right versions of Marko to work with ebay-ui. I did regenerate the whole thing since it was easier.

DylanPiercey:
Scratch that, it is a timeout issue.

DylanPiercey:
I discovered this morning that the preserving of the individual cards when rendering was only working in Marko 4, updating the PR with a patch.

DylanPiercey:
I fixed the issue above by moving the two properties stored on the widget instance (preserveItems and nativeScrolling) to an object on the state called config. This works because nested objects will not trigger a re-render when updated with Marko unless you explicitly `setStateDirty`. This works in both v3 and v4.

DylanPiercey:
Why did this get closed?

ianmcburnie:
Probably closed by accident. @seangates? I'll reopen.

seangates:
@DylanPiercey @ianmcburnie My fault. Complete accident.

DylanPiercey:
@seangates no problem 😄. Also just FYI for everyone I am using this branch to test the CI so please don't merge just yet.

yomed:
Somebody give this man a raise.

seangates:
Hear, hear!

senthilp:
Duly noted.

Issue #212, opened 2018-06-12
ebay-menu: does not close on pressing escape while focus on main button


abiyasa:
# Bug Report

## Description

**Actual**: The ebay-menu closes only when you press escape key while the focus is on the menu item. If the focus is on the main button, it does not close when you press escape key.

**Expected**: The menu should closes when you press escape key whether the focus is on the menu item or on the main button.

**How to reproduce**

- Click the main menu in order to open the menu
- Press Esc key
- The menu does not close

## Workaround

None

ianmcburnie:
FYI I know we do this on the skin ds6 site, but it's not strictly a requirement (see W3C design patterns and MIND patterns), more of a nice-to-have. The only requirement for button is ENTER and SPACEBAR. We will probably add ESC too though as it does feel useful and doesn't seem to cause any issues.

abiyasa:
Thanks! I didn't know the escape key behavior is different between menu & select. 
I agree that this should be nice-to-have thing instead of a bug.

PR #246, opened 2018-07-10
Menu: close the Menu when esc button is pressed.

ajay2507:
## Description
In order to handle the esc button on ebay-menu. ebay-menu uses ebay-button so I have added `handleEscapeKeydown` on `handleKeydown` to handle esc key.

## Context
The menu should closes when you press escape key whether the focus is on the menu item or on the main button. In order to acheive this,
On `handleEscapeKeydown` method, if the element has  `aria-expanded` attribute, then make it false to close the menu.

## References
https://github.com/eBay/ebayui-core/issues/212

## Screenshots
Before after pressing esc key, menu doesn't close.
ebay-menu-before
After making the change, menu closes when you press esc key
ebay-menu-after

### DIFF
--- a/src/components/ebay-button/index.js
+++ b/src/components/ebay-button/index.js
@@ -68,6 +68,13 @@ function handleKeydown(e) {
     eventUtils.handleActionKeydown(e, () => {
         this.handleClick();
     });
+    // handle escape key for ebay-menu
+    eventUtils.handleEscapeKeydown(e, () => {
+        this.el.focus();
+        if (this.el.hasAttribute("aria-expanded")) {
+            this.el.setAttribute("aria-expanded", "false");
+        }
+    });
 }
 
 module.exports = markoWidgets.defineComponent({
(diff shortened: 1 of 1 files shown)


yomed:
@abiyasa Thanks for the PR. The button is starting to become aware of variants like `expand` in #247, but we still don't handle actual expand/collapse logic in this component. Once #247 is in, we might want to consider introducing a new `button-escape` event for the `variant==="expand"` button. Then the `<ebay-menu>` could listen for that event, and handle the collapse logic itself. Roughly:
```js
if (isExpandVariant) {
    handleEscapeKeydown(() => {
        emitAndFire('button-escape');
    });
}
```

@eBay/ebayui Thoughts?

ianmcburnie:
As mentioned in #212, there is no hard requirement for `ESC` to close a menu when focus is on the button (as `SPACEBAR` and `ENTER` do this). So this "issue" does have low priority and we shouldn't force it in if we don't feel like it's the right time or right approach. I would rather wait and go with the approach suggested by @yomed .

ajay2507:
@yomed  @ianmcburnie okay makes sense. Will wait until #247 gets merged.

yomed:
Hi @ajay2507, are you still interested in working on this? You would use the `0.7.0` branch as a basis, and send the PR there. Let me know if you have any questions.

ianmcburnie:
The PR will probably be for 0.8.0 as we are hoping to release 0.7.0 today.

PR #247, opened 2018-07-10
Menu: add priority attribute for button

### LINKED ISSUE #219: Menu: Support "priority" attribute

yomed:
## Description

- add `variant` to `<ebay-button>` for supporting prefixed Skin classes
- add button tests
- add `priority` to `<ebay-menu>`
- add menu tests
- update docs
- fix demo page ordering for 10+ examples

## Context

Skin uses different classes for specific types of button. I.e. `.expand-btn .expand-btn--primary` should be used instead of `.btn .btn--primary`. Since the Menu uses the `expand-btn`, it's simpler to just have `<ebay-button>` support this Skin pattern itself. With that in place, the Menu can use `variant="expand"` and the classes for `size`/`priority` will be correct.

Now that the Button component has some knowledge of the `expand` variant, https://github.com/eBay/ebayui-core/pull/246 makes more sense contextually.

## References

Fixes #219 

## Screenshots

screen shot 2018-07-10 at 10 47 07 am


src/components/ebay-button/index.js
```diff
@@ -15,33 +15,35 @@ function getTemplateData(state, input) {
...
+        variant = 'fake';
+        tag = 'a';
+        model.href = href;
+    } else {
+        tag = 'button';
     }
 
-    if (size === 'small' || size === 'large') {
-        classes.push(`btn--${size}`);
+    if (href || variant === 'expand' || variant === 'cta') {
```
seangates: Why not have `variant === 'fake' || ...` and then you can remove the `variant = 'fake';` assignment you make?

seangates: Ah, because you might not have the variant with an `href` version. Gotcha. Never mind.

yomed: It's a little weird -- `fake` is a considered a variant in Skin terms, but we use `href` as a proxy for it here. I was hoping to avoid a user doing something like 
```marko
<ebay-button variant="fake"> <!-- invalid, no href -->
```

seangates: Yup. As a developer, I can see how that would confuse things.

ianmcburnie:
We could go with new components if that makes life simpler:

* `ebay-fake-button`
* `ebay-expand-button`

yomed:
For now, this change is the simplest option. Depending on how we decide to handle the CTA button in the future, we may want to revisit how to handle the expand button.

ianmcburnie:
Okay. Simple is good. :-)

PR #248, opened 2018-07-11
Market Sans for DS6 only;

seangates:
## Description
Modified `browser.json` for the demo so that DS4 doesn't pull in Market Sans.

## Context
DS4 wasn't originally designed to use Market Sans, and it isn't widely adopted by teams who are on DS4. Therefore, we should be testing the DS4 components without it.

### DIFF
--- a/demo/browser.json
+++ b/demo/browser.json
@@ -3,7 +3,6 @@
         "require: marko",
         "require: marko-widgets",
         "@ebay/skin/less",
-        "@ebay/skin/marketsans",
         "./atom-light-syntax.css",
         "./style.less",
         "../src/components/ebay-breadcrumb",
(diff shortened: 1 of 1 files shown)


demo/browser.json
```diff
@@ -17,6 +16,12 @@
         "../src/components/ebay-notice",
         "../src/components/ebay-pagination",
         "../src/components/ebay-select",
-        "../src/components/ebay-radio"
+        "../src/components/ebay-radio",
+        {
+            "if-flag": "ds6",
```
yomed: Wouldn't the flag be `skin-ds6`? Did you verify this locally?

ianmcburnie: Good catch Yoni. I think you're right but will let Sean verify.

seangates:
@yomed Fixed the flag. One last review?

PR #259, opened 2018-07-17
Implement infinite scrolling / autoplay carousel.

DylanPiercey:
## Description
Adds `autoplay` and related attributes for an autoplaying carousel.

In this PR there was also some issues resolved around canceling native scroll transitions and a rounding error when determining the slide sizes.

# Notes
* Currently the native scrolling is disabled for autoplay carousels (can still navigate via controls).
* New events and autoplay related attributes are mentioned in the updated README.

## References
Implements #216

## Screenshots
image


DylanPiercey:
Note I plan on adding more tests with this PR, will do so tomorrow morning.

src/components/ebay-carousel/README.md
```diff
@@ -20,15 +20,24 @@ Name | Type | Stateful | Description
...
+`items-per-slide` | String | No | automatically fit a number of items for each carousel slide and enable slide controls
+`gap` | String | No | override the margin between carousel items in pixels (default: "16")
 
 ### Additional Attributes for when items-per-slide is set.
 Name | Type | Stateful | Description
 --- | --- | --- | ---
 `accessibility-status` | String | No | status text (default: "Showing Slide {currentSlide} of {totalSlides} - Carousel")
 `accessibility-current` | String | No | pagination current slide text (default: "Current Slide {currentSlide} - Carousel")
 `accessibility-other` | String | No | pagination other slide text (default: "Slide {slide} - Carousel")
+`autoplay` | Boolean | No | adds autoplay controls to the carousel and automatically goes to the next slide every 4 seconds
```
yomed: Was there any need for config for the `interval` timing?

src/components/ebay-carousel/examples/10-autoplay/template.marko
```diff
@@ -0,0 +1,19 @@
+<style>
```
yomed: Regarding the numbers in the folders -- did you prefer this to sorting?
https://github.com/eBay/ebayui-core/blob/1343c7b2c895669d812694fce693fbc69a7eb842/demo/index.js#L51

src/components/ebay-carousel/examples/10-autoplay/template.marko
```diff
@@ -0,0 +1,19 @@
...
+    .demo7-card {
```
yomed: The demo class setup isn't great here to begin with, but you're probably redeclaring the `demo7-...` classes from above.

src/components/ebay-carousel/index.js
```diff
@@ -128,21 +151,26 @@ function onRender() {
...
-            const offset = getOffset(state);
+            const offset = hasOverride ? offsetOverride : getOffset(state);
+            // Listen for future scroll events and snap to the closest point.
+            this.cancelScrollHandler = onScrollEnd(listEl, handleScrollEnd.bind(this));
             // Animate to the new scrolling position and emit update events afterward.
             this.cancelScrollTransition = scrollTransition(listEl, offset, this.emitUpdate);
         }
 
+        if (autoplay && !paused) {
+            this.autoPlayTimeout = setTimeout(() => this.nextEl.click(), 4000);
```
yomed: (minor) `autoplay` is used as 1 word elsewhere

src/components/ebay-carousel/style-touch.less
```diff
@@ -19,6 +19,12 @@
         }
     }
 
+    &__autoplay {
+        .carousel__list {
```
yomed: (minor) Can leave out the nesting here

src/components/ebay-carousel/style.less
```diff
@@ -67,6 +69,35 @@
...
+        padding: 0;
+        background-color: rgba(0,0,0,0.45);
+        height: 40px;
+        width: 40px;
+        border-radius: 50%;
+        text-align: center;
+        color: @ds6-color-g201-gray;
+        box-sizing: border-box;
+
+        &&& svg {
```
yomed: I think you were aiming for 3 play and 3 pause, but it ends up like 
```css
.carousel__play.carousel__play.carousel__play svg, 
.carousel__play.carousel__play.carousel__pause svg, 
.carousel__play.carousel__pause.carousel__play svg, 
.carousel__play.carousel__pause.carousel__pause svg, 
.carousel__pause.carousel__play.carousel__play svg, 
.carousel__pause.carousel__play.carousel__pause svg, 
.carousel__pause.carousel__pause.carousel__play svg,
.carousel__pause.carousel__pause.carousel__pause svg
```

src/components/ebay-carousel/style.less
```diff
@@ -67,6 +69,35 @@
...
     }
 
+    &__play, &__pause {
+        z-index: 2;
+        bottom: 16px; // accounts for padding.
+        right: 0;
+        margin-right: 19px;
+        margin-bottom: 19px;
+        position: absolute;
+        background: 0;
```
yomed: Is this needed with `background-color` below?

src/components/ebay-carousel/README.md
DylanPiercey: I based the interval on the homepage interval. However it makes it easier to test to make this configurable so I updated the `autoplay` attribute to be either a Boolean or Number. If a number is set that becomes the interval.

src/components/ebay-carousel/examples/10-autoplay/template.marko
DylanPiercey: The ui-marketplace uses folder sorting order which is why I made this update.

src/components/ebay-carousel/index.js
DylanPiercey: Will switch to all being `autoplay`. 👍

src/components/ebay-carousel/style-touch.less
DylanPiercey: I need the nesting here though since `carousel__list` is a child of `carousel__autoplay` (which is on the root element).

src/components/ebay-carousel/style.less
DylanPiercey: This was just c&p'd from homepage. Will see if we actually need it.

src/components/ebay-carousel/style.less
DylanPiercey: The more `&`'s the more frustrated I am with css. (Will try to clean this up)

src/components/ebay-carousel/examples/10-autoplay/template.marko
DylanPiercey: Good catch thanks!

src/components/ebay-carousel/README.md
senthilp: We should not make the `interval` configurable, as that can lead to abuse. If 4 second is what was standardized then we will stick with it.

marko.json
```diff
@@ -37,12 +37,17 @@
         "@index": "string",
         "@type": "string",
         "@slide": "string",
+        "@autoplay": "boolean",
+        "@paused": "boolean",
```
ianmcburnie: Setting a `paused` property feels a little odd to me. I would have thought we'd use `play()` and `pause()` methods similar to the HTML video tag. I would imagine a `paused` attribute would be read-only (again, like the video tag) if we are trying to be HTML-like. @senthilp thoughts?

DylanPiercey: I took inspirited from the video tags `muted` property. We could expose methods for this as well if we want though.

DylanPiercey: Also I am exposing the stateful `paused` property on the element (which the video tag also does). However the video element exposes 10^10 ways of configuring it's state so 🤷‍♂️. Previously I think we decided against having methods for boolean toggles like this.

senthilp: Stateful `paused` makes sense, very similar to `video` tag's `paused`. This is sort of an in-between situation where there is no right or wrong.

ianmcburnie: @senthilp Video tag's `paused` is read-only.

ianmcburnie: ... and of course this is a carousel, not video, I know I know ;-) But trying to find some common ground.

ianmcburnie: I think `muted` is just an attribute for initial render, and not stateful:

`muted - Whether to mute the media resource by default`

https://www.w3.org/TR/html51/semantics-embedded-content.html#the-video-element

DylanPiercey: I am okay with either option. Making `paused` not stateful and switching to `play()` and `pause()` methods is also fine, could make the docs slightly more confusing though.

ianmcburnie: But `paused` will make no sense on 99% of our carousels, right? Only the ones with autoplay.

DylanPiercey: That is correct.

ianmcburnie: I would remove `paused` in order to fix this anomaly.

ianmcburnie: Actually, let me think on that last comment a bit more ;-)

ianmcburnie: My conclusion is... that they should have named it `automute` rather than `muted`. Lol.

I can see the attraction and simplicity of just a single stateful property to control playback (especially as we have the power to do that in our Marko layer on top of HTML), but if we are striving to be like the HTML API then we should keep `paused` as a readonly property, and add play() and pause() methods. This will be most familiar for users of the video tag. Not saying it's better, just more familiar ;-)

In all honesty I don't mind either way, but it feels like as we've been strict with the API up to now, we may as well continue on this path. I think after v1 we should asses all of the components again (properties vs methods) as a whole and see if it's still the right approach.

DylanPiercey: Are we fine with allowing `paused` to be set programmatically as attributes when using marko but just removing the property? Otherwise we are stuck with changing the state imperatively which is lame 😕.

src/components/ebay-carousel/README.md
DylanPiercey: @senthilp at first I agreed with you on this (which is why my original implementation hardcoded the 4s interval) however after giving it some thought I think this might not be the right approach in this case. The interval for the carousel needs to be finely tuned to the content that is in the carousel such that the user has enough time to properly consume that content. I imagine that could change enough to warrant having it be configurable.

marko.json
senthilp: @ianmcburnie I tried the `muted` attribute on YouTube `video` tag and it is stateful. I was able to mute and unmute by doing `$0.muted = true` and `$0.muted = false` 😄 .

src/components/ebay-carousel/README.md
senthilp: @DylanPiercey Agree.

marko.json
ianmcburnie: Hmm, that didn't work for me in Safari. Let me try again...

ianmcburnie: Adding and removing the attribute is not stateful though. Is that what you meant? Only setting it programmatically works for me (i.e. $0.muted = true and $0.muted = false).

senthilp: Yes, adding/removing does not work. Only setting it programmatically works.

src/components/ebay-carousel/index.js
```diff
@@ -205,12 +280,30 @@ function handleDotClick(originalEvent, target) {
...
+/**
+ * Toggles the play state of an autoplay carousel.
+ *
+ * @param {MouseEvent} originalEvent
+ */
+function togglePlay(originalEvent) {
+    const { state: { config, paused } } = this;
+    config.preserveItems = true;
+    this.setState('paused', !paused);
+    if (paused) this.nextEl.click();
```
yomed: Did you intend for the `carousel-next` event to be called by this (and the other time .click() is used)? I think we only wanted to call those events from human interaction, so that they can be used for tracking purposes.

src/components/ebay-carousel/style-touch.less
DylanPiercey: I misunderstood, added a commit to remove the unneeded nesting. Thanks!

src/components/ebay-icon/symbols/pause.marko
```diff
@@ -0,0 +1 @@
+<symbol id="icon-pause" viewBox="0 0 16 16"><g fill="currentColor" fill-rule="evenodd"><path d="M0 0h6v16H0zM10 0h6v16h-6z"></path></g></symbol>
```
yomed: Should these be added to the `ebay-icon/examples`?

src/components/ebay-carousel/index.js
```diff
@@ -181,12 +209,59 @@ function emitUpdate() {
...
+    let offsetOverride;
+
     config.preserveItems = true;
+    this.isMoving = true;
+
+    // When we are in autoplay mode we overshoot the desired index to land on a clone
+    // of one of the ends. Then after the transition is over we update to the proper position.
+    if (autoplayInterval) {
+        const LEFT = -1;
+        const RIGHT = 1;
```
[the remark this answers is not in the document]
DylanPiercey: Good call, will do.

src/components/ebay-icon/symbols/pause.marko
DylanPiercey: I'm not sure. They are not in skin yet so maybe not?

yomed: Ah ok, let's leave them out then.

src/components/ebay-carousel/index.js
DylanPiercey: In the code above it is triggered by clicking the play button and so I think it makes sense, for the actual autoplay part I am not sure, should we have a separate event? Should it not emit any special events? cc @ianmcburnie @senthilp?

marko.json
yomed: I think `paused` is fine as a property since it has a direct (and useful) mapping to the state. Previously we discussed using methods for either 1) dealing with non-primitive data types, or 2) acting on the component without a clear mapping to state.

src/components/ebay-carousel/index.js
ianmcburnie: I would prefer:

```js
if (paused) {
   this.gotoNextSlide(); // (or watever our method is called to move to next slide)
}
```

 i.e. call our public or internal API rather than simulate click.

yomed:
@DylanPiercey Not related to specific code, but I think you may need to watch out for autoplaying while resizing the browser.

src/components/ebay-carousel/index.js
DylanPiercey: Yeah I agree. Okay I'll make this change.

DylanPiercey:
I think I've addressed everything, let me know if I've missed anything 😄.

seangates:
@DylanPiercey Please fix the conflicts, then I'll look once more, and barring any crazy stuff I see, I'll approve and merge to get the pre-release out.

seangates:
@DylanPiercey Sweet. Looking now ...

DylanPiercey:
Let me try to fix the tests in the ci

src/components/ebay-carousel/index.js
```diff
@@ -80,7 +80,8 @@ function getTemplateData(state) {
 
         // Account for users providing a style string or object for each item.
         if (typeof style === 'string') {
-            item.style = `${style};flex-basis:${itemWidth};margin-right:${marginRight};transform:${transform}`;
+            item.style = `${style};flex-basis:${itemWidth};margin-right:${marginRight}`;
+            if (transform) item.style += `transform:${transform}`;
```
seangates: This passed linting?

DylanPiercey: 👍

Issue #47, opened 2018-03-15
Tests: Improve integration and v4 testing


yomed:
Porting over some notes from the old repo:
- Improve cleanup in failure case
- Use `npm pack`
- Verify no errors in browser console
- Move main functionality to standalone package
- Investigate improvements to Marko to allow multiple versions to be installed at the same time
- Improve local debugging of testing against v4

ianmcburnie:
Closing, as I'm not sure how relevant this still is and it's been on the backlog a long time.

PR #260, opened 2018-07-17
Fix not accounting for multiple src -> dist replacements.

DylanPiercey:
## Description
Currently the build script will only replace the first instance of `src/components` found in each `browser.json`. `ebay-menu` recently got updated to have two and this broke the build.

## Context
This caused the marketplace build to fail (and broke the `ebay-menu` in general on npm). In this PR all instances are replaced.

Note: We should merge/publish this soon as currently the 0.7.0-0 prerelease is broken.


yomed:
Just as a note, we should probably use npm-pack in our integration tests to catch these types of issues #47

PR #265, opened 2018-07-18
Update Marko CLI

DylanPiercey:
## Description
I released a new version of Marko cli so we can stop relying on the canary release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "lasso-marko": "^2.4.2",
     "lintspaces-cli": "^0.6.0",
     "marko": "^3",
-    "marko-cli": "3.2.0-alpha.86e5e3fd",
+    "marko-cli": "^3.2.0",
     "marko-dynamic-tag": "^1.0.0",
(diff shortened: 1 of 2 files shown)


ianmcburnie:
@seangates any reason to keep this back till 0.8.0 given that it's dev env related?

ianmcburnie:
Actually, it does seem to update babel runtime in yarn.lock, if that's something that has an impact on our actual components then maybe we do leave it for now. @DylanPiercey?

DylanPiercey:
@ianmcburnie the yarn.lock can never break the published components, only our builds / dev environments. Also in this case the the `@babel/runtime` is additive. There still exists the old `babel-runtime` in our lock.

DylanPiercey:
I mean to say it can't cause a regression to anything except our dev environment (and the build is passing). The lock file only locks down the files for us during dev. When people install our modules they get the latest tree which has nothing to do with our lock file.

ianmcburnie:
Oh, duh, it's under 'devDependencies' - silly me!

PR #266, opened 2018-07-18
Remove dialog placeholder element.

### LINKED ISSUE #262: ebayui-dialog class attribute
# Bug Report

## Description
When i set a `class` for `<ebayui-dialog>` i expect this class to be rendered on the root element of ebayui-dialog, so i can control the style (if needed) from my component. (using `> .dialog` etc...). Right there is another element rendered between my class dialog and my component (.dialog-placeholder).

## Workaround
The current workaround is to couple my styles with the `.dialog-placeholder` that comes from ebayui-dialog. And i don't like coupling.

image

DylanPiercey:
## Description
This removes the extra `.dialog-placeholder` element for the `ebay-dialog` component.

## Context
This placeholder was added back when the plan was to implement the dialog using a portal to the document body. As I have mentioned before I think It makes the most sense to hold off on that implementation (also it's working fairly well as is) until we upgrade to Marko v4. On top of that when we upgrade to Marko v4 it no longer requires that I have this placeholder anyway.

## References
Fixes #262

### DIFF
--- a/src/components/ebay-dialog/template.marko
+++ b/src/components/ebay-dialog/template.marko
@@ -1,12 +1,10 @@
 
-<div class="dialog-placeholder" w-bind>
-    <div w-id="dialog" class=data.dialogClass hidden=!data.open role="dialog" w-preserve-attrs="hidden" ${data.htmlAttributes} w-onclick="handleDialogClick">
-        <div w-id="window" class=data.windowClass role="document">
-            <button w-id="close" class="dialog__close" type="button" aria-label=data.ariaLabelClose w-onclick="handleCloseButtonClick">
-                <ebay-icon type="inline" name="close"/>
-            </button>
+<div w-bind class=data.dialogClass hidden=!data.open role="dialog" w-preserve-attrs="hidden" ${data.htmlAttributes} w-onclick="handleDialogClick">
+    <div w-id="window" class=data.windowClass role="document">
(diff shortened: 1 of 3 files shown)


ianmcburnie:
Testing has already started on 0.7.0, so this might have to wait until 0.8.0.

src/components/ebay-dialog/template.marko
```diff
@@ -1,12 +1,10 @@
 
-<div class="dialog-placeholder" w-bind>
-    <div w-id="dialog" class=data.dialogClass hidden=!data.open role="dialog" w-preserve-attrs="hidden" ${data.htmlAttributes} w-onclick="handleDialogClick">
-        <div w-id="window" class=data.windowClass role="document">
-            <button w-id="close" class="dialog__close" type="button" aria-label=data.ariaLabelClose w-onclick="handleCloseButtonClick">
-                <ebay-icon type="inline" name="close"/>
-            </button>
+<div w-bind class=data.dialogClass hidden=!data.open role="dialog" w-preserve-attrs="hidden" ${data.htmlAttributes} w-onclick="handleDialogClick">
```
PrashantAshok: 👍 . The additional wrapper was making me add `inline-block` to prevent wrapping issues.

ianmcburnie:
Will merge into 0.8.0 branch when ready.

PR #267, opened 2018-07-18
CI: fix on scroll end grouped

DylanPiercey:
## Description
Ensures that the timeouts are setup before simulated scroll transitions run.

## Context
This caused the CI to fail intermittently because previously the simulated scroll animation frame could (if the browser is running slowly) finish before the `setTimeout` of `0`.

Moving the transition to be after the timeout ensures that the timeout is always called first.


seangates:
But don't merge until you're merging to 0.8.0, please. 😁

ianmcburnie:
@seangates This feels safe for 0.7.0 though, right?

seangates:
@ianmcburnie Is this simply to fix the build, or is it something that Hemnath needs?

seangates:
Sorry, I was just trying to protect the testing efforts. Carry on.

ianmcburnie:
I don't believe this has any impact on testing efforts, so I'm moving to 0.7.0 and merging.

PR #269, opened 2018-07-18
Demo: use 2-digit numbers for example folders

### LINKED ISSUE #261: Demo: examples can show out of order
# Bug Report

## Description

We should update component `example` folders to use 2 digit numbering like `01 02 03` rather than `1 2 3`. In the latter case, 10 is considered before 1 in folder ordering. This is only strictly needed once a component hits 10 examples, but we might as well update them all for consistency and future-proofing. We should also remove the current `sort` used in `demo/index.js`.

yomed:
## Description

Changes demo `examples/` to use `01 02 03` for numbering in order to have consistent folder ordering.

## Context

Note: I changed all components except for `carousel`, since #259 takes care of that.

## References

Fixes #261


ianmcburnie:
I guess this is safe to go into 0.7.0 too as it's just related to the demos.

PR #272, opened 2018-07-18
Adding <span> to match Skin updates;

seangates:
## Description
Added the empty span to use for the icon.

## Context
The current Skin implementation uses a `::after` pseudo-element, but that implementation had limitations which were undesirable. Skin has been updated to use a `<span>` with a background SVG, or `<svg>` with a foreground SVG. This allows for much better control, and in the case of the foreground SVG, full customization of the icon.

## References
Related to https://github.com/eBay/skin/issues/222

### DIFF
--- a/src/components/ebay-select/template.marko
+++ b/src/components/ebay-select/template.marko
@@ -12,4 +12,5 @@
             ${option.label}
         </option>
     </select>
+    <span class="select__icon"/>
 </span>


ianmcburnie:
@seangates Can this change go safely in now before we upgrade to skin v6?

seangates:
No, this will break stuff.

seangates:
Well ... wait ... let me test that notion.

seangates:
Testing in IE now ...

seangates:
Okay, I'm eating my earlier words. It appears to not have any affect on the component. I don't think this will matter, even if it does go in before 6.0.0.

ianmcburnie:
I guess it's just weird to have that <span> lying around and people might ask about it. Maybe we'll just hold it back until we start using skin v6 pre-release.

seangates:
Sure, I just didn't want to forget about this span being needed. I mean, I don't even mind closing this PR.

sendlo:
Looks good. But per Ian's comment, I guess we should leave this open until we're ready to cut the pre release.

seangates:
Pre-release of Skin with ebay-select is out, though. So we need it now.
Also, the small markup won’t hurt anything.

Otherwise, Hemnath won’t be able to test until ebayui 1.0.0.

PR #273, opened 2018-07-19
Menu: support icon in label and no toggle

### LINKED ISSUE #217: Menu: label can't be an icon; chevron toggle can't be hidden
# Feature Request

## Description

Currently, the ebay-menu does not let the developer customize the icon that is to the right of the label.  With the introduction of ebay-icon, I would like to be able to choose contextually relevant icons for the dropdown menu open/closed states.

yomed:
## Description

- adds `icon` attribute 
- adds `no-toggle-icon` attribute 
- adds `accessibility-text` attribute for both icon and no-label cases
- removes `marko-dynamic-tag` in favor of official Marko inclusion (which works for our case since our example components are very simple)
- fix demo ordering for menu

## Context

These fulfill the use case outlined in #217.

This uses a transformer for some magic in taking a dynamic icon name and converting it into an actual icon. This ran into some issues in Marko v4 which does not supported preserved transcluded content at the top level, but we have some quirky workarounds thanks to @DylanPiercey.

Note: Even with this in, we still need some minor fixes from Skin to be fully operational -- https://github.com/eBay/skin/issues/261

## References

Fixes #217

## Screenshots

screen shot 2018-07-18 at 6 32 54 pm
screen shot 2018-07-18 at 6 32 59 pm
screen shot 2018-07-18 at 6 33 04 pm


src/components/ebay-menu/test/test.server.js
```diff
@@ -107,7 +108,19 @@ describe('menu', () => {
...
         const $ = testUtils.getCheerio(context.render(input));
         expect($(labelSelector).length).to.equal(0);
-        expect($('.expand-btn__icon').length).to.equal(1);
+        expect($('svg.expand-btn__icon').length).to.equal(1);
+    });
+
+    test('renders with icon', context => {
+        const input = { icon: 'settings', iconTag: { renderBody: mock.iconRenderBody } };
+        const $ = testUtils.getCheerio(context.render(input));
+        expect($('div.expand-btn__icon').html()).to.equal('icon');
```
DylanPiercey: This is breaking the CI since marko v4 is outputting some placeholder comments here. I think a good fix would be to change to `.text()` instead of `.html()`?

yomed: Oops, thanks

seangates:
Could we split the demo changes from the component changes, and merge the demo changes first? Doesn't seem like it would affect existing demos to merge those changes first, then do the component changes as a second PR, in isolation.

yomed:
I'd prefer to keep these together, since this feature necessitates the demo changes.

seangates:
@yomed I'll look at this again once the conflicts are fixed. 👍

yomed:
@DylanPiercey @seangates Conflicts resolved.

seangates:
Just ran into [this](https://github.com/eBay/skin/issues/261) (Skin issue 261), even without clicking on anything:

**Before** applying `flex-shrink: 0;':
image

**After** applying `flex-shrink: 0;':
image

yomed:
@seangates Correct, they're related parts of that flexbox issue (`shrink` issues with more content, and `align-self` issues when changing the border).

seangates:
We also need one of these:

```
.expand-btn__icon:only-child {
    margin: 0;
}
```

seangates:
Yeah, I'm just documenting them. I (likely) need to move them over to the other issue.

PR #274, opened 2018-07-19
Textbox component

dasarianudeep:
## Description
Develop reusable component for input textbox , texarea

## Screenshots
ebayui-textbox


src/components/ebay-textbox/index.js
```diff
@@ -0,0 +1,31 @@
+const markoWidgets = require('marko-widgets');
+const processHtmlAttributes = require('../../common/html-attributes');
+const template = require('./template.marko');
+
+function getInitialState(input) {
+    const htmlAttributes = input.invalid ?
+        Object.assign({}, processHtmlAttributes(input), { 'aria-invalid': 'true' }) : processHtmlAttributes(input);
```
yomed: IMO The user's `HtmlAttributes` should be left alone and sent to the template from `processHtmlAttributes(input)`. You can position your own HTML attributes (`aria-invalid`) before theirs, so that ours take precedence. I imagine it would be very uncommon to have the user attempt to set this, but if are worried about that, we would need to handle it across all components.

src/components/ebay-textbox/index.js
```diff
@@ -0,0 +1,31 @@
...
+    let classes = { 'textbox__control': true, 'textbox__control--fluid': input.fluid };
+    if (input.class) {
+        classes = Object.assign({ }, classes, { [input.class]: true });
```
yomed: I think it's a little simpler and cheaper to use an array for this like 
```js
classes = ['...', input.class];
if (fluid) {
    classes.push('fluid');
}
```

marko.json
```diff
@@ -182,5 +182,17 @@
...
     },
-    "<ebay-select-option>": {}
+    "<ebay-select-option>": {},
+    "<ebay-textbox": {
+        "renderer": "./src/components/ebay-textbox/index.js",
+        "@*": "expression",
+        "@html-attributes": "expression",
+        "@value": "string",
+        "@class": "string",
+        "@placeholder": "string",
```
yomed: Similar for this one, it seems like it would just be a pass-through. You should still show these attributes in examples since they are required for most usage, but we don't need to handle it through Marko attributes.

src/components/ebay-textbox/test/test.server.js
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable no-unused-expressions */
```
yomed: Is this because of `to.be.true`? I would suggest removing this, and instead be consistent in your `expect` usage below. Use `expect(a).to.equal(b)`, rather than `to.eql()` or `.to.be`

src/components/ebay-textbox/test/test.server.js
```diff
@@ -0,0 +1,59 @@
...
+    expect(Boolean($(inputSelector).is(':disabled'))).to.be.true;
+});
+
+test('renders a input textbox with invalid/error state', context => {
+    const input = { invalid: true };
+    const $ = testUtils.getCheerio(context.render(input));
+    expect(Boolean($(inputSelector).prop('aria-invalid'))).to.be.true;
+});
+
+test('renders input textbox with a custom class', context => {
```
yomed: Remove this test, and instead use the `testCustomClass` util. Put it alongside the `testHtmlAttributes` test at the bottom. You'll see its usage in other component tests.

src/components/ebay-textbox/test/test.server.js
```diff
@@ -0,0 +1,59 @@
...
+const inputSelector = `input.textbox__control`;
+const textareaSelector = `textarea.textbox__control`;
+
+test('renders default input textbox', context => {
+    const $ = testUtils.getCheerio(context.render());
+    expect($(rootSelector).length).to.eql(1);
+    expect($(inputSelector).length).to.eql(1);
+});
+
+test('renders input textbox with some value', context => {
```
yomed: Remove, since this will be covered by `testHtmlAttributes` after the `marko.json` changes.

src/components/ebay-textbox/test/test.server.js
```diff
@@ -0,0 +1,59 @@
...
+    const input = { value: 'this is texbox' };
+    const $ = testUtils.getCheerio(context.render(input));
+    expect($(inputSelector).val()).to.equal('this is texbox');
+});
+
+test('renders input textbox with a placeholder', context => {
```
yomed: Remove, since this will be covered by `testHtmlAttributes` after the `marko.json` changes.

ianmcburnie:
This will go into 0.8.0 branch when ready.

src/components/ebay-textbox/index.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+    return {
+        htmlAttributes,
+        classes,
+        value: input.value,
+        tag: input.fluid ? 'div' : 'span',
+        textBoxTag: Boolean(input.multiline) ? 'textarea' : 'input',
+        disabled: Boolean(input.disabled),
+        placeholder: input.placeholder
```
PrashantAshok: can't `placeholder` be part of `htmlAttributes`?

dasarianudeep: makes sense

ianmcburnie:
Moving this back to 1.0.0 to keep 0.8.0 focussed on HP tasks.

src/components/ebay-textbox/index.js
seangates: @dasarianudeep I believe you should address this. Let us know if you have questions.

src/components/ebay-textbox/index.js
dasarianudeep: may not be

src/components/ebay-textbox/README.md
```diff
@@ -0,0 +1,44 @@
+# ebay-textbox
+
+## ebay-textbox Usage
+
+Default input textbox:
+
+```marko
+<ebay-textbox />
```
yomed: Keep this most basic example in the `README`, but remove the other ones. We want to avoid repeating documentation that is already covered in the `Attributes` section below, and the `examples/` directory.

src/components/ebay-textbox/README.md
```diff
@@ -0,0 +1,44 @@
...
+```marko
+<ebay-textbox multiline />
+```
+
+## ebay-textbox Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`class` | String | No | define custom class
+`placeholder` | String | No |
```
yomed: Remove `placeholder` here -- it's an HTML attribute.

src/components/ebay-textbox/index.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+    return {
+        classes,
+        htmlAttributes: processHtmlAttributes(input),
+        rootClass: input.class || '',
+        value: input.value,
+        tag: input.fluid ? 'div' : 'span',
+        textBoxTag: Boolean(input.multiline) ? 'textarea' : 'input',
+        disabled: Boolean(input.disabled),
+        placeholder: input.placeholder,
```
yomed: You don't need to process `placeholder` here or in the template. This will get handled through `htmlAttributes`.

src/components/ebay-textbox/test/test.server.js
```diff
@@ -0,0 +1,50 @@
...
+const inputSelector = `input.textbox__control`;
+const textareaSelector = `textarea.textbox__control`;
+
+test('renders default input textbox', context => {
+    const $ = testUtils.getCheerio(context.render());
+    expect($(rootSelector).length).to.equal(1);
+    expect($(inputSelector).length).to.equal(1);
+});
+
+test('renders input textbox with a placeholder', context => {
```
yomed: Can remove this test, as it's handled by `testHtmlAttributes`.

demo/browser.json
```diff
@@ -17,6 +17,7 @@
         "../src/components/ebay-pagination",
         "../src/components/ebay-select",
         "../src/components/ebay-radio",
+        "../src/components/ebay-textbox",
```
yomed: Please also add `ebay-textbox` (alphabetically) to the root `README.md`
https://github.com/eBay/ebayui-core#components

src/components/ebay-textbox/README.md
```diff
@@ -0,0 +1,44 @@
...
+`fluid` | Boolean | No | defaults to false
+`disabled` | Boolean | Yes | defaults to false
```
yomed: All of our boolean attributes default to `false`, so no need to mention. Can describe the behavior of the `true` case for these instead.

src/components/ebay-textbox/examples/7-multiline/template.marko
```diff
@@ -0,0 +1 @@
+<ebay-textbox multiline />
```
yomed: Please name the example folders with a leading `0`, as in `01 02 03`. Also, fix the numbering so `06` isn't skipped.

src/components/ebay-textbox/index.js
```diff
@@ -0,0 +1,31 @@
...
+function getInitialState(input) {
+    const classes = ['textbox__control'];
+    if (input.fluid) {
+        classes.push('textbox__control--fluid');
+    }
+    return {
+        classes,
+        htmlAttributes: processHtmlAttributes(input),
+        rootClass: input.class || '',
+        value: input.value,
```
yomed: `value` is an HTML attribute, so we don't need to process it separately here (or in the template).

src/components/ebay-textbox/index.js
```diff
@@ -0,0 +1,31 @@
...
+        tag: input.fluid ? 'div' : 'span',
+        textBoxTag: Boolean(input.multiline) ? 'textarea' : 'input',
```
yomed: We're treating `textbox` as one word, so just try to stay consistent, as in `textboxTag` rather than `textBoxTag`.

src/components/ebay-textbox/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+function getInitialState(input) {
+    const classes = ['textbox__control'];
+    if (input.fluid) {
+        classes.push('textbox__control--fluid');
+    }
+    return {
+        classes,
+        htmlAttributes: processHtmlAttributes(input),
+        rootClass: input.class || '',
```
yomed: For this, we usually use the pattern `['textboxt', input.class]`. No need to worry about falsey values, since Marko will handle that.

marko.json
```diff
@@ -183,5 +183,15 @@
...
     },
-    "<ebay-select-option>": {}
+    "<ebay-select-option>": {},
+    "<ebay-textbox>": {
+        "renderer": "./src/components/ebay-textbox/index.js",
+        "@*": "expression",
+        "@html-attributes": "expression",
+        "@class": "string",
+        "@invalid": "boolean",
+        "@disabled": "boolean",
```
DylanPiercey: Should we make disabled a pass through html attribute and remove it from here and in the component/template?

yomed: 👍 Let's do that for `disabled` here too. In other cases, we use `disabled` for internal behavior which is why it is an explicit attribute, but we don't have that yet for this component.

demo/browser.json
dasarianudeep: Done

src/components/ebay-textbox/test/test.server.js
```diff
@@ -0,0 +1,32 @@
...
+    const $ = testUtils.getCheerio(context.render(input));
+    expect(Boolean($(inputSelector).prop('aria-invalid'))).to.equal(true);
+});
+
+test('renders a textarea element', context => {
+    const input = { multiline: true };
+    const $ = testUtils.getCheerio(context.render(input));
+    expect($(textareaSelector).length).to.equal(1);
+});
+
```
yomed: Would be great if you could also cover the `fluid` use case so that this code is covered: https://coveralls.io/builds/18237755/source?filename=src%2Fcomponents%2Febay-textbox%2Findex.js#L8

src/components/ebay-textbox/README.md
```diff
@@ -0,0 +1,19 @@
...
+<ebay-textbox />
+```
+
+## ebay-textbox Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`class` | String | No | define custom class
+`fluid` | Boolean | No |
+`disabled` | Boolean | Yes |
```
yomed: Can remove `disabled` attribute from here.

src/components/ebay-textbox/test/test.server.js
dasarianudeep: Done

dasarianudeep:
@ianmcburnie @yomed Thanks Guys. Appreciate your timely help and feedback.

PR #281, opened 2018-07-21
Expand ebay-icon capabilities to stamp custom inline icons

dsathyakumar:
## Description
Expand `ebay-icon` capabilities to stamp custom inline icons.

## Context
- Currently `<ebay-icon/>` stamps inline SVG icons that are available in `/symbols` folder within it
- More often, there are repeated custom icons given, which currently cannot utilize the one time stamping inside `<body>`, as done by `<ebay-icon/>`.
- Now, adding a custom icon is as simple as
>`<ebay-icon type="inline" name="my-awesome-icon" custom="true"/>` 

(this causes the transformer to look within the project root for a `/symbols/my-awesome-icon.marko`)

## Screenshots
image

**PS**: I'm a newbie to this project 🙂, pls excuse if the PR doesn't meet your standards!


src/components/ebay-icon/index.js
```diff
@@ -26,6 +26,8 @@ function init() {
 }
 
 function getTemplateData(state, input, out) {
+    const custom = input.custom || 'false';
+    const isCustom = custom === 'true';
```
senthilp: Looks like `custom` is a boolean attribute. As per our [guidelines](https://medium.com/@senthil_hi/building-a-ui-component-in-2017-and-beyond-1f6d5c4d464#f7e3), if an attribute is boolean, just the presence of the attribute on an element represents the true value. Can you please update it accordingly? This should also be reflected in docs.

src/components/ebay-icon/template.marko
```diff
@@ -6,6 +6,11 @@
     <svg w-bind ${data.htmlAttributes} ${data.accessibilityAttributes} focusable="false" class=data.classes>
         <defs w-id="defs" w-body if(data.renderDefs)/>
         <title if(data.accessibilityText) id=data.titleId>${data.accessibilityText}</title>
-        <use xlink:href="#icon-${data.name}"/>
+        <if(data.isCustom)>
+            <use iscustom=data.isCustom xlink:href="#${data.name}"/>
+        </if>
+        <else>
+            <use iscustom=data.isCustom xlink:href="#icon-${data.name}"/>
```
senthilp: Do you still need `iscustom=data.isCustom` in the else block?

yomed:
The stamping functionality is definitely useful outside of `<ebay-icon>`, but I think that while we're in the official `<ebay-icon>` component, we should limit ourselves to the design-sanctioned list of icons.

If we wanted to share the stamping, I think we would need to abstract out the stamping to its own utility/component.

dsathyakumar:
🤔 seems strange that we would want to limit the list of icons & have functionality in two places.

seangates:
@dsathyakumar It does not seem strange when we want to limit the icons to a very specific set.

That said, and to re-iterate what @yomed said, we should abstract the stamping capability. Then `<ebay-icon>` could stamp a _limited set of icons_, and any other component could stamp their own set (or not limit the icon set, whichever they prefer).

seangates:
Created issue #285 and closing.

PR #284, opened 2018-07-24
Carousel: Patches from the autoplay branch.

DylanPiercey:
## Description
There were three small patches I made in the autoplay branch which it turns out are causing issues for the myebay team. This PR extracts those fixes so that we can release them more quickly.

## Fixes
* Rounding issue that could happen due to floating point calculations sometimes making it so you could not go to the very end.
* The wrong offset was used for item widths which could cause issues when rerendering/resizing.


DylanPiercey:
It was decided to wait for 0.8.

Issue #303, opened 2018-08-02
Button: does not respect "type" attribute in input


CestDiego:
Use Case:
```
<ebay-button type="submit"/>
```

Expected behavior

```
<button type="submit/>
```

Actual behavior

```
<button type="button"/>
```

This seems to be a limitation of Marko3

PR #304, opened 2018-08-02
Button: allow type override

CestDiego:
## Description
Add ability to override button type

## Context
Solves https://github.com/eBay/ebayui-core/issues/303


ianmcburnie:
This can go into 0.8.0 release today, bypassing QA cycle if we deem it safe enough.

mlrawlings:
You will need to add the `type` attribute here as well: [marko.json#L18-L30](https://github.com/CestDiego/ebayui-core/blob/320009380c0237f48d1801e52df7db56e532159d/marko.json#L18-L30)

Otherwise it will come through under `htmlAttributes`

src/components/ebay-button/template.marko
```diff
@@ -3,7 +3,7 @@
     w-body
     w-onclick="handleClick"
     w-onkeydown="handleKeydown"
-    type="button"
+    type=(input.type || "button")
```
mlrawlings: Use `data` instead of `input` since we're in Marko 3

mlrawlings:
Since this attribute is being handled specially, it probably makes sense to add a quick test for it as well

CestDiego:
Addressed comments here: https://github.com/cestdiego/ebayui-core/commit/613d73b

CestDiego:
```

{ Error: An unknown server-side error occurred while processing the command.
    at url("") - driver.js:200:141
  details: undefined,
  message: 'unknown error: DevToolsActivePort file doesn\'t exist\n  (Driver info: chromedriver=2.40.565383 (76257d1ab79276b2d53ee976b2c3e3b9f335cde7),platform=Linux 4.14.12-041412-generic x86_64)',
  type: 'RuntimeError',
  seleniumStack: 
   { type: 'UnknownError',
     message: 'An unknown server-side error occurred while processing the command.',
     orgStatusMessage: 'unknown error: DevToolsActivePort file doesn\'t exist\n  (Driver info: chromedriver=2.40.565383 (76257d1ab79276b2d53ee976b2c3e3b9f335cde7),platform=Linux 4.14.12-041412-generic x86_64)' } }
```

not sure why this error is happening to me.

src/components/ebay-button/index.js
```diff
@@ -47,6 +47,7 @@ function getTemplateData(state, input) {
     }
 
     model.tag = tag;
+    model.type = input.type;
```
yomed: We keep most of the logic in the renderer, so let's do 
`model.type = input.type || 'button'`

src/components/ebay-button/test/test.server.js
```diff
@@ -26,6 +26,18 @@ Object.keys(properties).forEach(property => {
     });
 });
 
+test('renders button with overriden type', context => {
```
yomed: typo -> `overridden`

src/components/ebay-button/test/test.server.js
```diff
@@ -26,6 +26,18 @@ Object.keys(properties).forEach(property => {
...
+    const input = { type: 'submit' };
+    const $ = testUtils.getCheerio(context.render(input));
+    expect($(`button.btn`).attr('type')).to.equal('submit');
+});
+
+test('renders button with default type', context => {
```
yomed: Maybe swap the order of these 2 tests. I.e. first test for default type, then test for overridden type.

marko.json
```diff
@@ -21,6 +21,7 @@
         "@html-attributes": "expression",
         "@class": "string",
         "@disabled": "boolean",
+        "@type": "string",
```
yomed: Move this below `class`, since it kind of belongs with the html attributes.

yomed:
@ianmcburnie I noticed that in Skin DS4 the button examples use `type=button`, while DS6 does not have a `type` at all. Should we just avoid adding a default `type`? Or should we default to `type=submit` since that shouldn't have an impact outside of forms?

CestDiego:
@yomed I resolved comments here: https://github.com/cestdiego/ebayui-core/commit/3aa5f90

marko.json
CestDiego: https://github.com/cestdiego/ebayui-core/commit/3aa5f90

src/components/ebay-button/index.js
CestDiego: https://github.com/cestdiego/ebayui-core/commit/3aa5f90

src/components/ebay-button/test/test.server.js
CestDiego: https://github.com/cestdiego/ebayui-core/commit/3aa5f90

src/components/ebay-button/test/test.server.js
CestDiego: https://github.com/cestdiego/ebayui-core/commit/3aa5f90

src/components/ebay-button/test/test.server.js
```diff
@@ -26,6 +26,18 @@ Object.keys(properties).forEach(property => {
     });
 });
 
+test('renders button with default type', context => {
+    const input = { };
+    const $ = testUtils.getCheerio(context.render(input));
+    expect($(`button.btn`).attr('type')).to.equal('button');
+});
+
+test('renders button with overriden type', context => {
```
sidsakhadeo: Can there be an additional test for `reset` or does that seem redundant since the test already checks for `submit`? 
Following up on that, since there are only 3 valid types `button`, `submit` and `reset` should there be a check for this in the renderer?

yomed: Good point, but I think we're treating this as roughly an HTML attribute pass-through, and we don't normally validate those. I.e. you can do `aria-checked="maybe"` and we won't stop you.

marko.json
```diff
@@ -34,8 +35,8 @@
         "@*": "expression",
         "@html-attributes": "expression",
         "@class": "string",
-        "@index": "string",
         "@type": "string",
+        "@index": "string",
```
yomed: Wrong component 😆

yomed:
I'm inclined to go ahead with this PR, which keeps `button` as the default `type`. Although we are differing from HTML which defaults to `submit`, I think `button` is most common use case now.

DylanPiercey:
Why don’t we just not have a default and allow the browser to assign the default (submit). This way we don’t confuse devs who are expecting out button to work like a normal button.

senthilp:
I prefer to keep it `button`, as it applies to the general eBay use case. For the default (submit), we mostly go with `<input type="submit">`

yomed:
I could go either way, too. But I think we should keep it as-is, at least for now. Removing the current default `type=button` is theoretically a semver breaking change.

yomed:
@CestDiego I'm not sure why the build is failing, but I wonder if it could be related to the PR branch being `master`.

yomed:
Running the builds on the same commit hash via #312 seems to have everything passing.

Issue #258, opened 2018-07-16
New Component: Tabs


ianmcburnie:
Placeholder. More details to follow...

User story: Marko implementation of Skin's [tab component](https://ebay.github.io/skin/#tab)

Mind Pattern: https://ebay.gitbooks.io/mindpatterns/content/disclosure/tabs.html

Keyboard Behaviour Module: https://github.com/makeup-js/makeup-roving-tabindex

ianmcburnie:
Incase you miss my update above, we should try and leverage the [roving-tab-index](https://github.com/makeup-js/makeup-roving-tabindex) module again for the keyboard behaviour.

ianmcburnie:
Moving this to 0.9.0.

PR #307, opened 2018-08-02
Tab Component

yomed:
## Description

- adds initial tab component
- has minor changes to common utils and test utils

## Context

This component follows a similar pattern to the carousel in terms of selection. A heading (and a panel, in case of non-fake use case) is selected via the `index` property on the root element. I started out using a boolean attribute on the heading itself, like `<ebay-tab-heading selected/>`, but ran into lots of issues. A single heading must be selected at all times, so the normalization gets very messy in the boolean case.

## References

#258 

## Screenshots

screen shot 2018-08-02 at 4 48 49 pm
screen shot 2018-08-02 at 4 48 55 pm
screen shot 2018-08-02 at 4 48 59 pm
screen shot 2018-08-02 at 4 49 03 pm


marko.json
```diff
@@ -192,5 +192,27 @@
             "@selected": "boolean"
         }
     },
-    "<ebay-select-option>": {}
+    "<ebay-select-option>": {},
+    "<ebay-tab>": {
+        "renderer": "./src/components/ebay-tab/index.js",
+        "transformer": "./src/common/transformer/index.js",
+        "@*": "expression",
+        "@html-attributes": "expression",
```
senthilp: Do we really need `@html-attributes`? `@*` should be sufficient right? This is a general comment, not specific to this component.

yomed: This was introduced to support dynamic pass-through attributes. More details here: https://github.com/eBay/ebayui-core/pull/154

src/components/ebay-tab/template.marko
```diff
@@ -0,0 +1,58 @@
...
+                </div>
+            </div>
+        </div>
+    </if>
+    <else>
+        <div w-id="items" class="tabs__items" role="tablist">
+            <for(i, item in data.items)>
+                <div
+                    aria-controls="${widget.id}-tabpanel-${i}"
+                    id="${widget.id}-default-tab-${i}"
```
sendlo: Minor - why are they all called "default-tab" (plus id and item number)?  I would see this and think whatever tab I'm looking at is the default.

src/components/ebay-tab/index.js
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/**
+ * Common processing of index change via both UI and API
+ * @param {Number} index
+ */
+function processStateChange(index) {
+    if (index >= 0 && index < this.state.items.length && index !== this.state.index) {
+        this.setState('index', index);
+        emitAndFire(this, 'tab-select', { index });
```
sendlo: documentation says it emits the original event too

src/components/ebay-tab/template.marko
yomed: Thanks, I'll update. This was from the Skin example.

src/components/ebay-tab/index.js
sendlo: To clarify:
https://github.com/eBay/ebayui-core/tree/258-tab-component/src/components/ebay-tab#ebay-tab-events

It looks like the documentation says the original event is passed as the data.  I think it makes sense to pass the original event and the element, similar to what we do in pagination.

yomed: We don't currently send the `originalEvent` for cases that can be triggered both programmatically and by UI events. So I'm going to leave it out of this component until we determine the general pattern.

senthilp: Agree. For now, we will skip sending `originalEvent ` for programmatically triggered events.

DylanPiercey:
I think everything looks good here. My only reservation is the name `ebay-tab-item` is a little generic. I'd personally prefer something like `ebay-tab-heading`. If every one is happy with the current naming though then that's fine with me 😄.

sendlo:
I like Dylan's recommendation of changing it to ebay-tab-heading too.  (I actually like ebay-tab-title.) Tab item could be confused with the content.

Issue #200, opened 2018-05-24
Radio: clicking corner generates an event with value: "undefined" - option is not selected


hramakumar:
## eBayUI Version: 0.5.0-0

when you click near the borders of the radio button. the click does not work.

![image](https://user-images.githubusercontent.com/13157115/40517486-68eee97e-5f6a-11e8-9b5a-ea9ebb22e21c.png)

video: https://www.useloom.com/share/0af190c2fde1434dbac2515ce5de26ad

ianmcburnie:
This should be resolve when we update to latest skin patch release.

hramakumar:
@ianmcburnie I already verified this issue while testing 0.5.0-2
This can be closed

ianmcburnie:
Thanks!

PR #133, opened 2018-05-02
Checkbox Component

PrashantAshok:
## Description
ebay-checkbox component which renders a checkbox

## References

Closes #121 
https://dsl.ebay.com/ds6/forms

PR #310, opened 2018-08-03
Switch Component

### LINKED ISSUE #280: New Component: Switch
This will be available for DS4 and DS6 in upcoming Skin v6 release. Should be a simple component to squeeze into v1.0.0.

Sneak peak:

screen shot 2018-07-20 at 2 48 27 pm

New markup:

```html
<span class="switch">
    <input role="switch" type="checkbox" />
    <span class="switch__inner"></span>
</span>
```

yomed:
## Description

- adds new switch component

## Context

Overall, code is very similar to the `checkbox` component.
Currently only supports DS4.

## References

Closes #280 

## Screenshots

screen shot 2018-08-03 at 4 23 35 pm

### DIFF
--- a/demo/browser.json
+++ b/demo/browser.json
@@ -17,7 +17,7 @@
         "../src/components/ebay-pagination",
         "../src/components/ebay-radio",
         "../src/components/ebay-select",
-        "../src/components/ebay-switch"
+        "../src/components/ebay-switch",
         {
             "if-flag": "skin-ds6",
             "dependencies": [
--- a/demo/browser.json
+++ b/demo/browser.json
@@ -15,8 +15,9 @@
         "../src/components/ebay-menu",
         "../src/components/ebay-notice",
         "../src/components/ebay-pagination",
-        "../src/components/ebay-select",
         "../src/components/ebay-radio",
+        "../src/components/ebay-select",
+        "../src/components/ebay-switch"
         {
             "if-flag": "skin-ds6",
             "dependencies": [
--- a/marko.json
+++ b/marko.json
@@ -192,5 +192,11 @@
             "@selected": "boolean"
         }
     },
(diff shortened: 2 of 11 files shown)


senthilp:
LGTM

yomed:
@DylanPiercey I agree that the event name seems off (was just trying to stay consistent with the others). 
@PrashantAshok Do you recall why those names were chosen? https://github.com/eBay/ebayui-core/pull/133#issuecomment-390764890

yomed:
Created follow-up issue #320 for the event naming.

PrashantAshok:
More details [here](https://github.com/eBay/ebayui-core/pull/134#issuecomment-390281822)

DylanPiercey:
@PrashantAshok regarding the linked conversation: How does `radio-change` conflict with the native `change`? That should have been fine?

sendlo:
It looks like @ianmcburnie specifically wanted to differentiate it from the dom event.  And then they landed on all the events across the components to be "x-select".

I think we have two options for all the "select" events on our components.  We either mimic the dom event names, meaning all events will get different names based on its HTML counterpart.  Or we keep a consistent naming convention like "x-select" across our components.

Unless anyone has a really strong feeling about this, I think keeping them as "x-select" is fine.

yomed:
I think `x-select` is fine for components where the behavior is only selection (breadcrumb, tab, pagination). But for components where the event is used for both selection and deselection (radio, checkbox, switch), we should use something else, like `x-change`.

PrashantAshok:
@yomed , @hramakumar - sorry for commenting on a closed PR, but switch (which radio/checkbox has) also has the spacing/`undefined` issue. [fyi](https://github.com/eBay/ebayui-core/issues/200). This would a fix in Skin for switch.

PR #313, opened 2018-08-06
Carousel: Improve test coverage

### LINKED ISSUE #309: Carousel: code coverage is low
## Description
Latest code coverage for the carousel component is pretty low (79%) and that's dragging overall coverage low.  
Looks like adding tests for handleScrollEnd will probably bring it up to standards.

DylanPiercey:
## Description
This PR adds additional tests for the mobile scrolling version of the carousel and the two utilities `on-scroll-end` and `scroll-transition` associated with mobile scrolling.

While adding these tests I noticed that in the scroll end case the `carousel-update` event was not firing and that fix is included in this PR.

## References
Resolves #309

### DIFF
--- a/src/common/test-utils/browser.js
+++ b/src/common/test-utils/browser.js
@@ -22,34 +22,49 @@ function testOriginalEvent(spy) {
 }
 
 /**
- * Simulates a touch based scroll event over 100ms.
+ * Simulates a touch based scroll event over 4 animation frames.
  *
  * @param {HTMLElement} el The element to scroll.
  * @param {number} to The new scrollLeft for the element.
+ * @param {function} cb A callback to call after the scroll.
  */
-function simulateScroll(el, to) {
+function simulateScroll(el, to, cb) {
     const { scrollLeft } = el;
     const distance = to - scrollLeft;
-    const duration = 100;
+    const frames = 4;
+    let frame = 0;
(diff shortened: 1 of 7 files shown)


src/common/test-utils/browser.js
```diff
@@ -21,7 +21,35 @@ function testOriginalEvent(spy) {
...
+/**
+ * Simulates a touch based scroll event over 100ms.
+ *
+ * @param {HTMLElement} el The element to scroll.
+ * @param {number} to The new scrollLeft for the element.
+ */
+function simulateScroll(el, to) {
+    const { scrollLeft } = el;
+    const distance = to - scrollLeft;
+    const duration = 100;
```
yomed: Was there a specific reason you changed this to 100 from the 10 it was previously? It looks like the failing tests might need higher timeouts to compensate.

DylanPiercey: Yeah I needed a higher duration to be able to cancel the scroll effectively. I'll mess around with the timings.

DylanPiercey:
Added another commit to this PR which improves the resiliency of some of the existing carousel related tests by no longer using arbitrary timeouts but hard coded animation frame counts which are reliable. @yomed I also changed the test scroll transition utility to always do the transition over 4 frames instead of 100ms so that its both faster (usually) and can more precisely tested.

PR #314, opened 2018-08-06
Combobox: move htmlAttributes to input tag

### LINKED ISSUE #288: Combobox: needs accessibility support
# Bug Report

## Description

ebay-combobox doesn't provide a way to specify accessibility text for voice-over use case.

I'm following https://www.coolfields.co.uk/2011/04/accessible-forms-should-every-input-have-a-label/ to provide accessibility to the dropdown. The <input> tag needs either an id or a title attribute. 
It would be nice to have a way to pass in a string as generated <input> tag's title attribute, or a way to specify id of the generated <input> tag.

## Workaround

No
## Screenshots

N/A

yomed:
## Description

- move `htmlAttributes` from root tag to input tag of combobox
- update README.md
- update tests

## Context

This is a typical pattern we use for components with a prominent `input` tag. Just to note, this is a breaking change, as we don't know if anyone was relying on its placement in the root tag.

## References

Fixes #288

### DIFF
--- a/src/components/ebay-combobox/README.md
+++ b/src/components/ebay-combobox/README.md
@@ -35,6 +35,8 @@ Name | Required | Type | Stateful | Description
 `selected` | n/a | Number | Yes | allows you to set the selected index option to `selected`
 `borderless` | No | Boolean | No | whether button has borders
 
+Note: For this component, `class` is applied to the root tag, while all other HTML attributes are applied to the `input` tag.
+
 ### ebay-combobox Events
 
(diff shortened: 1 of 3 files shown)


sendlo:
I'm thinking hold off for 1.0.0 pre-release due to breaking changes.  Any objections?

PrashantAshok:
imo, breaking changes pre 1.0.0 should be ok. would be good to get this in, since htmlAttrs now has the right placement.

PR #318, opened 2018-08-07
Breadcrumb: Fixing accessibility to match Skin changes;

seangates:
## Description
- Changes the `<span>` to `<a>` with no href
- The role in hijax scenarios now stays an anchor since it is not actionable
- The leaf `<a>` with no href now has `aria-current="location"`

## Context
This change more closely aligns with a11y guidelines, and keeps all the leaf and branch nodes the same element type, allowing focus, as well as providing the proper `aria-current`.

## References
Fixes https://github.com/eBay/skin/issues/296

## Screenshots

**Before**
screen shot 2018-08-06 at 4 52 33 pm

**After**
image

P.S.: @yomed, I actually believe the last two tests are not correct, although they pass. Should they have `li a:not([href])` as the selector?

### DIFF
--- a/src/components/ebay-breadcrumb/index.js
+++ b/src/components/ebay-breadcrumb/index.js
@@ -8,18 +8,16 @@ function getTemplateData(state, input) {
     const inputItems = input.items || [];
     const hijax = input.hijax || false;
     const items = inputItems.map((item, index) => {
-        let tag = 'a';
         let ariaCurrent = null;
         let role;
         const href = item.href || null;
         const current = (inputItems.length - 1 === index);
         let shouldHandleClick = true;
         if (current && !href) {
-            tag = 'span';
-            ariaCurrent = 'page';
+            ariaCurrent = 'location';
             shouldHandleClick = false;
         }
-        if (hijax) {
+        if (hijax && href) {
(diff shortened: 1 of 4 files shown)


yomed:
Do I need to install some version of Skin v6 to see this locally?

seangates:
Yes. Skin@6.0.0-1.

seangates:
But, this could wait until eBayUI 1.0.0. I just have nowhere to point the change in the meantime, except to 0.9.0 branch.

seangates:
Rebased. Please review once more.

msendlakowski:
@seangates According to the breadcrumb pattern, the last crumb should be a span:
https://ebay.gitbooks.io/mindpatterns/content/navigation/breadcrumbs.html

Can't we just style the span to match the anchors?

msendlakowski:
Oh weird.  I've always thought anchors without hrefs are bad for accessibility.  I'll definitely have to ask him about that.  Thanks.

seangates:
@msendlakowski Yeah, it was new to me, too. But apparently without the `href` they simply become a `span` and are treated as inline text that can be styled. 🤷‍♂️

Issue #317, opened 2018-08-07
pagination in ie11 throws an error


martin-ebay:
# Bug Report

## Description
trying to use the pagination buttons in ie11 throws an error.

## Workaround
The fix is to fix the conditional in emitAndFire.js as it is checking

```
if (window.CustomEvent) {
   event = new CustomEvent(.....
```
the window.CustomEvent object exist in ie11 but ie11 does not support `new CustomEvent`.

yomed:
Better feature check: https://github.com/Modernizr/Modernizr/blob/5eea7e2a213edc9e83a47b6414d0250468d83471/feature-detects/customevent.js#L23

yomed:
Thanks for the bug report and PR -- Fix is out in `0.9.0`

PR #319, opened 2018-08-07
Fix CustomEvent usage in IE11

martin-ebay:
## Description

small try-catch wrap in emitAndFire for if statement that evals true in ie11 even though the contents of the if statement do not work in ie11

## Context
to make pagination work in ie11 as the conditional check results in a false positive

## References
https://github.com/eBay/ebayui-core/issues/317

### DIFF
--- a/src/common/emit-and-fire/index.js
+++ b/src/common/emit-and-fire/index.js
@@ -7,7 +7,7 @@
 function emitAndFire(widget, eventName, eventArg) {
     const originalEmit = widget.emit;
     let event;
-    if (window.CustomEvent) {
+    if ('CustomEvent' in window && typeof window.CustomEvent === 'function') {
         event = new CustomEvent(eventName, { detail: eventArg });
     } else {
         event = document.createEvent('CustomEvent');


src/common/emit-and-fire/index.js
```diff
@@ -7,8 +7,14 @@
 function emitAndFire(widget, eventName, eventArg) {
     const originalEmit = widget.emit;
     let event;
+    //if window.CustomEvent evals true on ie11 but does not support new CustomEvent
```
seangates: I know it's small, but could you move this comment above the try statement? It seems more relevant the closer it is to the code you wrote.

src/common/emit-and-fire/index.js
```diff
@@ -7,8 +7,14 @@
...
     if (window.CustomEvent) {
-        event = new CustomEvent(eventName, { detail: eventArg });
+        try {
```
senthilp: Since we are already checking `window.CustomEvent` why do we need a `try/catch`.

src/common/emit-and-fire/index.js
```diff
@@ -7,8 +7,14 @@
...
+            event = new CustomEvent(eventName, { detail: eventArg });
+        } catch(e) {
+            event = document.createEvent('CustomEvent');
```
senthilp: These lines are repeated
```
event = document.createEvent('CustomEvent');
event.initCustomEvent(eventName, true, true, eventArg);
```

martin-ebay:
reworked to use the test modernizer uses and removed duplicate code.

yomed:
Thanks @martin-ebay, code looks good. Would you mind actually making the PR from a branch that isn't named `master` in your fork? Our CI has some issues at the moment related to that.

martin-ebay:
@yomed Just created a new branch with the fix and a new PR https://github.com/eBay/ebayui-core/pull/323 since I can't change the branch on this PR.

Closing this PR

yomed:
Thanks, it turns out that maybe it's an issue from all forks? (cc @DylanPiercey)
I'm going to re-open this, get the build running from an upstream branch, and we should be set.

PR #321, opened 2018-08-07
Carousel: Allow uneven slides when `items-per-slide` is set.

### LINKED ISSUE #316: Carousel: Items are not visible when items count is less than items-per-slide
# Bug Report

## Description
Items are not visible when the items count is less than `items-per-slide`;

```marko
<ebay-carousel type="discrete" items-per-slide="5" gap="0">
    <ebay-carousel-item for(item in data.items)>
		content
    </ebay-carousel-item>
</ebay-carousel>
```

If `data.items` length is less than 5(which is `items-per-slide`), the component would not displaying any items in `carousel`.

https://github.com/eBay/ebayui-core/blob/d2d28a6d9d35f205a8a2d53987db9dd91289ea66/src/components/ebay-carousel/index.js#L40

If items length is 4 and `items-per-slide` is 5, above logic sets the total item count is 0 could be the possible reason.  

## Workaround
Have conditional `item-per-slide` and overwrite style to set the `width` of each item.

```marko
<ebay-carousel type="discrete" items-per-slide=(data.items.length > 5 ? 5 :data.items.length ) gap="0">
```

DylanPiercey:
## Description
Currently the carousel automatically truncates non packed slides from the end when `items-per-slide` is set. In this PR that no longer happens and partial slides are allowed. There was also a check which was off by one which I have fixed.

## References
Fixes #316

### DIFF
--- a/src/components/ebay-carousel/index.js
+++ b/src/components/ebay-carousel/index.js
@@ -36,8 +36,6 @@ function getInitialState(input) {
     const { items, itemsPerSlide } = state;
     if (itemsPerSlide) {
         state.classes.push('carousel--slides');
-        // Remove any extra items when using explicit itemsPerSlide.
-        items.length -= items.length % itemsPerSlide;
         // Only allow autoplay option for discrete carousels.
         if (input.autoplay) {
             const isSingleSlide = items.length <= itemsPerSlide;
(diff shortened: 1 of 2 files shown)


DylanPiercey:
Ah yes lame. I thought this was going to be an easy one but there is something strange going on. I'll take a look and try to gauge if this would be a quick fix.

DylanPiercey:
Updated the PR. I believe it is working in all cases now. The issue was that the `getNextIndex` was going further than a slide (its strategy is to find the next hidden item and move a slide width in the requested direction) which was working correctly except that we automatically round the index to the nearest slide when `items-per-slide` is set. I updated the PR to avoid going a slide left in the items per slide case because it'll automatically rounded to the appropriate slide. This ensures that the items on each discrete slide stay the same, which we need for the slide controls and is not the case for the continuous carousel which greedily moves as much as possible in the requested direction.

DylanPiercey:
Thanks for adding the test @yomed 😄

PR #328, opened 2018-08-08
Process style attribute explicitly for pass-through

### LINKED ISSUE #132: Allow style pass through with marko helper
From original PR #102:

# Description
Add style as a pass through 

## Context

Given that marko treats `class` and `style` differently. We should allow the style to also be explicitly defined in the template, this way we don't run into trouble when using the ebay-button inside a nested component.

```
// this is button wrapper file
<ebay-button ...input />
```

```
// this is the user of the  button-wrapper
$ const style = { color: red }
<button-wrapper fluid style=style > some text <button-wrapper />
```

This doesn't add `style` to the `input['*']` object. And if we otherwise did:
```
// this is button wrapper file
<ebay-button style=input.style />
```

What we get is:

image

yomed:
## Description

- added `style` as an explicit attribute to every tag
- updated renderers/templates to process `style`
- update testUtils
- update docs around pass-through of class/style vs other attributes
- minor cleanup and reordering

## Context

Currently, `style` gets processed as a normal HTML attribute. However, Marko has special treatment for `class` and `style` to allow object/array input rather than just string. To allow users to use the Marko helpers, we have to process it explicitly.

## References

Fixes #132

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -5,12 +5,15 @@
         "@*": "expression",
         "@html-attributes": "expression",
         "@class": "string",
+        "@style": "string",
         "@heading-text": "string",
         "@heading-level": "string",
         "@hijax": "boolean",
         "@items <item>[]": {
             "@*": "expression",
             "@html-attributes": "expression",
+            "@class": "string",
+            "@style": "string",
             "@href": "string"
         }
     },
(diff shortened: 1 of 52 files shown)


sendlo:
@yomed  these changes look backwards compatible and pretty safe.  Do you feel these can go in the 0.9.0 release since we have to cut a second pre-release with other bugfixes?

PR #332, opened 2018-08-09
rebase with master to pick up Dylan's travis changes

sendlo:
## Description
Rebase with master

## Context
So this can be tested along with the 0.9.0 pre release before its final.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+sudo: required
+addons:
+  chrome: stable
+
 language: node_js
 node_js:
   - "8"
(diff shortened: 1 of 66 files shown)


yomed:
Maybe simpler to leave this one in master and pick it up later. It's has no external impact.

sendlo:
I thought about it, but that means it doesn't get any testing before its released.  I agree it shouldn't effect anything external, but I'm playing it safe this way.

yomed:
It won't get testing anyway. There's nothing to test beyond what Dylan already did in creating the PR from his fork.

sendlo:
Yeah, I understand.  I didn't mean that it needs to be tested, but I don't think its good practice to publish a module that hasn't been signed off by QE as a whole unified build.  Even if he won't see any difference with or without Dylan's change.

PR #333, opened 2018-08-10
Menu: use onRender for update via parent state

### LINKED ISSUE #327: Menu: Emits incorrect element on client side rerender
# Bug Report

## Description
**Repro Steps**
1. First time server render, `event.el` emitted has the right `el` that was clicked. 
2. On subsequent client side re-render, `event.el` emitted has the incorrect (older) `el` that was clicked. 

Looks like `this.itemsEls` needs to be recalculated on re-render (onRender) as well. A quick test seems to fix the issue. Relevant code [link](https://github.com/eBay/ebayui-core/blob/79841fb66900ad1e87848d9acdfcd1f4b3cea12b/src/components/ebay-menu/index.js#L143)

## Workaround
None

yomed:
## Description

- replace `init` with `onRender`
- add test case for menu items update via parent state
- minor updates to related code

## Context

We currently do client initialization via `init`, and set vars that we don't expect to change, like `this.itemEls`. However, when updating the items via parent state, `init()` does not get called again, which causes `itemEls` to be out of date, resulting in bad data in the events. I've fixed this by using `onRender` instead, with an internal code split depending on `firstRender`.

I added a `TODO` for an issue that's a related bug, but less important for now.

## References

Fixes #327

### DIFF
--- a/src/components/ebay-carousel/test/test.browser.js
+++ b/src/components/ebay-carousel/test/test.browser.js
@@ -99,14 +99,13 @@ describe('given the carousel starts in the default state with items', () => {
         });
     });
 
-    // simulate parent state change
-    describe('when index is updated via state', () => {
+    describe('when index is updated via parent state', () => {
         let updateSpy;
         beforeEach(done => {
             updateSpy = sinon.spy();
             widget.on('carousel-update', updateSpy);
             widget.subscribeTo(list).once('transitionend', done);
-            widget.setState('index', 1);
+            widget.setProps({ index: '1', items: mock.sixItems });
         });
 
         it('then it emits the marko update event', () => {
--- a/src/components/ebay-menu/index.js
+++ b/src/components/ebay-menu/index.js
@@ -160,22 +160,24 @@ function onRender(event) {
             });
         }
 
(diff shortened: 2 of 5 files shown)


sendlo:
@DylanPiercey this looks good to me, but can you give your blessing as the resident Marko expert.  No `init` function, using `onRender` instead.

sendlo:
@yomed   the Expander and rovingTabIndex will be recreated each render.  Will that duplicate the events they create?  Or will it have any other side effects like memory leaks?

DylanPiercey:
Just looking into @sendlo's concern and I think it is valid. AFACT this would cause a memory leak. IMO this should be addressed in the `makeup-x` repo's which should expose some cleanup events.

yomed:
Yeah, I'll move `expander` to just the first render. But `rovingTabIndex` would need to be recreated for rerenders where items are updated via parent state. Would be nice if it could be handled without having to keep track of what kind of render we're doing.

yomed:
I can keep them both inside `firstRender` for now, and put `FIXME`s for later. It isn't any worse off than before.

PR #341, opened 2018-08-13
Carousel: Peek functionality

DylanPiercey:
## Description
This PR updates the `items-per-slide` option of the carousel to accept fractions to display part of the next slide. IE: `items-per-slide="2.5"` shows two items and half of the third item.

Note: this also automatically hide's the `dot` controls if we are doing partial slides as per the design system.

## Context
Originally I was going to create a separate `peek` attribute, however I think just using item's per slide is intuitive enough. Also you can only have a `peek` style carousel if `items-per-slide` is set, so this saves us some checks.

## Screenshots

image

image

### DIFF
--- a/src/components/ebay-carousel/README.md
+++ b/src/components/ebay-carousel/README.md
@@ -66,3 +66,4 @@ Notes:
 
 * The `carousel` will manipulate the `tabindex` property of nested focusable elements inside `<ebay-carousel-item>`.
 * The `autoplay` carousel currently does not support native scrolling and will use transforms instead.
+* The `items-per-slide` attribute can be set to a float such as `3.5` to show 3 items, and half of the 4th item. This also automatically enables the `no-dots` attribute.
--- a/src/components/ebay-carousel/examples/11-items-per-slide-partial-item/template.marko
+++ b/src/components/ebay-carousel/examples/11-items-per-slide-partial-item/template.marko
@@ -1,0 +1,25 @@
+<style>
+    .demo11-card {
+        color: #cdf4fd;
+        background: #a1208b;
+        font-size: 24px;
+        font-weight: bold;
+        height: 150px;
+        line-height: 150px;
+        text-align: center;
+    }
+</style>
+<ebay-carousel items-per-slide="3.25">
+    <ebay-carousel-item class="demo11-card">Card 1</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 2</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 3</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 4</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 5</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 6</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 7</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 8</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 9</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 10</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 11</ebay-carousel-item>
+    <ebay-carousel-item class="demo11-card">Card 12</ebay-carousel-item>
+</ebay-carousel>
--- a/src/components/ebay-carousel/index.js
+++ b/src/components/ebay-carousel/index.js
@@ -20,7 +20,7 @@ function getInitialState(input) {
         gap: input.gap || 16,
         noDots: input.noDots,
         index: parseInt(input.index, 10) || 0,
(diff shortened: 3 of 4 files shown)


src/components/ebay-carousel/README.md
```diff
@@ -61,3 +61,4 @@ Notes:
 
 * The `carousel` will manipulate the `tabindex` property of nested focusable elements inside `<ebay-carousel-item>`.
 * The `autoplay` carousel currently does not support native scrolling and will use transforms instead.
+* The `items-per-slide` attribute can be set to a float such as `3.5` to show 3 items, and half of the 4th item.
```
yomed: Should probably document that using a float will disable dots.

src/components/ebay-carousel/examples/11-items-per-slide-partial-item/template.marko
```diff
@@ -0,0 +1,25 @@
+<style>
+    .demo8-card {
```
yomed: `demo11`

msendlakowski:
@DylanPiercey  Is this needed?  There is no peek requirement for the "items-per-slide" carousel in the invision doc I've seen.
https://ebay.invisionapp.com/share/CWN9X1JF6MA#/screens/275500542

And the home page is only showing this when a specific "items per slide" is not set.  And that's already supported implicitly by setting the appropriate item widths.

We can check with Joseph, but I don't think we want a peek on this type of carousel.

DylanPiercey:
@msendlakowski this is to achieve the `Scaling Carousel` effect in that document. If you look at the home page as an example you can see that it works basically exactly the same as an `items-per-slide` carousel in terms of scaling with a fixed peak percentage and finally some breakpoints to change that. The only difference (besides peaking) between the peak and `items-per-slide` carousel is that the dot navigation controls do not show.

DylanPiercey:
@msendlakowski your right. The problem arises when you have margins. Then these calculations get more complicated which is part of the benefit of us taking care of it for them.

You can see the calculation we are currently doing [here](
https://github.com/eBay/ebayui-core/pull/341/files#diff-c3d5dce966b3694ee609f5d9eabf542fR79).

msendlakowski:
Its a good solution, but it forces applications to add logic in their renderers to determine "items-per-slide" at various break points, if they want the "peek".   And since "items-per-slide" isn't stateful they can't recalculate it if the window size changes.

If they use the CSS-only solution they can put this logic behind media queries and keep their JS cleaner. 

If the rest of the team thinks this is the best solution I'm OK with it, since it doesn't exclude the CSS-only solution.  But I agree with Yoni about the docs.  We should make it very clear that "items-per-slide" no longer means that the paginator dots will always show.  In fact, can we add a section in the docs explaining both ways to create the "peek", in case someone doesn't want to control it via JS?

DylanPiercey:
Technically anything that we deem as not stateful can still easily be updated by the consuming Marko components just not through the `el.itemsPerSlide` style API. Also as far as using a pure CSS solution goes we have a few options:

1) Users have to manually do all of these calculations at different break points in their own css to match the itemsperslide and peak they need.
2) We provide a less mixin which does this which they have to use on top of using the component.
3) We can hard code in a set number of `itemsPerSlide` and `peaks` combinations with a set gap.

The first two aren't really viable.
The last one is how we originally had the carousel set up however a user wanted the ability to change the carousel gap. Since the width calculations for the items are based on the gap it would be impossible (or at least size inefficient) to pre-generate all possible valid combinations of `items-per-slide`, `peak` and `gap`.

I have tried (and continue to explore) a css solution where the gap is not needed for the item width calculations which would make this simpler but I have not found one.

FTR after talking with Tom he mentioned that the homepage carousel also updates it's breakpoints with JS instead of CSS.

msendlakowski:
Really?  I see the home page using this for creating the peek in their carousel.
```
@media screen and (min-width: 960px) {
    .hl-standard-carousel__cell {
        -ms-flex:0 0 18.18181818%;
        flex: 0 0 18.18181818%;
        max-width: 18.18181818%;
        width: 18.18181818%
    }
}
```
Are they changing their design?  I just saw your message on slack.  We can continue there.

msendlakowski:
OK.  Talked with Dylan.  We're going to keep this.  Though teams could still use the CSS-only method if they don't need pixel perfect control over the peek.

src/components/ebay-carousel/examples/11-items-per-slide-partial-item/template.marko
```diff
@@ -0,0 +1,25 @@
...
+        background: #a1208b;
+        font-size: 24px;
+        font-weight: bold;
+        height: 150px;
+        line-height: 150px;
+        text-align: center;
+    }
+</style>
+<ebay-carousel items-per-slide="3.25">
+    <ebay-carousel-item class="demo8-card">Card 1</ebay-carousel-item>
```
yomed: Although these technically work as-is, they should also be `demo11`. One of these days we'll come up with a better way to handle these classes.

DylanPiercey: Whoops, will fix.

DylanPiercey:
Rebased with 1.0.0 😄

PR #349, opened 2018-08-14
Icon: Use relative path for icon symbol requires.

DylanPiercey:
## Description
Currently the ebay icon automatically includes the absolute path of the symbol to use at compile time in order to ensure that only the used symbols are sent to the browser. Lasso also creates a comment for the actual require path for every require. This is fine, however the resource server is generating the hash for these files based off of their unminified source which includes the absolute path comment. The issue occurs when there are multiple machines with different file paths and since the resource server is using the source with these comments to generate the hash, two or more bundles are created.

## Context
This changes the ebay-icon to always use a relative path to the `symbol` file which won't cause different comments to be created on different machines.


msendlakowski:
@yomed can I get your blessing on this and I'll get this out in a patch release

PR #350, opened 2018-08-14
Demo: update layout

yomed:
## Description

- move navigation to left side
- add design system toggle

## Context

The component list is getting long, so easier to navigate from the left. The toggle should help with local debugging now that we need to go back and forth between ds4/ds6 frequently.

## Screenshots

screen shot 2018-08-14 at 3 19 12 pm
screen shot 2018-08-14 at 3 19 09 pm


yomed:
Oh damn, ok let me fix that.

PR #353, opened 2018-08-15
Menu: collapse from button escape

### LINKED ISSUE #212: ebay-menu: does not close on pressing escape while focus on main button

yomed:
## Description

- create `button-escape` event
- use `button-escape` to collapse menu
- update docs
- add tests

## Context

A more generic event would be `button-keydown`, but we're already firing `button-click` when action buttons are pressed. So, I created `button-escape` instead.

## References

Fixes #212 
Follow up to #246

### DIFF
--- a/src/components/ebay-button/README.md
+++ b/src/components/ebay-button/README.md
@@ -22,4 +22,5 @@ Name | Type | Stateful | Description
 
 Event | Data | Description
 --- | --- | ---
-`button-click` | `{ originalEvent }` | click
+`button-click` | `{ originalEvent }` | click or action key pressed (space and enter)
+`button-escape` | `{ originalEvent }` | escape key pressed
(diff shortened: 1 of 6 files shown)


DylanPiercey:
I feel like maybe we don't need or shouldn't have the `button-escape` event. From the users perspective the important event is `menu-collapse` right? I'd personally vote against adding hooks like this unless specifically requested.

msendlakowski:
@DylanPiercey I was thinking the same thing, but otherwise the menu component has to have knowledge of the structure of the button component.  I think that's worse.

msendlakowski:
@DylanPiercey are you OK with the fix as is? Do you have a better suggestion for handling the escape event on the menu button?

PR #356, opened 2018-08-16
Menu: change class on non-toggle icon

### LINKED ISSUE #355: ebay-menu with icon label flips both the icon and chevron when opened
# Bug Report

## Description
Same as the title of this issue.

```css
.button.expand-btn[aria-expanded=true] .expand-btn__icon {
    transform: rotate(180deg);
}
```
this css applies to both the icon label and the chevron. Some icons appear to move when they are flipped due to this. It should only flip the chevron, if it exists.

## Screenshots
icon-jump

yomed:
## Description

- update menu non-toggle icon class from `expand-btn__icon` -> `btn__icon`

## Context

There isn't an exact pattern that's currently in Skin (there's a button with an icon, but no an expand button with one). But this seems to make sense, because the `expand-btn__icon` class is used to target the toggle for rotation. So now we can avoid that.

The movement issue is not fixed yet, but will be fixed in the next Skin pre-release.

## References

Fixes #355

### DIFF
--- a/src/components/ebay-menu/template.marko
+++ b/src/components/ebay-menu/template.marko
@@ -2,7 +2,7 @@
     <ebay-button w-id="button" class=data.buttonClass variant="expand" size=data.size priority=data.priority aria-expanded=String(data.expanded) aria-haspopup="true" aria-label=data.a11yText>
         <span class="expand-btn__cell">
             <!-- Convoluted markup to satisfy both Skin and Marko -->
-            <div if(data.icon) w-preserve-body class="expand-btn__icon">
+            <div if(data.icon) w-preserve-body class="btn__icon">
                 <span if(data.iconTag) w-body=data.iconTag body-only-if(true)/>
             </div>
             <span if(data.text)>${data.text}</span>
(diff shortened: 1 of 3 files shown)


src/components/ebay-menu/template.marko
```diff
@@ -2,7 +2,7 @@
     <ebay-button w-id="button" class=data.buttonClass variant="expand" size=data.size priority=data.priority aria-expanded=String(data.expanded) aria-haspopup="true" aria-label=data.a11yText>
         <span class="expand-btn__cell">
             <!-- Convoluted markup to satisfy both Skin and Marko -->
-            <div if(data.icon) w-preserve-body class="expand-btn__icon">
+            <div if(data.icon) w-preserve-body class="btn__icon">
                 <span if(data.iconTag) w-body=data.iconTag body-only-if(true)/>
```
seangates: `body-only-if(true)`?

yomed: Marko magic. Hopefully can clean this up after this lands https://github.com/marko-js/marko/pull/1094

PR #358, opened 2018-08-16
Icon: DS4 support.

### LINKED ISSUE #256: Icon: svg files are manually copied from Skin
# Feature Request

## Description

It would be nice to have a build tool to automatically import svg icon files from Skin. This can help reduce errors from manually copying and pasting the markup. It can also help us keep up to date when there are changes to Skin. The 2 main areas are: 1) the svg files themselves, and 2) references to the icons in the `examples/` .

### LINKED ISSUE #315: ebay-icon: DS4 icons are not part of the symbols library
# Bug Report

## Description
Currently, the only icons in the `ebay-icon/symbols` directory are for DS6. We need to be able to support different design systems' icons. One solution is to possibly use the `skin-ds#` Lasso flag which gives us this information. I don't believe we need Skin to be tied to the components, though, which puts us in a tough spot.

Please let me know your thoughts on how to proceed.

## Workaround
None.

DylanPiercey:
## Description
This PR.... Well... Here it is.

First of all this adds a script which automatically pulls the icons from skin (both ds4 and ds6) and outputs a structure similar to whats mentioned in #315 which fixes #256 🎊!

This PR also changes the transform to inline *both* the ds6 and ds4 version of the icon. Then the `browser.json` remaps symbols `ds6 => ds4` in ds4 mode and `ds4 => ds6 in ds6` mode. Ultimately this means that only one symbols is sent to the browser based on your lasso flags.

The hacky part is how I am handling this on the server. What happens is that on the server side the icon component looks at a hidden lasso global in the Marko output stream which contains the lasso flags. It then checks to see if the skin flag matches in the component before grabbing one of the symbols that were passed in. The transform now adds a hidden `_themes` attribute which passes in both symbols for the `ebay-icon` to pick from.

## Concerns
I hope we never have to support three design systems simultaneously because that would make the code larger per usage which is unfortunate although this could be improved in a Marko v4 implementation.

If you are rendering with ds4 flags and the icon you ask for does not exist it falls back to ds6, this is kind of strange but hard to avoid (this also happens the other way around).

Since there are a lot of naming differences between icons there is still some manual effort that would be required to upgrade unless we normalize them further.

## References
Fixes #315 
Fixes #256


seangates:
389 files ... [gulp]

yomed:
Pure magic. A few side notes:
1) Can you update the CONTRIBUTING.md file?
2) What should we do with the `examples/` now?

src/components/ebay-icon/template.marko
```diff
@@ -4,7 +4,24 @@
...
+                <%
+                    var theme;
+
+                    if (process.browser) {
+                        theme = data.themes.filter(Boolean)[0];
+                    } else {
+                        // Actual magic, please do not try to replicate.
+                        var lassoContext = out.global["lasso/LassoRenderContext"];
+                        var flags = lassoContext && lassoContext.data.config.flags;
+                        theme = data.themes[!flags || ~flags.indexOf("skin-ds6") ? 1 : 0];
```
msendlakowski: `~` magic even

DylanPiercey: Originally this was copied in more and I was trying to save bytes. I could make this check better but I don't think it's a big deal. If your against it then I can update the PR.

msendlakowski:
@DylanPiercey Can you add a newline at the end of the browser.json generation?  Its passing the lint check, but I think lint rules are expecting a newline and my editor is automatically adding them.

DylanPiercey:
Re @yomed 

1. Can do 😄
2. I forgot to mention this in my concerns. I need feedback regarding this. If you look at the examples now in ds4 mode it looks pretty bad.

src/components/ebay-icon/template.marko
msendlakowski: I'm not a fan since its obscure, but its not an issue.  I just don't see it often.

DylanPiercey:
@msendlakowski will do re new line on `browser.json`.

seangates:
By examples, you mean the marketplace?

seangates:
This?

image

DylanPiercey:
Rebased and addressed all the PR comments besides what to do about examples. Maybe we can hash that out tomorrow during the standup or something.

DylanPiercey:
Just realized that this removes the play/pause button for the autoplay carousel. We should probably get that in skin before we merge this PR.

seangates:
@DylanPiercey - Let me know how I can support you. Either you can open an issue and I'll jump on it, or you can create a PR and I'll review it. Either way works for me.

PR #362, opened 2018-08-20
Pagination: fix responsive behavior

### LINKED ISSUE #360: Pagination display bug
# Bug Report

## Description
when implemented pagination with 9 items there are issues with responsive / mobile displays.

Say on mobile 5 page items are visible out of 9, and page 5 is the active page.  On mobile you will see 1-5 instead of 3-7 as expected.

It appears as the pagination component is only applying hidden attributes to items starting from the end of the list.

The flag needs distributed evenly between the beginning and end of the list.

## Workaround
Current work around is to limit the number of items being rendered into the pagination component based on current page size.  This is not ideal and good bit of extra work to implement.

martin-ebay:
## Description
added an onUpdate function so that the component would update with window resizes and data updates from client side updates without page reloads.

Also adjusted how the page numbers that are displayed are calculated to keep the currently selected item in the middle if possible.

## Context
This was to fix the issue of having 9 page results with page 7 selected as current on a mobile view with only 5 page buttons visible.  The current page button was hidden along with the buttons after it and instead you saw the buttons 1-5.

## References
Fixes #360


seangates:
@martin-ebay First things first, do you have any tests written for the additional code?

martin-ebay:
I did not add any new test.

This primarily updates the refresh function, and is focused on changing which items have the hidden attribute.

The only thing outside of that was that I added an onUpdate function to trigger the refresh.  I had an instance where I was making an ajax call then updating which page item has the current flag, but it would not update without the onUpdate.

martin-ebay:
I'm wondering if the style has to do with the fact that the contents are added externally, and can be updated externally as well.

As for the setTimeout, I have it set to 0 which just forces the call to the end of the js call execution.  I'm doing this because when the onUpdate triggers if I call the refresh function the dom elements don't exist yet.  If I don't do this the items won't exist and I can not get offsetWidth values used for the logic in the refresh function.

As far as the looking for the current element in state I was not seeing the current value when I tried checking the state.  I'm guessing this maybe has something to do with the current being set by adding the "current" attribute to the page item from outside of the component when rendering.  I would prefer to use state for this, just not sure if it is available or updating correctly.

martin-ebay:
I will add a reference for the setTimeout and will clear it onBeforeUpdate and onDestroy shortly

DylanPiercey:
@martin-ebay yeah the `setTimeout` is fine and makes sense to use here, just that Marko will update effectively using `setImmediate` at the start of the next event queue where as `setTimeout` is going to fire after within usually `5ms` which is slower. Again the chances of the component being removed in between these two is small unless the user is doing something strange but better to avoid the chance for these subtle frustrating bugs.

Also regarding the `current` item, AFAICT it was under `items[*].current` but that could be wrong. I'd say the way this works right now is fine but if you are able to use the state that would be better.

I'd personally update the item state based off of screen size here instead of directly mutating the DOM. (You would still have to read the dom to calculate what's visible, but let Marko actually render the update). Not something that we need to do now though. I think this is fine as is.

martin-ebay:
I would be up for making those updates to utilize state more, but maybe in additional PR when I have more time to dedicate to this.

seangates:
I've reset the base branch from `master` to the `1.0.0` release branch. This causes a conflict, so @martin-ebay will need to rebase.

I've also requested @msendlakowski to take one last look at this and give the second review.

Thanks, all! let me know if anyone has questions.

src/components/ebay-pagination/index.js
```diff
@@ -76,27 +76,71 @@ function init() {
...
+    if (totalPages - current < rangeRight) {
+        start -= (rangeRight - (totalPages - current));
+    }
+
+    if (start < 0) {
+        end -= start;
+    }
+
+    for (let i = 0; i < totalPages; i++) {
+        if (i < start) {
```
yomed: I think you can simplify this section, since `removeAttribute` will be a no-op if the attribute isn't there.
```js
if (i < start || i > end) {
    this.pageEls[i].setAttribute('hidden', true);
} else {
    this.pageEls[i].removeAttribute('hidden');
}
```

martin-ebay:
I have added the improvements suggested.  I have not added test yet as I could not figure out how to get the refresh function or window.resize to trigger after resizing the root element

src/components/ebay-pagination/index.js
```diff
@@ -76,27 +76,63 @@ function init() {
...
-    const adjustedNumPages = Math.min(constants.maxPagesAllowed,
-        Math.max(numPagesAllowed, constants.minPagesRequired));
+    let current = 0;
     const totalPages = this.pageEls.length;
-
-    // Let's show all the pages that we can.
-    for (let i = 5; i < adjustedNumPages && i < totalPages; ++i) {
-        if (this.pageEls[i].hasAttribute('hidden')) {
-            this.pageEls[i].removeAttribute('hidden');
+    for (let i = 0; i < this.state.items.length; i++) {
```
PrashantAshok: Minor: use `some`?
```
this.state.items.some((num, i) => {
  if (num.current) {
    current = i;
    return true;
  }
});
```

martin-ebay:
small bug fixed, and 9 unit test added to pagination component for testing the visible items at different widths, with different "current" items, near beginning, middle, and near end.

PR #367, opened 2018-08-22
Pagination: Button mode

### LINKED ISSUE #299: Pagination: add support for button tag and pure client-side behaviour
# Feature Request

## Description

I've ran into a few cases where pure client-side pagination is being used. The backend service does not have any URLs for the href, therefore anchor tags cannot be used (at least not without abusing them with invalid href values). For example, when pagination through feedback results.

Of course for pages like SRP, even when re-rendering results on the client, there should always be a URL. But for some experiences there may not be.

So I think we need to support button tag too... somehow.

## Context

Depends on: https://github.com/eBay/skin/issues/293

DylanPiercey:
## Description
This change coincides with https://github.com/eBay/skin/pull/328 which now allows the pagination styles to contain buttons for items and prev/next.

If the `href` attribute is omitted on any item it now becomes a button (previously this made it disabled and so this is a breaking change).

Also because this functionality heavily overlaps with the existing `hijax` functionality I have completely removed that, added new tests and simplified things.

## References
Fixes #299 

Note:
Currently to view this PR you will have to manually link the skin branch mentioned above.


src/components/ebay-pagination/README.md
```diff
@@ -46,6 +45,10 @@ Event | Data | Description
 Name | Type | Stateful | Description
 --- | --- | --- | ---
 `disabled` | Boolean | No | Previous/next button is disabled or not
-`href` | String | No | for link that looks like a menu-item
+`href` | String | No | for link that looks like a menu-item, by omitting the item will switch to a button
```
yomed: Looks like this description may have gotten mangled. Maybe:
`for links; omitting the href will switch to using a button`

DylanPiercey: Ah good catch.

Issue #374, opened 2018-08-24
Pagination not displaying enough items when selecting last couple of pages


martin-ebay:
# Bug Report

## Description
If you are using the pagination--fluid option and have selected a page near the end of the total pages the wrong number of items are displayed. And at a few resolutions the too many items where visible and where wrapping into a hidden part of the view.

## Workaround
pull in the latest ebay/skin once https://github.com/eBay/skin/pull/350 has been merged in and release.

a bit of code updates to the pagination refresh function to utilize better values after css updates.

PR #375, opened 2018-08-24
Pagination: improve hidden overflow logic

martin-ebay:
## Description
when using pagination--fluid and selecting a page number near the totalPage number the displayed number of pages would be missing a few items.

This update fixes the problem yet is dependent upon https://github.com/eBay/skin/pull/350 getting merged in and released to ebay/skin

## Context
to fix a bug introduced by my last update that I missed.

## References
fix for https://github.com/eBay/ebayui-core/issues/374

### DIFF
--- a/src/components/ebay-pagination/index.js
+++ b/src/components/ebay-pagination/index.js
@@ -65,6 +65,7 @@ function getTemplateData(state) {
 
 function init() {
     this.pageContainerEl = this.el.querySelector('.pagination__items');
+    this.pageContainerEl.style.flexWrap = 'nowrap';
     this.pageEls = this.pageContainerEl.children;
     this.containerEl = this.el;
     this.previousPageEl = this.el.querySelector('.pagination__previous');
@@ -91,44 +92,40 @@ function refresh() {
     for (let i = 0; i < this.state.items.length; i++) {
         if (this.state.items[i].current) {
             current = i;
-        } else {
-            // remove all hidden attribues to get accurate widths
-            this.pageEls[i].removeAttribute('hidden');
         }
+        this.pageEls[i].removeAttribute('hidden');
     }
+
     const totalPages = this.pageEls.length;
-    const pageNumWidth = this.pageEls[current].offsetWidth + constants.margin;
-    const containerWidth = this.containerEl.offsetWidth - pageNumWidth * 2;
-    const numPagesAllowed = Math.floor((containerWidth) / (pageNumWidth));
-    const adjustedNumPages = Math.min(constants.maxPagesAllowed - 1,
-        Math.max(numPagesAllowed, constants.minPagesRequired));
+    const pageNumWidth = this.pageEls[0].children[0].offsetWidth + constants.margin;
+    const numPagesAllowed = (((this.pageContainerEl.offsetWidth) / pageNumWidth));
+    const adjustedNumPages = Math.floor(Math.min(constants.maxPagesAllowed,
+        Math.max(numPagesAllowed, constants.minPagesRequired)));
 
     let start = 0;
     let end = adjustedNumPages;
-    const rangeLeft = Math.floor(adjustedNumPages * 0.5);
+    let rangeLeft = Math.floor(adjustedNumPages * 0.5);
     const rangeRight = Math.floor(adjustedNumPages * 0.5);
+
+    if (rangeLeft + rangeRight + 1 > adjustedNumPages) {
+        rangeLeft -= 1;
+    }
 
     start = current - rangeLeft;
     end = current + rangeRight;
-    if (end > totalPages) {
-        start -= (end - totalPages);
-    }
 
     if (totalPages < constants.maxPagesAllowed) {
         end = totalPages;
     }
 
-    if (totalPages - current < rangeRight) {
-        start -= (rangeRight - (totalPages - current));
+    if (current + rangeRight >= totalPages) {
+        end = totalPages;
+        start = end - adjustedNumPages;
     }
 
-    if (start < 0) {
-        end -= start;
+    if (start <= 0) {
+        end = adjustedNumPages - 1;
         start = 0;
-    }
-
-    if (end - start < constants.minPagesRequired && end === totalPages && start > 0) {
-        start = end - constants.minPagesRequired;
     }
 
     for (let i = 0; i < totalPages; i++) {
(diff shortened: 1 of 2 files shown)


src/components/ebay-pagination/index.js
```diff
@@ -91,46 +91,42 @@ function refresh() {
...
+        this.pageEls[i].removeAttribute('hidden');
     }
+
     const totalPages = this.pageEls.length;
-    const pageNumWidth = this.pageEls[current].offsetWidth + constants.margin;
-    const containerWidth = this.containerEl.offsetWidth - pageNumWidth * 2;
-    const numPagesAllowed = Math.floor((containerWidth) / (pageNumWidth));
-    const adjustedNumPages = Math.min(constants.maxPagesAllowed - 1,
-        Math.max(numPagesAllowed, constants.minPagesRequired));
+    const pageNumWidth = 56;
```
msendlakowski: If we're hardcoding the width in CSS why do we have to do it here too?  Can't we reliably pick it up from the element now?

martin-ebay: hmm so far when measuring the first element they keep showing up 2 px wider on the 640px wide test.  Trying to track down why.  When I test this in a live view they never report anything other than 48, but in the test they keep showing up 50.  I wouldn't think this was possible with the updated css.

martin-ebay:
ok figured out my issue, had to measure the child element as the li element itself was not a set width.

yomed:
@martin-ebay It looks like the pagination tests are failing in this branch. Also, when I pulled it locally, I see the pagination incorrectly hiding the last element on page load (on the first example in the local demo page).

martin-ebay:
@yomed do you know if skin has been updated.  This needs to pull in the latest from skin for the test to pass.  I was waiting for a new release of 6.0.0 skin that had the css updates for pagination.

However the initial item hiding on the test page shouldn't be happening.  Something must be off when not using the pagination--fluid style option.  I'm curious if there is a valid use case for not using the pagination-fluid by default as it is responsive.

I'll look into the hiding of the last item when not in fluid this afternoon.

msendlakowski:
@martin-ebay all skin PRs regarding pagination have been merged, but not published yet.  Does this PR need to wait for skin?

martin-ebay:
it does. As well as I'm working on the fix @yomed mentioned.  I have it working locally but wanted to test one more thing inline with the pagination--fluid option, but some staging service calls are failing right now blocking me from the final check.

Once I can do that final check I'll make a new PR for skin as there are some required updates, and a small update to this PR to make the pagination work as expected with and without pagination--fluid.

msendlakowski:
@yomed Sean is going to merge and publish the skin changes from @martin-ebay and I'm going to publish a final pre-release of ebayui after integrating that.  Do you have any more comments on this PR or can I add this to the pre-release?

msendlakowski:
@martin-ebay  I ran tests locally using your changes and a few were broken.  I've updated skin.  Can you rebase and fix your tests.

martin-ebay:
will get those fixed up this morning

yomed:
@martin-ebay I still see the issue mentioned earlier -- pagination incorrectly hiding the last element on page load (e.g. on the first example in the local demo page).

martin-ebay:
@yomed are you look at this with the latest from skin 6.0.0?  I will check locally here in a few, but I didn't look when I fixed the unit test.

yomed:
@martin-ebay Oh ok, let me do that. @seangates Is it cool if we just do the skin upgrade in this PR?

src/components/ebay-pagination/index.js
```diff
@@ -65,6 +65,7 @@ function getTemplateData(state) {
 
 function init() {
     this.pageContainerEl = this.el.querySelector('.pagination__items');
+    this.pageContainerEl.style.flexWrap = 'nowrap';
```
yomed: Any particular reason this is here rather than in CSS/HTML?

src/components/ebay-pagination/index.js
```diff
@@ -91,46 +92,42 @@ function refresh() {
     for (let i = 0; i < this.state.items.length; i++) {
         if (this.state.items[i].current) {
             current = i;
-        } else {
-            // remove all hidden attribues to get accurate widths
-            this.pageEls[i].removeAttribute('hidden');
         }
+        this.pageEls[i].removeAttribute('hidden');
```
yomed: Did you feel that the comment wasn't needed for this?

src/components/ebay-pagination/index.js
```diff
@@ -91,46 +92,42 @@ function refresh() {
...
+        Math.max(numPagesAllowed, constants.minPagesRequired)));
 
     let start = 0;
     let end = adjustedNumPages;
-    const rangeLeft = Math.floor(adjustedNumPages * 0.5);
+    let rangeLeft = Math.floor(adjustedNumPages * 0.5);
     const rangeRight = Math.floor(adjustedNumPages * 0.5);
 
+    if (rangeLeft + rangeRight + 1 > adjustedNumPages) {
+        rangeLeft -= 1;
```
yomed: Make sure that this new logic path is covered with a new test case.

src/components/ebay-pagination/index.js
martin-ebay: @seangates told me that skin had to wrap and hide elements that wrap using the overflow of the container.  Using wrap messes with the math, so I'm overriding it in ebayui as a progressive enhancement.

src/components/ebay-pagination/index.js
martin-ebay: I can add one back in if needed.

src/components/ebay-pagination/index.js
martin-ebay: I will write a new test for this case after lunch.  just need to add a new width

seangates:
@msendlakowski I'm okay with upgrading Skin in the PR, as it is necessary for this PR to work correctly.

@msendlakowski Any reservations?

seangates:
@martin-ebay You might just need to rebase.

src/components/ebay-pagination/index.js
martin-ebay: there is no a new test at width 440px when the fifth element is selected.  It only works when the current element is not near the beginning or end.

martin-ebay:
ok all I've rebased, added a new test to show the lopsided distribution when an even number of elements are visible with item in the middle selected.

Let me know if there is anything else

Issue #403, opened 2018-09-05
Select: event has incorrect selected value


martin-ebay:
# Bug Report

## Description
selecting the first element when the second element is already selected returns second element as the component value instead of the first as expect in the on-change event.

## Workaround
need to update the setSelectedOption function and pull the _this2.update() out of the map loop.

PR #404, opened 2018-09-05
Select: fix value in change event

martin-ebay:
## Description
updated the call to update in the setSelectedOption map loop to prevent it from calling early

## Context
fix for issue #403


seangates:
Should `this.update()` come after the `setState`?

yomed:
Thanks for the PR. Please include a test case that was failing previously.

martin-ebay:
@yomed 2 unit test added along with a second set of mock data

msendlakowski:
@yomed  The unit test isn't fully testing the issue, but @martin-ebay is out of time.  Are you OK if we merge and publish this patch fix quickly.  And I'll follow up with a better test.

yomed:
@martin-ebay Thanks for the attempt on the test. Since it appears very difficult to simulate this test, and because your tests do not fail before the fix, we decided to go ahead with the fix by itself.

Issue #410, opened 2018-09-13
Carousel: Items are not re-rendering(seeing old state) when parent state is updated


sraone:
# Bug Report
Items are not re-rendering(seeing old state) when parent state is updated

## Description
I am using a `carousel` items are bound from the state. When the state is replaced with new items, `carousel` still showing old items.

`template.marko`

```marko
<div w-bind>
    <button type="button" w-on-click="handleReloadClick">Reload</button>
    <ebay-carousel>
        <ebay-carousel-item for(item in data.items)>${item}</ebay-carousel-item>
    </ebay-carousel>
</div>
```

`index.js`

```javascript
module.exports = require('marko-widgets').defineComponent({
    template: require('./template.marko'),

    getInitialState() {
        return {
            items: [1, 2, 3, 4, 5, 6]
        }
    },

    handleReloadClick() {
        this.setState('items', [10, 20, 30]);
    }
});
```

## Screenshots

with initial state

![image](https://user-images.githubusercontent.com/1853563/45511675-7d2b6f80-b752-11e8-8a3c-1ea8b7cdbbb0.png)

with new state

![image](https://user-images.githubusercontent.com/1853563/45511750-af3cd180-b752-11e8-90b6-b337b7e70586.png)

But it should display `10, 20, 30`.

yomed:
@DylanPiercey @mlrawlings It looks like the `renderBody` doesn't get updated correctly on the new input. Starting with `[1, 2]` and updating to `['a', 'b', 'c', 'd']` will result in `1 2 c d`.

sraone:
@msendlakowski I don't think its marko issue because items are rendering as expected if I remove `carousel` component.

msendlakowski:
Thanks.  I'll take a closer look.

seangates:
@sraone Can you try the `1.1.0` branch and let us know if the change worked for you?

If you need help with how to test this branch let me know. I can assist.

ianmcburnie:
Fixed in v1.1.0 branch.

PR #413, opened 2018-09-15
w-preserve-body-if in carousel component has to be true or false

msendlakowski:
## Description
Changed `w-preserve-body-if` to false rather than undefined
Fixes https://github.com/eBay/ebayui-core/issues/410

## Context
Currently it was passed as undefined for state changes, but the Marko "if" check tested for an absolute false, not just falsy.

I'm not sure if this is going to cause any other issues with carousel rerendering. So it would be good if @DylanPiercey could review.

Also, I couldn't figure out a clean way to add a test, since I couldn't identify a parent component where I could change the state.

## Screenshots
screen shot 2018-09-14 at 5 29 20 pm

### DIFF
--- a/src/components/ebay-carousel/template.marko
+++ b/src/components/ebay-carousel/template.marko
@@ -31,7 +31,7 @@
                     style=item.style
                     w-body=item.renderBody
                     aria-hidden=(!item.fullyVisible && 'true')
-                    w-preserve-body-if(config.preserveItems)/>
+                    w-preserve-body-if(!!config.preserveItems)/>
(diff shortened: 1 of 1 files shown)


DylanPiercey:
Looks good to me. For future reference Marko 4 does a truthy check (https://github.com/marko-js/marko/blob/master/src/components/taglib/preserve-tag-browser.js#L12) and Marko 3 does the literal false check (https://github.com/marko-js/marko-widgets/blob/master/taglib/preserve-tag.js#L31).

PR #417, opened 2018-09-17
Fix backtick in button README

seangates:
Fixing the button readme.

### DIFF
--- a/src/components/ebay-button/README.md
+++ b/src/components/ebay-button/README.md
@@ -12,7 +12,7 @@ Name | Type | Stateful | Description
 --- | --- | --- | ---
 `priority` | String | No | "primary" / "secondary" (default) / "none"
 `size` | String | No | "small" or "large" (default: medium)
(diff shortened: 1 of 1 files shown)


src/components/ebay-carousel/template.marko
```diff
@@ -31,7 +31,7 @@
                     style=item.style
                     w-body=item.renderBody
                     aria-hidden=(!item.fullyVisible && 'true')
-                    w-preserve-body-if(config.preserveItems)/>
```
ianmcburnie: Is this change intentional? Same goes for the yarn.lock update. Looks like a couple of other commits got brought along.

seangates:
Maybe the target needs to change?

seangates:
Looking at it, those were rebased from master. Let me rebase to 1.1.0.

ianmcburnie:
Changing target to master, because it might be a while till `1.1.0` is released.

seangates:
Okay, let me fix this again now.

PR #424, opened 2018-09-20
8 pill component

msendlakowski:
## Description
Add pill support for button component.
Add truncate utility. 
Add expand example that was missing.

## Context
I tried to add the pill support in a way that the user didn't have to add the inner content.  I wanted them to create it like:
```
<ebay-button href="#" priority="secondary" pill>Single line</ebay-button>
```

Instead I ended up with:
```
<ebay-button href="#" priority="secondary" pill>
    <span class="fake-btn__cell">
        <span>Single line</span>
    </span>
</ebay-button>
```

I did this because I ran into a problem when button content was dynamically updated. The text content didn't get updated. I can give more details if needed.

This seems to be the current expectations though.  The expand button requires the user to create it as:
```
<ebay-button variant="expand" priority="primary">
    <span class="expand-btn__cell">
        <span>Expand</span>
        <ebay-icon type="inline" name="chevron-down-bold" class="expand-btn__icon" no-skin-classes/>
    </span>
</ebay-button>
```

I think we should eventually see if we can simplify these cases for the users.

### A11y
The truncated button results in the following HTML structure:
```
<span class="fake-btn__cell">
    <span>
        <span class="clipped">Lots of text that should be truncated at 2 lines</span>
        <span aria-hidden="true">Lots of text that should be trunca...</span>
    </span>
</span>
```

## References
https://github.com/eBay/skin/pull/379

## Screenshots
screen shot 2018-09-20 at 1 50 52 pm

### DIFF
--- a/src/components/ebay-button/README.md
+++ b/src/components/ebay-button/README.md
@@ -6,6 +6,14 @@
 <ebay-button>text</ebay-button>
 ```
 
+```marko
+<ebay-button pill>
+    <span class="btn__cell">
+        <span>Pill text</span>
+    </span>
+</ebay-button>
+```
+
 ## ebay-button Attributes
 
 Name | Type | Stateful | Description
--- a/src/components/ebay-button/README.md
+++ b/src/components/ebay-button/README.md
@@ -18,6 +18,7 @@ Name | Type | Stateful | Description
 `disabled` | Boolean | Yes |
 `partially-disabled` | Boolean | No
 `variant` | String | No | optional, to alter Skin classes: "expand" / "cta"
+`pill` | Boolean | No | for buttons/links that look like a pill (DS6 only)
 
 ## ebay-button Events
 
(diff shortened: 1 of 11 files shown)


src/components/ebay-button/README.md
```diff
@@ -18,6 +18,7 @@ Name | Type | Stateful | Description
 `disabled` | Boolean | Yes |
 `partially-disabled` | Boolean | No
 `variant` | String | No | optional, to alter Skin classes: "expand" / "cta"
+`pill` | Boolean | No | for buttons/links that look like a pill (DS6 only)
```
senthilp: @msendlakowski In the `README.md` can you also add a simple example of the markup for the pill use case, along with the simple `<ebay-button>` tag?

ianmcburnie:
Let's reopen this when Skin work on button and pill is done.

PR #428, opened 2018-10-02
Textbox: add basic native events

### LINKED ISSUE #427: Textbox - Support for emitting events
# Bug Report

## Description
When I add events on the ebay-textbox tag, the event doesn't get triggered. 
`<ebay-textbox placeholder="Search for something" name="search-items" on-input('filterItems')  fluid/>`
The `input` event doesn't get triggered

yomed:
## Description

- add `change` `input` `focus` `blur`
- update docs and tests

## Context

This is not the full list of native DOM events that could potentially need forwarding, but this should hold us over until we decide if we want to handle _all_ events.

## References

Fix #427

### DIFF
--- a/src/components/ebay-textbox/README.md
+++ b/src/components/ebay-textbox/README.md
@@ -15,3 +15,12 @@ Name | Type | Stateful | Description
 `fluid` | Boolean | No |
 `multiline` | Boolean | No | renders a multi-line texbox if true
 `invalid` | Boolean | No | indicates a field-level error with red border if true
+
+## ebay-textbox Events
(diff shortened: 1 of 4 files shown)


PrashantAshok:
@yomed - can you please backport the changes to 0.9.2? Shouldn't break anything, since new func added.

seangates:
@PrashantAshok What's preventing updating to v1?

PrashantAshok:
There are breaking changes in v1 (event name, attrs etc.). It's on roadmap, hopefully soon.

PR #429, opened 2018-10-02
Ensure marko treats nested transformed tags as components.

DylanPiercey:
## Description
This change adds a `renderer` for each of the nested elements in the `marko.json` which forces them to be treated at a component. Previously they were treated as elements and subtle differences between the compilation of the two was causing these not to be keyed properly in Marko 4 when inside of a for loop.

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -1,7 +1,6 @@
 {
     "<ebay-breadcrumb>": {
         "renderer": "./src/components/ebay-breadcrumb/index.js",
-        "transformer": "./src/common/transformer/index.js",
         "@*": "expression",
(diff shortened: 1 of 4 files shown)


DylanPiercey:
My initial PR here was only partially successful. I've updated it to take a different approach to the transforms and instead I am now transforming the children directly, instead of finding them from the parent element transform.

This allows the transform to run earlier which causes things like `<for>` loops to be transformed properly.

I removed the test that checks for nested transforms since it's no longer applicable.

---

Ideally we would get a patch release out for this ASAP.

DylanPiercey:
Will try rebasing 👍

Issue #432, opened 2018-10-03
Carousel: touch scroll will not allow scroll to last item


scttdavs:
# Bug Report

## Description

When there is not a full page of items on the last page (let's say only 1 where we show 2 items per page), I cannot touch scroll to it. Tapping the next button works however. Touch scrolling between other pages works correctly, it is only the partial last page that has an issue.

## Workaround

Tapping next button?

## Screenshots

![snap-grid](https://user-images.githubusercontent.com/3631216/46440413-3c12f380-c731-11e8-8b81-3993bbbc889b.gif)

ianmcburnie:
Fix went out in 1.0.7.

PR #433, opened 2018-10-05
Carousel: Sizing & Native scroll animation

### LINKED ISSUE #431: Carousel items allowed to grow too big.
# Bug Report

## Description

We have titles in our carousel items that need to be a certain number of lines (let's say just one line) and we truncate it. The width is set to 100% on these elements. But the carousel items are styled with flex-basis by ebayui, not width. So the elements are allowed to grow to however long the truncated text would be.

Adding a `width` to the carousel item with the same value as flex-basis fixes it, but I understand this may adversely affect touch scrolling behavior (the snapping to grid?).

## Workaround

No known one. Adding width seemed to fix it, but it may break touch scrolling, and the value of flex-basis is dynamically generated by the component.

## Screenshots

screen_shot_2018-10-03_at_4 18 59_pm

This is what it should look like (for example, after `width` is added)
image

DylanPiercey:
## Description
If desired I can separate these out into two PR's.

1. switched to using `width` instead of `flex-basis` to size `carousel-items`. This resolves #431 with no apparent change in appearance.

2. added an additional timeout to the `onScrollEnd` handler. Chrome doesn't seem to be working properly with the previous 1 animation frame polling, not sure if this is something that has changed recently but causes the scroll animation to start too early.

# Details

For the timeout change I just noticed while fixing #431 that this was happening. This could be what is causing #432 and I will confirm with @scttdavs.

## References
fixes #431

### DIFF
--- a/src/common/test-utils/browser.js
+++ b/src/common/test-utils/browser.js
@@ -39,9 +39,13 @@ function simulateScroll(el, to, cb) {
             if (++frame > frames) {
                 triggerEvent(el, 'touchend');
                 el.scrollLeft = to;
-                // Allow two frames for the on scroll to finish.
+                // Allow two frames and a timeout for the on scroll to finish.
                 if (cb) {
-                    waitFrames(2, cb);
+                    requestAnimationFrame(() => {
+                        setTimeout(() => {
+                            requestAnimationFrame(cb);
+                        }, 64);
+                    });
                 }
 
                 return;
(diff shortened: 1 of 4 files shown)


ianmcburnie:
As discussed, let's separate out these two fixes. Fix 1 will go into v1.1.0, fix 2 can go out as a v1.0.7 patch.

ianmcburnie:
Maybe switching to using width instead of flex-basis can be applied as a modifier/attribute if we are worried about it breaking touch scrolling?

ianmcburnie:
Going to close this PR, because as discussed it seems we should put one fix out as a patch, and one out as a minor.

PR #438, opened 2018-10-15
Carousel: Switch from flex-basis to width

### LINKED ISSUE #431: Carousel items allowed to grow too big.

DylanPiercey:
## Description
This PR swaps out `flex-basis` usage for `width`. This allows for items within the carousel to size themselves relative to the width of the carousel item, which is currently not possible with flex-basis.

## References
Fixes #431

### DIFF
--- a/src/components/ebay-carousel/index.js
+++ b/src/components/ebay-carousel/index.js
@@ -82,8 +82,6 @@ function getTemplateData(state) {
         a11yStatusText = state.a11yStatusText
             .replace('{currentSlide}', slide + 1)
             .replace('{totalSlides}', totalSlides);
-    } else {
-        itemWidth = 'auto';
     }
 
     items.forEach((item, i) => {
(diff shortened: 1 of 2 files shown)


ianmcburnie:
I'll hold off on merging until 1.0.7 is merged to master, released and then rebased back into 1.1.0.

PR #441, opened 2018-10-15
Carousel: Native scrolling snap fixes

### LINKED ISSUE #432: Carousel: touch scroll will not allow scroll to last item

### LINKED ISSUE #439: Carousel: In mobile occasionally the final scroll position can be off by a bit.
# Bug Report
Currently we determine when the native scrolling has finished by polling the `scrollLeft` of the carousel element. In chrome it (no longer?) seems to be updating this every animation frame. This means that the final resting position of the carousel can be off by however much inertial scrolling is remaining.

## Description
Open the chrome dev tools or chrome on a mobile device and swipe through the carousel a bunch.

DylanPiercey:
## Description
This PR fixes an issue where the carousel would snap to the incorrect place when items-per-slide was set in mobile. The new implementation a properly checks for the closest slide instead of just the closest item.

The also fixes an issue we were seeing in Chrome where the scroll position was not being updated quick enough and was causing false positives in our `onScrollEnd` handler. To fix this I added an additional artificial timeout for the scroll position polling.

## References
* fixes #432 
* fixes #439

### DIFF
--- a/src/components/ebay-carousel/index.js
+++ b/src/components/ebay-carousel/index.js
@@ -329,7 +329,6 @@ function handleScrollEnd(scrollLeft) {
     // Always update with the new index to ensure the scroll animations happen.
     config.preserveItems = true;
     this.setStateDirty('index', closest);
-    this.once('update', this.emitUpdate);
     emitAndFire(this, 'carousel-scroll', { index: closest });
 }
 
--- a/src/components/ebay-carousel/index.js
+++ b/src/components/ebay-carousel/index.js
@@ -135,7 +135,9 @@ function init() {
         onRender.call(this);
     });
     observer.observeRoot(this, ['index']);
-    if (autoplayInterval) observer.observeRoot(this, ['paused']);
+    if (autoplayInterval) {
+        observer.observeRoot(this, ['paused']);
+    }
 
     if (!autoplayInterval && getComputedStyle(this.listEl).getPropertyValue('overflow-x') !== 'visible') {
         config.nativeScrolling = true;
@@ -150,7 +152,9 @@ function onRender() {
     const hasOverride = offsetOverride !== undefined;
 
     // Do nothing for empty carousels.
-    if (!items.length) return;
+    if (!items.length) {
+        return;
+    }
 
     // When there is an offset override (used for infinite scroll) we reset it
     // after rendering to restore the expected carousel state.
@@ -184,7 +188,9 @@ function onRender() {
         if (autoplayInterval && !paused) {
             const moveRight = this.move.bind(this, RIGHT);
             this.autoplayTimeout = setTimeout(() => {
-                if (this.isMoving) return this.once('carousel-update', moveRight);
+                if (this.isMoving) {
+                    return this.once('carousel-update', moveRight);
+                }
                 moveRight();
             }, autoplayInterval);
         }
@@ -216,8 +222,12 @@ function onRender() {
 function cleanupAsync() {
     clearTimeout(this.autoplayTimeout);
     cancelAnimationFrame(this.renderFrame);
-    if (this.cancelScrollHandler) this.cancelScrollHandler();
-    if (this.cancelScrollTransition) this.cancelScrollTransition();
+    if (this.cancelScrollHandler) {
+        this.cancelScrollHandler();
+    }
+    if (this.cancelScrollTransition) {
+        this.cancelScrollTransition();
+    }
     this.cancelScrollHandler = this.cancelScrollTransition = undefined;
 }
 
@@ -237,7 +247,9 @@ function emitUpdate() {
  * @param {HTMLElement} target
  */
 function handleMove(originalEvent, target) {
-    if (this.isMoving) return;
+    if (this.isMoving) {
+        return;
+    }
     const { state } = this;
     const direction = parseInt(target.getAttribute('data-direction'), 10);
     const nextIndex = this.move(direction);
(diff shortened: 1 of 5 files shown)


src/components/ebay-carousel/index.js
```diff
@@ -88,7 +88,7 @@ function getTemplateData(state) {
 
     items.forEach((item, i) => {
         const { style, transform } = item;
-        const marginRight = i !== items.length && `${gap}px`;
```
DylanPiercey: This is technically unrelated, but the intent of this code was to remove the unneeded additional margin from the end which is was not doing because of the lack of the `-1`. This seemed like a extremely low risk change that didn't warrant an issue or another PR.

src/components/ebay-carousel/index.js
```diff
@@ -286,37 +286,27 @@ function handleScrollEnd(scrollLeft) {
     }
 
     const { state } = this;
-    const { config, items } = state;
```
DylanPiercey: This change was primarily to switch from checking for the closest `carousel-item` to instead looking for the closest group of items when `items-per-slide` is used.

src/components/ebay-carousel/utils/on-scroll-end/index.js
```diff
@@ -40,17 +41,19 @@ module.exports = function onScrollEnd(el, fn) {
     // Finally after the touch end we poll the scroll state every animation frame
     // to determine when inertial scrolling has stopped.
     function checkScrollEnded(lastOffset) {
-        frame = requestAnimationFrame(() => {
```
DylanPiercey: An additional 64ms timeout seems to resolve the issue in even the slowest devices. (Tested on the Moto G5 Plus).

src/components/ebay-carousel/utils/scroll-transition/index.js
```diff
@@ -21,7 +21,7 @@ module.exports = function scrollTransition(el, to, fn) {
                 return fn();
             }
 
-            el.scrollLeft = easeOut(delta / duration) * distance + scrollLeft;
```
DylanPiercey: This is also not a necessary change however easeOut was chosen with the intent that the carousel would still be moving when this animation happened. Since we are currently waiting for the scrolling to stop before animating `ease-in-out` looks much better. Again this is a very low risk change.

src/components/ebay-carousel/index.js
```diff
@@ -286,37 +286,27 @@ function handleScrollEnd(scrollLeft) {
...
+    let low = 0;
+    let high = Math.ceil(totalItems / itemsPerSlide) - 1;
 
-    // If we are closer to the end than the closest item, then we just go to the end.
-    if (Math.abs(maxOffset - scrollLeft) < Math.abs(closestOffset - scrollLeft)) {
-        closest = items.length - 1;
+    while (high - low > 1) {
+        const mid = Math.floor((low + high) / 2);
+        if (targetLeft > items[mid * itemsPerSlide].left) low = mid;
+        else high = mid;
```
seangates: I know this is shorthand, but eBay's code style convention is to use normal `if` blocks. I know that kinda sucks, but ... I'd like to keep it consistent everywhere.

DylanPiercey: This is how it was before, so we should probably lint for this. Either way I'll fix it.

DylanPiercey:
Alright I added a commit to switch all inline statements to use brackets.
I also added a test for the group snapping. In doing so I found out that we were actually emitting `carousel-update` twice in the native scrolling mod. Once when the scroll ended and again after the scroll animation ended. I removed a single line to fix this as well as it was messing with my tests.

src/components/ebay-carousel/index.js
ianmcburnie: I know it's done now, but for future reference I'd prefer us not to touch unrelated lines of code (like syntax changes) in bug fixes. The "lines changed" in the PR is now significantly more than it should be. Sounds like nitpicking, but anything to help speed up PR reviews, and reduces future merge conflicts, is a godsend to me!!!

Anyhow, changes look good. 

Anyway, changes look good.

DylanPiercey: I agree. Again the linter should catch this (or maybe there should be no stylistic rules ALA prettier 🙏). My only other options were to not address the PR comment (which was valid), or to only update a portion of the file which would make it inconsistent.

seangates: When making the comment, I didn't realize there were others in the file. Oh well.

senthilp: Agree with @DylanPiercey, this is a linter's job. We have explicitly set the [`curly`](https://eslint.org/docs/rules/curly) linter rule to [allow](https://github.com/eBay/eslint-config-ebay/blob/96f55c9b3eafb4c5ac07e536dbfc271a2d1a55f4/rules/bp.js#L14) this option. This is part of [eslint-config-ebay](https://github.com/eBay/eslint-config-ebay), which every team is using. Maybe we should have a ebayui specific override if needed.

seangates: @senthilp I didn't know that the linter allowed that! If that's the case, then I rescind my comment, and I apologize to @DylanPiercey for all this confusion.

senthilp: No worries. Personally, I like your suggestion. The rule should be like `"curly": [2, "all"]`. But the general consensus was to allow single statement `if` conditions to be without braces.

PR #471, opened 2018-11-05
Fixing the LESS variables because of Skin bundle changes;

### LINKED ISSUE #470: Carousel: update LESS variables because of Lasso changes in Skin v7
This issue is created to track an issue which arises by splitting the DS6 and DS4 variables in Skin based on the `skin-ds6` Lasso flag. The issue is described in https://github.com/eBay/skin/pull/406/files.

Therefore, we must update the carousel to support the appropriate variables with whichever design system is used.

The issue was found when attempting to switch to DS4 in the demo app of this project.

seangates:
**Note:** This should not be merged until Skin v7 is updated and available for the 2.0.0 branch to use. 

## Description
- adds split variable files per Design System
- creates new, generic variables in order to provide capability for switching design systems

## Context
In Skin v7 there was an update which changed the way variables are delivered, essentially splitting out the DS4 and DS6 variables based on the `skin-ds6` Lasso flag. This had detrimental results for the carousel, as it was relying solely on the DS6 Less variables. When you switched the demo page (which is part of this project) to DS4 you would receive an error. Even though the carousel is intended for use only on DS6 pages, we need to support the demo app rendering, and therefore, the DS4 variables.

## References
Fixes #470

### DIFF
--- a/src/components/ebay-carousel/browser.json
+++ b/src/components/ebay-carousel/browser.json
@@ -12,6 +12,18 @@
                 "@ebay/skin/icon"
             ]
         },
+        {
+            "if-not-flag": "skin-ds6",
+            "dependencies": [
+                "./variables-ds4.less"
+            ]
+        },
+        {
+            "if-flag": "skin-ds6",
+            "dependencies": [
+                "./variables-ds6.less"
+            ]
+        },
         "./style.less",
         {
             "if-flag": "touch",
--- a/src/components/ebay-carousel/style-no-touch.less
+++ b/src/components/ebay-carousel/style-no-touch.less
@@ -11,14 +11,14 @@
         }
 
         &:hover:not([aria-disabled="true"]) {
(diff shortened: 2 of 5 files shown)


ianmcburnie:
@seangates , I don't plan on v1.1.0 using Skin 7. v2.0.0 will use Skin 7. So I will change the target branch to 2.0.0 as soon as we create it.

seangates:
@ianmcburnie Updated the description to state the v2.0.0 branch should be used.

ianmcburnie:
Thanks Sean. Oh, and thanks for the fix too! Nice work!

seangates:
I'm now thinking this should be in 1.2.0, as it's preventing me from testing the badge properly. Thoughts?

ianmcburnie:
The LESS changes aren't going out until Skin 7.0.0. We aren't taking the until coreUI 2.0.0 due to the breaking changes it might cause.

PR #472, opened 2018-11-05
New pill component (issue #8)

seangates:
## Description
- adds new pill component based on button
- adds a truncation utility

## Context
The new pill component is based on the button component, as in Skin the pill is simply a modifier class for a button. By using the button we have an easy way to simply extend functionality, and make it easy for a developer to drop in the pills to their app.

The pill also has truncation built in, based on the width of the component. A truncation utility is now provided for other components to use as well, based on Design System requirements.

Lastly, the pill has accessible capabilities. For the `<button>` version an `aria-pressed` attribute will toggle between true and false. For an `<a>` version of the pill we use an additional `<span>` to add a "Selected" string to the button. Also, during truncation we add the full text of the button as clipped text for screen readers, while hiding the truncated version of the button text.

## References
Completes #8.

## Screenshots

image

### DIFF
--- a/src/common/multiline-ellipsis/index.js
+++ b/src/common/multiline-ellipsis/index.js
@@ -1,0 +1,60 @@
+/**
+ * Adds an ellipsis to a el based on the height and width
+ * specified in CSS rules
+ */
+function truncate(el) {
+    if (!el) {
+        return;
+    }
+
+    const text = el.innerText;
+
+    if (this.isTextOverflowing(el)) {
+        const checkFunc = i => {
+            el.innerText = text.substring(0, i);
+            return this.isTextOverflowing(el) ? -1 : 0;
+        };
+        const len = this.binarySearch(text.length - 1, checkFunc);
+        const ellipsisAmount = 2;
+        const truncatedText = text.substring(0, len).slice(0, -ellipsisAmount);
+        el.innerHTML = `<span aria-hidden="true">${truncatedText}…</span><span class="clipped">${text}</span>`;
+    }
+}
+
+function isTextOverflowing(el) {
+    const currentOverflow = el.style.overflow;
+    if (!currentOverflow || currentOverflow === 'visible') {
+        el.style.overflow = 'hidden';
+    }
+
+    const isOverflowing = el.clientWidth < el.scrollWidth || el.clientHeight < el.scrollHeight;
+    el.style.overflow = currentOverflow;
+    return isOverflowing;
+}
+
+function binarySearch(length, callback) {
+    let low = 0;
+    let high = length - 1;
+    let best = -1;
+    let mid;
+
+    while (low <= high) {
+        mid = ~~((low + high) / 2);
+        const result = callback(mid);
+        if (result < 0) {
+            high = mid - 1;
+        } else if (result > 0) {
+            low = mid + 1;
+        } else {
+            best = mid;
+            low = mid + 1;
+        }
+    }
+    return best;
+}
+
+module.exports = {
+    truncate,
+    isTextOverflowing,
+    binarySearch
+};
--- a/src/common/multiline-ellipsis/index.js
+++ b/src/common/multiline-ellipsis/index.js
@@ -14,7 +14,7 @@ function truncate(el) {
             el.innerText = text.substring(0, i);
             return this.isTextOverflowing(el) ? -1 : 0;
         };
-        const len = this.binarySearch(text.length - 1, checkFunc);
+        const len = binarySearch(text.length - 1, checkFunc);
         const ellipsisAmount = 2;
         const truncatedText = text.substring(0, len).slice(0, -ellipsisAmount);
         el.innerHTML = `<span aria-hidden="true">${truncatedText}…</span><span class="clipped">${text}</span>`;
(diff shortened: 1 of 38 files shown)


senthilp:
@seangates We should be adding a `README.md` for this right?

ianmcburnie:
Reminder: This will go into 2.0.0 branch when we cut it.

src/common/multiline-ellipsis/index.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    const text = el.innerText;
+
+    if (this.isTextOverflowing(el)) {
+        const checkFunc = i => {
+            el.innerText = text.substring(0, i);
+            return this.isTextOverflowing(el) ? -1 : 0;
+        };
+        const len = this.binarySearch(text.length - 1, checkFunc);
```
DylanPiercey: Seems a little strange to use `this` here. I think it would be better to just use the hoisted `binarySearch` function as is. This means that if the truncate function is bound for some reason it doesn't break.

src/common/multiline-ellipsis/index.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function binarySearch(length, callback) {
+    let low = 0;
+    let high = length - 1;
+    let best = -1;
+    let mid;
+
+    while (low <= high) {
+        mid = ~~((low + high) / 2);
```
DylanPiercey: For consistency with the binary search in the carousel code and for general readability we should probably change `~~` to `Math.floor(`.

src/components/ebay-pill/index.js
```diff
@@ -0,0 +1,80 @@
...
+    } else {
+        if (selectedTextSpan) {
+            pillBtn.removeChild(selectedTextSpan);
+        }
+    }
+}
+
+function init() {
+    const pillBtn = this.el.querySelector(pillSelector);
+    pillBtn.addEventListener('click', () => {
```
DylanPiercey: These events won't be cleaned up when the component is destroyed. We should be using `this.subscribeTo` like in other components which will allow Marko to cleanup the event handlers when the component is destroyed.

Eg:

```
pillBtn.addEventListener('click', ...)
```

Should be:

```
this.subscribeTo(pillBtn).on('click', ...)
```

 for all of these handlers.

seangates: Yeah, good catch. Changing.

src/common/multiline-ellipsis/index.js
seangates: Sure.

src/common/multiline-ellipsis/index.js
seangates: Done.

seangates:
That makes sense, too. Fixed. Please check again, @ianmcburnie.

src/components/ebay-pill/examples/02-selected/template.marko
```diff
@@ -1 +1 @@
-<ebay-pill checked=true>Pill</ebay-pill>
\ No newline at end of file
+<ebay-pill pressed=true>Pill</ebay-pill>
```
ianmcburnie: Minor, but wondering why we are showing `pressed=true` instead of just `pressed`?

DylanPiercey:
I agree with @ianmcburnie that we should probably change `pressed=true` to `pressed` for the examples. Otherwise LGTM.

seangates:
@DylanPiercey @ianmcburnie Updated! Changed to simply `pressed`.

src/common/multiline-ellipsis/index.js
```diff
@@ -0,0 +1,59 @@
...
+function truncate(el) {
+    if (!el) {
+        return;
+    }
+
+    const text = el.innerText;
+
+    if (this.isTextOverflowing(el)) {
+        const checkFunc = i => {
+            el.innerText = text.substring(0, i);
```
DylanPiercey: I think it should be safe to switch to `textContent` instead of `innerText` which is going to be faster since it `innerText` requires layout to be completed.

seangates: Let me try it out (for the sake of speed).

My only worry is whether someone adds any hidden text to a button (not saying it's going to happen, but someone might throw something in there):
>`innerText` is aware of styling and won’t return the text of “hidden” elements, whereas `textContent` does.

seangates: So far, in my tests, I'm not seeing a material performance gain using `textContent`.

Also, from my tests, my worry about pulling in hidden content is actually unfounded. Therefore, it comes down to the performance difference between `innerText` and `textContent`, of which I can't confirm the improvement.

DylanPiercey: It can be a bit of a pain to profile this but `textContent` is definitely much faster in real world apps. When you have `innerText` reflows *must* happen before you can read the value which can cause jank. `textContent` on the other hand does not have this issue.

Relevant excerpt from the MDN link:
>Since innerText takes into account CSS styling, reading this property triggers a reflow, ensuring up-to-date computed style. textContent doesn’t. Reflows can be computationally expensive thus should be avoided when possible.

seangates: Updated per your feedback, @DylanPiercey.

seangates:
retest

Issue #456, opened 2018-10-24
Textbox: add icon capabilities


seangates:
# Feature Request

## Description

The textbox has been spec'd in DS4 & DS6 with prefix and postfix icons. These prefix and postfix icons should be added to the component and allowed to be configured by the developer.

## Context

This capability is already available in Skin for DS4 and DS6.

## Screenshots

![image](https://user-images.githubusercontent.com/105656/47453647-831d6300-d78a-11e8-903c-38eec06640a4.png)

ianmcburnie:
Merged to 1.3.0 branch.

PR #478, opened 2018-11-14
Enhance textbox with icon support (issue #456)

rachelvillamin:
## Description
- Added support to display an icon prefix or postfix for single-line textbox
- Updates <ebay-textbox> with icon-related attributes

## Context
Textbox with Icon is already supported in Skin for DS4 and DS6:
 ```
<span class="textbox">
    <svg aria-hidden="true" class="textbox__icon" focusable="false" width="16" height="16">
        <use xlink:href="#icon-mail"></use>
    </svg>
    <input class="textbox__control" type="text" placeholder="placeholder text" />
</span>

<span class="textbox textbox--icon-end">
    <input class="textbox__control" type="text" placeholder="placeholder text" />
    <svg aria-hidden="true" class="textbox__icon" focusable="false" width="16" height="16">
        <use xlink:href="#icon-mail"></use>
    </svg>
</span>
```
## Screenshots
screen shot 2018-11-16 at 2 46 07 pm

screen shot 2018-11-16 at 2 46 25 pm

### DIFF
--- a/src/components/ebay-textbox/README.md
+++ b/src/components/ebay-textbox/README.md
@@ -15,8 +15,8 @@ Name | Type | Stateful | Description
 `fluid` | Boolean | No |
 `multiline` | Boolean | No | renders a multi-line texbox if true
 `invalid` | Boolean | No | indicates a field-level error with red border if true
-`icon-name` | String | No | name of the icon from skin
-`icon-position` | String | No | Can be "prefix" / "postfix"
+`icon` | String | No | name of the icon from skin
+`icon-position` | String | No | Can be "prefix" / "postfix". If not specified or if the value passed is neither "prefix" nor "postfix",  the `icon-position` is set by default to "prefix"
 
 ## ebay-textbox Events
 
--- a/src/components/ebay-textbox/README.md
+++ b/src/components/ebay-textbox/README.md
@@ -15,6 +15,8 @@ Name | Type | Stateful | Description
 `fluid` | Boolean | No |
 `multiline` | Boolean | No | renders a multi-line texbox if true
 `invalid` | Boolean | No | indicates a field-level error with red border if true
+`icon-name` | String | No | name of the icon from skin
+`icon-position` | String | No | Can be "prefix" / "postfix"
 
 ## ebay-textbox Events
 
--- a/src/components/ebay-textbox/examples/09-ds6-suported-icon/template.marko
+++ b/src/components/ebay-textbox/examples/09-ds6-suported-icon/template.marko
@@ -1,0 +1,1 @@
+<ebay-textbox icon-name="messages" placeholder="email" />
--- a/src/components/ebay-textbox/examples/09-ds6-suported-icon/template.marko
+++ b/src/components/ebay-textbox/examples/09-ds6-suported-icon/template.marko
@@ -1,1 +1,0 @@
-<ebay-textbox icon-name="messages" placeholder="email" />
--- a/src/components/ebay-textbox/index.js
+++ b/src/components/ebay-textbox/index.js
@@ -2,7 +2,6 @@ const markoWidgets = require('marko-widgets');
 const emitAndFire = require('../../common/emit-and-fire');
 const processHtmlAttributes = require('../../common/html-attributes');
 const template = require('./template.marko');
-let initIcons = false;
 
 function getInitialState(input) {
     const classes = ['textbox__control'];
@@ -20,9 +19,9 @@ function getInitialState(input) {
         rootClass: rootClasses,
         style: input.style,
         classes,
-        initIcons: false,
-        iconName: input.iconName,
-        displayIcon: input.iconName && !Boolean(input.multiline),
+        icon: input.icon,
+        iconTag: input.iconTag && input.iconTag.renderBody,
+        displayIcon: input.icon && !Boolean(input.multiline),
         iconPostfix,
         tag: input.fluid ? 'div' : 'span',
         textboxTag: Boolean(input.multiline) ? 'textarea' : 'input',
@@ -32,13 +31,6 @@ function getInitialState(input) {
 
 function getTemplateData(state) {
     return state;
-}
-
-function init() {
-    if (!initIcons) {
-        initIcons = true;
-        this.setState('initIcons', initIcons);
-    }
 }
 
 function handleEvent(originalEvent, eventName) {
@@ -51,7 +43,6 @@ const handleFocus = function(originalEvent) { this.handleEvent(originalEvent, 'f
 const handleBlur = function(originalEvent) { this.handleEvent(originalEvent, 'blur'); };
 
 module.exports = markoWidgets.defineComponent({
-    init,
     template,
     getInitialState,
     getTemplateData,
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

File: src/components/ebay-textbox/template.marko [hunk 1]
```diff
@@ -1,4 +1,81 @@
+<if(data.initIcons)>
+<span style="display: none">
```
senthilp: @rachelvillamin Thanks for the PR. Does including all the icon tags here cause a maintenance problem down the line? Meaning, developers should remember to update this list when adding a new icon.

src/components/ebay-textbox/template.marko
```diff
@@ -1,4 +1,81 @@
...
+        <ebay-icon type="inline" name="add"/>
+        <ebay-icon type="inline" name="arrow-move" if(data.someFalsyProperty)/>
```
dasarianudeep: @rachelvillamin Rather than adding markup for every icon name, you can maintain an array and loop it to create multiple icons.
lets say
```
<for (name in iconnames_array)>
   <ebay-icon type="inline" name="${name}" />
</for>
```
Something like this wouldn't bloat the template.

ianmcburnie:
This seems suitable for v1.2.0 release. Changing target branch.

src/components/ebay-textbox/template.marko
seangates: This line should be tabbed (4 spaces) at the beginning of the line. (Whitespace is off.)

rachelvillamin: I had to declare the icons explicitly or the icons will not appear. I think this is a limitation on ebay-icon.

screen shot 2018-11-15 at 11 06 06 am

src/components/ebay-textbox/template.marko
rachelvillamin: I don't think this will work as the icon needs to be stamped initially as per ebay-icon documentation. See my comment above. And also screenshot below:
screen shot 2018-11-15 at 11 10 29 am

src/components/ebay-textbox/template.marko
senthilp: @DylanPiercey What do you think?

src/components/ebay-textbox/template.marko
DylanPiercey: All of the approaches mentioned so far will cause all icons to be shipped to the browser in the JS bundle. We need to use a transformer to avoid that, similar to what we did for the `ebay-menu` icon. https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu/transformer.js

src/components/ebay-textbox/template.marko
rachelvillamin: Not sure why I didn't get a lint error. This will be fixed in my next commit.

src/components/ebay-textbox/examples/09-ds6-suported-icon/template.marko
```diff
@@ -0,0 +1 @@
+<ebay-textbox icon-name="messages" placeholder="email" />
```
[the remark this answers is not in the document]
rachelvillamin: Sorry for the mispelling :P And I'm not sure why I didn't get a lint error. This will be fixed in my next commit for this and the other files as well.

src/components/ebay-textbox/template.marko
rachelvillamin: Cool. Thanks Dylan! I'll look at ebay-menu and follow the same approach :)

src/components/ebay-textbox/template.marko
rachelvillamin: Thanks Dylan. I will use a transformer. Declaring the ebay icons explicitly was the only way I could get it to work yesterday. I didn't know about transformers :) Thanks for your help!

rachelvillamin: hi @DylanPiercey,
I have implemented rendering the icons using transformer. It's working. Not sure if I properly implemented it though :P Need your feedback. Thanks!

src/components/ebay-textbox/index.js
```diff
@@ -2,17 +2,28 @@ const markoWidgets = require('marko-widgets');
...
 const processHtmlAttributes = require('../../common/html-attributes');
 const template = require('./template.marko');
+let initIcons = false;
 
 function getInitialState(input) {
     const classes = ['textbox__control'];
     if (input.fluid) {
         classes.push('textbox__control--fluid');
     }
+    const iconPostfix = input.iconPosition === 'postfix';
```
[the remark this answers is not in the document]
seangates: Using a string as the `input` is great, but you should also have a boolean for the `data` passed to the template.

E.g. both `data.iconPostfix` and `data.iconPrefix`.

But, one of them should be the default, which is basically all I was asking.

src/components/ebay-textbox/README.md
```diff
@@ -15,6 +15,8 @@ Name | Type | Stateful | Description
 `fluid` | Boolean | No |
 `multiline` | Boolean | No | renders a multi-line texbox if true
 `invalid` | Boolean | No | indicates a field-level error with red border if true
+`icon-name` | String | No | name of the icon from skin
```
seangates: This is now simply `icon` and not `icon-name`.

src/components/ebay-textbox/index.js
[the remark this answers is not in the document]
DylanPiercey: What do you all think of just having `icon-left=“icon-name”` and `icon-right=“icon-name”` and just erroring if they have both? (Alternatively could be `icon-prefix` and `icon-postfix`).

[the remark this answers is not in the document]
seangates: That's a great idea, actually. Two birds, one stone, as they say.

src/components/ebay-textbox/template.marko
rachelvillamin: This is done. I'm now using a transformer. Thanks Dylan for our help especially with the body-if-only work-around.

src/components/ebay-textbox/index.js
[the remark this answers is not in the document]
rachelvillamin: Didn't implement this as per the discussion in our sprint demo/retro last weekend.

src/components/ebay-textbox/README.md
rachelvillamin: Done :)

rachelvillamin:
Hi guys,
Need your help. Not sure why one of the test is failing:

screen shot 2018-11-20 at 2 51 55 pm

@DylanPiercey do you have any idea? This works in my local and also it was passing in travis-ci/pr build.

Thanks in advance :)

DylanPiercey:
Looks like it’s just a timing issue. Restarting the build and it should be good 🙂.

src/components/ebay-textbox/transformer.js
```diff
@@ -0,0 +1,35 @@
+const ebayUIAttributeTransformer = require('../../common/transformers/ebayui-attribute');
+
+// Transforms an `icon` attribute into an `<ebay-menu:icon>` tag
```
seangates: I think you mean:
```
// Transforms an `icon` attribute into an `<ebay-icon>` tag.

rachelvillamin: Done :)

Issue #19, opened 2018-03-05
New component: Badge


ianmcburnie:
Design doc:
https://ebay.invisionapp.com/share/CWN9X1JF6MA#/screens/276179313

Related issue: https://github.com/eBay/skin/issues/80

ianmcburnie:
Merged to 1.3.0 branch.

PR #479, opened 2018-11-15
New badge component (issue #19)

seangates:
## Description
- creates a new `<ebay-badge>`
- updates `<ebay-menu>` and `<ebay-button>` with badge attributes

## Context
From the Skin documentation on Badge:
>A badge is a visual indicator added to an element to convey quantity, newness or both. Badges are intended to remind a user of previous actions taken, or to alert them of new actions that they should consider.

The usage of badge can be standalone, in a menu, or on an icon button. The way it's been implemented means we can quickly and easily expose the badge in many use cases, giving developers wide latitude, but also helping them do it quickly for menu and icons.

## References
Completes #19 

## Screenshots
image


seangates:
@ianmcburnie The intent of `a11y-text` is to place that text inline for both a standalone badge component and the menu component. I waffled on whether or not the L10N folks would frown on a string that did not have the number included. I'll do what's best here, but maybe how I think about the two attributes simply comes from a different place.

Given the example in Skin, when a badge is placed inside an icon button the `aria-label` takes precedence (or so I thought).

seangates:
@ianmcburnie Okay, I've made updates based on your feedback. Please check and let me know.

marko.json
```diff
@@ -1,4 +1,14 @@
 {
+    "<ebay-badge>": {
+        "renderer": "./src/components/ebay-badge/index.js",
+        "@*": "expression",
+        "@html-attributes": "expression",
+        "@class": "string",
+        "@style": "string",
+        "@type": "string",
+        "@number": "expression",
+        "@a11y-text": "string"
```
ianmcburnie: The property is listed here as `a11y-text`, but the readme is asking for `aria-label`.

seangates: Oh, I think I forgot to remove that. Let me get it fixed.

seangates:
@ianmcburnie Fixed. One more pass, please?

src/components/ebay-button/browser.json
```diff
@@ -2,7 +2,10 @@
     "dependencies": [
         {
             "if-not-flag": "ebayui-no-skin",
-            "path": "@ebay/skin/button"
+            "dependencies": [
+                "@ebay/skin/actionable",
+                "@ebay/skin/button"
+            ]
         },
         "require: marko-widgets",
```
rachelvillamin: Hi Sean,
Do we need to declare a dependency on ebay-badge here? Or is it unnecessary?

seangates: Great callout. I believe I do. Will update.

seangates:
@rachelvillamin Please check again. 👍

seangates:
@ianmcburnie Updates! Please check once more.

PR #481, opened 2018-11-16
Dialog: fix scroll restoration with multiple quick renders.

### LINKED ISSUE #480: Dialog: Unable to scroll on body after quickly re-rendering dialog multiple times
# Bug Report

## Description
I’m using `ebay-dialog` full screen and (1) if I re-render dialog soon after it's first opened and (2) then close the dialog by clicking (X) button, scrolling on main page/body is still blocked. 

## Screenshots
Screenshot of body styles after has already been opened and then re-rendered again:
css_dialog_opened

Screenshot of body styles after clicking (X) button to close dialog:
css_dialog_closed

DylanPiercey:
## Description
Currently if you render multiple dialogs open at the same time (probably shouldn't do that) or you render the same dialog open really quickly (we're talking during the update of the dialog) it will cause the previous body styles to be in the scroll lock state. The effect is that the user won't be able to scroll any more.

## Context
The fix is simple which is to just check that we have not already prevent scrolling before we store the `previousStyles` or lock the scrolling.

## References
fixes #480

### DIFF
--- a/src/common/body-scroll/index.js
+++ b/src/common/body-scroll/index.js
@@ -18,30 +18,32 @@ module.exports = { prevent, restore };
  * scrolling.
  */
 function prevent() {
-    const { body } = document;
-    const { pageXOffset, pageYOffset } = window;
-    const { width, height, marginTop, marginLeft } = getComputedStyle(body);
-    let styleText = 'position:fixed;overflow:hidden;';
-    previousPosition = [pageXOffset, pageYOffset];
-    previousStyles = body.getAttribute('style');
-    styleText += `height:${height};`;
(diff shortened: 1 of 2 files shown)


ianmcburnie:
Thanks. We could potentially put this out as a patch release, but as it's seems like an edge case and would require a round of QA, I'm happy to bundle it in with v1.2.0.

DylanPiercey:
@ianmcburnie I think it's fine if this waits until 1.2.0. I've chatted with the developer who was having the issue and the solution we have come up with no longer causes this edge case to appear.

Still think we should address it though 😄.

Issue #67, opened 2018-03-23
New Component: Tooltip


senthilp:
https://ebay.invisionapp.com/share/CWN9X1JF6MA#/screens/299234476

Some additional thoughts:

- The component should allow the application to suggest a position to open the tooltip, but use JS to override that in order to keep the tooltip on page.
- A11y guidelines:
    - Tooltip: https://ebay.gitbooks.io/mindpatterns/content/disclosure/tooltip.html
    - Infotip: https://ebay.gitbooks.io/mindpatterns/content/disclosure/bubblehelp.html
    - Tourtip: Guidelines need to be established. They are unique since they are usually meant to be open when the page is loaded and grab the user's focus.
- Based on my implementation of tourtips in the past
    - It looks bad if there are many tourtips open on the page at the same time. They can even overlap each other. It would be good to add some management to prevent this.
    - Tourtips usually only show up once and then the user never sees them again (using local storage).  We could create this functionality to prevent applications from having to duplicate it.  Applications will probably also want ways to reset the timeout period.

ianmcburnie:
This task will be unblocked by first pre-release of skin 7.0.0. At the time of writing, DS4 tooltip is done in skin, and that would be enough to start this task.

seangates:
Completed in #499.

PR #499, opened 2018-12-14
Tooltip Component

seangates:
## Description
- creates a new tooltip component

## Context
The Tooltip component is actually the conglomeration of three components: tooltip, infotip, and tourtip. Because they all use very similar styles and functionality, they have share quite a bit of code, and therefore, are used with a single component API.

Also, because the accessibility requirements change between each, we have simplified the API where necessary, in order to enable developers to more easily focus on the content.

## References
Completes #67 

## Screenshots
image
image
image


seangates:
@ianmcburnie Squashed, just for you!

seangates:
@DylanPiercey @ianmcburnie I realized today that the README's are kinda messed up. I'll get them fixed.

src/components/components/ebay-tooltip-base/index.js
```diff
@@ -0,0 +1,173 @@
...
+}
+
+function handleTooltipClose() {
+    this.setState('expanded', false);
+    this.expander.collapse();
+    emitAndFire(this, 'tooltip-close');
+}
+
+function alignOverlay() {
+    const overlay = this.el.querySelector(`.${this.state.type}__overlay`);
```
DylanPiercey: Since all of these styles are now static it would be best we just did this mapping in `getTemplateData`. It will simplify the logic (no need to re apply this on init or expand) and will mean that it will work with server side rendering for tooltips that should open by default.

src/components/components/ebay-tooltip-base/index.js
```diff
@@ -0,0 +1,173 @@
...
+
+    if (!overlay) {
+        return;
+    }
+
+    if (this.state.styleTop || this.state.styleLeft || this.state.styleRight || this.state.styleBottom) {
```
DylanPiercey: Is overriding the positioning something we currently have a use case for? I feel that we have covered the cases we need too and that exposing this API allows users to deviate from the DS and is a little clunky.

src/components/components/ebay-tooltip-overlay/index.js
```diff
@@ -0,0 +1,21 @@
+const emitAndFire = require('../../../common/emit-and-fire');
+const template = require('./template.marko');
+
+function getInitialState(input) {
```
DylanPiercey: We can just omit `getInitialState` and `getTemplateData` if they just return the first argument.

src/components/components/ebay-tooltip-overlay/template.marko
```diff
@@ -0,0 +1,34 @@
+<div
+    class=`${data.type}__overlay`
```
DylanPiercey: Since we are not transpiring the Marko templates with babel this template literal will cause issues in older browsers if teams do not have babel configured to run against Marko files (only in Marko 3 and older versions of Marko 4).

Here I think we should use the Marko specific syntax `class="${data.type}__overlay"` instead until we upgrade to Marko 4.

src/components/components/ebay-tooltip-overlay/template.marko
```diff
@@ -0,0 +1,34 @@
...
+    role=(data.type === 'tourtip'?'region':'tooltip')
+    aria-labelledby=(data.type === 'tourtip'?`tourtip-label-${widget.id}`:'')
```
DylanPiercey: Similar to the above template literal comment, there are a few of these that should probably change.

src/components/ebay-infotip/index.js
```diff
@@ -0,0 +1,44 @@
+const processHtmlAttributes = require('../../common/html-attributes');
+const emitAndFire = require('../../common/emit-and-fire');
+const template = require('./template.marko');
+
+function getInitialState(input) {
+    const iconTag = input.iconTag;
```
DylanPiercey: I would prefer that we return a new object here instead of mutating input. By returning an object it is clear which state is actually going to be passed to the template and this would make it consistent with the rest of the components in the repo. Mutating the input can also cause undesired effects when rerendering.

src/components/ebay-infotip/index.js
```diff
@@ -0,0 +1,44 @@
...
+    input.hostSelector = '.infotip__host';
+    input.overlaySelector = '.infotip__overlay';
+    input.iconTag = iconTag && iconTag.renderBody;
+    input.expanded = false;
+    input.expandInit = false;
+
+    return input;
+}
+
+function getTemplateData(state) {
```
DylanPiercey: Similar to above, we should be able to remove this `getTemplateData` function.

src/components/ebay-infotip/template.marko
```diff
@@ -0,0 +1,32 @@
...
+        class="infotip"
+        ${data.htmlAttributes}
+        w-on-expander-expand='handleExpand'
+        w-on-expander-collapse='handleCollapse'
+        w-bind>
+        <button
+            class='infotip__host icon-btn'
+            aria-describedby=`infotip-${widget.id}`
+            aria-label=data.ariaLabel>
+            <div if(data.icon) body-only-if(true)>
```
DylanPiercey: Here we should replace the `<div if(data.icon) body-only-if(true)>` with `<if(data.icon)>` as they achieve the same output.

src/components/ebay-infotip/test/test.browser.js
```diff
@@ -0,0 +1,47 @@
+const sinon = require('sinon');
+const expect = require('chai').expect;
+const testUtils = require('../../../common/test-utils/browser');
+const renderer = require('../');
+
+describe('given the default infotip', () => {
```
DylanPiercey: We should probably test for that the tooltip is visible instead of just testing for the emitted events.

src/components/ebay-tooltip/index.js
```diff
@@ -0,0 +1,36 @@
...
+    input.htmlAttributes = processHtmlAttributes(input);
+    input.hostSelector = '.tooltip__host';
+    input.overlaySelector = '.tooltip__overlay';
+    input.expanded = false;
+    input.expandInit = false;
+
+    return input;
+}
+
+function getTemplateData(state) {
```
DylanPiercey: Should be able to remove this `getTemplateData` function.

src/components/ebay-tooltip/index.js
```diff
@@ -0,0 +1,36 @@
+const processHtmlAttributes = require('../../common/html-attributes');
+const emitAndFire = require('../../common/emit-and-fire');
+const template = require('./template.marko');
+
+function getInitialState(input) {
```
DylanPiercey: Prefer to return a new object here as per above comment.

src/components/ebay-tourtip/index.js
```diff
@@ -0,0 +1,39 @@
+const processHtmlAttributes = require('../../common/html-attributes');
+const emitAndFire = require('../../common/emit-and-fire');
+const template = require('./template.marko');
+
+function getInitialState(input) {
```
DylanPiercey: Prefer to return an object here as per above comment.

src/components/ebay-tourtip/index.js
```diff
@@ -0,0 +1,39 @@
...
+    input.location = input.location || 'bottom';
+    input.htmlAttributes = processHtmlAttributes(input);
+    input.hostSelector = '.tourtip__host';
+    input.overlaySelector = '.tourtip__overlay';
+    input.expanded = true;
+
+    return input;
+}
+
+function getTemplateData(state) {
```
DylanPiercey: Should be able to remove this `getTemplateData` function.

src/components/ebay-tourtip/index.js
```diff
@@ -0,0 +1,39 @@
...
+    return state;
+}
+
+function init() {
+    this.setState('expandInit', true);
```
DylanPiercey: What is the purpose of waiting for the component to mount before making it expanded?

seangates: It's because it's location is wrong and jumps after it's loaded.

src/components/components/ebay-tooltip-base/index.js
seangates: This is really for fine tuning. We could remove it, but I'm 100% sure developers will ask for it.

src/components/components/ebay-tooltip-base/index.js
seangates: `this.el` is `undefined` should I be using `document.querySelector` (or something else)?

seangates: Yeah, I don't understand how this will work in getTemplateData, because I need a DOM element which has already been added to the page. getTemplateData doesn't seem to have any Marko DOM element references for me to tack onto. Am I missing something?

DylanPiercey: `this` is not available in `getTemplateData`. In this case you should only need `state` which you can access like:

```
getTemplateData(state, input) { ... }
```

src/components/ebay-infotip/index.js
DylanPiercey: We are using `Object.assign` elsewhere, so we can use that here also.

src/components/components/ebay-tooltip-base/index.js
DylanPiercey: I'm saying we can create the style data in `getTemplateData` and forward that to the template and assign the `style` attribute there, rather than directly mutating the overlay.

seangates: As a string/object/array. Okay, I think I get it. I was moving that direction.

src/components/ebay-tourtip/index.js
seangates: You know, I think this is mitigated by my refactor. It appears I don't need it anymore. I'll look at this in the morning.

src/components/ebay-infotip/test/test.browser.js
seangates: Yeah, I was having some trouble with this earlier. I'll look at it in the morning.

seangates:
@DylanPiercey I've a little work to do on tests, and removing the `expandInit` for the Tourtip, but this should be in a much better place.

src/components/components/ebay-tooltip-overlay/template.marko
```diff
@@ -0,0 +1,35 @@
+<div
+    class="${data.type}__overlay"
+    role=(data.type === 'tourtip'?'region':'tooltip')
```
ianmcburnie: Only classic tooltip should have role of tooltip, i.e. not the Infotip.

src/components/components/ebay-tooltip-base/index.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function init() {
+    const isHostPresent = this.el.querySelector(`.${this.state.type}__host`);
+
+    if (isHostPresent) {
+        this.expander = new Expander(this.el, {
+            hostSelector: `.${this.state.type}__host`,
+            contentSelector: `.${this.state.type}__overlay`,
+            focusManagement: this.state.type === 'tourtip' ? null : 'focusable',
```
ianmcburnie: Let's not do any focus management when this thing opens. Let's keep focus on the host. A problem with setting focus on the close button of an Infotip is that if a screen reader navigates forward they will miss the content of the Infotip (due to the close button being at the end).

seangates: 👍  That's an easy change.

src/components/components/ebay-tooltip-overlay/template.marko
seangates: Cool. Easy change.

seangates: So ... what role should the infotip have?

ianmcburnie: No role.

seangates:
@DylanPiercey Wondering if we can forego the test change, and merge this? Or do you feel that test change is important enough to hold up merging? (I only ask because I've covered everything else, and I need to move on to some other things.)

DylanPiercey:
@seangates I'm not entirely sure yet why the span isn't receiving the `on-expander-expand` event [here](https://github.com/eBay/ebayui-core/pull/499/files#diff-882aa8a2032cf4088fc8973d537654e1R7). However I do think that we should change the top level tooltips to just listen to `tooltip-expand` and `tooltip-close` from the `ebay-tooltip-base` directly. I think this cleans things up and it also seems to fix the issue and that we should update all top level components to receiving the events from the tooltip base, instead of relying on receiving it from makeup-expander (only the base needs to do that).

seangates:
@DylanPiercey I'll try that. The events were likely just not set up correctly in the reshuffle. I'll let you know.

ianmcburnie:
Can you add a reference to tooltip from the main README. Easy to miss!

src/components/ebay-tourtip/README.md
```diff
@@ -0,0 +1,43 @@
+# ebay-tourtip
+
+## ebay-tourtip Example
+
+```marko
+<ebay-tourtip a11y-close-text="Dismiss tourtip" aria-label="Important information">
```
ianmcburnie: Minor, but I don't think aria-label should be present in this example.

seangates:
@ianmcburnie Updated. Also merged in recent 2.0.0 branch updates.

seangates:
@DylanPiercey This should do it. 🤞

seangates:
I've found an issue with the `ebay-tooltip` and have been working to resolve it. In a nutshell, the host is on a span and not on a focusable element, making the accessibility incorrect.

Additionally, there was feedback provided through a demo to the team today, as follows (rough notes):

- Infotip close button focus returns to host button
- Change tooltip default icon
- Check aria-describedby only for tooltip
- Location changed “pointer”
- Remove custom icon button example
- Default => Button tooltip
- Pull out styles css from flyout location
- Attribute to disable hover on tooltip

seangates:
@ianmcburnie @DylanPiercey All buttoned up and ready to fly! 🦅

seangates:
😴  now.

@ianmcburnie and @DylanPiercey, please take a moment for review once more.

@mkchang Feel free to chime in, too!

src/components/components/ebay-tooltip-base/index.js
```diff
@@ -30,6 +32,11 @@ function init() {
         if (!hostAriaDescribedBy && this.el.parentElement) {
             host.setAttribute('aria-describedby', `${this.el.parentElement.id}-overlay`);
         }
+
+        if (focusableEls[0]) {
```
DylanPiercey: If the component is rerendered with new content then this list of focusable elements can become out of date.

yarn.lock
```diff
@@ -5,21 +5,18 @@
 "@babel/code-frame@7.0.0-beta.51":
   version "7.0.0-beta.51"
   resolved "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.51.tgz#bd71d9b192af978df915829d39d4094456439a0c"
-  integrity sha1-vXHZsZKvl435FYKdOdQJRFZDmgw=
```
PrashantAshok: Curios, why is the `integrity` getting removed? Maybe a older yarn version?

mkchang: Yeah I believe yarn 1.10+ adds the integrity hashes, I've been switching to 1.13 for ebayui-core work

seangates: Let me double check.

seangates: Yeah, that's way too many changes. Let me clean it up.

src/components/components/ebay-tooltip-base/index.js
seangates: Okay. Let me work on that.

yarn.lock
seangates: Good catch there, @PrashantAshok! It should all be sorted now.

seangates:
@ianmcburnie @DylanPiercey Okay, I'm hoping this is the last change. I move the expander initialization to `onRender`, and also added the `hostContainerClass` to the options of the expander.

Please let me know asap if you have questions.

src/components/components/ebay-tooltip-base/index.js
```diff
@@ -30,10 +29,10 @@ function onRender() {
     const isTooltip = this.state.type === 'tooltip';
 
     if (this.host) {
-        this.expander = new Expander(this.el, {
+        this.expander = new Expander(this.el.querySelector(`.${this.state.type}`), {
```
DylanPiercey: should probably use `this.el.getElementsByClassName(this.state.type)[0]` which will be much faster and maybe a bit more clear in this case?

seangates:
@ianmcburnie This should be good to go now!

ianmcburnie:
LGTM!!! :-)

Issue #490, opened 2018-12-06
badge: we display negative, decimal numbers and 0- pls do the validation to check all nos are positive whole numbers


hramakumar:
# Bug Report

## Description
we display negative, decimal numbers and 0- pls do the validation to check all nos are positive whole numbers
repro:
go to https://go/ebayui/test
look at the badge section
search for labels badge_negative and badge_0
## Screenshots
![image](https://user-images.githubusercontent.com/13157115/49605665-fc8da100-f945-11e8-81dc-5db5851229b5.png)

it also accepts decimal numbers. WIll post that screenshot later

ianmcburnie:
Fixed by @BhavinPatel04 and merged to 2.0.0 branch. Closing.

PR #502, opened 2018-12-19
Added a check for badge number to be a positive whole number

BhavinPatel04:
## Description
Updated the check condition so the badge number is always a positive whole number

## Context
Badge number always has to be a positive whole number

## References
issue https://github.com/eBay/ebayui-core/issues/490


src/components/ebay-badge/index.js
```diff
@@ -4,7 +4,7 @@ const template = require('./template.marko');
 function getInitialState(input) {
     const number = Number(input.number);
     const imgRole = (input.type !== 'menu' && input.type !== 'icon');
-    const showBadge = !isNaN(number);
+    const showBadge = !isNaN(number) && Number.isInteger(number) && number >= 0;
```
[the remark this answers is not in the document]
BhavinPatel04: Ok. `string` numbers are converted to `Number` in line 5 right? Also, the requirement is to only show whole numbers I think.
Yes, I too think we should add a warning in these situations for the dev. But showing zero would be misleading I think.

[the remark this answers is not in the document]
seangates: >Curious what others think on just making the badge number zero if it is negative?

Yes, we should change it to `0` if the number is negative.

Also, we should round decimal numbers to their nearest whole number.

Some examples to go by:

```
3 => 3
0 => 0
-5 => 0
5.7 => 6
3.1 => 3
139 => 99+
```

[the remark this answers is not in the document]
seangates: (Sorry, I needed to edit my response as I didn't respond how I wanted. Please see above.)

seangates:
Please also add a test for any new logic you add to the component.

src/components/ebay-badge/index.js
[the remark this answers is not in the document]
BhavinPatel04: Cool. I shall edit the fix to round the decimal numbers.

src/components/ebay-badge/index.js
```diff
@@ -2,9 +2,9 @@ const processHtmlAttributes = require('../../common/html-attributes');
 const template = require('./template.marko');
 
 function getInitialState(input) {
-    const number = Number(input.number);
+    const number = Math.round(input.number);
     const imgRole = (input.type !== 'menu' && input.type !== 'icon');
-    const showBadge = !isNaN(number);
+    const showBadge = !isNaN(number) && number >= 0;
```
DylanPiercey: Technically we could change this to just `const showBadge = number >= 0;` since `NaN` is not greater than, less than or equal to any number.

It might be more clear to leave it as is though, curious what others think?

seangates: We could certainly simplify it. Simplicity usually wins the day.

ianmcburnie:
Hi @BhavinPatel04 I updated the target branch to 2.0.0 (we are targeting this fix for v2 release). It looks like you'll need to update your branch.

BhavinPatel04:
@ianmcburnie, I have updated the branch with 2.0.0 code

src/components/ebay-badge/index.js
BhavinPatel04: I have changed this logic as per @DylanPiercey' s comments

src/components/ebay-badge/index.js
```diff
@@ -2,9 +2,9 @@ const processHtmlAttributes = require('../../common/html-attributes');
 const template = require('./template.marko');
 
 function getInitialState(input) {
-    const number = Number(input.number);
+    const number = Math.round(input.number);
     const imgRole = (input.type !== 'menu' && input.type !== 'icon');
-    const showBadge = !isNaN(number);
+    const showBadge = number >= 0;
```
DylanPiercey: I just noticed this, don't we want to hide it when `number` is `0`?

BhavinPatel04: Ah.. you are right. fixing it with `number > 0`

seangates: Good catch!

PR #505, opened 2019-01-10
Carousel: Fix styles when minified with yui-compressor

### LINKED ISSUE #504: Touch carousel does not work after deploying
# Bug Report

## Description
The touch version of the carousel works locally but not after deploying. It seems the compressor the resource server uses affects some css that is needed. See the diff below.

After reproducing locally, applying the correct uncompressed style inline fixes it.

## Screenshots
image

DylanPiercey:
## Description
Currently the resource server improperly minifies the css for the carousel because it uses `yui-compressor` under the hood. This change adds spacer comments to ensure that the `@supports` rule can still be parsed properly after minification.

## References
Fixes #504

### DIFF
--- a/src/components/ebay-carousel/style.less
+++ b/src/components/ebay-carousel/style.less
@@ -150,10 +150,11 @@
  * Not all browsers that have implemented scroll snapping have done so in a way that we can leverage.
  * Currently the below media query ensures that the browser supports exactly what we need.
  */
+/* autoprefixer: off */
 @supports (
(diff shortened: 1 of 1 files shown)


ianmcburnie:
We already have a 1.3.1 release. This should be going into a 1.3.2 branch (cut from master).

ianmcburnie:
I merged this to 1.3.2 branch. For some reason GH didn't auto close it... so closing manually.

PR #509, opened 2019-01-14
Fix issue with YUI minification

DylanPiercey:
## Description
This is a followup PR to #505. The previous PR appeared to be working but after testing with @scttdavs it turns out it was not enough. We came up with another hack/solution which does appear to work perfectly by creating two preserved comments with a space in the middle.

It also looks like I forgot to update the `@supports` in the `touch` styles, so this PR includes that as well.
## References
Fixes #505

### DIFF
--- a/src/components/ebay-carousel/style-touch.less
+++ b/src/components/ebay-carousel/style-touch.less
@@ -22,14 +22,16 @@
     }
 }
 
-@supports not (
-    (-webkit-scroll-snap-coordinate: 0 0) or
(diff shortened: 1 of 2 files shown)


DylanPiercey:
Hold off on this, looks like this change is not working either 😢.

DylanPiercey:
Alright the PR should be good to go now. @scttdavs found a work around that has been confirmed to be working and I've updated the PR accordingly.

DylanPiercey:
Just confirmed on my local. This change does work (both for touch and desktop styles).

Issue #500, opened 2018-12-17
Replace occurrences of background icons (span tag) with foreground icons (svg tag)


ianmcburnie:
Any internal dependency on the skin background icon module causes a perf hit due to the significant CSS payload it imports. Let's convert any occurrences of background icons (i.e. the span) tag to foreground icons (i.e. the svg tag).

ianmcburnie:
Let's see if we can fix this in a 1.3.x or 1.4 release. Otherwise, 2.0.0 it will have to be.

seangates:
Completed in #511.

PR #515, opened 2019-01-28
1.4.0 branch merge before release

seangates:
## Description
- changes for icon (removing inclusion for the docs) in the 1.4.0 branch

## References
See #500


seangates:
Since we're asking Inception to update anyway, we might as well include 6.3.3. Easy enough.

seangates:
@ianmcburnie That's a good callout. Let me see if I can find a good place for it. Possibly in the icon section as a note about background icons?

PR #527, opened 2019-02-02
Bumped Skin to v6.3.5

ianmcburnie:
I feel like we should be using the tilde, i.e `~6.3.5` instead of `6.3.5`. Inception use the tilde for skin too. Any objections if I change it?


ianmcburnie:
Actually, I'm incorrect. Inception is not using tilde in latest v3 branch.

PR #549, opened 2019-02-13
529 clone combobox to listbox

### LINKED ISSUE #529: Convert ebay-combobox to ebay-listbox
The `ebay-combobox` component is going to be repurposed from a readonly "custom select" to an editable combobox. However, we still need a "custom select", and as the current logic of `ebay-combobox` matches up very closely with the W3C's new listbox pattern (an alternative way of creating a custom select), let's clone what we have and call it `ebay-listbox`.

I think the only logic that will need updating is that it should use active-descendant type kb navigation instead of roving-tabindex kb navigation.

seangates:
## Description
- creates a new listbox based on the original combobox
- now uses the `makeup-active-descendant` to conform to A11Y specs

## Context
As part of the effort to solidify the A11Y components and their varied differences, we've created the listbox.

## References
Fixes #529


ianmcburnie:
Looking good except missing the following:

1. `aria-activedescendant` attribute is not being applied to element with role=listbox
1. `aria-selected` is not being applied to elements with role=option

I suspect both these issues might be caused by improper elements passed to the activeDescendant constructor. Please check the example in skin 7 branch.

PR #552, opened 2019-02-15
Adding missing icon symbols;

seangates:
## Description
- re-run of `scripts/import-svg`

## Context
Some SVG symbols were missing.


src/components/ebay-icon/symbols/left-justified/ds4.marko
```diff
@@ -1,3 +1,3 @@
-<symbol id="left-justified" viewBox="0 0 24 17">
+<symbol id="icon-left-justified" viewBox="0 0 24 17">
```
ianmcburnie: Weird that "icon-" prefix was missing from many of these. But this is correct now, right?

seangates: Yes, the icon prefix in the ID is needed. Certainly disconcerting it was missing.

PR #555, opened 2019-02-15
Checkbox and radio now use <ebay-icon>;

### LINKED ISSUE #554: Checkbox and Radio are always showing DS6 icons
Details to come... (but Sean is working on a patch right now for 1.x)

seangates:
## Description
- now uses `<ebay-icon>` instead of directly using SVG symbols

## Context
This is a bug because the DS6 checkbox was being used for DS4 and could not be switched out.

## References
Fixes #554

### DIFF
--- a/src/components/ebay-checkbox/browser.json
+++ b/src/components/ebay-checkbox/browser.json
@@ -7,6 +7,7 @@
                 "@ebay/skin/field"
             ]
         },
+        "../../../src/components/ebay-icon",
         "require: marko-widgets",
         "require: ./index.js"
     ]
--- a/src/components/ebay-checkbox/template.marko
+++ b/src/components/ebay-checkbox/template.marko
@@ -9,11 +9,7 @@
         disabled=data.disabled
         ${data.htmlAttributes} />
     <span class="checkbox__icon" hidden>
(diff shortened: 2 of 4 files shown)


ianmcburnie:
Shouldn't the icons be blue for ds4?

screen shot 2019-02-15 at 11 50 29 am

seangates:
We can (and probably should) default them to blue, yes. We could do that in either Skin (forced for all users of Skin) or within the component. I am fine with either.

seangates:
Within seems more correct, though, the more I think about it.

ianmcburnie:
I'll add the fill back to the skin symbols. Not sure when/why those went missing, but an easy fix. For now, rather than waiting to run the script again, you can just manually update the coreui symbols I guess.

seangates:
Okay, I can do that as part of this since it's a patch for 1.x. Sounds good!

seangates:
Having gone through conversations in https://github.com/eBay/skin/pull/561 and offline, this PR will stand as is. Updating colors here will have the same issue as updating skin. Colors have already been customized by teams using this component, and we would break them in 1.x, which is not correct.

Therefore, we only want to fix the icon, and not the color. This PR is now as is.

PR #560, opened 2019-02-19
Fix providing id attribute to ebay button

### LINKED ISSUE #559: Button: does not support passing button id value
# Bug Report

## Description
The `<ebay-button>` component does not allow an ID to be passed in. This is similar to the textbox input issue: #520.

## Workaround
No known workaround.

## Screenshots
screen shot 2019-02-19 at 1 23 41 pm

DylanPiercey:
## Description
Similar to https://github.com/eBay/ebayui-core/pull/558 however accessing `widget.id` is not possible on the same tag or above a `w-bind`. This PR reaches into Marko 3's internals somewhat to get what the id would be before that point.

Ultimately this allows custom `id`'s to be set on `ebay-button`.

## References
Fixes #559

### DIFF
--- a/src/components/ebay-button/index.js
+++ b/src/components/ebay-button/index.js
@@ -3,9 +3,10 @@ const emitAndFire = require('../../common/emit-and-fire');
 const eventUtils = require('../../common/event-utils');
 const processHtmlAttributes = require('../../common/html-attributes');
 const observer = require('../../common/property-observer');
+const getWidgetId = require('../../common/get-marko-3-widget-id');
 const template = require('./template.marko');
 
-function getInitialState(input) {
+function getInitialState(input, out) {
     const href = input.href;
     const priority = input.priority || 'secondary';
     const size = input.size;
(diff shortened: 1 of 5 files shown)


src/components/ebay-button/index.js
```diff
@@ -77,6 +78,7 @@ function getInitialState(input) {
 
     return {
         htmlAttributes: processHtmlAttributes(input),
+        id: input.id || getWidgetId(out),
```
seangates: Are there any possibility of binding issues? Does the custom ID get used as the Marko v3 element's ID through the lifecycle? Or, does it matter?

DylanPiercey: It's not possible to access the id in any pre render lifecycle method so this won't cause any issues. There is one case that I didn't account for in my hacky utility that I am fixing now which is why I put this up as a draft.

After the CI is passing this should be okay.

seangates: I pulled the branch and it doesn't seem to be working for me. I simply did this:
```
<ebay-button priority="primary" id="button-test">text test</ebay-button>
```

seangates: Never mind. I was looking at the wrong element in my inspector!

seangates:
@ianmcburnie Please double-check this when you get a moment.

PR #565, opened 2019-02-26
Carousel: Disable scroll snapping for FireFox

DylanPiercey:
## Description
Currently in Firefox the carousel component shows scroll bars when scroll snapping is supported. Because there is not straight forward way to hide this scroll bar (perhaps we can look at this more in the future) this PR opts to disable to scroll snapping in FireFox and fall back to transition based implementation used in browsers without scroll snap support (IE/Edge).

### DIFF
--- a/src/components/ebay-carousel/style.less
+++ b/src/components/ebay-carousel/style.less
@@ -148,6 +148,16 @@
  * Not all browsers that have implemented scroll snapping have done so in a way that we can leverage.
  * Currently the below media query ensures that the browser supports exactly what we need.
  */
+
+@supports (-moz-appearance:none) {
+    // Explicitly disable support in firefox since we can't easily hide the scrollbar.
+    .carousel__list {
+        &&&& {
+            overflow-x: visible;
+        }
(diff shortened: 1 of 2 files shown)


src/components/ebay-carousel/style.less
```diff
@@ -148,6 +148,16 @@
  * Not all browsers that have implemented scroll snapping have done so in a way that we can leverage.
  * Currently the below media query ensures that the browser supports exactly what we need.
  */
+
+@supports (-moz-appearance:none) {
+    // Explicitly disable support in firefox since we can't easily hide the scrollbar.
+    .carousel__list {
+        &&&& {
```
seangates: Why is this repeated in this way? Is that a FireFox feature (bug) to actually match this selector?

DylanPiercey: It is basically just `.carousel__list.carousel__list.carousel__list.carousel__list`. I'm just forcing a certain specificity instead of using `!important` so that it can be overwritten if needed.

PR #566, opened 2019-02-26
Combobox: new editable combobox

### LINKED ISSUE #22: Combobox: add ARIA 1.1 autocomplete list functionality
Combobox pattern:

https://www.w3.org/TR/wai-aria-practices-1.1/#combobox

For first version let's implement autocomplete with manual selection.

Screen Shot 2019-03-27 at 9 42 01 AM

Example:

https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html

Helpful plugins:

https://github.com/makeup-js/makeup-active-descendant

seangates:
## Description
- add editable combobox

## Context
Adds an editable combobox

## References
Closes #22 

## Screenshots
image


seangates:
Just bumped Skin's version to prerelease v7.1.0-2 as a companion PR (https://github.com/eBay/skin/pull/598) was merged to v7.1.0 branch over there.

seangates:
@ianmcburnie It should. Let me create a quick test and confirm.

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,179 @@
...
+        if (this.activeDescendant) {
+            this.activeDescendant.destroy();
+        }
+
+        if (this.expander) {
+            this.expander.cancelAsync();
+        }
+    },
+    onDestroy() {
+        this.activeDescendant.destroy();
```
DylanPiercey: We may need to add the same guards for `this.activeDescendant` and `this.expander` in the `onDestroy` as well since you could render an empty combobox, and then remove it which would cause an error.

I had a similar issue in the dialog and opted to create a `_release` method (https://github.com/eBay/ebayui-core/blob/8e2d0e4987e1b7245097227429b2bed301fac36f/src/components/ebay-dialog/index.js#L134) which was then called in both places: https://github.com/eBay/ebayui-core/blob/8e2d0e4987e1b7245097227429b2bed301fac36f/src/components/ebay-dialog/index.js#L28 & https://github.com/eBay/ebayui-core/blob/8e2d0e4987e1b7245097227429b2bed301fac36f/src/components/ebay-dialog/index.js#L33

Might make sense to do something similar here.

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,179 @@
...
+            || '';
+
+        emitAndFire(this, 'combobox-change', {
+            currentInput: this.state.currentValue,
+            selectedOption: this.state.options[this.state.selectedIndex],
+            selectedValue,
+            options: this.state.options
+        });
+    },
+    filterOptionsDisplay() {
```
DylanPiercey: I feel that the name of this method no longer aligns with what it does (simply seems to expand and collapse now).

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,179 @@
...
+                    activeDescendantClassName: 'combobox__option--active',
+                    autoInit: selectedIndex === -1 ? -1 : 0,
+                    autoReset: -1
+                }
+            );
+
+            this.expander = new Expander(this.el, {
+                autoCollapse: true,
+                expandOnFocus: true,
+                expandOnClick: this.state.readonly && !this.state.disabled,
```
DylanPiercey: If `this.state.disabled` is true, do we still need to even setup the expander?

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,179 @@
...
+    },
+    handleComboboxKeyUp(originalEvent) {
+        const selectedEl = this.getEl('options').querySelector('.combobox__option--active');
+        let newValue = this.getEl('input').value;
+
+        eventUtils.handleUpDownArrowsKeydown(originalEvent, () => {
+            if (!this.expander.isExpanded() && this.getEls('option').length > 0) {
+                this.expander.expand();
+            }
+            this.moveCursorToEnd();
```
DylanPiercey: Would this cause issues where the user is trying to go back and edit the input and it would always move them to the end?

ianmcburnie:
@seangates The local combobox demo page does not work for me. I see a JS runtime error.

marko.json
```diff
@@ -89,6 +89,29 @@
...
+        "@borderless": "boolean",
+        "@disabled": "boolean",
+        "@options <option>[]": {
+            "@*": "expression",
+            "@html-attributes": "expression",
+            "@class": "string",
+            "@style": "string",
+            "@value": "string",
+            "@text": "string",
+            "@selected": "boolean"
```
ianmcburnie: Let's use `autofill` instead of selected

ianmcburnie:
For the 3rd example, "options with values" when I select  "Option 2 - July Sale (paused)", the value of the combobox appears to be set to the text, not the value.

Screen Shot 2019-04-12 at 1 52 54 PM
Screen Shot 2019-04-12 at 1 53 07 PM

ianmcburnie:
@seangates I've been thinking about that selected/autofill use case that you've added. Won't it conflict with the `value` attribute of the combobox itself? How about if we just use value instead of selected/autofill attribute?

marko.json
ianmcburnie: Actually, see this comment: https://github.com/eBay/ebayui-core/pull/566#issuecomment-482728718

src/components/ebay-combobox/index.js
seangates: This is only for the up and down arrows, which usually move you all the way to the beginning or all the way to the end, respectively. If we didn't have this, when a user would press the up arrow the cursor would always move to the very beginning. Maybe I have to handle the up and down arrows separately in this use case?

FWIW: Right and left arrows would still allow them to traverse the content as normal.

src/common/event-utils/index.js
```diff
@@ -11,6 +11,19 @@ function handleKeydown(keyCodes, e, callback) {
     }
 }
 
+// inverse of found keys
+function handleInverseKeydown(keyCodes, e, callback) {
```
DylanPiercey: This is not a huge deal, but I feel that this function should be called something like: `handleNotKeydown`, since keys other than keydown keys are not actually the inverse or opposite of the key down keys.

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,179 @@
...
+    moveCursorToEnd() {
+        const currentInput = this.getEl('input');
+
+        if (currentInput) {
+            const len = currentInput.value.length;
+            currentInput.setSelectionRange(len, len);
+        }
+    },
+    handleComboboxKeyUp(originalEvent) {
+        const selectedEl = this.getEl('options').querySelector('.combobox__option--active');
```
DylanPiercey: To avoid using a querySelector, I'd prefer to see:
```
const selectedEl = this.getEls('option')[this.state.selectedIndex];
```

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,179 @@
...
+        this.emitChangeEvent();
+        this.expander.collapse();
+    },
+    setSelectedIndex(index = 0) {
+        const newIndex = index || findIndex(this.state.options, option => option.text === this.state.currentValue);
+
+        this.setState('selectedIndex', newIndex);
+    },
+    emitChangeEvent() {
+        const selectedValue =
```
DylanPiercey: We are repeating ourselves a bit here, I would store the selected option in a var and have something like

```js
    emitChangeEvent() {
        const selectedOption = this.state.options[this.state.selectedIndex];
        const selectedValue = selectedOption && selectedOption.value || '';

        emitAndFire(this, 'combobox-change', {
            currentInput: this.state.currentValue,
            selectedOption,
            selectedValue,
            options: this.state.options
        });
    },
```

DylanPiercey:
Left some minor nitpicks, otherwise seems to be working for more 👍

src/components/ebay-combobox/index.js
```diff
@@ -0,0 +1,179 @@
...
+            this.state.options[this.state.selectedIndex]
+            && this.state.options[this.state.selectedIndex].value
+            || '';
+
+        emitAndFire(this, 'combobox-change', {
```
ianmcburnie: To avoid possible confusion with a text input's `onchange` event (i.e. it fires when input loses focus), how about we name this `combobox-autocomplete` instead?

marko.json
```diff
@@ -89,6 +89,29 @@
         "@style": "string",
         "@disabled": "boolean"
     },
+    "<ebay-combobox>": {
```
ianmcburnie: We'll need an option for autocomplete too. By default a combobox has no autocomplete.

src/components/ebay-combobox/index.js
seangates: But, this doesn't change util the the option is 'clicked'. E.g. makeup-activedescendant doesn't change the `selectedIndex`.

src/components/ebay-combobox/index.js
ianmcburnie: Just to clarify, we should only fire a `combobox-change` event when the combobox loses focus. If you are already doing that, then please ignore my comment.

PR #567, opened 2019-02-26
Update for preserving menu content on re-render;

### LINKED ISSUE #564: Menu: Shows incorrect data on client side rerender
# Bug Report

## Description
On client side rerender, the menu shows incorrect data. 

## Workaround

None

## Possible fix
AFAIK child DOM nodes are preserved when rerendering the component when using `w-preserve-body`. Removing it from this [line](https://github.com/eBay/ebayui-core/blob/97c8459791d1413097ee534878fdf4d87833c25c/src/components/ebay-menu/template.marko#L36) seems the fix the issue.

## Screenshots

Ideally after rerender it should show the same content as on page load. 
**On load**
screen shot 2019-02-22 at 11 17 10 am

**After rerender**
screen shot 2019-02-22 at 11 17 22 am
_[Ignore the styling issues]_

seangates:
## Description
- removes `w-preserve-body` from outer span
- removes `body-only-if(true)` from inner span

## References
Fixes #564


src/components/ebay-menu/template.marko
```diff
@@ -33,9 +33,9 @@
                 w-onkeydown="handleItemKeydown"
                 w-preserve-attrs="tabindex,data-makeup-index"
                 ${item.htmlAttributes}>
-                <span w-preserve-body>
+                <span>
                     <span body-only-if(!item.badgeNumber) aria-hidden='true'>
-                        <span body-only-if(true) w-body=item.renderBody/>
+                        <span w-body=item.renderBody/>
```
DylanPiercey: I thought we didn't want this span in the output, has this changed?

seangates: Unless you know another way, the only way I found to preserve that content was to remove this along with the `w-preserve-body` on the outer `<span>`.

seangates: Yes, I'd prefer not to have this span, though.

DylanPiercey: Okay, if the `body-only-if` true is messing it up then I'd say lets go with this for now.

src/common/closest-parent/index.js
```diff
@@ -0,0 +1,32 @@
+// https://developer.mozilla.org/en-US/docs/Web/API/Element/matches
+
+function closestParent(el, selector) {
+    let elem = el;
+
+    // Element.matches() polyfill
```
ianmcburnie: Do we need this polyfill? Browser support seems solid.

seangates: We could certainly entertain not doing the polyfill. I could shorten it to this:
```
if (!Element.prototype.matches) {
  Element.prototype.matches = Element.prototype.msMatchesSelector ||
                              Element.prototype.webkitMatchesSelector;
}
```

ianmcburnie: Your call. I just looked at MDN and CanIUse and it seemed well supported. I see now though that IE11 only has partial support.

src/common/closest-parent/index.js
```diff
@@ -0,0 +1,32 @@
+// https://developer.mozilla.org/en-US/docs/Web/API/Element/matches
+
+function closestParent(el, selector) {
```
ianmcburnie: Sorry to be totally OCD here, but should this module name be a bit more descriptive, like `closestAncestor` or `closestAncestorMatch` or something like that? It does iterate upwards through ancestors, right?

ianmcburnie:
Just a reminder for us that we also need to bump skin to 7.0.2 in coreui 2.0.5 so that we can pickup the fake button fix.

src/common/closest-parent/index.js
seangates: To me "parents" and "ancestors" mean the same thing. 🤷‍♂️  But stating `match` does make sense, as that's what it's doing.

seangates: There's actually this, too: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

I think this is better, since it's already part of the JS spec. Thoughts?

ianmcburnie: Parents are direct ancestors, one node up ;-)

ianmcburnie: Closest is better if you can find a polyfill for it.

seangates: Yeah, working on updating the PR.

seangates:
@ianmcburnie Please check again, as I'm now using a polyfill from `npm`.

PR #569, opened 2019-02-27
Placing floating label init behind load event;

### LINKED ISSUE #557: Textbox: floating label text clashes with browser autofill text
Screenshot:

screen shot 2019-02-19 at 12 29 39 pm

Last time we checked, the browser doesn't give us an event for when this autofill takes place.

Possible solutions:

1. The underlying `makeup-floating-label` module has a `refresh()` method. We could expose that and allow app to handle the logic (maybe calling the method after some arbitrary delay).
2. Add a polling mechanism (possibly opt-in via an attribute) that will check for any non-user entered text (I guess we could detect this based on whether the textbox has fired a focus event or not)

Other ideas?

seangates:
## Description
- moves floating label init behind the window's load event

## Context
In conjunction with changes made in the `makeup-floating-label` (https://github.com/makeup-js/makeup-floating-label/pull/2) which detects the autofill before applying the proper class, this change will wait until the page is ready and autofill has completed.

## References
Fixes #557 

### Before

screen shot 2019-02-19 at 12 29 39 pm

### After

image

### DIFF
--- a/src/components/ebay-textbox/index.js
+++ b/src/components/ebay-textbox/index.js
@@ -59,7 +59,7 @@ function onRender() {
     if (this.state.floatingLabel && !this.floatingLabel && document.readyState === 'complete') {
         this.initFloatingLabel();
     } else if (this.state.floatingLabel) {
-        window.addEventListener('load', this.initFloatingLabel);
+        window.addEventListener('load', this.initFloatingLabel.bind(this));
     }
 }
 
--- a/src/components/ebay-textbox/index.js
+++ b/src/components/ebay-textbox/index.js
@@ -57,17 +57,17 @@ function getTemplateData(state) {
 
 function onRender() {
     if (this.state.floatingLabel && !this.floatingLabel && document.readyState === 'complete') {
-        this.initFloatingLabel();
-    } else {
-        window.addEventListener('load', this.initFloatingLabel);
+        this.initFloatingLabel(this);
+    } else if (this.state.floatingLabel) {
+        window.addEventListener('load', this.initFloatingLabel(this));
     }
 }
 
-function initFloatingLabel() {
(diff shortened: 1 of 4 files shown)


ianmcburnie:
`makeup-floating-label@0.0.3` is now published.

ianmcburnie:
How do you feel about this going out in patch vs minor release? I'm okay with patch, I think.

seangates:
Patch is fine for me, too.

ianmcburnie:
Okay, I'll just wait for you to update this branch with latest `makeup-floating-label` before we proceed.

seangates:
@ianmcburnie Bumped `makeup-floating-label`. Please check at your convenience.

seangates:
Tested mobile, Safari on iOS and Chrome on Android. No regressions visible.

ianmcburnie:
Ok, this patch is now good to go. Please also update to Skin 7.0.3.

src/components/ebay-textbox/index.js
```diff
@@ -55,9 +55,19 @@ function getTemplateData(state) {
...
-        this.floatingLabel = new FloatingLabel(this.el);
+function onRender() {
+    if (this.state.floatingLabel && !this.floatingLabel && document.readyState === 'complete') {
+        this.initFloatingLabel(this);
+    } else if (this.state.floatingLabel) {
+        window.addEventListener('load', this.initFloatingLabel(this));
+    }
+}
+
+function initFloatingLabel(_this) {
```
DylanPiercey: Since we are using `this.initFloatingLabel` we shouldn't need to pass the `_this` parameter (just use this as is).

seangates: I can't reproduce the testing issue which caused me to do this. Reverting.

Issue #575, opened 2019-03-06
Textbox: type attribute is hard-coded


seangates:
Presently, we hardcode the type to text, even though there are other data types which are available (e.g. `number`, `email`, `date`, etc.). We should decide if we want to allow all data types, a limited set, or create other components _on top of textbox_ to cast other types.

In src/components/ebay-textbox/template.marko

```
type="text"
```

Posted via CodeStream

ianmcburnie:
For 2.1.0 let's open it up to a few other types that will work safely with our current component.

seangates:
Merged.

PR #579, opened 2019-03-06
Textbox: Additional types

seangates:
## Description
- add allowance for a "password" type

## Context
When working with users one of the immediate uses for textbox is as a password field. This allows support for that.

## References
Relates to #575

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -355,6 +355,7 @@
         "@id": "string",
         "@class": "string",
         "@style": "string",
+        "@type": "string",
         "@invalid": "boolean",
         "@fluid": "boolean",
         "@multiline": "boolean",
--- a/src/components/ebay-textbox/examples/01-default/template.marko
+++ b/src/components/ebay-textbox/examples/01-default/template.marko
@@ -1,1 +1,1 @@
-<ebay-textbox value="ebay-textbox" />
+<ebay-textbox value="ebay-textbox"/>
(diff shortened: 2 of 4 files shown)


ianmcburnie:
#575 has this marked for 2.1.0 release because technically it adds new functionality rather than fixing a bug. Could be persuaded otherwise if we feel this is urgent, but I was under the impression there was a workaround.

seangates:
The workaround is after the page has loaded and pretty ... hacky. It requires a developer to modify the DOM, and not use Marko.

seangates:
Also, this PR only includes the password type. I still believe #575 to be for 2.1.0, with thi PR addressing the immediate need.

ianmcburnie:
Ok, it doesn't seem to be strictly following semver rules though. Just so we are all aware. 

Heads up that code coverage goes down slightly too for that uncovered code path, but maybe you are leaving that for when we add the other types too.

seangates:
Yes, I saw the coverage drop and will address it in the full-featured PR.

DylanPiercey:
Curious why we are not just allowing `type` to pass through?

seangates:
Per a comment from @ianmcburnie in our standup I went with "just allow password".

ianmcburnie:
We could do that, but then it opens it up to type=checkbox, type=radio, etc. which make no sense.

senthilp:
@ianmcburnie How will types like `email`, `tel` etc. work? Should we check on a whitelist then?

ianmcburnie:
We'll just go through them one by one, in different browsers and devices and see which ones work "as is". I imagine most of the obvious ones will be fine.

PR #591, opened 2019-03-18
Floating label refresh using onUpdate

### LINKED ISSUE #581: Textbox floating label does not come down when setting textbox value to empty
## eBayUI Version: x.x.x
2.0.6-1

## Description
Textbox floating label does not come down when textbox value is updated to empty by state change.

## Screenshots
screen shot 2019-03-07 at 1 58 05 pm

seangates:
## Description
- add the `onUpdate` lifecycle method, and call `initFloatingLabel`

## Context
When updating the state from a parent component, the `value` gets changed to an empty string. This should trigger a refresh of the floating label, but the textbox component wasn't doing anything during the update part of the lifecycle.

## References
Fixes #581 

### Before
screen shot 2019-03-07 at 1 58 05 pm

### After
image

### DIFF
--- a/src/components/ebay-textbox/index.js
+++ b/src/components/ebay-textbox/index.js
@@ -74,6 +74,10 @@ function onRender() {
     }
 }
 
+function onUpdate() {
+    this.initFloatingLabel();
+}
+
 function initFloatingLabel() {
     if (!this.floatingLabel) {
         this.floatingLabel = this.el && new FloatingLabel(this.el);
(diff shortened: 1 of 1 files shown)


ianmcburnie:
We should cover this with a unit test if possible.

seangates:
Would we update the `input` values dynamically? How would I do that?

ianmcburnie:
My first thought was that `onUpdate` should be covered by the framework (i.e. we trust the framework lifecycle and it's own tests). Then I thought, what if somebody were to remove this `onUpdate` method, we would have a bug again with no way to catch it, right?

seangates:
Hopefully we won't remove code without knowing what it's supposed to do. ;-)

But, I see your point. I will look at what can be done to create a test. It's an interesting use case, to be sure.

PR #593, opened 2019-03-19
Listbox: invalid options preservation fix

### LINKED ISSUE #592: List-box: Invalid options preservation
I was using a set of `listbox` which was supposed to be updated dynamically based on the selection of others. It seems `w-preserve` prevents the refreshing of the options.

chtitoon:
This is addressing the issue described in #592


ianmcburnie:
Thanks @chtitoon. We have a 2.0.6 patch going out soon, this can most likely go out in that, so I updated the target branch.

PR #602, opened 2019-03-29
601 infotip expander

### LINKED ISSUE #601: Infotip: makeup-expander does not get re-attached
# Bug Report

## Description
When re-rendering the `<ebay-infotip/>` the makeup-expander does not get re-initialized/re-attached to the component. This is also likely to affect the `<ebay-tooltip/>` and `<ebay-tourtip/>`, but I haven't confirmed that yet.

## Workaround
None.

DylanPiercey:
## Description
Fixes the issue where `infotip` and `tourtip` were not expanding (or closing) properly after a rerender and adds tests.

## References
Fixes #601

### DIFF
--- a/src/components/ebay-infotip/template.marko
+++ b/src/components/ebay-infotip/template.marko
@@ -6,9 +6,11 @@
         w-on-base-expand="handleExpand"
         w-on-base-collapse="handleCollapse">
         <span
+            w-preserve-attrs="class,id"
             class="infotip"
             ${data.htmlAttributes}>
             <button
+                w-preserve-attrs="aria-expanded,aria-controls,aria-described-by"
                 class="infotip__host icon-btn"
                 aria-label=data.ariaLabel>
                 <if(data.icon)>
--- a/src/components/ebay-infotip/test/test.browser.js
+++ b/src/components/ebay-infotip/test/test.browser.js
@@ -19,30 +19,43 @@ describe('given the default infotip', () => {
 
     afterEach(() => widget.destroy());
 
-    describe('when the host element is clicked', () => {
-        let spy;
-        beforeEach(() => {
-            spy = sinon.spy();
-            widget.on('tooltip-expand', spy);
-            testUtils.triggerEvent(host, 'click');
+    thenItCanBeOpenAndClosed();
+
+    describe('after it is rerendered', () => {
+        before(() => {
+            widget.setStateDirty('test');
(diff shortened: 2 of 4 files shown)


DylanPiercey:
Note: `ebay-combobox` and `ebay-listbox` require a similar treatment.

src/components/ebay-infotip/test/test.browser.js
```diff
@@ -19,30 +19,43 @@ describe('given the default infotip', () => {
...
     afterEach(() => widget.destroy());
 
-    describe('when the host element is clicked', () => {
-        let spy;
-        beforeEach(() => {
-            spy = sinon.spy();
-            widget.on('tooltip-expand', spy);
-            testUtils.triggerEvent(host, 'click');
-        });
+    thenItCanBeOpenAndClosed();
```
seangates: Do we not need one for the initial render?

src/components/ebay-infotip/template.marko
```diff
@@ -6,9 +6,11 @@
         w-on-base-expand="handleExpand"
         w-on-base-collapse="handleCollapse">
         <span
+            w-preserve-attrs="class,id"
```
seangates: Presently, preserving the class doesn't do much since the class is always just `"infotip"`. Any HTML class attribute passed in is never used (lines 10-11 👇). @melquan and I uncovered this and I've created an issue for it: #604.

DylanPiercey: Makeup expander is what is changing the class which is why I've added this.

src/components/ebay-infotip/test/test.browser.js
DylanPiercey: This line is what is testing the initial render.

https://github.com/eBay/ebayui-core/pull/602/files/3bc8f3b64af7c0b7fa61126544811f7eb1450cc7#diff-eab84de4b65d4fa0081b4cd28ee4eccfR30 is testing the rerender.

src/components/ebay-infotip/template.marko
seangates: Ah, right. Still we have the same problem on the HTML attribute, though.

src/components/ebay-infotip/test/test.browser.js
seangates: Must have missed that second one. Thanks!

PR #603, opened 2019-03-29
Refactor ebay-icon to keep the DOM more inline with what Marko expects

DylanPiercey:
## Description
While digging into #601 I discovered that the `ebay-icon` was mutating the DOM on the first render which I thought may be causing the issue. It turns out this was fine But I have created a fix to avoid this anyways.

I also refactored the template to use standard Marko syntax instead of inline scripts which will make the code slightly better after a future migration to Marko 4.


ianmcburnie:
@DylanPiercey Is this safe for patch release? Or do you prefer to wait to put it in 2.1.0?

DylanPiercey:
This should be safe for a patch release.

PR #608, opened 2019-04-02
Ensure Marko generated ID's on elements touched by makeup-expander

### LINKED ISSUE #601: Infotip: makeup-expander does not get re-attached

DylanPiercey:
## Description
Turns out that #602 was not the full solution for #601. I isolated the issue fixed in that PR which I confirmed working, but after testing out the release with the app that reported this issue it looked like the issue was not fixed.

The issue solved in this PR is similar but only effects Marko 3 apps. Specifically it has to do with that `makeup-expander` is setting the `id` attribute which is special in Marko 3. In Marko 3 these ID's are used to move nodes around and since the Marko template disagreed with makeup the nodes were always being destroyed in that app which caused the expander to become out of sync.

Luckily makeup-expander only assigns an ID if one does not already exist (https://github.com/makeup-js/makeup-expander/blob/master/src/index.js#L75) and so adding Marko generated ID's to these root elements allows things to stay in sync. 

## References
Fixes #601

### DIFF
--- a/src/components/ebay-infotip/template.marko
+++ b/src/components/ebay-infotip/template.marko
@@ -6,7 +6,8 @@
         w-on-base-expand="handleExpand"
         w-on-base-collapse="handleCollapse">
         <span
-            w-preserve-attrs="class,id"
+            w-id="expander-root"
+            w-preserve-attrs="class"
             class="infotip"
             ${data.htmlAttributes}>
             <button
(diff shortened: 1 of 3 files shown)


seangates:
@DylanPiercey Thoughts on pre-releasing this as `2.0.9-0` first to be doubly sure?

DylanPiercey:
@seangates this should be fine to release, I've tested with their app. If we did want to prerelease and test it with their app that is fine by me though.

PR #610, opened 2019-04-03
Textbox: Event when floating label is present and component is updated;

### LINKED ISSUE #594: Textbox: emit an event when textbox is updated by Marko, or the floating label is refreshed
# Feature Request

## Description
When the textbox is updated, or when the floating label is initialized or "refreshed", it should emit an event with the current value of the textbox. This allows a developer to handle scenarios such as validation, or enabling/disabling form controls. Specifically, this is to account for the use case when browser autofill occurs, the floating label has been refreshed, but the normal DOM `change` event isn't emitted from the textbox.

## Context
The identity team is using the textbox, and need a way to listen for when the value of the textbox is not an empty string, and to validate the input, before enabling the submit button on the form.

seangates:
## Description
- new event `textbox-floating-label-init` was added

## Context
When the textbox is updated, it re-initializes the floating label. Because of the three-way relationship with `makeup-floating-label`, this component, and Marko, we send out an event whenever the parent component forces an update on the textbox component we need to know about it. This helps developers to know when the floating label was refreshed/re-initialized.

## References
Closes #594.

### DIFF
--- a/src/components/ebay-textbox/README.md
+++ b/src/components/ebay-textbox/README.md
@@ -27,4 +27,4 @@ Event | Data | Description
 `textbox-focus` | `{ originalEvent, value }` |
 `textbox-blur` | `{ originalEvent, value }` |
 `textbox-keydown` | `{ originalEvent, value }` |
-`textbox-floating-label-init` | `{ null, value }` |
+`textbox-floating-label-init` | `{ originalEvent: null, value: null }` |
--- a/src/components/ebay-textbox/README.md
+++ b/src/components/ebay-textbox/README.md
@@ -27,3 +27,4 @@ Event | Data | Description
 `textbox-focus` | `{ originalEvent, value }` |
 `textbox-blur` | `{ originalEvent, value }` |
 `textbox-keydown` | `{ originalEvent, value }` |
+`textbox-floating-label-init` | `{ null, value }` |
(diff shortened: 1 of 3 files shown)


src/components/ebay-textbox/README.md
```diff
@@ -26,3 +26,4 @@ Event | Data | Description
 `textbox-input` | `{ originalEvent, value }` |
 `textbox-focus` | `{ originalEvent, value }` |
 `textbox-blur` | `{ originalEvent, value }` |
+`textbox-floating-label-init` | `{ null, value }` |
```
ianmcburnie: It feels a little odd, as an end user, to receive a null value as first param.

DylanPiercey: It would actually (by the looks of it) be an object with `{ originalEvent: null, value: null }`. I think the docs are wrong here.

ianmcburnie: Aha! Thanks Dylan. Yes, let's fix those docs.

seangates:
@ianmcburnie @DylanPiercey Updated docs! One more quick check would be great. Thanks for your patience.

src/components/ebay-textbox/README.md
```diff
@@ -27,3 +27,4 @@ Event | Data | Description
 `textbox-focus` | `{ originalEvent, value }` |
 `textbox-blur` | `{ originalEvent, value }` |
 `textbox-keydown` | `{ originalEvent, value }` |
+`textbox-floating-label-init` | `{ originalEvent: null, value: null }` |
```
ianmcburnie: Don't the events above also have this signature?

PR #618, opened 2019-04-08
Carousel: Use media queries instead of lasso flags for touch devices

DylanPiercey:
## Description
This PR refactors the carousel styling to use media queries instead of lasso flags to target touch devices.

## Context
We want to move away from lasso flags and in this case the only place we are using the `touch` flag in ebayui-core and skin was here.

Also is wasteful to have this flag since it means an entirely new set of bundles need to be created for it with lasso which is causing warmup issues for some teams.

Also we do not get much, if any, file size savings here by putting both sets of functionality into one less file.


ianmcburnie:
Nice! This should probably go into [2.4.0](https://github.com/eBay/ebayui-core/projects/26#card-19988113) which is going to be carousel-only related changes.

ianmcburnie:
@DylanPiercey It's worth checking that these changes are going to play nicely with this feature request for v2.4.0: https://github.com/eBay/ebayui-core/projects/26

DylanPiercey:
👍 it should make it easier for one of them and not really change the other.

src/components/ebay-carousel/style.less
```diff
@@ -141,25 +176,42 @@
     }
 }
 
+@media (hover: none) {
```
ianmcburnie: @DylanPiercey What are we doing for IE11 again? I don't think this media query is supported.

DylanPiercey: This isn’t supported in ie11 and won’t run. It’s fine though because it is meant to detect any device that doesn’t support hover (mobile devices) and ie11 does.

ianmcburnie: Thanks!

src/components/ebay-carousel/README.md
```diff
@@ -11,8 +11,6 @@
 ```
 
 This component will bundle different resources depending on Lasso flags provided by your application.
```
seangates: Not sure this comment is still needed, either, since all other components also handle the `skin-ds6` flag and you've removed the `touch/no-touch` flags from just this one component.

PR #625, opened 2019-04-12
624 refactor

### LINKED ISSUE #624: Refactor components to use recommended data organization & other best practices
# Refactor

### State vs Input vs Computed
Currently most components do not have a clear strategy for what parts of data is state vs input vs computed which can make digging into the components a bit more tricky.

Marko 3 does not have the most clear separation of these concepts but it is possible to at least make the intent clear by explicitly only including actual state (and not input) in `getInitialState`.

Computed values typically go in `getTemplateData` in Marko 3, however in Marko 4 the recommendation is to inline these computed values into the template and to extract out helper functions if necessary which can be imported.

In Marko 3 `input` is usually just merged in (which makes it easy to loose track of) but I think a refactor to make the intent clear will improve maintainability and make it easier to have an optimized migration to Marko 4.

### Minimizing state
In some of our components (as mentioned above) we put things that are not state, into the state object (mostly because of Marko 3 limitations). Because the boundaries are blurred here I think some components have been mutating the input which is not ideal or recommended and further complicates the state management.

### querySelector's
It is never recommended to use `querySelector` in Marko (3 or 4) however we are in a few components. Where these selectors are passed to makeup there is not much we can do, but in the other cases we can improve by using the recommended api (w-id="key" & widget.getEl("key")). This is not only going to improve performance, but also decouple us from these specific class names.

### Stateless components
A large number of the ebayui components should be stateless (that is they can never update by themselves such as button, badge, etc). However all of these components still are incorrectly using API's like `getInitialState`. These components should be refactored to either not have component logic at all, or only have logic for handling events (since many forward events and that's it).

DylanPiercey:
## Description
Implements the refactor described in #624

Every component has been refactored except for `ebay-menu` and `ebay-carousel`.

I got half way through the menu refactor and realized that it's programatic api is really holding it back. I think it would make more sense to tackle that when/if we remove the programatic api.

For the carousel it is quite complicated and would take some time to refactor (which I'd be willing to do) however the carousel implementation could benefit a lot from some of the Marko 4 features and so I think a more holistic refactor should take place when we eventually upgrade.

## References
Closes #624

---

I suspect it will take some time to look over these changes (lots got moved around and so the PR is pretty big) so I opted to put up the PR now.

I am still testing these changes in Marko 4 though and this PR is not yet ready to merge.


ianmcburnie:
As 2.2.0 (combobox) and 2.3.0 (menu filtering/flags) are time sensitive, let's hold this back until 2.4.0.

DylanPiercey:
Fixed the issues when testing against Marko 4. Once the current in flight PR's land I'll resolve any merge conflicts.

PR #643, opened 2019-04-24
Autofill change for Chrome;

### LINKED ISSUE #639: combobox: when address auto fill in on in chrome- it displays the address over the options
# Bug Report

## Description
When address auto fill in on in chrome- it displays the address over the options
Chrome is not respecting the `autofill="off"` attribute

I see this happening only when combobox has options with special characters
 This will block all German and other language users
I am filing this as a non-blocker as of now But this might me a blocker
Lets discuss about this

seangates:
## Description
- Fixes a found issue with the expander being closed and pressing the `Enter` key
- Change for autofill because of Chrome

## References
Fixes #639

### DIFF
--- a/src/components/ebay-combobox/template.marko
+++ b/src/components/ebay-combobox/template.marko
@@ -26,7 +26,7 @@
             aria-haspopup='listbox'
             w-on-blur='handleComboboxBlur'
             w-on-keyup='handleComboboxKeyUp'
-            autocomplete="off"
+            autocomplete="new-password"
             w-preserve-attrs="aria-controls,aria-activedescendant,aria-owns,aria-expanded"
             ${processHtmlAttributes(data)}/>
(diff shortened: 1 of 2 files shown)


src/components/ebay-combobox/template.marko
```diff
@@ -26,7 +26,7 @@
             aria-haspopup='listbox'
             w-on-blur='handleComboboxBlur'
             w-on-keyup='handleComboboxKeyUp'
-            autocomplete="off"
+            autocomplete="new-password"
```
ianmcburnie: What does this "new-password" value mean/do? Needs a description in the PR and a comment in the code.

seangates: Actually, this may not be working as expected anyway. Let me confirm it even works as expected. I'll close the PR for now.

PR #653, opened 2019-05-02
Menu: added simple menuitem shortcut via keypress #606

### LINKED ISSUE #606: Menu: move focus to menuitem when character key is pressed
# Feature Request

## Description

For long menus, providing a quick filter ability is useful. For example, pressing the key 'k' would move menu focus to the first menu item beginning with the letter 'k'

## Context

https://www.w3.org/TR/wai-aria-practices-1.1/#menu

## Screenshots

Screen Shot 2019-04-01 at 3 12 20 PM

ianmcburnie:
# Fixes #606

Adds simple keyboard shortcut to menuitems. For example, pressing 'c' should move the keyboard focus to the first menu item starting with the letter 'C' (case insensitive). It recognizes a single character only. Not multiple in quick succession.

### Functionality

* Modified focus management of menu (via underlying makeup-expander) so that keyboard focus always moves to the first item when the menu is expanded, regardless of input device.  This is probably the biggest change. This is required so that mouse user can click with mouse, then press a shortcut key. 
* Added new `nodelist-utils`, containing a method `findNodeWithFirstChar`
* Added `handleKeypress` function on menu items which handles shortcut logic

### Tests

* Added browser tests for `nodelist-utils`
* Added browser tests for menuitem keypress and shortcut logic
* Added `key` param to `triggerEvent` function in `test-utils/browser.js`. Seems preferable to check for `event.key` in this situation (i.e. printable chars only)
* Updated menu mock data to use characters instead of numbers. This is for testing of shortcut logic (didn't break existing tests)

### Examples

* Added simple icon example to menu examples (not strictly related to this PR, but felt it was missing)
* Added simple sprite example to menu examples which contains country flags and names. This can be used to manually test the shortcut feature

### Deps

* Updated to latest makeup-expander (v0.7.1)

## Notes

<strike>The method `findNodeWithFirstChar` currently retains the empty string key in its map. I could easily filter this out, but it doesn't seem to do any harm right now and actually lets us find the first node with empty innerText (could be useful..)</strike>

The sprite example could potentially mislead developers into thinking we still use sprites at eBay. We try not to these days, except for the specific scenario of country flags (until somebody rustles up some nice SVG versions).

### DIFF
--- a/src/common/nodelist-utils/index.js
+++ b/src/common/nodelist-utils/index.js
@@ -1,0 +1,16 @@
+// returns first node in list with matching first character (case-insenstive)
+function findNodeWithFirstChar(nodeList, char) {
+    // e.g. <li>A</li><li>B</li><li></li><li>B</li><li>A</li>
+
+    const keyValuePairList = [...nodeList].map((el, index) => ({ [el.innerText.charAt(0).toLowerCase()]: index }));
+    // e.g. [{a: 0}, {b: 1}, {'': 2}, {b: 3}, {a: 4}]
+
+    const charMap = Object.assign(...keyValuePairList.reverse());
+    // e.g. {a: 0, b: 1, '': 2}
+
+    return charMap[char];
+}
+
+module.exports = {
+    findNodeWithFirstChar
+};
--- a/src/common/nodelist-utils/index.js
+++ b/src/common/nodelist-utils/index.js
@@ -1,14 +1,6 @@
-// returns first node in list with matching first character (case-insenstive)
+// returns index of first node in NodeList with matching first character (case-insenstive)
 function findNodeWithFirstChar(nodeList, char) {
-    // e.g. <li>A</li><li>B</li><li></li><li>B</li><li>A</li>
-
-    const keyValuePairList = [...nodeList].map((el, index) => ({ [el.innerText.charAt(0).toLowerCase()]: index }));
-    // e.g. [{a: 0}, {b: 1}, {'': 2}, {b: 3}, {a: 4}]
-
-    const charMap = Object.assign(...keyValuePairList.reverse());
-    // e.g. {a: 0, b: 1, '': 2}
-
-    return charMap[char];
+    return [...nodeList].findIndex((el) => el.innerText.charAt(0).toLowerCase() === char);
 }
 
 module.exports = {
--- a/src/common/nodelist-utils/test/test.browser.js
+++ b/src/common/nodelist-utils/test/test.browser.js
@@ -1,0 +1,89 @@
+const expect = require('chai').expect;
+const NodeListUtils = require('../');
+const findNodeWithFirstChar = NodeListUtils.findNodeWithFirstChar;
+
+document.body.innerHTML = `
+    <ul id="list1">
+        <li>Ringo</li>
+        <li>Paul</li>
+        <li>George</li>
+    </ul>
+    <ul id="list2">
+        <li>Bruce</li>
+        <li>Bruce</li>
+        <li>Bruce</li>
+    </ul>
+    <ul id="list3">
+        <li></li>
+        <li></li>
+        <li></li>
+    </ul>
+`;
+
+describe('NodeListUtils.findNodeWithFirstChar', () => {
+    describe('given three nodes with different first letters (R, P, G)', () => {
+        const nodeList = document.querySelectorAll('#list1 li');
+
+        describe('when "r" is input', () => {
+            test('then 0 is returned', () => {
+                expect(findNodeWithFirstChar(nodeList, 'r')).to.equal(0);
+            });
+        });
+
+        describe('when "p" is input', () => {
+            test('then 1 is returned', () => {
+                expect(findNodeWithFirstChar(nodeList, 'p')).to.equal(1);
+            });
+        });
+
+        describe('when "g" is input', () => {
+            test('then 2 is returned', () => {
+                expect(findNodeWithFirstChar(nodeList, 'g')).to.equal(2);
+            });
+        });
+
+        describe('when "j" is input', () => {
+            test('then undefined is returned', () => {
+                expect(findNodeWithFirstChar(nodeList, 'j')).to.equal(undefined);
+            });
+        });
+
+        describe('when "" is input', () => {
+            test('then undefined is returned', () => {
+                expect(findNodeWithFirstChar(nodeList, '')).to.equal(undefined);
+            });
+        });
+    });
+
+    describe('given three nodes with same first letters (B, B, B)', () => {
+        const nodeList = document.querySelectorAll('#list2 li');
+
+        describe('when "b" is input', () => {
+            test('then 0 is returned', () => {
+                expect(findNodeWithFirstChar(nodeList, 'b')).to.equal(0);
+            });
+        });
+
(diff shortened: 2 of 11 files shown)


src/common/test-utils/browser.js
```diff
@@ -5,11 +5,13 @@ const expect = require('chai').expect;
  * @param {HTMLElement} el
  * @param {String} type
  * @param {Number} keyCode
+ * @param {String} key
  */
-function triggerEvent(el, type, keyCode) {
+function triggerEvent(el, type, keyCode, key) {
```
agliga: I'd move to having an options argument since there are more than 3 arguments. Or have the third argument be an extraEventOptions and then use Object.assign(event, extraEventObject);
That way if we want to add any more items it's a simple object. 
However, if we aren't planning on extending the event object, this could also be fine.

src/components/ebay-menu/examples/16-sprites/template.marko
```diff
@@ -0,0 +1,17 @@
+<style>
+    span.flag {
+        background-image: url('https://ir.ebaystatic.com/pictures/aw/pics/cmp/ds3/sprds3_21.png');
```
agliga: nit: I thought background images were not good for accessibility? Would this cause a problem there 
Isn't img tag a better way? The support might not be there fully, but there is object-position for img and sprites. 
I know it's just an example and not a big deal, but I think some team might end up copying this and then using it in their code.

src/common/test-utils/browser.js
ianmcburnie: I agree an options argument would be better. We should tackle that in a separate PR sometime.

agliga: 👍

src/components/ebay-menu/examples/16-sprites/template.marko
ianmcburnie: Yeah, I should have added this to the PR notes. The reason I am using a sprite is because that's what our global header are going to use. We don't have the flags in any other format. There is no issue with accessibility here, because the name of the country is displayed in text alongside the image. The screen reader will simply ignore the icon span. I think at some point we would get SVG versions of the flags and then create an actual country selector component. Until then this example will have to suffice. Let's monitor this example and see if it does cause confusion like you say.

src/common/nodelist-utils/index.js
```diff
@@ -0,0 +1,16 @@
+// returns first node in list with matching first character (case-insenstive)
+function findNodeWithFirstChar(nodeList, char) {
+    // e.g. <li>A</li><li>B</li><li></li><li>B</li><li>A</li>
+
+    const keyValuePairList = [...nodeList].map((el, index) => ({ [el.innerText.charAt(0).toLowerCase()]: index }));
```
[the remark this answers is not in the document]
ianmcburnie: Not entirely sure what you mean. Remember though it needs to return the index position of the node. Not sure if that is apparent. So long as it fullfils this contract I'm okay with the internals as they are for now, unless I'm doing something very, very silly... which is why I suck at code interviews ;-)

src/common/nodelist-utils/test/test.browser.js
```diff
@@ -0,0 +1,89 @@
+const expect = require('chai').expect;
+const NodeListUtils = require('../');
+const findNodeWithFirstChar = NodeListUtils.findNodeWithFirstChar;
+
+document.body.innerHTML = `
```
DylanPiercey: Overwriting the document body could mess with other tests. Any changes we make to the document should be undone via the mocha before/after hooks.

In this case I don't think we actually need the dom nodes to be attached to work with them here and so we should be able to just used a detached div to build the lists, eg:

```javascript
const CONTAINER = document.createElement("div");
CONTAINER.innerHTML = `
  ....
`
```

And then use `CONTAINER.querySelector` instead of `document.querySelector` which avoids the problem entirely.

ianmcburnie: Good call. Will fix.

src/common/nodelist-utils/index.js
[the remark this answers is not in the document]
agliga: I think it might be cleaner to do this (that is if it does the same thing)
return [...nodeList].findIndex((el) => el.innerText.charAt(0).toLowerCase() === char);

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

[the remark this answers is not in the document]
ianmcburnie: Thanks Andrew! Good call! I'd forgotten about `findIndex` (so many goodies in JavaScript these days). It doesn't quite work for the empty inner text scenario (the tests fail), but I'll see if I can rejig things and make use of `findIndex`.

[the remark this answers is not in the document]
agliga: 👍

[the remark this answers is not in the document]
ianmcburnie: Ah silly me, just needed to change tests to check for `-1` instead of `undefined`. MUCH simpler now. Thanks.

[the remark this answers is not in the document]
ianmcburnie: It's so simple I could potentially just remove the utility and tests altogether and just inline this.

src/common/nodelist-utils/test/test.browser.js
ianmcburnie: So THAT'S what caused the element scroll test to break in my local env!!

src/common/nodelist-utils/index.js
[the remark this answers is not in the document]
ianmcburnie: It does feel slightly overkill as a util method now, but I'm going to leave it in so that A) Listbox component (and maybe some others) can use it and B) we don't attempt to write it again in future in a long-winded fashion like I did :-)

ianmcburnie:
Not sure what's going on with the CI tests for this. Anyone have an idea? These test work locally but fail on CI:

https://github.com/eBay/ebayui-core/pull/653/files#diff-76bd13e1f625654e8bfe529f4e2fcc2aR198

I'm going to merge this PR in order to unblock our pre-release (@agliga approved before my previous typo commit). I have another branch to try and figure out the tests with some trial and error:

https://github.com/eBay/ebayui-core/tree/fix-ci

DylanPiercey:
@ianmcburnie the issue is likely a race condition. For most tests we should avoid using `setTimeout` and instead actually hook into the Marko lifecycle for when the component updates. The browser stack browsers are extremely slow and can cause those timeouts to not run in the same order as in our machines.

ianmcburnie:
@DylanPiercey Ok, thanks. I will probably require your assistance with that. I saw a lot of `delay` calls and `setTimeout` calls throughout our test code. Maybe at some point we sweep through and convert to Marko lifecycle wherever makes sense.

DylanPiercey:
@ianmcburnie yeah we should be using `setTimeout` more sparingly. It is rare that the thing we are waiting for is actually a timeout, usually it is a component update, dom event, or animation frame.

In this case we'd want to wait for the component to update which we can do via something like

```
before((done) => {
  component.once("update", done); // Don't run the tests until after Marko has finished updating.
  triggerClickEventThatCausesComponentToUpdate(el);
})
```

PR #657, opened 2019-05-06
Menu: fix shortcut when CAPS-LOCK enabled

### LINKED ISSUE #656: menu: keyboard type-does not work when cap lock is on
# Bug Report

## Description
repro:
go to menu
click on the menu
Type a letter with caps lock on
The matching option is not selected.

can repro it in market place

ianmcburnie:
# Fixes #656 

Now calling `toLowerCase()` on the char that is passed. Also updated tests to check for this.

### DIFF
--- a/src/common/nodelist-utils/index.js
+++ b/src/common/nodelist-utils/index.js
@@ -1,6 +1,6 @@
 // returns index of first node in NodeList with matching first character (case-insenstive)
 function findNodeWithFirstChar(nodeList, char) {
-    return [...nodeList].findIndex((el) => el.innerText.charAt(0).toLowerCase() === char);
+    return [...nodeList].findIndex((el) => el.innerText.charAt(0).toLowerCase() === char.toLowerCase());
 }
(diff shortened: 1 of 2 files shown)


src/common/nodelist-utils/index.js
```diff
@@ -1,6 +1,6 @@
 // returns index of first node in NodeList with matching first character (case-insenstive)
 function findNodeWithFirstChar(nodeList, char) {
-    return [...nodeList].findIndex((el) => el.innerText.charAt(0).toLowerCase() === char);
+    return [...nodeList].findIndex((el) => el.innerText.charAt(0).toLowerCase() === char.toLowerCase());
```
agliga: Just a quick question, is there a possibility either el.innerText or char will be null? It could cause an error if it does.

ianmcburnie: If called purely as a utility, then yes. From menu component as it stands, I don't believe so.

PR #658, opened 2019-05-06
Merge 2.3.0 branch to master prior to final release

ianmcburnie:
This release only contains the menu filtering feature.

### DIFF
--- a/src/components/ebay-menu/test/test.browser.js
+++ b/src/components/ebay-menu/test/test.browser.js
@@ -125,12 +125,14 @@ describe('given the menu is in the expanded state', () => {
     let root;
     let button;
     let firstItem;
+    // let secondItem;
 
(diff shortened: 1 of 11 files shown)


src/components/ebay-menu/test/test.browser.js
```diff
@@ -186,6 +188,25 @@ describe('given the menu is in the expanded state', () => {
         });
     });
 
+    // these two tests are failing on CI.
```
ianmcburnie: Ah, I forgot to go and fix this for the CI. I might not get to this today, so if we are okay leaving this in for now I will fix as part of v2.4.0 (promise!) - which should also go out this week.

seangates: I'm okay if you wait to fix this in 2.4.

PR #661, opened 2019-05-09
Upgraded backwards compatible devDeps

ianmcburnie:
Just taking the opportunity to upgrade some devDeps before we do the 2.4.0 release.

I specifically left out upgrading makeup deps and corejs to reduce chance of any possible regressions (we have enough going on with the refactor). Let's do those, and test those, in 2.5.0.

Before:

Screen Shot 2019-05-08 at 5 17 15 PM

After:

Screen Shot 2019-05-08 at 5 17 43 PM


ianmcburnie:
Travis is most likely still on an older version of Chrome.

PR #662, opened 2019-05-10
Fix build scripts

DylanPiercey:
## Description

The current build script only updates the `src` => `dist` paths in the `marko.json` for `renderer` and `transformers`.

In this PR I made the build script more generic and just replace `./src/` with `./dist/` in the entire `marko.json` to avoid this issue in the future.

I also cleaned them up a little bit to avoid using string concat for building paths, and to remove a section of code related to updating the `src` => `dist` within the `browser.json` for individual components. This is not needed because all dependencies of components should be within the `src` folder anyways, with relative paths, or a node_module.

## Context
My refactor in 2.4.0 changed some of the components to no longer need a renderer. Since their config now just points to a template the build script was not updating the `marko.json` path properly.

### DIFF
--- a/scripts/postpublish.js
+++ b/scripts/postpublish.js
@@ -1,23 +1,14 @@
 const fs = require('fs');
+const path = require('path');
 const del = require('del');
-
-const pathPrefix = `${ __dirname }/..`;
+const rootDir = path.join(__dirname, '..');
 
 // remove files created by prepublish
-del.sync([`${pathPrefix}/*.browser.json`, `${pathPrefix}/dist`]);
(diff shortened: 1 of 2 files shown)


scripts/postpublish.js
```diff
@@ -1,23 +1,14 @@
 const fs = require('fs');
+const path = require('path');
 const del = require('del');
-
-const pathPrefix = `${ __dirname }/..`;
+const rootDir = path.join(__dirname, '..');
```
seangates: Saving Windows users, one `path.join` at a time.

PR #671, opened 2019-05-15
Hide scrollbar using negative margin

### LINKED ISSUE #668: Scrollbar Under Carousel Inconsistent on iPad Safari (Touch)
# Bug Report

## Description

For the `<ebay-carousel>` component, carousels inconsistently show / don't show a scrollbar when scrolled by touching and dragging on iOS Safari, iPad, latest versions.

You can reproduce this bug by loading the ebay UI Marketplace's ebay-carousel demo page in the iOS Simulator set to use an iPad.

Find a touch enabled carousel and swipe to the left, and repeat for other carousels on the page.  Some of them will show a scrollbar at the bottom while in motion, and others won't.  

**Important**: This bug is hard to reproduce.  You may need to refresh and repeat several times to find one that does not show a scrollbar, as it seems it's far more likely to show a scrollbar than not.  Another place that's a bit easier to see it happen is the ebay signed in homepage, which has many carousels. 

When you refresh the ebayui demo page, you may see that a different set of carousels have a scrollbar when scrolling, while others do not.

Carousels which show the scrollbar on scroll will continue to show the scrollbar on scroll, and ones which do not show it will never start showing it [edit: until the page is refreshed].

## Workaround

No known workaround

## Screenshots

The blue carousel happened to not show a scrollbar in this runthrough:

[scrollbar-not-showing-on-a-carousel.mov.zip](https://github.com/eBay/ebayui-core/files/3179988/scrollbar-not-showing-on-a-carousel.mov.zip)

DylanPiercey:
## Description
Uses an alternative cross browser approach to hide the scroll bar on the carousel.

I have not been able to reproduce the issue described in #668, however this may be a fix.
@trg I have not been able to reproduce the issue just yet, but the plan is to get out a prerelease where you can see if this alternative approach resolves the issue.

## References
Potentially resolves #668

### DIFF
--- a/src/components/ebay-carousel/style.less
+++ b/src/components/ebay-carousel/style.less
@@ -163,13 +163,12 @@
 
                 &.carousel__dot--active::after {
                     opacity: 1;
-                    background-color: @carousel-dots-after-active-background-color;
                 }
 
                 &::after {
                     opacity: 0;
                     transition: opacity @ebay-carousel-transition-time @ebay-carousel-transition-function;
-                    background-color: @carousel-dots-after-background-color;
+                    background-color: @carousel-dots-after-active-background-color;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


DylanPiercey:
@seangates 80px to both cover the height of the scroll bar and the dot controls.

trg:
Thanks @DylanPiercey!

PR #676, opened 2019-05-17
Textbox: Fix disable capability;

### LINKED ISSUE #669: Textbox: disabled attribute is not working
# Bug Report

## Description
unable to disable a text box.
## Workaround

https://www.loom.com/share/72720438a5af47f584c91ffe5d598f41

 <ebay-textbox disabled />

seangates:
## Description
- adds state with `disabled` property to widget
- adds discrete disabled attribute to text input in the template
- new tests to cover the component when it's disabled to prevent events from firing

## References
Fixes #669

### DIFF
--- a/src/components/ebay-textbox/index.js
+++ b/src/components/ebay-textbox/index.js
@@ -6,6 +6,11 @@ module.exports = require('marko-widgets').defineComponent({
     getWidgetConfig(input) {
         return { floatingLabel: input.floatingLabel };
     },
+    getInitialState(input) {
+        return Object.assign({}, input, {
+            disabled: Boolean(input.disabled)
+        });
+    },
     init(config) {
         this.config = config;
         this.initFloatingLabel();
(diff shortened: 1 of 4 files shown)


src/components/ebay-textbox/index.js
```diff
@@ -38,9 +43,11 @@ module.exports = require('marko-widgets').defineComponent({
 
 function forwardEvent(eventName) {
     return function(originalEvent, el) {
-        emitAndFire(this, `textbox-${eventName}`, {
-            originalEvent,
-            value: (el || this.el.querySelector('input, textarea')).value
-        });
+        if (!this.state.disabled) {
```
ianmcburnie: I'm confused as to why these events would get forwarded in the first place if the DOM element is disabled.

seangates: Essentially, our tests “fake” the click, focus, blur, etc. If I were to try and write the test for a disabled state without this check it would always fail.

Should I remove the test in order to remove this check?

ianmcburnie: Thanks for explaining. I don't think we should be adding logic to our code purely for the purpose of our test setup ... if that makes sense.

PR #677, opened 2019-05-20
ebay-menu: Added disabled state option & <ebay-menu-label> subcomponent

### LINKED ISSUE #683: Menu: Add support for HTML in button
We have received a couple of requests to provide HTML support for the button.

@pcanella has kindly provided a PR: https://github.com/eBay/ebayui-core/pull/677

pcanella:
## Description
Adding a `disabled` state to the component that will enable the built in `ebay-button` and thus disable the component both on server side render as well as client side state change.

Additionally, adding a custom label subcomponent named `<ebay-menu-label>` in order to customize the initial "label" that the user sees. Previously, you could only customize `<ebay-menu-item>` subcomponents, but we saw a need in our use case where we had to display a sprite next to the label, but were unable to. Adding this to the existing component seemed to be the right move here. 

Fixes #683 

## References
N/A unless you're interested in our JIRA user story :smile:

## Screenshots
Disabled State with custom menu:
Screen Shot 2019-05-20 at 12 19 09 PM

Normal State with custom menu:
Screen Shot 2019-05-20 at 12 20 18 PM


ianmcburnie:
Let's target this for 2.6.0 please.

pcanella:
@seangates gotcha! Also, should I set the branch to merge into as 2.4.0 or 2.5.0 or something else?

pcanella:
@ianmcburnie When is 2.6.0 schedule to be released?

pcanella:
@seangates updated to 2.4.0 with no troubles 👍  and applied your suggestion! Let me know if I can do anything else in the meantime

seangates:
Cool. Thanks. I'll have a look when I get a free moment.

ianmcburnie:
Incremented target to 2.5.0, but this will go into 2.6.0 once that branch is ready.

src/components/ebay-menu/examples/01-basic/template.marko
```diff
@@ -1,4 +1,9 @@
 <ebay-menu text="eBay Menu">
+    <ebay-menu-label>
+        <span class="flag" style="background-position: 0 -25px">
```
agliga: Why is the background-position style needed? 
I would recommend in examples to try to stray away from using tags and such that generally aren't recommended to use (I know style is one you generally should stray away from using).

src/components/ebay-menu/index.js
```diff
@@ -94,7 +94,9 @@ function getInitialState(input) {
         priority: input.priority,
         expanded: false,
         items,
-        checked: checkedItems
+        checked: checkedItems,
+        customLabel: input.label,
```
agliga: I would keep consistency between both the variable passed in and the state/data variable defined. 
So either change it to be either:
checked: checked,
label: input.label

OR
checkedItems: checkedItems,
customLabel: input.customLabel

That way it's easy to see where the variables come from

src/components/ebay-menu/examples/01-basic/template.marko
ianmcburnie: We should leave the basic example untouched, for backwards compatibility. A new example should be created showing how to use this new feature.

src/components/ebay-menu/index.js
ianmcburnie: @agliga I think the label is the only code here. We should leave the `checked: checkedItems` part untouched to prevent risk of regression/breaking change.

seangates:
Merged to a branch we have control over. Will continue in the new `menu-disabled-state` branch.

PR #684, opened 2019-05-28
Combobox: expanded attribute

### LINKED ISSUE #659: Combobox: provide the functionality to expand and show all options and collapse the listbox with an outer control
# Feature Request

## Description
The current combobox shows only the filtered options in the listbox upon expansion when there is a default value.  There is a request to show all options on the click of an outer control (similar to a traditional select dropdown); and collapse the listbox when the outer control is clicked again.

eBayUI version: 2.2.0

seangates:
## Description
- adds an `expanded` attribute
- presence of the `expanded` attribute forces the listbox to open

## Context
There was a need to allow force the listbox open. When combined with a change like `autocomplete="none"` it will allow you to show all options when it opens.

## References
Closes #659


src/components/ebay-combobox/examples/05-show-all-options/template.marko
```diff
@@ -0,0 +1,7 @@
+<ebay-combobox name="example2text" value="Basic Offer" autocomplete="list" expanded>
```
ianmcburnie: Shouldn't `autocomplete` be set to "none"?

ianmcburnie: ...which essentially just turns this into a default combobox.

seangates: But then that would disable filtering of the options when they type.

ianmcburnie: So I'm confused as to the purpose of this new example. It's name "show all options", but if autocomplete is set to list then we should only show the last of those 3 options.

seangates: `expanded` only affects the initial state of the listbox (all options) when it the expander expands it. But, while it's expanded it can still be filtered when a person begins typing.

That is in contrast to when `expanded=false`, and the expander expands the listbox, it shows a filtered set of results.

That's the only change: the initial list of the listbox options. `expander` doesn't affect filtering, typing, keyboard, etc. If we were to set `autocomplete="none"`, we would lose filtering.

seangates: Therefore, that's what this example shows.

ianmcburnie: Expanded and filtered should be completely independent of each other.

ianmcburnie: To be clear, this example shows what to me is an invalid state. While autocomplete is set to list, filtering should be on. If we want to show all options while combobox has an initial expanded state (not sure why this would ever be the case, comboboxes are always collapsed until input receives focus), autocomplete should be set to "none".

seangates:
@ianmcburnie Please have a look now. I believe this aligns with what you're looking for.

seangates:
Thanks, @agliga. I'm reworking it at the moment. Hopefully it will be the last rework. 😉

src/components/ebay-combobox/index.js
```diff
@@ -63,6 +75,22 @@ module.exports = require('marko-widgets').defineComponent({
                 simulateSpacebarClick: true
             });
         }
+
```
DylanPiercey: I feel like this logic should be simplified to:

```js
if (wasToggled) {
  if (isExpanded) {
    this.expander.expand();
  } else {
    this.expander.collapse();
  }
}
```

The first render check should not be needed since `wasToggled` would be true. This means we could remove https://github.com/eBay/ebayui-core/pull/684/files#diff-69c445b9b29be957915e2a875fc5472dR47, the only change we'd have to make is to make sure that `wasExpanded` defaults to false (currently would be undefined). 

https://github.com/eBay/ebayui-core/pull/684/files#diff-69c445b9b29be957915e2a875fc5472dR45
Would become:

```js
const wasExpanded = this.expanded || false;
```

src/components/ebay-combobox/index.js
```diff
@@ -30,24 +30,36 @@ module.exports = require('marko-widgets').defineComponent({
...
+    onRender(opts) {
+        const { expanded: wasExpanded } = this;
+        const isExpanded = this.expanded = this.state.expanded;
+        const isFirstRender = (opts && opts.firstRender);
+        const wasToggled = isExpanded !== wasExpanded;
 
         if (!this.state.disabled && this.state.options.length) {
+            const selectedIndex = this.getSelectedIndex(this.state.options, this.state.currentValue);
+
+            let autoInit = selectedIndex === -1 ? -1 : 0;
```
DylanPiercey: I would rather see this as:

```js
const autoInit = selectedIndex === -1 || this.state.autocomplete === 'none' ? -1 : 0;
```

src/components/ebay-combobox/index.js
seangates: Cool. Thanks for the explanation. Fixing.

src/components/ebay-combobox/index.js
seangates: Sure. For clarity I did:

```javascript
const autoInit = (selectedIndex === -1 || this.state.autocomplete === 'none') ? -1 : 0;
```

seangates:
@DylanPiercey Updated.

src/components/ebay-combobox/index.js
```diff
@@ -63,6 +73,18 @@ module.exports = require('marko-widgets').defineComponent({
                 simulateSpacebarClick: true
             });
         }
+
+        if (isExpanded && !wasExpanded) {
```
agliga: Wont this code not do anything?

This assumes that isExpanded is true and wasExpanded is false. That means that wasToggled will also be true (since isExpanded !== wasExpanded) which will execute line 81.

seangates: It will do something, but you're right that it is basically duplicated in the following if block(s).

PR #692, opened 2019-06-03
Replaced Object.assign and other polyfills with functions

### LINKED ISSUE #654: Upgrade to core-js v3
Let's upgrade to v3 of core-js at our earliest opportunity

https://github.com/zloirock/core-js

After upgrading, we might be able to drop the following polyfills/transforms:

* `"babel-plugin-transform-object-assign"`
* `"nodelist-foreach-polyfill"`

UPDATE: We'll probably need to leave the polyfills in. Babel/CoreJS brings in quite a lot of stuff when targeting IE11.

DylanPiercey:
## Description
Upgrade core-js to v3, removes nodelist-foreach polyfill (just uses Array.prototype.forEach.call) and removes `element-closest` polyfill which can be avoided by receiving root element via the second argument to the Marko event handler.

## References
Resolves #654


src/common/html-attributes/index.js
```diff
@@ -18,7 +19,7 @@ function processHtmlAttributes(input = {}) {
 
     let obj = htmlAttributes || wildcardAttributes;
     if (htmlAttributes && wildcardAttributes) {
-        obj = Object.assign(wildcardAttributes, htmlAttributes);
```
ianmcburnie: Hmm, maybe I'm missing the point, but using `core-js-pure` seems like going back to the day when we just wrote our own JavaScript functions to solve missing features of JavaScript. I thought the point of Babel was so that we could use modern JavaScript syntax *today*. I know this solves global pollution problem, but does it come at a readability/cognitive cost? I don't have a huge problem with this direction, but I think we should all have a consensus as team before moving forward. @senthilp Any thoughts on this?

ianmcburnie:
This should probably go into 2.7.0, so that we don't hold up the teams eagerly awaiting the releases of 2.5.0 and 2.6.0.

src/common/html-attributes/index.js
DylanPiercey: We’ve had this discussion previously. The problem with using Babel to automatically inline polyfills is that it includes way more than we need, even with the auto polyfilling. At the end of the day we’d have to explicitly list all the polyfills which doesn’t seem all that different than the approach we have here to me.

ianmcburnie: Ok. Good point. This does solve the bloat. Also, I guess it's not like we are doing this for *all* modern features of JavaScript, it's just a few. If we document this, and why we are doing it, we should be fine. But let's run it by the group at next stand up.

ianmcburnie: p.s. I think listing them all is *slightly* different, because at least we then get to keep the `Object.assign` syntax.

seangates: Personally, I thought we were doing a very good job to avoid any JS that might be a problem cross-browser except in a few cases.

Do we know how much global JS code reduction we gain by doing this?

DylanPiercey: @seangates it is a significant amount. By default it’s going to pull in polyfills for symbols & iterables + some other stuff.

ianmcburnie: If the savings are significant, then I'm totally on board with this. Let's all get on the same page at next standup.

senthilp: Sorry, I am late for this conversation. @DylanPiercey this goes against our original idea of writing modern JavaScript. We (especially the selling C2C team) are thinking about using deferential loading for creating JS bundles. Using `core-js` will not benefit them. Currently, 88% of eBay customers are in browsers that support all of ES6/7. Adding a dependency on a new polyfill will not be ideal.

Also, let's say tomorrow we drop IE11 support, should we come back and change this code? Maybe I am missing something, we can discuss it further. 

@ianmcburnie let's wait on this PR.

DylanPiercey: The alternative here as far as I can tell is to bring in a large amount of polyfills (although automatically) which I don't think consuming teams would be happy with.

At the end of the day, if we drop ie11 tomorrow we will still have to update this package and redeploy it because we will have to update our `Browserslist` config and rebuild our scripts with babel. When doing that it doesn't seem unreasonable to me that we also re-evaluate which shims are still required.

senthilp: Valid arguments. We will discuss it further.

ianmcburnie:
Senthil's out this week, but I think we were all in agreement to move forward with Dylan's change after discussion. I've approved this PR and I say we merge it to 2.7.0 to unblock ourselves and keep up the pace!

PR #693, opened 2019-06-04
683 - Menu disabled state

### LINKED ISSUE #683: Menu: Add support for HTML in button

seangates:
**Recreated from a PR where we did not have control of the branch: #677**

## Description
Adding a `disabled` state to the component that will enable the built in `ebay-button` and thus disable the component both on server side render as well as client side state change.

Additionally, adding a custom label subcomponent named `<ebay-menu-label>` in order to customize the initial "label" that the user sees. Previously, you could only customize `<ebay-menu-item>` subcomponents, but we saw a need in our use case where we had to display a sprite next to the label, but were unable to. Adding this to the existing component seemed to be the right move here. 

Fixes #683 

## References
N/A unless you're interested in our JIRA user story :smile:

## Screenshots
Disabled State with custom menu:
Screen Shot 2019-05-20 at 12 19 09 PM

Normal State with custom menu:
Screen Shot 2019-05-20 at 12 20 18 PM


ianmcburnie:
Thanks for recreating this PR @seangates.

First of all, let's restore the 1st example as to how it was before. We should add a new example for this HTML label case. And an example for the disabled case too maybe.

seangates:
@ianmcburnie I think it's in a better spot.

ianmcburnie:
@seangates I'm seeing a regression in the menu shortcut key behaviour on my local env (it's not jumping to "Denmark" when I press "n"). Seems to be working fine in 2.5.0 on marketplace. Can you double check, please?

seangates:
Sure, I’ll jump on it.

PR #697, opened 2019-06-05
Fix for when multiline is chosen and there is no value;

### LINKED ISSUE #675: combobox: editable combobox select from option does not work in ie 11
# Bug Report

## Description
editable combobox - not able to select from the option list

### LINKED ISSUE #696: textbox: when text area does not have any value by default. it shows false
# Bug Report

## Description
when there are no default values in a text area it shows false.
Reproducible in marketplace

seangates:
## Description
- adds logic check for presence of value

## References
Fixes #696

### DIFF
--- a/src/components/ebay-textbox/template.marko
+++ b/src/components/ebay-textbox/template.marko
@@ -51,7 +51,7 @@
         w-onfocus="handleFocus"
         w-onblur="handleBlur"
         ${processHtmlAttributes(data)}>
-        <if(data.multiline)>
+        <if(data.multiline && data.value)>
             ${data.value}
         </if>
     </>


DylanPiercey:
How does this fix #675?

seangates:
`data.value` is `false` if no value is supplied. We don't want to provide a value to the multiline `<textarea>` if no value was sent.

seangates:
Oh, whoops. I see what you mean. That's the wrong issue number. Let me fix that.

seangates:
@DylanPiercey This fixes #697.

DylanPiercey:
#697 is this PR :p

DylanPiercey:
I think this is meant for https://github.com/eBay/ebayui-core/issues/696?

Just to be clear, if you set

```marko
<div>
  ${false}
</div>
```

Marko will not render the `false`. If the `false` is getting through, that means it is being cast to a string (eg: `${'false'}`). This PR won't fix that though, it'd need to be fixed further up the line.

DylanPiercey:
Disregard my last message, apparently this behavior changed in Marko 4. This check is necessary to be consistent :\

seangates:
Moving forward. Will publish the patch.

PR #698, opened 2019-06-06
Fix middle and prepend editing

### LINKED ISSUE #641: Combobox: unable to edit text in the middle or pre-pend text to an existing text
# Bug Report

## Description
Repor;
Click on a combobox
type some text
Try adding/deleting some text in the middle of the currently added text
The cursor automatically shifts to the right
https://www.loom.com/share/3b19029c733c4683bc91efa4ad0b3cb9

## Workaround
The user has to delete the entire text and start again

seangates:
## Description
- removes `moveCursorToEnd`

## Context
Moving the cursor caused more problems than it solved.

## References
Fixes #641


ianmcburnie:
@seangates Seems like we should close this based on QA comments, then? Maybe go back to the drawing board on this later.

PR #711, opened 2019-06-14
DevDeps: upgraded patch, minor and majors

ianmcburnie:
Updated all devDeps except marko-prettyprint which was having issues.

Commits are atomic, so we can merge this one without squashing.


ianmcburnie:
Reverted lasso-babel-transform to v1. Build is now passing.

Issue #710, opened 2019-06-14
combobox: throws a JS error when typing in the textbox if there are no options


melquan:
# Bug Report

## Description
When there are no options and user starts typing in the textbox of the combobox, a JS error is thrown and the the entered value cannot be emitted.

## Screenshots
![image](https://user-images.githubusercontent.com/19419138/59518119-db1feb80-8e79-11e9-8994-2e44868a4789.png)

![image](https://user-images.githubusercontent.com/19419138/59518333-4ff32580-8e7a-11e9-8999-dbbf5f670453.png)

hramakumar:
#651

melquan:
@hramakumar I thought there was already a ticket filed but didn't find it.  Sorry I overlooked it!

hramakumar:
no problem. Let's get this fixed.
We will keep you posted

hramakumar:
verified

ianmcburnie:
Fixed in 2.6.0. Closing.

PR #712, opened 2019-06-14
Combobox: fixes logic with no options;

### LINKED ISSUE #651: combobox: when there are no options -no expansion-exception in expand event
# Bug Report

## Description
typing on the combobox throws an exception in expand event

## Screenshots
image

seangates:
## Description
- adds no options example
- proper options length check in render
- expander check in keyUp handler
- expander check in `toggleListbox` function
- moves change event to make it work properly with the `blur`

## Context
Where there were no options available there would be errors. Also, the blur was not sending the change event when there were no options.

## References
Fixes #651 & #710

### DIFF
--- a/src/components/ebay-combobox/examples/05-no-options/template.marko
+++ b/src/components/ebay-combobox/examples/05-no-options/template.marko
@@ -1,0 +1,1 @@
+<ebay-combobox name="example2text" autocomplete="list"></ebay-combobox>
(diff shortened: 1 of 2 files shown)


seangates:
@ianmcburnie There's a problem in one of the logic statements which @melquan uncovered. I've since fixed it and pushed another commit. Can you please re-review just that one change?

Issue #689, opened 2019-05-30
Select: Support optgroup


melquan:
# Feature Request
Support optgroup for Select 

## Description
We have a need to show groupings in our dropdowns and would like to request the support of `optgroup` in the select component

## Screenshots
In our current flow:
![image](https://user-images.githubusercontent.com/19419138/58643042-5ea2df80-82b3-11e9-814b-e65a6f547ccf.png)

ianmcburnie:
Going to split this into two tickets. One for select, one for combobox.

ianmcburnie:
Done. Will be available in 2.7.0.

PR #719, opened 2019-06-19
Select: add optgroup

seangates:
## Description
- adds an `<optgroup>` using a matched `optgroup` string

## References
Completes #689 

## Screenshots
59790957-ef068b80-9285-11e9-8b36-3dd98ad8f899

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -335,7 +335,8 @@
             "@style": "string",
             "@value": "string",
             "@text": "string",
-            "@selected": "boolean"
+            "@selected": "boolean",
+            "@optgroup": "string"
         }
     },
     "<ebay-select-option>": {
--- a/src/components/ebay-select/examples/06-optgroup/template.marko
+++ b/src/components/ebay-select/examples/06-optgroup/template.marko
@@ -1,0 +1,14 @@
+<ebay-select name="formFieldName">
+    <ebay-select-option value="pre" text="Option PRE"/>
+    <ebay-select-option optgroup="Group 1" value="1" text="Option 1"/>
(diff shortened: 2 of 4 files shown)


DylanPiercey:
@ianmcburnie that is possible in Marko 3 and I'd recommend doing that now to avoid additional churn with the Marko 4 implementation.

ianmcburnie:
@DylanPiercey I agree about the churn. However, Sean was having issues with nested tags. @seangates ?

seangates:
@DylanPiercey Let me chat with you offline.

seangates:
@ianmcburnie (et al),

After trying the additional markup, which more closely matches the markup of a native `<optgroup>`, it appears to not work with a `<ebay-select-option>` tag for both root level and inside of a `<ebay-select-optgroup>` tag.

It _does_ work if we were to have three sub tags `<ebay-select-option>`, `<ebay-select-optgroup>`, and `<ebay-select-optgroup-option>`. But I don't think we want to have the API be that inflexible.

Lastly, we certainly could use only two tags with Marko v4, though, as Dylan and I confirmed (another nail in v3's coffin).

In conclusion, the question becomes: do we want to implement this as is (the present code with an optgroup label, which works) and then change the API in Marko v4 (because we can do it better; e.g. the right way with just two tags), or do we punt it to Core v4 and do it correctly with just the two tags once Marko v4 components are in place?

ianmcburnie:
@seangates I'm happy proceeding with what you have, in order to unblock selling team, and then we can change the API to be more what we would like in v4.

seangates:
Sounds good. Proceeding.

Issue #700, opened 2019-06-06
menu and listbox: change events are triggered even when clicking an already selected option


hramakumar:
# Bug Report

## Description
In menu and listbox the change events are triggered even when clicking an already selected option from the list. Ideally, we need to have it triggered only when there is a change in the menu option

hramakumar:
with 2.6.0 its fixed in ebay select but the issue is still present in combobox and listbox

ianmcburnie:
I don't even think this was targeted as a fix for 2.6.0, so not sure how it got fixed. Currently we have it for 2.8.0. We could maybe bring it forward to 2.7.0 though.

seangates:
@hramakumar The combobox change event only happens on "blur". For combobox, are you listening to another event besides `combobox-change`? Maybe `combobox-select`?

seangates:
If the above is the case, then I will remove the combobox from this issue.

seangates:
Per conversation, I will look at `listbox` too. Also, `combobox` works as expected.

ianmcburnie:
This is in 2.7.0 pre-release. Closing.

Issue #713, opened 2019-06-17
Menu: make independent of button


ianmcburnie:
The MIND patterns have been updated to make menu independent of any button. This has the following benefits:

* Menu can be embedded directly in the page without any button (similar to how we made form controls independent of any label)
* Better support for mWeb, which is also often embedded directly on page without button

Examples:

* https://ebay.github.io/mindpatterns/input/menu/index.html
* https://ebay.github.io/mindpatterns/input/menu-button/index.html
* https://ebay.github.io/mindpatterns/input/combobox/index.html

Related tickets:

* https://github.com/eBay/skin/issues/630

msendlakowski:
@ianmcburnie  We should have time to do this in 4.1.0 if its still high priority.

ianmcburnie:
Done in 4.2.0.

PR #720, opened 2019-06-19
Menu: fix event firing when radio item already selected

seangates:
## Description
- adds a check for existing index selection before proceeding with checking the item

## Context
When a user would click an already-checked radio menu item we would still fire the `menu-change` event. Adding this check prevents the component from even trying to go through the "change" flow if the item is already selected.

## References
Relates to #700

### DIFF
--- a/src/components/ebay-menu/index.js
+++ b/src/components/ebay-menu/index.js
@@ -195,11 +195,6 @@ function onRender(event) {
     }
 }
 
-/**
- * Internal marko function, can be triggered from both makeup and API
- * http://v3.markojs.com/docs/marko-widgets/javascript-api/#setstatedirtyname-value
- * @param {Boolean} expanded
- */
 function update_expanded(expanded) { // eslint-disable-line camelcase
     if ((expanded && this.buttonEl.getAttribute('aria-expanded') === 'false') ||
         (!expanded && this.buttonEl.getAttribute('aria-expanded') === 'true')) {
@@ -207,10 +202,6 @@ function update_expanded(expanded) { // eslint-disable-line camelcase
     }
 }
 
-/**
- * Common processing after data change via both UI and API
- * @param {Array} itemIndexes
- */
 function processAfterStateChange(itemIndexes) {
     const itemIndex = itemIndexes[(itemIndexes.length - 1)];
     const itemEl = this.itemEls[itemIndex];
@@ -244,11 +235,6 @@ function processAfterStateChange(itemIndexes) {
     }
 }
 
-/**
- * Handle normal mouse click for item
- * @param {MouseEvent} e
- * @param {HTMLElement} itemEl
- */
 function handleItemClick(e, itemEl) {
     const itemElIndex = getItemElementIndex(itemEl);
     if (this.getCheckedList().indexOf(itemElIndex) === -1) {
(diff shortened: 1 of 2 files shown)


src/components/ebay-menu/test/test.browser.js
```diff
@@ -188,25 +188,6 @@ describe('given the menu is in the expanded state', () => {
         });
     });
 
-    // these two tests are failing on CI.
```
ianmcburnie: Were these commented tests removed on purpose? I'd like to keep them around and fix at some point.

src/components/ebay-menu/index.js
```diff
@@ -250,7 +250,12 @@ function processAfterStateChange(itemIndexes) {
  * @param {HTMLElement} itemEl
  */
 function handleItemClick(e, itemEl) {
-    this.setCheckedItem(getItemElementIndex(itemEl), true);
+    const itemElIndex = getItemElementIndex(itemEl);
+    if (this.getCheckedList().indexOf(itemElIndex) === -1) {
```
ianmcburnie: For this first conditional block, do we need to check that it's a radio and not a checkbox?

Also, are these code comments correct? It says getCheckedList is only for checkboxes, not radios.

```js
/**
 * Determine currently checked items (for checkbox case)
 * @returns {Array} checked indexes
 */
function getCheckedList() {
```

src/components/ebay-menu/test/test.browser.js
seangates: I removed them as they were commented out a year ago. We shouldn't have commented tests. We should fix them or remove them.

ianmcburnie: Couldn't have been a year ago. These tests are for the menu shortcut which I did a month or two back.

ianmcburnie: I'll create an issue and copy the code into it.

src/components/ebay-menu/index.js
seangates: >do we need to check that it's a radio and not a checkbox?

We have a logical case for emitting an event that is neither a `checkbox` or `radio` type, hence I didn't limit it to the radio use case: https://github.com/eBay/ebayui-core/blob/92c456f10fbfc0712d61a9e551db83e68e74af09/src/components/ebay-menu/index.js#L242-L243

>are these code comments correct?

Probably not! Let me review the comment throughout.

src/components/ebay-menu/test/test.browser.js
seangates: Hmmm. My editor must have been lying to me then. But, yeah, we shouldn't have commented code (tests or otherwise). Moving it to an issue is a great idea.

ianmcburnie: https://github.com/eBay/ebayui-core/issues/721

src/components/ebay-menu/index.js
ianmcburnie: Ok, it's a little tricky to follow. I agree with @DylanPiercey that this component would benefit from a refactor when we go to MarkoV4. I think when we separate it from button (#713) is going to help too.

seangates: Yes, for sure. It didn't get refactored so it has lots of "ins and outs".

PR #722, opened 2019-06-24
Carousel: now using CSS from Skin

### LINKED ISSUE #718: Carousel: use styles from Skin
Currently, carousel is the only component which doesn't have its styles pulled from skin. This presents some problems for global header team who need to wrap the classes with an extra layer of specificity (like they do with other skin modules).

Depends on: https://github.com/eBay/skin/issues/131

ianmcburnie:
Closes #718 

This PR replaces all local carousel CSS with Skin CSS.

@DylanPiercey Please do a quick local smoke test.

@hramakumar I will push another 2.7.0 prerelease containing this change.


ianmcburnie:
Thanks. Will take a look.

ianmcburnie:
@DylanPiercey I'm also seeing this issue in the 2.7..0 branch without this set of changes. Seem like the shadow dom isn't being populated after a few clicks:

Screen Shot 2019-06-24 at 5 09 17 PM

Also notice how `xlink:href` and `href` attributes become out of sync. Maybe an issue with `ebay-icon`. I wonder if one of the dev dep updates caused this? Do you mind taking a quick look?

DylanPiercey:
@ianmcburnie it looks like this issue is isolated to Marko v3 and related to my refactoring to a dynamic tag in prep for Marko 4.

For some reason when just the icon piece gets updated it uses the wrong namespace.

A simple fix for this is to force the entire icon to be replaced when it is going to be updated by Marko (3) which involves adding a `w-id` on any icon that does not have one.

I think the simplest way forward would just be to enforce this via a transform by adding:

```javascript
        if (!el.hasAttribute('w-id')) {
            // can be removed in Marko 4
            el.setAttributeValue('w-id', builder.literal(`use_icon_${iconName}`));
        }
```

[Here](https://github.com/eBay/ebayui-core/blob/92144ace5f3552eefe53deaef8182fe9078c44cf/src/components/ebay-icon/transformer.js#L24)

ianmcburnie:
That seems to do the trick, thanks.

PR #723, opened 2019-06-26
Increased browser support for tests

### LINKED ISSUE #497: CI: get browserstack tests running on Firefox and Safari
# Feature Request

Like the title says, let's get our browserstack tests running on Firefox and Safari. Currently they only run on Chrome.

Notes: https://github.com/eBay/ebayui-core/blob/master/marko-cli.js#L29

(IE not compiling properly, because we don’t use `lasso-babel-transform` when running tests, according to Dylan).

DylanPiercey:
## Description
This PR updates our tests to run in the following browsers:
* Chrome 49
* Firefox latest
* Safari latest

Other browsers may not be possible with the current setup since some of them do not implement the features needed via the WebDriver spec. Still digging into to these to verify, however thought this was a good base to work off of.

## References
Closes #497


ianmcburnie:
p.s. I changed the scope of #497 to just firefox and safari. We can create a new ticket for IE and/or Edge if you think those are close too.

PR #728, opened 2019-06-28
Rename ebay-listbox to ebay-listbox-button and ebay-menu to ebay-menu…

### LINKED ISSUE #725: Menu: rename to ebay-menu-button
As a segway to v4, let's rename the current `ebay-menu` to `ebay-menu-button`. API will remain unchanged in this version and menu will still be coupled to button.

In v4 we will completely separate the menu from the button #713 

Doing it like this means we get to Markov4 quicker, but can go back and add an `ebay-menu` to CoreUI v3.1 if it is asked for.

### LINKED ISSUE #726: Listbox: rename to ebay-listbox-button
As a segway to v4, let's rename the current `ebay-listbox` to `ebay-listbox-button`. API will remain unchanged in this version and listbox will still be coupled to button.

In v4 we will completely separate the menu from the button #690 

Doing it like this means we get to Markov4 quicker, but can go back and add an `ebay-listbox` to CoreUI v3.1 if it is asked for.

DylanPiercey:
## Description
This PR renames the `ebay-listbox` => `ebay-listbox-button` and `ebay-menu` => `ebay-menu-button`.

## Context
I've done this via a transform and so currently both of the names for these components are supported, however the old name now logs a deprecation warning.

Although this is registered as a transform, after we upgrade to Marko 4 we can just change the `transform` field in the `marko.json` to `migrate` and we will get automated migrations for this change 😄.

## References
Fixes #725 
Fixes #726


ianmcburnie:
Thanks. This may be a teeny bit premature as this change is slated for v3.0.0 release. We still have to get through some backlog items in v2.8.0 (and possibly a v2.9.0). So just make sure to keep the branch up to date.

I like the automated migrations!

DylanPiercey:
@ianmcburnie the nice thing about this is that it is not a breaking change and so we don't necessarily need to wait for 3.0.0.

It would be a breaking change if we removed support for the older names, but this keeps both in place with a deprecation warning so that we can remove the old names when we see fit (which would be if we decided to back port the new listbox and menu components).

ianmcburnie:
Ah okay. But don't we have to make changes in two places - ebay-menu & ebay-menu-button? Sorry, I haven't fully looked through the changes yet!

DylanPiercey:
The existing names are now just registered as transforms (https://github.com/eBay/ebayui-core/pull/728/files#diff-e56f5fdf7a730e2c503dd8217ef07299R199) which do a rename to their new names. That's how both are supported at the same time.

ianmcburnie:
Nice! The magic of transforms. Okay, let's target this one for 2.8.0 (just to reduce the QA scope of 2.7.0 a little - as Hemnath is close on that one).

I guess now we don't need a CoreUI v3.0.0 release purely for these renames. Although if there are any other breaking changes we'd like to make before we upgrade to Marko v4, we could use the v3.0.0 release for that. We can all discuss on Monday when Sean is back.

ianmcburnie:
@DylanPiercey I just changed the target branch to 2.8.0, looks like your feature branch will now need rebasing.

ianmcburnie:
Hmm, something still doesn't seem right. 35 commits. 95 changed files...

DylanPiercey:
Should be good now, pending CI.

PR #747, opened 2019-07-09
Issues fixed for 2.8.0

### LINKED ISSUE #551: Notice: update type "priority" to type "attention"
On the playbook and skin side we changed the terminology from "priority" to "attention".

We can update in a backwards compatible way, leaving the old "priority" value in there until next major version (or squeeze this change into 2.0.0 now).

### LINKED ISSUE #695: Menu: giant icon in examples
Not sure of the impact of this, may just be our example page, but should be fixed nonetheless:

Screen Shot 2019-06-05 at 8 21 04 AM

agliga:
## Description

Added two fixes to 2.8.0. Fixed priority to be attention and changed ebay-menu icon to be settings

## Context

Priority was changed to attention in skin. The microphone icon was not available in ds4 so changed it to a generic settings ones

## References

Fixes #695
Fixes #551


src/components/ebay-notice/template.marko
```diff
@@ -3,7 +3,7 @@
 </script>
 
 <var type=(data.type || "page")/>
-<var status=(data.status || "priority")/>
+<var status=(data.status || "attention")/>
```
ianmcburnie: Can we also check for `priority` for backwards compatibility? We can remove that alias in v3.

agliga: So not sure if that is needed. The status is passed directly to skin to render that given priority. So passing priority will just pass it directly to skin. 
The only thing that was changed here was the default value.

ianmcburnie: Ah, "priority" was the default? In that case you're right, it's not needed.

ianmcburnie:
Conflict needs fixing though.

src/components/ebay-menu/examples/12-icon-without-toggle/template.marko
```diff
@@ -0,0 +1,10 @@
+/**
+ * Note: Due to limitations in Marko v4, this tag must be within a parent component.
+ * Below we have added an empty class to force this to be the case.
+ */
+class {};
+<ebay-menu icon="settings" a11y-text="Microphone" no-toggle-icon>
```
seangates: `a11y-text="Settings"`?

src/components/ebay-menu-button/examples/12-icon-without-toggle/template.marko
```diff
@@ -3,7 +3,7 @@
  * Below we have added an empty class to force this to be the case.
  */
 class {};
-<ebay-menu-button icon="mic" a11y-text="Microphone" no-toggle-icon>
+<ebay-menu icon="settings" a11y-text="Settings" no-toggle-icon>
```
seangates: Oh, this changes from `<ebay-menu-button ... >` to `<ebay-menu ... >`, which is probably not correct.

agliga: Good catch!

Issue #752, opened 2019-07-10
Button: classname coming as ${variant}-btn when rendered


pranaykondekar:
button component have invalid classnames when passed a variant.
Rendered markup is coming as 
`<button class="${variant}-btn ${variant}-btn--no-text ${baseClass}--${size} ${variant}-btn--secondary" type="button">
  <svg class="expand-btn__icon" xmlns="http://www.w3.org/2000/svg" focusable="false" aria-hidden="true"><use xlink:href="#icon-chevron-down-bold"></use></svg>
</button>`

ianmcburnie:
Closing as invalid.

PR #753, opened 2019-07-10
fix: fix for button class name template string

pranaykondekar:
#752 
Template string of baseClass and sizeClass in button component
## Context
Class name was getting rendered as 
`<button class="${variant}-btn ${variant}-btn--no-text ${baseClass}--${size} ${variant}-btn--secondary" type="button"> <svg class="expand-btn__icon" xmlns="http://www.w3.org/2000/svg" focusable="false" aria-hidden="true"><use xlink:href="#icon-chevron-down-bold"></use></svg> </button>`


DylanPiercey:
@pranaykondekar was this in a Marko 3 or 4 app? What is the exact Marko version you are using?

This PR updates the component to use the Marko 4 syntax for dynamic strings in attributes (in Marko 3 these get passed through as template literals in the final output which is not supported by all the browsers we are targeting).

pranaykondekar:
@DylanPiercey We are on marko 4.12.3

DylanPiercey:
@pranaykondekar the ebayui-core components are written in Marko v3 and work in Marko v4 thanks to the compatibility layer. The compatibility layer has seen a huge number of improvements since 4.12.3 (we  are currently on 4.18.8) and you'd need to upgrade Marko to resolve the issue. If there are issues in upgrading Marko or you find any regressions I am happy to help.

pranaykondekar:
Thank you @DylanPiercey . Upgrading marko to latest version solved the issue. Closing this PR.

Issue #749, opened 2019-07-09
Button: add a fake-link variation


seangates:
# Feature Request: Add a fake link variation to the button component

## Description
Presently the button only supports a fake button variation (an anchor styled as a button) using the `<ebay-button>`'s `href` attribute. We should conversely add a "fake link" variation which is a button which is styled to look like a link.

## Context
Some teams need to mask an action in what looks like a link. For example, some text which is on its own in a page but needs to appear clickable. A visual "button" style would be too prominent, but a link style would be add the appropriate affordance the item is clickable.

This variation is already [available in Skin](https://ebay.github.io/skin/ds6/index.html#link-fake) but should be added to the button component.

## Screenshots

This example is designed to open a dialog when clicking on the price, but it really shouldn't look like a button:
![image](https://user-images.githubusercontent.com/105656/60942126-bc591d00-a29e-11e9-82e9-3932901f14a8.png)

PR #755, opened 2019-07-11
feat: support for fake-link button variant

pranaykondekar:
#749 
Support for `fake-link` button variant

fake-link


seangates:
Oh, yeah. Docs!

PR #757, opened 2019-07-12
Combobox: Fix IE11 event order issue

### LINKED ISSUE #675: combobox: editable combobox select from option does not work in ie 11

seangates:
## Description
- add `mousedown` listener for option "click"
- amend check for "clicked" item in blur listener

## Context
In IE11 when `click` is used it's too late in the event processes as `blur` fires first and nullifies the handling for the selected option. This change detects the click earlier with the `mousedown` event and then handles the blur processing correctly.

## References
Fixes #675


ianmcburnie:
@seangates It feels a little jumpy/quick-to-respond. Can you check for mousedown, then do the action when mouse is released?

seangates:
@ianmcburnie I've updated the processing to just set a `optionClicked` flag, then use it later in processing, retaining the "click" event handling.

ianmcburnie:
Does this need rebasing? Seeing a lot of commits and files changed.

seangates:
@ianmcburnie All better!

PR #759, opened 2019-07-15
Changed all icons to have a viewBox to make them larger

### LINKED ISSUE #615: Icon: "missing" icon does not have viewbox
# Bug Report

## Description
ds4 icons are broken

## Screenshots
https://www.useloom.com/share/909d4da8ac3f4b8db57327521ac7987c

agliga:
## Description
Made icon larger by increasing viewBox

## Context
This allows users to see if there is an icon error when switching between DS4/6

## References
Fixes #615 

## Screenshots
image

### DIFF
--- a/src/components/ebay-icon/symbols/alert/ds6.marko
+++ b/src/components/ebay-icon/symbols/alert/ds6.marko
@@ -1,4 +1,4 @@
-<symbol id="icon-alert">
+<symbol id="icon-alert" viewBox="0 0 16 16">
     <g fill="#111820" fill-rule="nonzero">
     <g transform="translate(5.5 4.5)">
         <circle cx="2.3" cy="6.1" r="1"/>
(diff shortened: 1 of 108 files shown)


agliga:
> Is the viewbox the same for all the icons, though? We have specific sizes for each icon, from what I remember working on them.

Yes because this is ONLY for the "missing" icons which are all the same and do not have a specific size

src/components/ebay-icon/symbols/alert/ds6.marko
```diff
@@ -1,4 +1,4 @@
-<symbol id="icon-alert">
+<symbol id="icon-alert" viewBox="0 0 16 16">
```
seangates: I see the change for all the icons, though, not just the missing one.

seangates: Okay, ignore me. I see now. These SVG files are for when they've used an old DS4 (now-reoved in DS6) icon, correct?

agliga: Yes, that or if there is a DS6 icon which does not exist in DS4

agliga:
Forced push in order to fix the unverified commit

PR #761, opened 2019-07-17
Carousel: Pause autoplay on user interaction

### LINKED ISSUE #586: Carousel: Autoplay should pause when slide content has keyboard focus (or mouse hover)
When carousel has mouse hover, or interactive slide content has keyboard focus, the autoplay should pause in order to give user sufficient time to activate that content. This could be handled at the app level with our current API, but as it will often get overlooked, let's see if we can handle it at the component level.

DylanPiercey:
## Description
Updates the carousel to prevent any slides from happening when in autoplay mode during a user interaction.

User interactions: Focused, hovered, touched.

## References
Fixes #586


seangates:
@ianmcburnie @DylanPiercey Build failed on these new tests.

seangates:
@DylanPiercey This appears to fail in the FF on Windows 10 test only.

DylanPiercey:
Yeah I saw. Will be taking a look today.

PR #768, opened 2019-07-19
Increase timeout for carousel autoplay tests

DylanPiercey:
## Description
As discussed earlier this PR increases the timeout for all of the carousel autoplay tests to avoid a race condition causing the CI to intermittently fail.

(Lets definitely wait for the CI to pass before merging this)


agliga:
Still failing? Maybe need more timeout

DylanPiercey:
Hmm, could be that `carousel-update` is actually called twice. Will dig in more in a second, don't merge this one :).

DylanPiercey:
I think this one is okay now.

PR #748, opened 2019-07-09
Remove imperative API's

DylanPiercey:
## Description
Removes all imperative API's and exposing properties/events on the DOM as described in #732.

## References
Implements #732 

This again is semver major, but I'll rebase as necessary.

Issue #733, opened 2019-07-02
Upgrade to Skin v8.0.0


ianmcburnie:
Expected breaking changes in Skin v8.0.0:

- [x] Skin `listbox` class being renamed to `listbox-button` (plus minor DOM & BEM changes too) (@agliga)
- [x] Skin `menu` class being renamed to `menu-button` (plus minor DOM & BEM changes too) (@agliga)
- [x] Pill is being separated into its own module (class becomes `pill` instead of `button--pill`) (@seangates)
- [x] Carousel classes `play` and `pause` were consolidated into a single `playback` class (@msendlakowski, @seangates)
- [x] CTA button was moved out of button into own `@ebay/skin/cta-button` module (@msendlakowski)
- [x] Expand button was moved out of button into own `@ebay/skin/expand-button` module (@agliga)
- [x] LESS variables and mixins were renamed & refactored. This may not impact coreui. (@seangates)

seangates:
@ianmcburnie @agliga @msendlakowski Just add your name next to the item you will complete, then check it off (using `[x]` in the description) once completed.

PR #794, opened 2019-07-30
Pill: migrate pill class names and remove ebay-button

seangates:
## Description
- updates pill class names
- removes `<ebay-button>` in favor of a native button
- updates some template logic lost when removing `<ebay-button>`
- updates tests accordingly

## Context
The pill now is it's own Skin CSS module. Accordingly, the pill behaves a bit differently than the `<ebay-button>` component.

**Note:** test failures were overcome by removing the imperative API event emitted through `emitAndFire`. This might cause some conflicts given the #748 changes, but is small enough I think it won't be a problem.

## References
Related to #733


src/components/ebay-pill/index.js
```diff
@@ -19,10 +18,9 @@ module.exports = require('marko-widgets').defineComponent({
         }
     },
     handleButtonClick(event) {
-        this.setState('pressed', !this.state.pressed);
-        emitAndFire(this, 'button-click', event);
-    },
-    handleButtonEscape(event) {
```
agliga: Why are we removing this? Just curious.

seangates: We aren't using the `<ebay-button>` anymore. That event we had to manually bubble, but we don't need to anymore. Note the changes in the template removing the `button-escape` event because we changed to a native button.

seangates:
@agliga I just rebased, if you would like to quickly look again.

seangates:
@agliga I noticed this too, and even mentioned it in Slack. Basically, there was a regression we missed in v2.7.0 (or prior, as I haven't had a chance to look farther back).

seangates:
@agliga Let me look see if I can fix it quick.

msendlakowski:
@seangates Are you still looking into fixing that overflow text issue, or is this ready to merge?

seangates:
@msendlakowski I stepped over to fix the checkbox to get the patch out for v2.8. I need to talk to @DylanPiercey about some changes he made back in v2.4 which were the source of the regression.

seangates:
@agliga No. 42 is max per Design System specs. The pill is a fixed width and fixed height.

Yeah, this is a stopgap truncation as the other solution (when re-implemented) was ... worse.

PR #796, opened 2019-07-30
Listbox and menu skin changes

agliga:
## Description
Fixed listbox and menu to work with new skin code

## References
Related to #733


src/components/ebay-listbox-button/template.marko
```diff
@@ -17,7 +17,8 @@
     <button
         w-id="button"
         class=[
-            "listbox__control",
+            "btn",
```
seangates: The `.expand-btn` class didn't provide the correct styles? Do we just need to put these classes in a different order in the class attribute?

agliga: Good catch! I had issues loading expanded button and I was trying something else. I'll revert it

PR #797, opened 2019-07-30
Use Marko 4 for demos and the primary dev environment

mlrawlings:
## Description

- Updates the dialog demos to use inline Marko 4 class components, rather than accessing the (deprecated) imperative APIs from the DOM
- Makes Marko 4 the default version of Marko for development
- Improves npm scripts for installing Marko versions and cleanup

## Context

- We have removed the imperative APIs for 3.0.0 (#748)
- Now that the demos use Marko 4, it makes sense to have that be the default devDependency
- The existing versions of `installMarkoV3`/`installMarkoV4` touch `package.json`/`yarn.lock`.  This was less of an issue when Marko 3 was the primary development version since no new versions were being published in that semver range.
- Commands that need cleanup now do so unconditionally even if the tests fail


seangates:
@mlrawlings This should go into the v3.0.0 branch. Sorry, but you have a conflict now.

mlrawlings:
Per offline discussion, changed the target back to `2.8.0`

PR #798, opened 2019-07-30
CTA: Move out of button and into its own tag

msendlakowski:
## Description
Move the CTA button variant out of `<ebay-button>` and into its own `<ebay-cta-button>` component. 

## Context
This simplifies the normal button and the CTA button logic. It follows the changes made to skin. It also sets us up to be ready for the more complex CTA buttons coming in DS 6.5.

## References
Related to https://github.com/eBay/ebayui-core/issues/733

## Screenshots
Screen Shot 2019-07-30 at 3 33 00 PM
Screen Shot 2019-07-30 at 3 33 12 PM

### DIFF
--- a/src/components/ebay-cta-button/index.marko
+++ b/src/components/ebay-cta-button/index.marko
@@ -1,0 +1,25 @@
+<script marko-init>
+    var getWidgetId = require("../../common/get-marko-3-widget-id");
+    var processHtmlAttributes = require("../../common/html-attributes");
+</script>
+
+<var size=(data.size || "medium")/>
+<var baseClass="cta-btn"/>
+<var sizeClass="${baseClass}--${size}"/>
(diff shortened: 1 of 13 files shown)


src/components/ebay-cta-button/index.marko
```diff
@@ -0,0 +1,25 @@
...
+    id=(data.id || getWidgetId(out))
+    class=[
+        data.class,
+        baseClass,
+        sizeClass
+    ]
+    style=data.style>
+
+    <span class="cta-btn__cell">
+        <span><invoke data.renderBody(out) if(data.renderBody)/></span>
```
seangates: Please use `w-body` on the span. No need for the `<invoke .../>`.
```suggestion
        <span w-body/>
```

seangates: Never mind. I think I'm mistaken here.

PR #803, opened 2019-08-01
Expand button as separate module

agliga:
## Description

Changed expand button to be it's own module

## Context

Added a separate module for the button and moved code out. 
Also in future will change the other buttons for menu and listbox to this module.

## References

#733


marko.json
```diff
@@ -155,6 +155,18 @@
...
         "@a11y-close-text": "string"
     },
+    "<ebay-expand-button>": {
+        "template": "./src/components/ebay-expand-button/index.marko",
+        "@*": "expression",
+        "@html-attributes": "expression",
+        "@id": "string",
+        "@class": "string",
+        "@style": "string",
+        "@type": "string",
```
msendlakowski: It looks like the type attribute isn't needed anymore.

PR #805, opened 2019-08-01
Listbox: close on select

### LINKED ISSUE #707: Listbox: collapse the expanded options after selecting an option
# Feature Request

## Description
The expanded options should collapse once an option is selected. It leaves the options expanded now, which the user has to click outside the listbox to have the options collapse.

seangates:
## Description
- adds close on click functionality
- fix the option be cast as a String for the emitted event data

## Context
When a user clicks on the listbox option we kept the listbox open. This isn't a great experience for mouse users, so we opted to close the dropdown when it was clicked with a mouse.

## References
Fixes #707

### DIFF
--- a/src/components/ebay-listbox-button/index.js
+++ b/src/components/ebay-listbox-button/index.js
@@ -58,6 +58,9 @@ module.exports = require('marko-widgets').defineComponent({
     handleCollapse() {
         this.emit('listbox-collapse');
     },
+    handleMouseDown() {
+        this.wasClicked = true;
+    },
     handleListboxChange(event) {
         const selectedIndex = parseInt(event.detail.toIndex, 10);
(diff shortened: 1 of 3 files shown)


src/components/ebay-listbox-button/index.js
```diff
@@ -58,6 +58,9 @@ module.exports = require('marko-widgets').defineComponent({
     handleCollapse() {
         this.emit('listbox-collapse');
     },
+    handleMouseDown() {
```
msendlakowski: This will only handle when its selected by a mouse click.  Don't we need the same behavior for keyboard?  Why not just put this in handleListboxChange?

seangates: All the behavior for keyboard is already handled. We only needed to handle the click from the mouse.

PR #816, opened 2019-08-02
Update icons to use arc style naming convention

DylanPiercey:
## Description
This PR builds on top of #815 to convert all icon symbol paths to use the [arc]() naming convention. This is a stepping stone toward consuming these components via Webpack.

Besides the renaming of the files I also updated the script to include the `missing icon` error directly into the output Marko templates instead of a standalone file because that will allow it to be picked up without using a browser function.


seangates:
@DylanPiercey Rebased (because of my stupidity in merging the other one first.) Target is now 3.0.0.

PR #820, opened 2019-08-06
Added option to change default icon from close to chevron-down

pcanella:
## Description
Just like the title says! We have added the option to change the `dialog-close` icon from the default "X" to the `chevron-down` icon. This will allow us to transition more easily into the modification of `ebayui-dialog` to support the `bottom drawer` pattern

## Screenshots
Screen Shot 2019-08-06 at 11 44 36 AM


DylanPiercey:
I kind of feel like this should not be an option and instead we should have a new type of dialog (pull up/docked or something like that). That way we could make sure it is only used when intended. Also we could ship the animations and what not that people are expecting with this variant of dialog.

pcanella:
@DylanPiercey Unfortunately, the New Buyer team has an immediate ask for a "bottom drawer" type component that looks like this and design system/ebayui is not totally ready for it (why we got this ask due to these constraints, I'm still not sure):


ezgif-2-62a6208cbf8f

seangates:
@DylanPiercey Yes. The new "drawer" will be post-6.5, though. We just don't have any bandwidth to provide the new "drawer" with 8.0.0.

This is a stop-gap since the icon is the only thing holding back @pcanella from delivering.

marko.json
```diff
@@ -153,7 +153,8 @@
         "@open": "boolean",
         "@type": "string",
         "@focus": "string",
-        "@a11y-close-text": "string"
+        "@a11y-close-text": "string",
+        "@icon": "string"
```
ianmcburnie: Pretty sure people are going to try any of our icon names here and be confused and file a bug when that doesn't work for them.

ianmcburnie: As @DylanPiercey says, I'm not really a fan of this stop gap either, we'll have to put in a boolean attribute like `chevron-down-icon`. It pollutes our API and could be misused. Maybe we can speak with design team and see if we can live with an X until BONC is built? If they pushback firmly, only then should we consider this option.

ianmcburnie: Actually, @pcanella maybe we can call the attribute `bottom-container` or `bonc`, and do some minimal work to set location to bottom and width to 100%?

msendlakowski: Last I knew, the core design team isn't even sure if this "partially closed" bonc is going to be a supported pattern.  
We said we could support this minimal icon change, but I was thinking something more specific to the chevron icon too.

Since we don't know if this is going to be an official pattern, what about @pcanella making their own custom version of the dialog, that they maintain for their repo.  We need to be really careful about adopting designs that are not official.   This is what killed off other UI libraries.

ianmcburnie: Ah yes, after viewing it again I see now the "partially closed" aspect that you mention. The button is very tied to the dialog in terms of transitions/animations. Which makes it very different from a bonc. Ignore my previous comment.

seangates: >Pretty sure people are going to try any of our icon names here and be confused and file a bug when that doesn't work for them.

This doesn't bother me too much.

Also, I see the reasoning behind waiting, but asking Patrick to do a whole lot of work up front, he'd still have to wait for our bottom tray, and then he'd still have a bit of work to implement the new bonc/bottom drawer/bottom dialog/whatever we call it.

I do, however, agree that we might change the attribute to more clearly relay what it's used for. Maybe `bonc-chevron=true` or similar?

ianmcburnie: I don't think this is a bonc though. It's something slightly different. At least that's the conclusion I came to after Mike's comment.

ianmcburnie: After all is said and done, perhaps leaving it as `icon` or `icon-type` but clearly documenting the allowed values in the README is best.

pcanella: Yeah, I agree with you guys that this isn't an ideal solution and that I should probably just fork `ebayui-dialog` and go from there. 

It's a bit offputting that I was told to use this pattern, as it doesn't exist in ebayUI as of yet. If we are requiring ebay devs to use ebayUI components and one isn't ready, is the proper solution really to just make it from scratch? Or should we, as end-users of ebayUI, push back on design and not build it until it's available, to spec and following a11y requirements?

I just don't want to take this in the wrong direction 😅

seangates: @pcanella Did you say you already had the bottom animation set up? (I thought I heard that somewhere.)

I agree with @ianmcburnie that we can leave it as `icon` or `icon-type` and make sure the docs are clear. Bigger fish are on the horizon.

ianmcburnie: @pcanella We have a roadmap of features and updates we are working through as part of DS6.5. All of our focus is on that roadmap and sticking to that roadmap (time & resources are tight). BONC *is* on that roadmap, but further towards the end (a few months away potentially).

pcanella: @ianmcburnie totally makes sense! Didn't mean to sound salty. I guess my question is, is it common for designers to give out mocks with future design systems patterns, or do they check in with the ebayUI team to make sure the pattern is supported first? 

I just don't want to get into a situation where I clone it myself, make it to our designer's spec and then have to re-do it once it comes to the ebayUI library because I'm not in compliance with something.

pcanella: @seangates Yeah, right now we have a dialog that's a full modal and uses the "slide" up animation to show it. I plan on doing the same thing, but adding the mask and not sliding 100% all the way up. It's not ideal, but since we already created that component (with ebayUI components), I figured I'd just modify that.

seangates: My experience is that domain designers like to be on bleeding edge of designs ... and ~never~ rarely check whether or not eBayUI components support their design.

But the other question is whether the Design System even has this design. (I don't think the design system has this bottom overlay.) And, until the DS supports it, eBayUI won't even consider creating it.

pcanella: @seangates  LOL at the strikthrough. But also, playbook DOES have these design (well, I think it's the same thing) here: 

https://playbook.corp.ebay.com/PlaybookApp#/executional-guidelines/product-design-system/patterns/bottom-container

seangates: You'll note that pattern was _mobile app specific_. They haven't yet created one that works across all platforms.

pcanella: Right right, our ask is only for mobile at the moment. I'll just create it from scratch I guess. I'll close this PR.

seangates:
Closing for the time being.

PR #830, opened 2019-08-14
DS 6.5 buttons

msendlakowski:
## Description
Update Button component to match the new DS6.5 styles.  Including support for the `delete` button.

## Context
Even though we only support "large" as a size override now, I left size as an attribute for backwards compatibility and just in case we need to add additional sizes back in later.

## References
https://github.com/eBay/skin/pull/756
https://github.com/eBay/skin/pull/749
https://github.com/eBay/skin/issues/700

## Screenshots
Screen Shot 2019-08-14 at 12 22 37 PM
Screen Shot 2019-08-14 at 12 22 46 PM

### DIFF
--- a/src/components/ebay-button/README.md
+++ b/src/components/ebay-button/README.md
@@ -10,15 +10,15 @@
 
 Name | Type | Stateful | Description
 --- | --- | --- | ---
-`priority` | String | No | "primary" / "secondary" (default) / "none"
-`size` | String | No | "small" or "large" (default: medium)
+`priority` | String | No | "primary" / "secondary" (default) / "delete" / "none"
+`size` | String | No | "large" (default: "none")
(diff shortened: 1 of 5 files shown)


src/components/ebay-button/README.md
```diff
@@ -10,15 +10,15 @@
 
 Name | Type | Stateful | Description
 --- | --- | --- | ---
-`priority` | String | No | "primary" / "secondary" (default) / "none"
-`size` | String | No | "small" or "large" (default: medium)
+`priority` | String | No | "primary" / "secondary" (default) / "delete" / "none"
+`size` | String | No | "large" (default: "none")
```
seangates: Do we still support specifying `small`? Or is that class completely removed from Skin?

PR #843, opened 2019-08-29
Filter menu;

### LINKED ISSUE #770: Filter Menu: new component
A menu specifically used for filtering. Potentially we need two variants:

* Form based (i.e. full page reload)
* ARIA based (i.e. AJAX/SPA)

Can be placed directly on page (no use case for this yet, but thinking ahead...)

Can be used inside of a filter menu button (our DS6.5 use case)

seangates:
## Description
- Adds the `ebay-filter-menu`
- Modifies the `ebay-filter-menu-button` to use the new `ebay-filter-menu`

## Context
The `ebay-filter-menu` can now be used as a standalone component, specifically for mobile-first apps. The API is similar to the `ebay-filter-menu-button`, for obvious reasons, but with some less attributes.

With regards to the `ebay-filter-menu-button` changes, many of the attributes now pass through, in order to power the `ebay-filter-menu`.

## References
Closes #770 

## Screenshots
image
image
image

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -190,6 +190,7 @@
     "@with-button": "boolean",
     "@form-name": "string",
     "@form-action": "string",
+    "@form-method": "string",
     "@a11y-footer-text": "string",
     "@items <item>[]": {
       "@*": "expression",
@@ -216,6 +217,7 @@
     "@pressed": "boolean",
     "@form-name": "string",
     "@form-action": "string",
+    "@form-method": "string",
     "@a11y-text": "string",
     "@a11y-footer-text": "string",
     "@items <item>[]": {
--- a/marko.json
+++ b/marko.json
@@ -186,7 +186,7 @@
     "@footer-text": "string",
     "@class": "string",
     "@style": "string",
-    "@with-button": "boolean",
+    "@class-prefix": "string",
     "@form-name": "string",
     "@form-action": "string",
     "@form-method": "string",
--- a/marko.json
+++ b/marko.json
@@ -178,6 +178,31 @@
     "@pressed": "boolean",
     "@aria-selected-text": "string"
   },
+  "<ebay-filter-menu>": {
+    "renderer": "./src/components/ebay-filter-menu/index.js",
+    "@*": "expression",
+    "@html-attrbibutes": "expression",
+    "@variant": "string",
+    "@text": "string",
+    "@footer-text": "string",
+    "@class": "string",
+    "@style": "string",
+    "@with-button": "boolean",
+    "@form-name": "string",
+    "@form-action": "string",
+    "@a11y-footer-text": "string",
+    "@items <item>[]": {
+      "@*": "expression",
+      "@html-attributes": "expression",
+      "@class": "string",
+      "@style": "string",
+      "@checked": "boolean",
+      "@value": "string"
+    }
+  },
+  "<ebay-filter-menu-item>": {
+    "transformer": "./src/common/transformers/attribute-tags/index.js"
+  },
   "<ebay-filter-menu-button>": {
     "renderer": "./src/components/ebay-filter-menu-button/index.js",
     "@*": "expression",
--- a/marko.json
+++ b/marko.json
@@ -183,7 +183,6 @@
     "@*": "expression",
     "@html-attrbibutes": "expression",
     "@variant": "string",
-    "@text": "string",
     "@footer-text": "string",
     "@class": "string",
     "@style": "string",
--- a/src/components/ebay-filter-menu-button/index.js
+++ b/src/components/ebay-filter-menu-button/index.js
@@ -54,11 +54,13 @@ module.exports = require('marko-widgets').defineComponent({
         });
     },
     handleMenuChange(e) {
-        this.setCheckedItem(this.getItemElementIndex(e.el), e.el);
+        const { el } = e;
+        this.setCheckedItem(this.getItemElementIndex(el), el);
     },
     handleItemKeydown(e) {
-        eventUtils.handleActionKeydown(e.originalEvent, () => {
-            this.setCheckedItem(this.getItemElementIndex(e.el), e.el);
+        const { el, originalEvent } = e;
+        eventUtils.handleActionKeydown(originalEvent, () => {
+            this.setCheckedItem(this.getItemElementIndex(el), el);
         });
     },
     handleFooterButtonClick() {
@@ -68,13 +70,16 @@ module.exports = require('marko-widgets').defineComponent({
         this.buttonEl.setAttribute('aria-pressed', (this.getCheckedItems().length > 0));
     },
     handleFormSubmit(e) {
-        this.emitComponentEvent('form-submit', null, e.originalEvent);
+        const { originalEvent } = e;
+        this.emitComponentEvent('form-submit', null, originalEvent);
     },
     handleExpand(e) {
-        this.emitComponentEvent('expand', null, e.originalEvent);
+        const { originalEvent } = e;
+        this.emitComponentEvent('expand', null, originalEvent);
     },
     handleCollapse(e) {
-        this.emitComponentEvent('collapse', null, e.originalEvent);
+        const { originalEvent } = e;
+        this.emitComponentEvent('collapse', null, originalEvent);
     },
     getItemElementIndex(itemEl) {
         const parentNode = itemEl.parentNode || itemEl.el.parentNode;
--- a/src/components/ebay-filter-menu-button/index.js
+++ b/src/components/ebay-filter-menu-button/index.js
@@ -1,7 +1,5 @@
 const assign = require('core-js-pure/features/object/assign');
 const Expander = require('makeup-expander');
-const scrollKeyPreventer = require('makeup-prevent-scroll-keys');
-const rovingTabindex = require('makeup-roving-tabindex');
 const eventUtils = require('../../common/event-utils');
 const template = require('./template.marko');
 
@@ -28,14 +26,6 @@ module.exports = require('marko-widgets').defineComponent({
                 autoCollapse: true,
                 alwaysDoFocusManagement: true
             });
-
-            if (this.state.variant !== 'form') {
-                // FIXME: should be outside of firstRender, but only when itemEls changes
-                this.rovingTabindex = rovingTabindex.createLinear(
-                    this.contentEl.querySelector('[role="menu"]'), 'div',
-                    { index: 0, autoReset: 0 }
-                );
-            }
         }
     },
     update_expanded(expanded) { // eslint-disable-line camelcase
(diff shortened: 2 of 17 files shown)


DylanPiercey:
I feel that there should be more attributes passed through to the form, eg `method`. We'll probably have some asks for those.

src/components/ebay-filter-menu/test/test.browser.js
```diff
@@ -0,0 +1,132 @@
...
+
+    describe('when an item is added via input from its parent and the new item is clicked', () => {
+        const newInput = mock.Basic_3Items;
+        const thirdItemText = newInput.items[2].renderBody.text;
+        beforeEach(async() => {
+            await component.rerender(newInput);
+            await fireEvent.click(component.getByText(thirdItemText));
+        });
+
+        it('then the new item is selected or something');
```
DylanPiercey: This probably shouldn't make it in right?

src/components/ebay-filter-menu/index.js
```diff
@@ -0,0 +1,80 @@
...
+    onRender(event) {
+        this.contentEl = this.el.querySelector('.filter-menu__menu, .filter-menu-button__menu');
+
+        if (this.state.withButton) {
+            this.contentEl = this.el;
+        }
+
+        if (event.firstRender) {
+            if (this.state.variant !== 'form') {
+                // FIXME: should be outside of firstRender, but only when itemEls changes
```
DylanPiercey: This is going to cause hard to track bugs, would be good to get this `FIXME` fixed!

src/components/ebay-filter-menu-button/index.js
```diff
@@ -59,20 +49,16 @@ module.exports = require('marko-widgets').defineComponent({
...
+        this.setCheckedItem(this.getItemElementIndex(e.el), e.el);
     },
-    handleItemKeydown(e, itemEl) {
-        eventUtils.handleActionKeydown(e, () => {
-            this.setCheckedItem(this.getItemElementIndex(itemEl), itemEl);
-        });
-
-        eventUtils.handleEscapeKeydown(e, () => {
-            this.buttonEl.setAttribute('aria-expanded', false);
+    handleItemKeydown(e) {
```
msendlakowski: Minor, but can we use destructuring in the function definition to define `el` and `originalEvent`?

src/components/ebay-filter-menu-button/index.js
```diff
@@ -81,19 +67,19 @@ module.exports = require('marko-widgets').defineComponent({
         this.buttonEl.setAttribute('aria-expanded', false);
         this.buttonEl.setAttribute('aria-pressed', (this.getCheckedItems().length > 0));
     },
-    handleFormSubmit(originalEvent) {
-        this.emitComponentEvent('form-submit', null, originalEvent);
+    handleFormSubmit(e) {
+        this.emitComponentEvent('form-submit', null, e.originalEvent);
+        e.originalEvent.preventDefault();
```
DylanPiercey: I don't think we should prevent default here. That should be left to the end user since they have access to the original event. As it stands there is no way to have the built in form behavior with this component.

msendlakowski: Destructuring here to get originalEvent would simplify and minify better too.

src/components/ebay-filter-menu/README.md
```diff
@@ -0,0 +1,55 @@
+# ebay-filter-menu-button
```
msendlakowski: Is the README TBD?  It looks like its a duplicate of `ebay-filter-menu-button`.

marko.json
```diff
@@ -178,6 +178,31 @@
     "@pressed": "boolean",
     "@aria-selected-text": "string"
   },
+  "<ebay-filter-menu>": {
+    "renderer": "./src/components/ebay-filter-menu/index.js",
+    "@*": "expression",
+    "@html-attrbibutes": "expression",
+    "@variant": "string",
+    "@text": "string",
```
msendlakowski: `text` isn't a valid option on ebay-filter-menu, right?  Its just for the button.  The examples also have text being passed.

src/components/ebay-filter-menu/template.marko
```diff
@@ -0,0 +1,55 @@
+<script marko-init>
+    var processHtmlAttributes = require("../../common/html-attributes");
+</script>
+
+<var baseClass = data.withButton ? 'filter-menu-button': 'filter-menu'/>
```
msendlakowski: I'm confused about why the `filter-menu` is aware of the `filter-menu-button`.  Shouldn't it be completely isolated, and just pass events back to the button?

src/components/ebay-filter-menu/index.js
seangates: Where should we be doing this for Marko 3 components?

src/components/ebay-filter-menu-button/index.js
seangates: This was left in by mistake! Fixing.

src/components/ebay-filter-menu/template.marko
seangates: Well ... it has to do with class names, and the way Skin was done. Let's discuss.

seangates:
@DylanPiercey Good call on `form-method`. I'll add that.

src/components/ebay-filter-menu/test/test.browser.js
```diff
@@ -0,0 +1,132 @@
...
+        beforeEach(async() => {
+            await fireEvent.click(firstItem);
+            await fireEvent.click(secondItem);
+        });
+
+        it('then it emits two menu-change events with correct data', () => {
+            const changeEvents = component.emitted('filter-menu-change');
+            expect(changeEvents).to.have.length(2);
+
+            const firstEventData = changeEvents[0][0];
```
agliga: Maybe simplify this to:
`const [[firstEventData], [secondEventData] = changeEvents;`
OR
`const [firstEventData, secondEventData] = changeEvents.flat();`

src/components/ebay-filter-menu/test/mock/index.js
```diff
@@ -0,0 +1,19 @@
+const assign = require('core-js-pure/features/object/assign');
+const { createRenderBody, getNItems } = require('../../../../common/test-utils/shared');
+
+exports.Basic_2Items = {
```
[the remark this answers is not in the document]
seangates: Straight copy from Dylan's tests refactor. Certainly could, but it would then be inconsistent, which is why I didn't do it in the first Place.

src/components/ebay-filter-menu/index.js
```diff
@@ -0,0 +1,80 @@
...
+        if (this.state.variant !== 'form') {
+            eventUtils.handleActionKeydown(e, () => {
+                this.setCheckedItem(this.getItemElementIndex(itemEl), itemEl);
+            });
+        }
+    },
+    handleFooterButtonClick(originalEvent) {
+        this.emitComponentEvent('footer-click', null, originalEvent);
+    },
+    handleFormSubmit(originalEvent) {
```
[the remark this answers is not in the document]
seangates: Yes, they bubble. `filter-menu` must be able to stand alone, and it's events all likely bubble through the `filter-menu-button`. Developers will likely not be listening to the `filter-menu` events because they will have only implemented the `filter-menu-button`, and be listening for its events.

Let me know if you still have any confusion. I'd be happy to talk through things.

src/components/ebay-filter-menu/test/mock/index.js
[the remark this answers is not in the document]
DylanPiercey: Some of these mock names got ridiculously long and became unreadable when using camel case. Separating things out with underscores seemed like a good compromise.

src/components/ebay-filter-menu/test/test.browser.js
```diff
@@ -0,0 +1,132 @@
+const { expect, use } = require('chai');
+const { render, fireEvent, cleanup } = require('@marko/testing-library');
+const { pressKey } = require('../../../common/test-utils/browser');
+const mock = require('./mock');
+const template = require('..');
+
+use(require('chai-dom'));
+afterEach(cleanup);
+
+/** @type import("@marko/testing-library").RenderResult */
```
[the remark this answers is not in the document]
seangates: @DylanPiercey Is this for TypeScript or something?

marko.json
seangates: Great catch. Yes, the text isn't needed at all.

src/components/ebay-filter-menu/README.md
seangates: Updated!

src/components/ebay-filter-menu/test/test.browser.js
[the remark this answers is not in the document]
DylanPiercey: @seangates it is a JSDoc which is read by the typescript compiler built into vscode so that we actually get autocompletions on the `component` var.

src/components/ebay-filter-menu/test/test.browser.js
seangates: Correct. Removing.

src/components/ebay-filter-menu/index.js
DylanPiercey: On every render (not just the first) we should be setting up the rovingTabIndex, then `onBeforeUpdate` we should destroy it.

src/components/ebay-filter-menu-button/index.js
seangates: Old habits ...

seangates: Updating.

src/components/ebay-filter-menu/index.js
seangates: Cool. Let me make those updates.

seangates:
@agliga I'm learning new tricks!

msendlakowski:
If we really want to get fancy...
```
handleItemKeydown({ el, originalEvent } = {}) {
```

msendlakowski:
I'm good with this, but before merging I'd like to wait until we can talk with Ian about whether or not we should completely separate the components in Skin as well.

src/components/ebay-filter-menu/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+```marko
+<ebay-filter-menu>
+    <ebay-filter-menu-item value="item 1">item 1</ebay-filter-menu-item>
+    <ebay-filter-menu-item value="item 2">item 2</ebay-filter-menu-item>
+    <ebay-filter-menu-item value="item 3">item 3</ebay-filter-menu-item>
+</ebay-filter-menu>
+```
+
+### ebay-filter-menu Attributes
```
msendlakowski: Do you want to update the documentation for `class-prefix` or just keep it as an internal attribute used by our components?

seangates: I can add it. Most people won't need to use it, though.

PR #848, opened 2019-09-04
Misc cleanup and fixes for filter menu in Marko 3

DylanPiercey:
## Description
There are a few changes in here, and a little bit of refactoring.

## Fixes
1. Fixes an issue where the events were not forwarded between `ebay-filter-menu` and `ebay-filter-menu-button`. This issue was because `w-body` on a nested tag is broken in Marko 3. Fixed by switching it to use `<span body-only-if(true) w-body=item.renderBody/>` 😞.
2. Item els were being thrown away in Marko 3 causing issues with the tests. Fixed by adding `w-id="item[]"` to each item in the `ebay-filter-menu`.

## Improvements
1. `expand` and `pressed` states did not look to actually be used for the `ebay-filter-menu-button`. This allowed me to delete some code and just allow the expander to handle the `aria-expanded` attribute alone as we do in some other components.
2. Refactored some method names to be more clear (setX => toggleX in some cases).
3. Calculated item indexes in the toggle methods instead of doing that calculation every where the method was called.
4. in some places we were using `querySelector` which was not necessary and so I switched them to use the recommended `getEl`.
5. Updated `marko-cli` because I fixed an issue where the stack trace was not printing properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "lasso-marko": "^2.4.7",
     "lintspaces-cli": "^0.7.1",
     "marko": "^4",
-    "marko-cli": "^7.0.14",
+    "marko-cli": "^7.0.16",
     "marko-prettyprint": "^1.5.2",
     "marko-widgets": "^7",
     "mocha": "^6.1.4",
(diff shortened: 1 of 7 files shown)


DylanPiercey:
Closing since it was decided to do this across multiple PR's.

PR #850, opened 2019-09-05
Filter menu improvements

seangates:
✅ @DylanPiercey approved

## Description
- used destructuring better
- removed a couple of dead code items
- used the `getEl` selectors
- improved overall readability

## Context
These changes are patterns we should follow for future components and will help maintenance and readability.

**Note:** we should consider improving the event forwarding between the filter button and the filter menu. Changing events in any significant way would be a breaking change and should probably wait.

## References
Related to the #849 PR


seangates:
@msendlakowski Well ... it's still there in the template. We just don't need it in the component JavaScript anymore.

Issue #858, opened 2019-09-07
Pagination- broken in iphone


hramakumar:
# Bug Report

## Description
pagination is broken in iphone. 
part of the text are truncated
![paginationbroken](https://user-images.githubusercontent.com/13157115/64467659-9cdc4600-d0cf-11e9-960f-f9a0a5cf6647.png)

hramakumar:
verified

PR #861, opened 2019-09-09
fixing pagination on small screens

msendlakowski:
## Description
Reduced the "minimum page" limit.
Removed "overflow:hidden" after calculating the items that are visible.

## Context
The primary problem was that small screens couldn't show 5 items, which was set as a minimum.  I removed that restriction because I didn't see it anywhere in the design system patterns.

With that change it looked better, but the border on some items were cut off due to `overflow: hidden`.  I think might be due to the browser not accounting for padding and margin when applying the flex.

Screen Shot 2019-09-09 at 3 52 32 PM

The overflow control is needed in the Skin version, but in ebayUI we use JS to calculate the hidden items and hide them individually.  So I remove `overflow:hidden` after calculating the hidden elements.  I add it back in before recalculating so the page UI doesn't jump.

## References
Fixes: https://github.com/eBay/ebayui-core/issues/858

## Screenshots
Screen Shot 2019-09-09 at 3 33 36 PM

### DIFF
--- a/src/components/ebay-pagination/index.js
+++ b/src/components/ebay-pagination/index.js
@@ -4,7 +4,7 @@ const template = require('./template.marko');
 
 const constants = {
     indexForNavigation: 2,
-    minPagesRequired: 5,
+    minPagesRequired: 3,
     maxPagesAllowed: 9,
     margin: 8
 };
@@ -88,6 +88,7 @@ function onUpdate() {
 
 function refresh() {
     let current = 0;
+    this.pageContainerEl.style.overflow = 'hidden';
     for (let i = 0; i < this.state.items.length; i++) {
         if (this.state.items[i].current) {
             current = i;
@@ -134,6 +135,7 @@ function refresh() {
             this.pageEls[i].removeAttribute('hidden');
         }
     }
+    this.pageContainerEl.style.overflow = 'visible';
 }
 
 /**


src/components/ebay-pagination/index.js
```diff
@@ -88,6 +88,7 @@ function onUpdate() {
 
 function refresh() {
     let current = 0;
+    this.pageContainerEl.style.overflow = 'hidden';
```
agliga: I think we should rethink this. 
I tried on local and I see what the change here does, but I don't understand why we need the hidden overflow at all. I read the comment but I can't seem to reproduce or see the issue stated there. Can you elaborate?

On the skin side the `.pagination__items` class has `overflow: hidden` but we are setting this here. Should we change that on the skin side?

It seems like the bigger issue is the height of the container is `48px` while the height of the button inside the container is also `48px`. I think we should use flex to grow the button to match the container height but leave enough room for the container.
I'm fine with the overflow fix, but I would say put the `overflow:visible` as the default on skin.

seangates:
What is the width which was used to test this?

src/components/ebay-pagination/index.js
msendlakowski: We need it in skin because otherwise we'll get this:
Screen Shot 2019-09-10 at 8 42 32 AM

I'm not sure how the height of the container would effect this. It's more about the width.

But I'm not thrilled with this solution either, so if anyone can come up with something better, I'm all for it.

msendlakowski:
> What is the width which was used to test this?

IPhone 8 size. 375px.

PR #863, opened 2019-09-11
Menu button split from menu.

### LINKED ISSUE #713: Menu: make independent of button

seangates:
## Description
- creates a new `<ebay-menu>` component
- updates all the JS and Marko for the `<ebay-menu-button>`, avoiding anti-patterns and aligning it with how the `<ebay-filter-menu>` and `<ebay-filter-menu-button>` behave

These changes do take some liberties with how the event api works to align things across our components. If we feel like it's enough of a change to warrant a "breaking change" I would advocate we publish this as 5.0.

The attribute APIs did not change, except for the addition of `value` to the `<menu-button-item>` for the `radio` and `checkbox` use cases.

## References
Closes #713


src/components/ebay-menu-button/template.marko
```diff
@@ -1,25 +1,38 @@
...
+            "${baseClass}__button",
+            (data.borderless && 'expand-btn--borderless')
+        ]
         variant="expand"
         size=data.size
         priority=data.priority
-        no-text=data.noText
-        aria-expanded=String(data.expanded)
+        no-text=(!data.text && !data.icon)
+        aria-expanded="false"
```
msendlakowski: Is this always expanded when loaded now?  I see there is an `expanded` boolean in the attributes that are passed in.  Is that still an option?

src/components/ebay-menu-button/template.marko
```diff
@@ -1,25 +1,38 @@
...
         aria-haspopup="true"
         aria-label=data.a11yText
         disabled=data.disabled
         w-on-button-escape="handleButtonEscape">
-        <span class=['expand-btn__cell', (data.customLabel && 'menu-button__control--custom-label')]>
-            <span if(data.customLabel) w-preserve-body>
-                <span w-body=data.customLabel.renderBody body-only-if(true)/>
+        <span class=['expand-btn__cell', (data.label && 'menu-button__control--custom-label')]>
+            <span if(data.label) w-preserve-body>
+                <span w-body=data.label.renderBody body-only-if(true)/>
```
msendlakowski: is `body-only-if(true)` a trick to render body content without the container?

src/components/ebay-menu/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+### ebay-menu Attributes
+
+Name | Type | Stateful | Description
+--- | --- | --- | ---
+`type` | String | No | Can be "fake" / "radio" / "checkbox"
+`reverse` | Boolean | No | expand menu flyout to the left
+`fix-width` | Boolean | No | constrain items container width to button width
+`size` | String | No | button size, "small" or "large" (default: medium)
```
msendlakowski: There is no more medium button.  Default is small and they can override with "large"

src/components/ebay-menu/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## ebay-menu-item Tag
+
+### ebay-menu-item Usage
+
+```marko
+<ebay-menu-item>item 1</ebay-menu-item>
+```
+
+### ebay-menu-item Attributes
```
msendlakowski: The new "value" attribute should be documented.

src/components/ebay-menu-button/template.marko
DylanPiercey: @msendlakowski it was always collapsed when loaded previously, but the aria attribute was off.

I brought up to @seangates that we should fix this and expose an `expanded` attribute (removed in this PR because it wasn't doing anything) that is actually linked to the expander instance.

src/components/ebay-menu-button/template.marko
DylanPiercey: @msendlakowski that is correct. Oh Marko 3...

src/components/ebay-menu/index.js
```diff
@@ -0,0 +1,101 @@
...
+        }
+    },
+    toggleItemChecked(itemEl) {
+        const itemIndex = indexOf(itemEl.parentNode.children, itemEl);
+        const item = this.state.items[itemIndex];
+        const currentIndex = this.state.items.findIndex(radioItem => radioItem.checked);
+
+        if (this.state.type === 'radio' && itemIndex !== currentIndex) {
+            this.state.items.forEach((eachItem, i) => {
+                eachItem.checked = i === itemIndex;
```
msendlakowski: (Minor) I know this is valid, but could you wrap the `i === itemIndex` in parens, to prevent people from getting confused if they're not as smart as you.  :)

src/components/ebay-menu/template.marko
```diff
@@ -0,0 +1,53 @@
...
+    <div
+        w-id="menu"
+        role=(!isFake && 'menu')
+        class=[
+            "${baseClass}__items",
+            (data.reverse && "${baseClass}__items--reverse"),
+            (data.fixWidth && "${baseClass}__items--fix-width")
+        ]>
+        <for(item in data.items)>
+            <var itemRole = (isRadio ? 'menuitemradio' : (isCheckbox ? 'menuitemcheckbox' : !isFake && 'menuitem'))/>
```
msendlakowski: I think this could end up with an empty itemRole, if isRadio is false and isCheckbox is false, and isFake is true.  Will that cause any problems?

src/components/ebay-menu/README.md
seangates: Okay, let me just update the docs. The string is a passthrough for the class sizing.

src/components/ebay-menu/template.marko
seangates: That's expected. That specific case it should not have a role. (E.g. a fake menu with links should not have `role` on any of the links or buttons.)

src/components/ebay-menu/README.md
seangates: Thanks for catching that. I will add it.

seangates:
@msendlakowski I'm making the updates for the event indexes here in a moment. Doc updates and the syntax update are done.

seangates:
@msendlakowski @agliga I've restored the indexes as part of the emitted events. We will no longer have breaking changes for this PR.

ianmcburnie:
Is the button in fake menu supposed to look like this:

Screen Shot 2019-09-17 at 3 59 29 PM

ianmcburnie:
Does reverse make sense for a non-button menu?

Screen Shot 2019-09-17 at 4 02 36 PM

ianmcburnie:
The reverse example doesn't seem to be working properly:

Screen Shot 2019-09-17 at 4 36 46 PM

seangates:
@ianmcburnie Let me look at that first screen shot as I'm fairly sure that was fixed.

The fixed width and reverse options are pass-through for the menu-button. Let me remove them from examples.

I'll look at that last one as to why it's not working, as I'm fairly sure that was working, too.

Maybe a bad merge (for the non-working screen-shots)?

seangates:
So, the #1 and #3 screen shots are working for me. Maybe you linked to the wrong branch? Let's try and figure it out tomorrow.

seangates:
@ianmcburnie Two updates here, and one over in [Skin](https://github.com/eBay/skin/pull/839), should address those issues you saw.

Let me know if you have any trouble seeing the update from Skin on the core side.

ianmcburnie:
Ok, I'm going to try again now.

ianmcburnie:
@seangates I'm still having intermittent issues with the JavaScript not working. When I remove this line it seems to work:

https://github.com/eBay/ebayui-core/blob/770-713-menus/src/components/ebay-menu-button/examples/09-icon-with-text/template.marko#L5

Reverse menu now looks to be working (I think I forgot to build skin after linking last time) as does the fake menu background color.

ianmcburnie:
@seangates Could you also remove `reverse` and `fix-width` from the `ebay-menu` README? And anything else that might not apply to a standalone menu.

seangates:
So, I could remove the `reverse` and `fixed-width` from the menu docs, but they would become "hidden features" of that component.

I'll restore the flags example, and ensure that key behavior is working.

ianmcburnie:
Reverse and fixed-width don't make any sense to me for a standalone menu. Unless I'm missing something?

seangates:
Not missing anything, except that they are available options which we won't document.

seangates:
@ianmcburnie Updated the key nav of options.

PR #864, opened 2019-09-11
Normalize to selected instead of pressed;

### LINKED ISSUE #859: Filters: Normalize naming for API around pressed/selected/active
# Feature Request

## Description
Presently we have a `pressed` and `a11y-active-text` attributes, and also talk about "selected" filters. Make sure that the API and documentation refer to "selected".

## Context
The present setup can be confusing. This work would be able to remove confusion in the API and nomenclature.

seangates:
## Description
- changes `pressed` to `selected`
- changes `a11y-active-text` to `a11y-selected-text`
- fixed an issue with `marko.json` where `aria-selected-text` was the attribute instead of `a11y-selected-text`

## Context
Because both `pressed` and `selected` were used in describing the active/selected/pressed state of the `<ebay-filter>`. Therefore, to help with developer experience, we updated the attribute and documentation to make things easier.

## References
Closes #859

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -175,8 +175,8 @@
     "@disabled": "boolean",
     "@href": "string",
     "@priority": "string",
-    "@pressed": "boolean",
-    "@aria-selected-text": "string"
+    "@selected": "boolean",
+    "@a11y-selected-text": "string"
   },
   "<ebay-filter-menu>": {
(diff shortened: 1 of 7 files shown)


msendlakowski:
This should force a major version update.  Let's discuss in standup, but I think its probably the right move.

PR #874, opened 2019-09-18
Integrate storybook for examples

DylanPiercey:
## Description
Sets up storybook to consume our existing examples and build a ds4 & ds6 version.

You can check out the output from this by cloning and then serving the `./docs` folder.
Once this is merged we can setup GitHub Pages to consume our docs folder.

It is setup to automatically build storybook whenever we run `npm version` to ensure that we update the docs anytime we do a release.

Screen Shot 2019-09-20 at 11 45 30 AM


seangates:
image

Looks like the `filter-menu` needs a fix in the `destroy()`. I'll check it out.

PR #876, opened 2019-09-18
Fix for filter-menu destroy;

seangates:
## Description
- fixes `onDestroy` (and `onBeforeUpdate`) lifecycle methods in the filter menu

## Context
Found during PR review of #874


src/components/ebay-filter-menu/index.js
```diff
@@ -31,12 +31,16 @@ module.exports = require('marko-widgets').defineComponent({
...
-        this.rovingTabindex.destroy();
-        scrollKeyPreventer.remove(this.contentEl);
+        this._handleDestroy();
     },
     onDestroy() {
-        this.rovingTabindex.destroy();
-        scrollKeyPreventer.remove(this.contentEl);
+        this._handleDestroy();
+    },
+    _handleDestroy() {
```
dkallus: I'm assuming this is a private function. Should we move this function outside of exports then? I'm not sure what's considered best practice for private functions.

seangates: The pattern is elsewhere. I'm not too concerned about it.

@ianmcburnie @DylanPiercey @agliga Opinions?

agliga: I think it's fine to have private functions like this. It's easier for organization 
Only thing I would do is either put them together at the top/bottom (depending on who you ask) of a class/export so they are grouped together. But that is mostly a linting thing.

seangates:
@msendlakowski Yeah, there's only one "variant" from the default.

Issue #877, opened 2019-09-19
DS6.5 Success Message (new component)


msendlakowski:
Requirements: https://github.com/eBay/skin/issues/816

ianmcburnie:
Success Messaging will be available in 4.5.0. Closing.

PR #879, opened 2019-09-20
Success-message: new component

msendlakowski:
## Description
Added the `success-message` component and tests.

## Context
Primary concern is the nested button.  I'm not sure if there is a better way to nest this and still have all the button events available.

## References
Fixes: https://github.com/eBay/ebayui-core/issues/877

## Screenshots
Screen Shot 2019-09-20 at 1 26 25 PM
Screen Shot 2019-09-20 at 1 26 40 PM
Screen Shot 2019-09-20 at 1 26 48 PM


src/components/ebay-success-message/index.marko
```diff
@@ -0,0 +1,25 @@
...
+        <span class="success-message__status" role="img" aria-label=data.iconA11yText></span>
+        <span>${data.title}</span>
+    </>
+
+    <p
+        class=["success-message__content", data.content.class]
+        style=data.content.style
+        if(data.content && data.content.renderBody)>
+        <invoke data.content.renderBody(out) />
+    </>
```
seangates: I know this is an easy convention, but we usually close non-dynamic tags so there's no confusion.
```suggestion
    </p>
```

ianmcburnie:
@msendlakowski take a look at the way @agliga is adding the CTA button for page notice module. The downside of what you currently have is that the developer has to know which class to add to the button.

ianmcburnie:
This is on hold due to design.

ianmcburnie:
@msendlakowski  I'll try and make sure to keep this branch updated, so that it doesn't get stale.

ianmcburnie:
Going to close this one and reopen when we have further updates from design. I'll keep the branch updated as we go.

PR #881, opened 2019-09-20
Updated notice to support guidance type

agliga:
## Description
* Changed notice to support guidance type
* Added support for blank status only for guidance
* Added cta button for notice

## Context
* Added cta button which switches between anchor or button depending if href is present
* If type is guidance, will only only allow blank status type. This will remove the icon too. This is to prevent any breaking changes.

TODO: Still need to add tests for new changes

## References
eBay/skin#817

## Screenshots
image

image


ianmcburnie:
README also needs updating.

agliga:
> Do we have an example with overflow menu button? Or should we put that in a new PR?

Not yet, waiting for overflow to be implemented

ianmcburnie:
Ah, good point. Do you think it will slot into here easily enough when ready? I imagine it will just be the ebay-menu-button with a variant on it's icon type, so you could test it out like that. Or we just merge this PR as is, and do that in a follow up.

src/components/ebay-notice/README.md
```diff
@@ -14,6 +16,21 @@ For the dismissible use case, if the parent of `<ebay-notice>` wants to close it
...
+```marko
+<ebay-notice a11y-heading-text="Attention" a11y-heading-tag="h3" status="attention" type="page">
+    <ebay-notice-content>
+        <p>Couldn't load all the items, please try again later.</p>
+    </ebay-notice-content>
+    <ebay-button priority="secondary" transparent>Try again</ebay-button>
+</ebay-notice>
+```
+
+## ebay-infotip Sub-tags
```
ianmcburnie: Typo. Should be `ebay-notice`

marko.json
```diff
@@ -385,7 +386,16 @@
     "@a11y-heading-text": "string",
```
ianmcburnie: Seems like heading-text is a no-op in the case of the default section type (aka light guidance), because it relies on an icon. I think this will also break the aria-labelledby connection. In this case we should use aria-label instead of aria-labelledby. It's getting a bit messy, but will ultimately be cleaned up when I refactor the HTML side, and we can always use aria-labelledby.

ianmcburnie:
@agliga README needs updating with new section variant.

marko.json
agliga: Done, I changed it so that if it's light guidance then it will set `aria-label` with the value of  `a11y-heading-text`. `aria-labelledby` will be null since status `widget.elId("status")` will be null/undefined

ianmcburnie:
4.3.0 branch is ready.

src/components/ebay-notice/examples/02-page-notice-cta/template.marko
```diff
@@ -0,0 +1,12 @@
...
+    <ebay-notice-content>
+        <p>
+            <strong>Error:</strong> Please take another look at the following:
+        </p>
+        <p>
+            <a href="#">Card number</a>,
+            <a href="#">Expiration date</a>
+        </p>
+    </ebay-notice-content>
+    <ebay-button priority="secondary" transparent>Button</ebay-button>
```
dkallus: For my knowledge, would placing `on-button-click` directly on `ebay-button` also work? If so, what's benefit of putting `on-button-click` on the parent component?

I know we had a discussion around this, but I didn't completely understand the problem at the time.

ianmcburnie: @dkallus Good catch. Yes, I thought the consensus was to place on-button-click directly on the button?

msendlakowski: Just making sure I understand.
By doing it this way we render all body content from inside `ebay-notice`.  Users can technically add anything they want and we'll render it.  But we're recommending a button. 

The other option was to use `ebay-notice-button` and make it a pseudo attribute of `ebay-notice`.  But in that design we couldn't put the event directly on the button.

This makes sense, though it opens the component up for some abuse.

agliga: Good catch. It was my mistake. I had it initially on the parent but didnt remove it. 
I will remove it.

agliga:
> Should we also add browser test for verifying on-button-click works correctly?

No need to since the button click is handled by `ebay-button`. Also if this was due to the `on-button-click` action on the parent, it has been removed

PR #888, opened 2019-09-26
Tabs: Fix manual activation;

### LINKED ISSUE #887: `ebay-tab`: manual activation doesn't seem to work
# Bug Report

## Description

1. go to [marketplace](http://app.uimarketplace.svc.32.tess.io/?tag=ebay-tab)
2. scroll to the bottom to see the `activation="manual"` example
3. click any tab, press left/right arrow to move focus to another tab
4. press `space` or `enter`
5. =(

## Workaround

Possible cause:
https://github.com/eBay/ebayui-core/blob/c019071ef7925e66186ee4b0eff68b3ccaa3c87f/src/components/ebay-tab/index.js#L40

in the function that handles `space`/`enter` presses:
```
eventUtils.handleActionKeydown(event, () => {
            event.preventDefault();

            if (this.state.activation === 'auto') {
                this.setIndex(el.dataset.index);
            }
});
```
condition is mistakenly inverted and should be
```this.state.activation === 'manual'```

### PR DESCRIPTION
**Pull request by seangates**
## Description
- fix logical test when `enter` or `spacebar` is pressed
- update tests to catch manual activation

## References
Fixes #887

### DIFF
--- a/src/components/ebay-tab/test/test.browser.js
+++ b/src/components/ebay-tab/test/test.browser.js
@@ -10,6 +10,20 @@ afterEach(cleanup);
 /** @type import("@marko/testing-library").RenderResult */
 let component;
 
+function thenItHasMovedToTab(index) {
+    it('then it emits the select event with correct data', () => {
+        const selectEvents = component.emitted('tab-select');
+        expect(selectEvents).has.length(1);
+
+        const [[eventArg]] = selectEvents;
+        expect(eventArg).has.property('index', index);
+    });
+
+    it(`then heading ${index + 1} is selected`, () => {
+        expect(component.getAllByRole('tab')[index]).has.attr('aria-selected', 'true');
+    });
+}
+
 describe('given tabs with first heading selected', () => {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/components/ebay-tab/test/test.browser.js [hunk 1]
```diff
@@ -85,6 +72,52 @@ describe('given tabs with first heading selected', () => {
...
+            await pressKey(component.getAllByRole('tab')[1], {
+                key: '(Space character)',
+                keyCode: 32
+            });
+        });
+
+        thenItHasMovedToTab(1);
+    });
+
+    function thenItHasMovedToTab(index) {
```
dkallus: There are 3 `thenItHasMovedToTab` functions, 1 in each major describe block. If they do the same thing, might be good to create a single function that can be used throughout the tests.

seangates: Sure let me work on that. I didn’t do an optimize sweep, but that’s a fair point.

seangates: @dkallus Updated.

Issue #890, opened 2019-09-27
ebay-menu-button: support overflow icon


ianmcburnie:
Section title and guidance module require an menu button with overflow icon. Let's see what it would take to add that to coreui.

I think we need to allow ebay-menu-button to swap out expand-button with an icon-btn.

PR #891, opened 2019-09-30
Menu Button: supporting overflow button

dkallus:
## Description
User can now show menu button as an overflow icon by setting a single attribute.

## Context
Section title and guidance module require a menu button with overflow icon. Showing overflow icon is already possible with existing implementation, however, user has to set `icon="overflow" priority="none" no-toggle-icon borderless` attributes and some additional styling. With this change, we can set a single attribute `variant="overflow"` to show the desired overflow icon.

I added 1 inline styling `style="border:0;"` to remove the border around the icon. Without this, we'll need a change to Skin.

Also, I added `expand-btn` class to the button since it's required for the menu to show/hide correctly.

I will change PR to 4.3.0 once branch is created.

## References
ebay-menu-button: support overflow icon #890

Depends on https://github.com/eBay/skin/pull/866 Menu Button: added generic class for toggling menu

## Screenshots
overflow3


src/components/ebay-menu-button/template.marko
```diff
@@ -15,35 +15,54 @@
...
-            <span if(data.label) w-preserve-body>
-                <span w-body=data.label.renderBody body-only-if(true)/>
+    <if(data.variant === 'overflow')>
+        <ebay-button
+            w-id="button"
+            class=[
+                "${baseClass}__button",
+                "expand-btn"
+            ]
+            style="border:0;"
```
seangates: I was initially concerned about this, but I see why you did this.

I guess the alternative was to add a special class `expand-btn--overflow` and create a new `style.less` file with that class and the `border: 0;` rule.

I'm not opposed to this given the desire to re-use as much as possible. This uses a lot less code and limits payload (albeit, it wouldn't have been a huge amount anyway).

seangates:
@dkallus So, developers have to remember to use `reverse` (if it's used on the right side)? Not a huge deal.

ianmcburnie:
`Also, I added expand-btn class to the button since it's required for the menu to show/hide correctly.`

I think we might want to change this situation. If not in this PR, then later in a refactor. What and where is the dependency?

dkallus:
@seangates Correc

> @dkallus So, developers have to remember to use `reverse` (if it's used on the right side)? Not a huge deal.

@seangates Correct. I'm fine with adding in reverse by default, but then there's no way for user to us not-reversed if they ever come across that scenario.

seangates:
@dkallus We could, in theory, just invert `reverse`, so that it works _opposite_ of their setting. 🤔

dkallus:
> `Also, I added expand-btn class to the button since it's required for the menu to show/hide correctly.`
> 
> I think we might want to change this situation. If not in this PR, then later in a refactor. What and where is the dependency?

@ianmcburnie 

> `Also, I added expand-btn class to the button since it's required for the menu to show/hide correctly.`
> 
> I think we might want to change this situation. If not in this PR, then later in a refactor. What and where is the dependency? I feel like using `icon-btn` is a better long term fix. Expand-btn is a very specific button, which made a lot of sense when menu was coupled to it's button, but not so much now.



> `Also, I added expand-btn class to the button since it's required for the menu to show/hide correctly.`
> 
> I think we might want to change this situation. If not in this PR, then later in a refactor. What and where is the dependency? I feel like using `icon-btn` is a better long term fix. Expand-btn is a very specific button, which made a lot of sense when menu was coupled to it's button, but not so much now.

@ianmcburnie It comes from `ebay-menu` styling. https://github.com/eBay/skin/blob/dcfcfc34f68d114aa1a045aa3143994e458ba3b6/src/less/menu-button/base/menu-button.less#L118-L121. We don't check if `.menu-button__button` is expanded, we check if `.expand-btn` is expanded.

ianmcburnie:
4.3.0 branch is now ready.

dkallus:
@agliga Can you take another look? I'm now using a single button and then switching between overflow vs default inside buttons' body. With this change, I also removed the browser overflow tests, since now we're using same button for existing flow vs overflow flow.

agliga:
> @agliga Can you take another look? I'm now using a single button and then switching between overflow vs default inside buttons' body. With this change, I also removed the browser overflow tests, since now we're using same button for existing flow vs overflow flow.

Much better!

ianmcburnie:
Let's make it reverse by default for overflow case.

dkallus:
@seangates I updated to use opposite value of "reverse" for overflow variant. Passing `reverse` attribute will correctly switch directions.

As for overflow example, I left overflow on the left side. I didn't find a simple way to align to right side without introducing a <div> wrapper.

PR #892, opened 2019-10-01
Tabs: manual fix

seangates:
## Description
- add a `handleArrowsKeydown` to event utils file
- fix the `direction` variable
- remove the `handleActionKeydown` guard for `manual` as it actually doesn't matter in the `auto` case

## Context
Issues found during testing the recent fix for the manual tabs.

## References
Found during #887 (#888) testing.

### DIFF
--- a/src/common/event-utils/index.js
+++ b/src/common/event-utils/index.js
@@ -41,6 +41,10 @@ function handleLeftRightArrowsKeydown(e, callback) {
     handleKeydown([37, 39], e, callback);
 }
 
+function handleArrowsKeydown(e, callback) {
+    handleKeydown([37, 38, 39, 40], e, callback);
+}
+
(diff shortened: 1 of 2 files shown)


ianmcburnie:
Is rovingtabindex module not able to handle this?

ianmcburnie:
Also, the ticket #887  doesn't mention anything about up and down arrow keys. It says left and right keys. To keep things simple for this PR, maybe just fix the left & right keys issue? Unless the ticket is inaccurate, of course.

seangates:
The ticket is not accurate.

seangates:
Description has been update.

The issue was found during testing of the fix, and therefore was "related to" that issue. Also, the issue found only was reproducible if the up or down arrow keys were pressed, and then the right and left arrows were pressed.

PR #899, opened 2019-10-03
Add a guard for combobox blur/change event;

### LINKED ISSUE #740: editable combobox- change event is triggered on every select and blur
# Bug Report

## Description
editable combobox xhange event is triggered even when the same item is selected again and again and also on every collapse.
https://www.loom.com/share/0f2a1494ff9244ef8f3863604c7afc18

seangates:
## Description
- adds an `valueChanged` property to check when the value of the input changed
- guards the `change` event so that when it blurs we know that the option had changed
- minor method name change for better clarity

## Context
Presently, the `blur` of the combobox input field fires a `combobox-change` every time, without any concern for whether the input changed from the original value. This change will only fire the `combobox-change` event when the value is different from what was there before, whether typing or whether selecting from the menu.

Also, the generic method for emitting events for the combobox was updated for better clarity.

## References
Fixes #740

### DIFF
--- a/src/components/ebay-combobox/index.js
+++ b/src/components/ebay-combobox/index.js
@@ -149,6 +149,7 @@ module.exports = require('marko-widgets').defineComponent({
 
         if (this.valueChanged) {
             this.emitComboboxEvent('change');
+            this.valueChanged = false;
         }
     },
     handleOptionMouseDown() {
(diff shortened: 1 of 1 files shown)


ianmcburnie:
See https://github.com/eBay/ebayui-core/issues/740#issuecomment-538027827

Can we not just fire our custom change event when the DOM change event fires?

seangates:
@ianmcburnie Will try the native event.

seangates:
So, I know now why we don't use the native event: it doesn't fire if the value of the `<input>` is changed via the API, only when the user changes it via keyboard entry.

Hence, we need to keep the current behavior and fire when the change event happens via both the native API and user input.

ianmcburnie:
Okay, thanks Sean.

seangates:
Also, I added one fix to reset the `valueChanged` property because when the checked item was clicked it was using a stale check after a blur event.

PR #908, opened 2019-10-08
Menu Button: Add scroll management attribute;

### LINKED ISSUE #906: Menu Button: add configuration to allow focus management to go to first item instead of checked item
We'll need a new enumerated attribute. We should keep the current behaviour as the default. This new behaviour is opt-in (we might make it the default in future, when we get the chance).

We are going to try and do this as a 2.9.0 minor version if a simple code change. If not, it will go in latest 4.x.

seangates:
## Description
- adds a `scroll-management` attribute to allow for limiting scroll behavior

## Context
When working with certain use cases the scroll behavior for radio or checkbox should be disabled. Specifically, when the label is changed by the developer they may also decide to limit the scroll behavior to surface the most relevant choices at the very top, even if the user has chosen something else.

**Note:** I made a mistaken push to 9.2.0 😞 and we can't force push changes there. Therefore, we'll have to live with two commits, unless you folks are really dead set on a clean history. (E.g. `focus-first` was the original attempt at a fix, but I went a different direction.)

## References
Fixes #906

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -253,6 +253,7 @@
         "@priority": "string",
         "@iconTag <icon>": {},
         "@disabled": "boolean",
+        "@scroll-management": "string",
         "@items <item>[]": {
             "@*": "expression",
             "@html-attributes": "expression",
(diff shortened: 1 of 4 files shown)


ianmcburnie:
Let's release 2.9.0 with this change, and also port it to latest 4.x

ianmcburnie:
Ok, thanks @seangates !

Issue #513, opened 2019-01-22
New Component: Section Title


ianmcburnie:
Title module has 3 subtypes: arrow, info and overflow:

![screen shot 2019-01-22 at 12 22 01 pm](https://user-images.githubusercontent.com/38065/51563207-9222b880-1e40-11e9-84dd-cbdf7fd6fe67.png)

Related Skin issue: https://github.com/eBay/skin/issues/524

PR #914, opened 2019-10-10
Section Title: new component

dkallus:
## Description
* new ebay-section-title component that supports title, subtitle, CTA, infotip, and overflow menu button

## Context
* depends on https://github.com/eBay/skin/pull/876

## References
New Component: Section Title #513

Section Title: new component https://github.com/eBay/skin/pull/876

## Screenshots
Sharing screenshots for main use cases. There are a few more that show combinations of size + cta + info + overflow together.

### Large screens
core_1
core_2
core_3
core_4

### Small screens
core_small_1
core_small_2
core_small_3
core_small_4


src/components/ebay-section-title/examples/01-title/template.marko
```diff
@@ -0,0 +1,3 @@
+<ebay-section-title>
```
agliga: I would actually think this is a bit too verbose for the default example. 
We need to check if section-title-title render body is there and then use that, otherwise use the default. I did something similar on `ebay-notice` with CTA button. 
I would prefer to do:
```
<ebay-section-title>
   Section Title
</ebay-section-title>
```

I think the nested one should only be used when there's other variations there. 
Or we can always have the title be on it's own and the other variations inside of tags (although I'm fine keeping it the way you have it for the other cases).

src/components/ebay-section-title/examples/03-cta/template.marko
```diff
@@ -0,0 +1,5 @@
+<ebay-section-title>
+    <ebay-section-title-title>Today’s Deals – All With Free Shipping</ebay-section-title-title>
+    <ebay-section-title-subtitle>Plus, guaranteed best prices.</ebay-section-title-subtitle>
+    <ebay-section-title-cta href="https://www.ebay.com">See All</ebay-section-title-cta>
```
agliga: So I would almost say that for CTA, we should just have an option on the `<ebay-section-title>` tag. Something like link and text. If there's no text, it just shows the arrow. If the link option is passed, then it has the `<ebay-section-title-cta>` markup.

src/components/ebay-section-title/examples/05-info/template.marko
```diff
@@ -0,0 +1,10 @@
+<ebay-section-title>
+    <ebay-section-title-title>Today’s Deals – All With Free Shipping</ebay-section-title-title>
+    <ebay-section-title-subtitle>Plus, guaranteed best prices.</ebay-section-title-subtitle>
+    <ebay-section-title-info>
```
agliga: Does this have different markup? Can we just put it inside the title or something?

src/components/ebay-section-title/examples/06-overflow/template.marko
```diff
@@ -0,0 +1,11 @@
+<ebay-section-title>
+    <ebay-section-title-title>Today’s Deals – All With Free Shipping</ebay-section-title-title>
+    <ebay-section-title-subtitle>Plus, guaranteed best prices.</ebay-section-title-subtitle>
+    <ebay-section-title-overflow>
```
agliga: I would maybe make this section generic and call it like title-right-side or something.

src/components/ebay-section-title/examples/07-all-giant/template.marko
```diff
@@ -0,0 +1,18 @@
+<ebay-section-title size="giant">
+    <ebay-section-title-title href="https://www.ebay.com">Today’s Deals – All With Free Shipping</ebay-section-title-title>
```
agliga: Why do we have two different links here? 
Again, adding it as an option on the top might be cleaner.

dkallus: I forgot to remove href on `title`. This was an earlier idea and currently doesn't do anything.

I will likely move to using attributes on `ebay-section-title` parent.

dkallus: done

src/components/ebay-section-title/examples/03-cta/template.marko
dkallus: @agliga I removed `cta` child and added `href` and `cta-text` to the main `ebay-section-tag` tag. Let me know if you any other suggestions for `cta-text` attribute name.

src/components/ebay-section-title/examples/01-title/template.marko
dkallus: I updated component so this now works:
```
<ebay-section-title>Section Title</ebay-section-title>
```

src/components/ebay-section-title/examples/06-overflow/template.marko
dkallus: I'll check with Ian and Sean for naming suggestions as well. Keeping it as overflow for now, but I'm fine with renaming.

src/components/ebay-section-title/examples/05-info/template.marko
dkallus: I tried this, but since I use a `<div class="section-title__info>` container around infotip, the container class still exists when infotip isn't passed in body (data.renderBody is always defined, even when infotip isn't passed).

Possible workaround: I can get rid of "section-title__info" container, and possibly "section-title__overflow" container as well. Instead, I would style the infotip and overflow directly like `.section-title .infotip` and `.section-title .menu-button`. Then `<ebay-section-title-info>` and `<ebay-section-title-overflow>` won't be required anymore. If I do this, I need to change Skin as well, so checking with team first.

Thoughts?

src/components/ebay-section-title/examples/06-overflow/template.marko
ianmcburnie: I can @agliga's point, but I think overflow is okay. It's anything that "overflows" the section title. Also open to other suggestions though.

We definitely can't have direction specific names, like left or right, due to RTL. Things like "start" and "end" are better (taking a cue from CSS flexbox and Grid). As we build more of these layout kind of components I think we can re-review and look at consistency.

src/components/ebay-section-title/examples/05-info/template.marko
ianmcburnie: @dkallus @agliga My thinking was that the containers would give us a bit more flexibility in terms of layout. If you think they are redundant and provide no benefit, then perhaps we could first merge your current skin PR, then you could post a follow up draft showing the proposed change without container elements?

dkallus: I'd prefer to keep the containers

PR #921, opened 2019-10-15
Carousel: Add a11y heading text as needed;

### LINKED ISSUE #764: Carousel: Continuous status text issues
# Bug Report

## Description
Currently with continuous carousels the `a11y-status-text` option is omitted since there is no `totalSlides` (no `items-per-slide` is set). After discussion it sounds like we also need another a11y attribute in order to accommodate this.

Essentially we need one a11y attribute for the `slide x of y` text, and another for a prepending heading (for the complete a11y text of `Heading - slide x of y` where the continuous variant would just use `Heading`.

seangates:
## Description
- add a new <strike>`a11yHeadingText`</strike> `a11yTitleText` attribute
- modify carousel `a11yStatusText` to include <strike>heading</strike> title
- new logic block for heading as `a11yStatusText` when `itemsPerSlide` is not provided

## References
Fixes #764

### DIFF
--- a/src/components/ebay-carousel/README.md
+++ b/src/components/ebay-carousel/README.md
@@ -16,6 +16,8 @@ This component will bundle different resources depending on Lasso flags provided
 
 Name | Type | Stateful | Required | Description
 --- | --- | --- | --- | ---
+`a11y-heading-text` | String | No | No | a11y heading text used to describe the carousel (continuous only)
+`a11y-heading-tag` | String | No | No | use h1-h6 when there isn't a visible heading before the carousel (continuous only) (default: "h2")
 `a11y-previous-text` | String | No | Yes | a11y text for previous control (default: "Previous Slide")
 `a11y-next-text` | String | No | Yes | a11y text for next control (default: "Next Slide")
 `index` | String | Yes | No | 0-based index position
--- a/marko.json
+++ b/marko.json
@@ -68,6 +68,8 @@
     "@a11y-next-text": "string",
     "@a11y-status-text": "string",
     "@a11y-status-tag": "string",
+    "@a11y-heading-text": "string",
+    "@a11y-heading-tag": "string",
     "@a11y-current-text": "string",
     "@a11y-other-text": "string",
     "@a11y-play-text": "string",
--- a/src/components/ebay-carousel/examples/13-a11y-text-continuous/template.marko
+++ b/src/components/ebay-carousel/examples/13-a11y-text-continuous/template.marko
@@ -1,0 +1,24 @@
+<style>
+    .demo1-card {
+        color: #0a1c6b;
+        background: #c2f5ff;
+        font-size: 24px;
+        font-weight: bold;
+        width: 200px;
+        height: 120px;
+        line-height: 120px;
+        text-align: center;
+    }
+</style>
+<ebay-carousel
+    a11y-heading-text="Top Products"
+    a11y-heading-tag="h2"
+    a11y-previous-text="Go to previous slide - Top Products"
+    a11y-next-text="Go to next slide - Top Products">
+    <ebay-carousel-item class="demo1-card">Card 1</ebay-carousel-item>
+    <ebay-carousel-item class="demo1-card">Card 2</ebay-carousel-item>
+    <ebay-carousel-item class="demo1-card">Card 3</ebay-carousel-item>
+    <ebay-carousel-item class="demo1-card">Card 4</ebay-carousel-item>
+    <ebay-carousel-item class="demo1-card">Card 5</ebay-carousel-item>
+    <ebay-carousel-item class="demo1-card">Card 6</ebay-carousel-item>
+</ebay-carousel>
--- a/src/components/ebay-carousel/index.js
+++ b/src/components/ebay-carousel/index.js
@@ -25,6 +25,8 @@ function getInitialState(input) {
         a11yNextText: input.a11yNextText || 'Next Slide',
         a11yStatusText: input.a11yStatusText || 'Showing Slide {currentSlide} of {totalSlides} - Carousel',
         a11yStatusTag: input.a11yStatusTag || 'span',
+        a11yHeadingText: input.a11yHeadingText,
+        a11yHeadingTag: input.a11yHeadingTag || 'h2',
         a11yCurrentText: input.a11yCurrentText || 'Current Slide {currentSlide} - Carousel',
         a11yOtherText: input.a11yOtherText || 'Slide {slide} - Carousel',
         a11yPauseText: input.a11yPauseText || 'Pause - Carousel',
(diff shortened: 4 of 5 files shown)


seangates:
Yeah, I could use some direction here. I'm not exactly sure what the A11Y consequences are, or how to address them.

ianmcburnie:
@seangates I think @DylanPiercey sums it up well here: https://github.com/eBay/ebayui-core/issues/764

My only concern is the use of the term 'heading' in the new attribute. I'd be happier if it was 'title'. We also need to be certain this is not a breaking change for 4.x.

seangates:
@ianmcburnie Yeah, I read that, but I'm not sure when the `title` gets applied.

Also, note that I changed the attribute/var names: https://github.com/eBay/ebayui-core/pull/921/commits/5b39f463965a7cee4b4d1bdbd4eaf19a4a7a2487

src/components/ebay-carousel/README.md
```diff
@@ -26,7 +27,8 @@ Name | Type | Stateful | Required | Description
 
 Name | Type | Stateful | Required | Description
 --- | --- | --- | --- | ---
-`a11y-status-text` | String | No | Yes | status text (default: "Showing Slide {currentSlide} of {totalSlides} - Carousel")
+`a11y-heading-text` | String | No | No | a11y heading text used in combination with `a11y-status-text` (default: "{heading} Showing Slide {currentSlide} of {totalSlides} - Carousel") _**Note**: a `-` will be appended to the heading automatically. There's no need for the developer or model data to add it._
```
ianmcburnie: Name and description needs updating from 'heading' to 'title'.

seangates: Ah, I thought that I had done this! (I think I know what happened, though.) I'll get that fixed.

src/components/ebay-carousel/examples/13-continuous-title/template.marko
```diff
@@ -0,0 +1,20 @@
+<style>
```
ianmcburnie: To be less vague, I think this file should be named `13-a11y-text-continuous`, in line with the other a11y example.

ianmcburnie:
Okay, I'll try and rephrase Dylan's comments. @DylanPiercey please chime in if this is incorrect.

So, we have to make sure of the following:

1. The existing "a11y text" example continues to work as is (i.e. no breaking changes). We would mark that example as deprecated.
1. The "a11y text" example, updated to use `a11y-title-text`, would leverage this new attribute by prefixing it before the status text (it would also need its `a11y-status-text` updated to remove the old title text, to avoid duplication). It might be best to create another new example with this behaviour.
1. A continuous carousel out in the wild as is, with no changes, would continue to function as is (i.e. it doesn't output any title or status at all)
1. A continuous carousel opting into the new `a11y-title-text` would simply render that text, verbatim, with no other text. You have already added an example of this in the PR. The odd/confusing bit here I think is that `a11y-status-tag` would still be used to set its container tag. We may have to live with this and address it in v5.

seangates:
>verbatim

So, no replacement of `{currentPage} of {totalPages}`?

ianmcburnie:
To my knowledge, a continuous carousel has no knowledge of total pages. Which is why we are just wanting to render a static title in that scenario.

seangates:
@ianmcburnie Thanks for the clarity. I'll rework this PR a little.

seangates:
@ianmcburnie @DylanPiercey I believe I have things in order, but I'm still not 100% sure.

- `a11y-title-text` is now always required (per our docs)
- `a11y-status-text` now is prepended with the `a11y-title-text`
- `a11y-title-text` is used within the `a11y-status-tag` as a sensible default when using the continuous carousel
- removed the `-` bit in the docs as it was confusing and really not necessary if they supply their own `a11y-status-text` (which is required for discrete carousels, and we provide a default)

Let me know if you see anything amiss and I'll try and jump on it first thing Friday.

src/components/ebay-carousel/examples/09-a11y-text/template.marko
```diff
@@ -11,10 +11,10 @@
     }
 </style>
 <ebay-carousel
-    items-per-slide="2"
     a11y-previous-text="Go to previous slide - Top Products"
     a11y-next-text="Go to next slide - Top Products"
-    a11y-status-text="Showing slide {currentSlide} of {totalSlides} - Top Products - Carousel"
+    a11y-title-text="Top Products"
+    a11y-status-text="{title} Showing slide {currentSlide} of {totalSlides} - Carousel"
```
ianmcburnie: Will this create an unwanted space if `a11y-title-text` is not provided?

seangates: Sure, but does that ever get seen, or read out by the screen reader?

ianmcburnie: Screen readers do have very fine-grained punctuation settings. I'm honestly not sure if or how this might be treated, so I think its safest just to not have it if we can. Sorry.

seangates: @ianmcburnie Adjusted. Let me know if the changes work.

ianmcburnie:
@seangates First example is showing an empty block:

Screen Shot 2019-10-21 at 8 23 20 AM

seangates:
Correct. No a11y text was provided in that example. Also, we don't have a default string for the title.

ianmcburnie:
We can drop aria-live="polite" when the carousel is continuous (as in "13-a11y-text-continuous"):

Screen Shot 2019-10-21 at 8 23 20 AM

ianmcburnie:
@seangates It is not correct. No HTML should be rendered when a11y text is not provided.

ianmcburnie:
@seangates 09-a11y-text doesn't look right:

Screen Shot 2019-10-21 at 8 31 28 AM

Looks like the example is missing the status text.

seangates:
I didn't realize you meant the HTML markup. (You edited your comment after the fact. 😉 )

ianmcburnie:
Surely it didn't look right even before I made my edit? There is no inner text in that HTML block.

ianmcburnie:
Here's another idea:

* `a11y-heading-text` - used when items-per-slide is not set, and design does not provide a visible heading (e.g "Top Products")
* `a11y-heading-tag` - used when items-per-slide is not set, and design does not provide a visible heading (e.g "h2")

seangates:
You still want to call it `heading`?

ianmcburnie:
To clarify, I'm *proposing* reverting all code changes in this PR, and going with the following two new attributes only:

* `a11y-heading-text` - used when items-per-slide is not set, and design does not provide a visible carousel heading (e.g "Top Products")
* `a11y-heading-tag` - used when items-per-slide is not set, and design does not provide a visible carousel heading (e.g "h2")

When `a11y-heading-text` is not set, then nothing should be rendered

We would only need one new example adding to the examples page.

The README would need updating to explain when to use `a11y-heading-text` and when to use `a11y-status-text`.

@DylanPiercey @agliga @seangates thoughts?

seangates:
I think that's fine. Seems more straightforward to me.

ianmcburnie:
@seangates Okay, let's go with that

seangates:
@ianmcburnie @DylanPiercey Updated and simplified.

PR #922, opened 2019-10-15
Add special handling for aria-invalid;

seangates:
## Description
- adds `aria-invalid` capability to the button

## References
Companion PR to https://github.com/eBay/skin/pull/884, as part of the fix for https://github.com/eBay/skin/issues/741

## Screenshots
image

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -291,6 +291,7 @@
     "@grow": "boolean",
     "@borderless": "boolean",
     "@disabled": "boolean",
+    "@aria-invalid": "string",
     "@options <option>[]": {
       "@*": "expression",
       "@html-attributes": "expression",
--- a/marko.json
+++ b/marko.json
@@ -291,7 +291,7 @@
     "@grow": "boolean",
     "@borderless": "boolean",
     "@disabled": "boolean",
-    "@aria-invalid": "string",
+    "@invalid": "boolean",
     "@options <option>[]": {
       "@*": "expression",
       "@html-attributes": "expression",
--- a/src/components/ebay-listbox-button/examples/07-aria-invalid/template.marko
+++ b/src/components/ebay-listbox-button/examples/07-aria-invalid/template.marko
@@ -1,0 +1,5 @@
+<ebay-listbox-button name="formFieldName" aria-invalid="true">
+    <ebay-listbox-button-option value="1" text="Option 1"/>
+    <ebay-listbox-button-option value="2" text="Option 2"/>
+    <ebay-listbox-button-option value="3" text="Option 3"/>
+</ebay-listbox-button>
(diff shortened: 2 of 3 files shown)


ianmcburnie:
Does this need to go in README?

Also, given that the only aria value that makes sense in this scenario is "true" (i.e. not "spelling" or "grammar"), should we use a boolean `invalid` attribute and abstract away the aria details? I know we did something similar with `autocomplete` & `aria-autocomplete` elsewhere and so I'm striving for consistency. I guess my only concern with `invalid` would be that it may get confused with built-in HTML form validation in some way.

seangates:
We did it for the combobox, and it wasn't boolean.

ianmcburnie:
So, `invalid="true"` then?

seangates:
For combobox it's `autocomplete="list"` or `autocomplete="none"` (or, really, any value you want other than "list").

For this, we're allowing `aria-invalid` to have any value set. If no value is set for the `aria-invalid` attribute, the attribute will be null and Marko won't add it to the markup.

ianmcburnie:
Personally, I think a boolean is a better choice here. It's always true or false.

seangates:
The `aria-invalid` spec from WCAG states that it's an enum.

ianmcburnie:
Yes, I stated that above, "true", "false", "spelling", "grammar". We only need true or false.

seangates:
I'm trying to keep it as close to a "passthrough" attribute as possible. Even though it's not technically a "passthrough attribute", it's being treated as such in code.

At least, that's how I was envisioning it.

@ianmcburnie, is your argument to favor developer ergonomics over being closer to the spec? Because, I can certainly see that. `invalid=myValidityDataVar` is easier than `aria-invalid=(myValidityDataVar && 'true')`.

ianmcburnie:
That was definitely the thinking. Abstracting it away means that devs don't have to go looking up the ARIA spec and wonder which value out of several to use. We use `invalid` in ebay-textbox too, and we did something similar for 'autocomplete' in combobox. So I'm also arguing this way for consistency. 

I know we are doing some direct ARIA pass throughs in some places though. In some places that totally makes sense. I think as part of v5 we could sweep over the API again and see how consistent we are being and decide one or approach or another, or just keep it on a case by case basis as we are doing now.

Might be something @abiyasa and @darkwebdev are interested in.

seangates:
Should `aria-invalid` also be added to the hidden select?

ianmcburnie:
I don't think a user ever interacts with the select directly, do they? We only use it for storing form data. So probably not necessary to add it there. If on the other hand a select is rendered on the server side, before JavaScript kicks in (i.e. progressive enhancement), then it would be needed.

darkwebdev:
I like the idea of `invalid` property. It would be even cooler to be able to use it without any value like `<ebay-listbox-button disabled>...</ebay-listbox-button>`, but I'm not sure if that's consistent with your codestyle.

seangates:
@ianmcburnie You know, that's an interesting thought, because at this point JS is required because the `<select/>` stays hidden 100% of the time.

@darkwebdev Please see my most recent commit. We have other places where booleans are used. E.g. `disabled` or `borderless`.

Issue #807, opened 2019-08-01
Listbox-button: pass name attribute to button


ZZS77:
# Feature Request

## Description
Need name attribute on listbox button to be able to be query selectable

ianmcburnie:
@ZZS77 can you confirm what you mean by query selectable?

PR #931, opened 2019-10-22
Listbox Button: Adding name attribute to the button

LokeshRishi:
## Description
Added name attribute to the button in `ebay-listbox-button`. The value should be provided as a part of the options as `buttonName`.

## References
Resolving #807

### DIFF
--- a/src/components/ebay-listbox-button/README.md
+++ b/src/components/ebay-listbox-button/README.md
@@ -34,6 +34,7 @@ Name | Type | Stateful | Required | Description
 `selected` | Number | Yes | n/a | allows you to set the selected index option to `selected`
 `borderless` | Boolean | No | No | whether button has borders
 `fluid` | Boolean | No | No | whether listbox width is 100%
+`buttonName` | String | No | Yes | used for the `name` attribute of the native `<button>`
 
 **Note:** For this component, `class` is applied to the root tag, while all other HTML attributes are applied to the `input` tag.
 
--- a/marko.json
+++ b/marko.json
@@ -295,6 +295,7 @@
     "@disabled": "boolean",
     "@fluid": "boolean",
     "@invalid": "boolean",
+    "@button-name": "string",
     "@options <option>[]": {
       "@*": "expression",
       "@html-attributes": "expression",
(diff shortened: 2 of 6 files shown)


src/components/ebay-listbox-button/README.md
```diff
@@ -34,6 +34,7 @@ Name | Type | Stateful | Required | Description
 `selected` | Number | Yes | n/a | allows you to set the selected index option to `selected`
 `borderless` | Boolean | No | No | whether button has borders
 `fluid` | Boolean | No | No | whether listbox width is 100%
+`buttonName` | String | No | Yes | used for the `name` attribute of the native `<button>`
```
ianmcburnie: @agliga @seangates Is this correct? `buttonName` instead of just `name`? I thought this would just be a pass through.

seangates: `name` is what it should be. And, we probably don't need to document it. We simply should handle it and place the attribute in the right place in the template.

So, the only changes should be:
- add the attribute to `marko.json`
- add the attribute to button in the template

Maybe we could even do without the tests? I can see that being good to know the name was set on the button instead of the wrapper, though.

LokeshRishi: @seangates `name` is already in use:
https://github.com/eBay/ebayui-core/blob/56b70df668de30d113de683b7cb86e5f72a4fe9c/src/components/ebay-listbox-button/template.marko#L65.

Do you think we should use the same option at both the places?

agliga: So from what it looks like is we probably have to move the name from the select to the button. 
Since the select listbox button has the value of the selected options, when submitting the form `button` will get those values and map it to name. 
If we have `select` have the same name, it will conflict and have two of the same keys. 
That or we have another option for name/selectName. Also moving name to button/changing that functionality might be breaking.

ianmcburnie: Ah, good point @agliga . Given that the purpose of the listbox-button is to submit form data, then the server should be getting the key/value pairs from the name of the select element. Not entirely sure why they want to query the button instead, but I think the PR as it currently stands should satisfy that and also be backwards compatible...?

Issue #926, opened 2019-10-18
Checkbox: support onFocus event


BhavinPatel04:
# Feature Request

## Description
We need to pass an `onFocus` event to the input

ianmcburnie:
@BhavinPatel04 I'm curious as to why onfocus is needed for a checkbox. Is it a tooltip of some sort?

BhavinPatel04:
@ianmcburnie We need to hide all the tooltips and toasts onFocus

ianmcburnie:
Okay, we'll add this. For parity we'll update radio component too.

ianmcburnie:
Done in 4.4.0-1

PR #935, opened 2019-10-23
Adding support for w-onfocus to checkbox and radio

LokeshRishi:
## Description
Updated `ebay-checkbox` and `ebay-radio` - adding support for `w-onfocus`

## References
Fix for #926


src/components/ebay-checkbox/index.js
```diff
@@ -7,5 +7,15 @@ module.exports = require('marko-widgets').defineComponent({
...
-    }
+    },
+    handleFocus: forwardEvent('focus')
 });
+
+function forwardEvent(eventName) {
+    return function(originalEvent, el) {
+        this.emit(`checkbox-${eventName}`, {
+            originalEvent,
+            value: (el || this.el.querySelector('input')).value
```
agliga: Should add checked too?

src/components/ebay-checkbox/index.js
```diff
@@ -7,5 +7,15 @@ module.exports = require('marko-widgets').defineComponent({
             value: originalEvent.target.value,
             checked: originalEvent.target.checked
         });
-    }
+    },
+    handleFocus: forwardEvent('focus')
 });
+
+function forwardEvent(eventName) {
```
agliga: Either we should have handleChange call into this too (which I think might be the better way to do it), or we remove this and have handleFocus so the same as handleChange but with a different event.

src/components/ebay-radio/index.js
```diff
@@ -10,5 +10,15 @@ module.exports = require('marko-widgets').defineComponent({
...
             });
         }
-    }
+    },
+    handleFocus: forwardEvent('focus')
 });
+
+function forwardEvent(eventName) {
+    return function(originalEvent, el) {
+        this.emit(`radio-${eventName}`, {
```
agliga: @ianmcburnie Should we have a check for disabled as we do in the handleClick event too?
Also same as above, I'd combine the click and focus and use forwardEvents.

ianmcburnie: I don't think a disabled radio can gain keyboard focus, so we probably wont run into that.

src/components/ebay-checkbox/index.js
ianmcburnie: Yep.

src/components/ebay-checkbox/index.js
ianmcburnie: Yes, for consistency, let's have them both use this new `forwardEvent` function

src/components/ebay-radio/index.js
```diff
@@ -1,14 +1,17 @@
...
-        const target = originalEvent.target;
+    handleClick: forwardEvent('change'),
+    handleFocus: forwardEvent('focus')
+});
 
-        if (!target.disabled) {
-            this.emit('radio-change', {
+function forwardEvent(eventName) {
+    return function(originalEvent, el) {
+        if (eventName !== 'change' || (eventName === 'change' && !el.disabled)) {
```
agliga: Why not do `if (eventName !== 'change || !el.disabled)`.
That way if eventName is change and it is disabled then it won't pass. Otherwise if event is not change it'll pass or if the el is not disabled. 

Also even if they can't focus, this seems like extra unnecessary logic, I would say if we're going with this route, then a simple `if (!el.disabled)` should be fine. It's not like they're going to trigger this event on a disabled radio anyways

src/components/ebay-checkbox/index.js
```diff
@@ -1,11 +1,16 @@
...
+    handleFocus: forwardEvent('focus')
+});
+
+function forwardEvent(eventName) {
+    return function(originalEvent, el) {
+        this.emit(`checkbox-${eventName}`, {
             originalEvent,
-            value: originalEvent.target.value,
-            checked: originalEvent.target.checked
+            value: (el || this.el.querySelector('input')).value,
```
agliga: Since this is on the input itself, does it make sense to change the value selector?
Also did you test that this is giving the right value?

src/components/ebay-radio/index.js
LokeshRishi: Agree. Let's keep it consistent with checkbox, I will remove the disabled check.

agliga: Don't remove the disabled check, it is needed. 
Add it for both cases imo

src/components/ebay-checkbox/index.js
LokeshRishi: Yes, and I have updated the focus event test to validate the value.

Issue #934, opened 2019-10-22
Carousel: add option to always show paddles


ianmcburnie:
Deals & events team has a case where slideshow carousel (i.e. one item per slide) has many slides, resulting in many dots, which does not look good. Ideally they would reduce the amount of items in the carousel, but this is not possible.

Removing the dots removes all kind of affordance (as the paddles only show up on hover in this case). Adding a small peek does not look good either.

So I'm proposing to add an option to always show paddles for this scenario.

Perhaps a `paddles` attribute with the following values to begin with:

* `auto` - automatically determined (default, current behaviour)
* `on` always show

<img width="412" alt="Screen Shot 2019-10-22 at 3 10 31 PM" src="https://user-images.githubusercontent.com/38065/67341432-7ee26d80-f4e4-11e9-81fd-ebae9e43b630.png">

sairam0903:
Can't wait to see this option for the Carousel

ianmcburnie:
So, I was under the impression that the above behaviour was by design on our end. But thinking about it, it doesn't really makes sense. A much simpler proposal from me would have been to just say "always show paddles when `no-dots` and `items-per-slide=1`. Sorry. We could possibly still keep this option as a secret option if we feel it's useful...

sairam0903:
@ianmcburnie So the solution for this issue is, when `items-per-slide=1` and `no-dots` is set, we always show the paddles without any secret option?

ianmcburnie:
@sairam0903 That's correct. Let us know if you foresee any issue with that.

sairam0903:
I think that should be fine, let us know the version to upgrade when this is out.

PR #936, opened 2019-10-23
Added show paddles always on carousel on nodots and no peek

agliga:
## Description
Changed to show paddles on `no-dots` and no peek

## References
#934

### DIFF
--- a/src/components/ebay-carousel/index.js
+++ b/src/components/ebay-carousel/index.js
@@ -19,6 +19,7 @@ function getInitialState(input) {
         config: {}, // A place to store values that should not trigger an update by themselves.
         gap: isNaN(gap) ? 16 : gap,
         noDots: input.noDots,
+        paddles: input.paddles || 'auto',
         index: parseInt(input.index, 10) || 0,
         itemsPerSlide: parseFloat(input.itemsPerSlide, 10) || undefined,
         a11yPreviousText: input.a11yPreviousText || 'Previous Slide',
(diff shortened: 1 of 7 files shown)


ianmcburnie:
@agliga Can you also just check that this case is covered appropriately: `items-per-slide`, `no-dots`, `autoplay`. It should not show paddles (as per screenshot). Because the play/pause is the affordance. Correct?

Screen Shot 2019-10-23 at 2 03 54 PM

agliga:
> @agliga Can you also just check that this case is covered appropriately: `items-per-slide`, `no-dots`, `autoplay`. It should not show paddles (as per screenshot). Because the play/pause is the affordance. Correct?
> 
> Screen Shot 2019-10-23 at 2 03 54 PM

Good catch, I'll update. It was showing the paddles on autoplay

src/components/ebay-carousel/index.js
```diff
@@ -118,6 +118,7 @@ function getTemplateData(state) {
         offset: hasOverride ? config.offsetOverride : offset,
         disableTransition: hasOverride,
         showPaddles: itemsPerSlide ? items.length > itemsPerSlide : true,
+        alwaysShowPaddles: state.noDots && state.peek === 0 && !state.autoplayInterval,
```
LokeshRishi: @agliga Why do we need a check for `autoplayInterval`?

agliga: This was brought up by Ian in a previous comment. 
The idea behind this fix is if we have a carousel with no peek and no dots we should show that this widget is a carousel. Otherwise people won't know unless they hover over it.
So when we have an autoplay carousel, we shouldn't show paddles because the autoplay will show that it is a carousel and automatically switch to the next slide and show the play/pause button.

Issue #889, opened 2019-09-26
Pagination: use foreground icons instead of background icons


ianmcburnie:
@seangates reports that pagination is using background icons. We should switch this over to use foreground icons, using `<ebay-icon>`, like we do elsewhere.

PR #938, opened 2019-10-25
Changed pagination component to use ebay-icon

agliga:
## Description
Changed pagination to use ebayui icons

## Context
There was some skin changes required.
I added pagination prev and next icons which render differently on ds4 or 6 to keep backwards compatibility.

## References
#889 
Skin change: https://github.com/eBay/skin/pull/893

## Screenshots
Screenshots are done with the skin changes applied 
DS6 with foreground images
image

DS4 with foreground images
image


agliga:
The tests are failing, this is because the skin version is the old version. 

However when I ran the tests with the new version of skin with the pagination inline images fix the tests pass locally. 
So once skin is updated to `9.4.0` then the tests will pass

Issue #939, opened 2019-10-25
ebay-menu-button-item: Attributes not being passed along to the rendered link


tws2000:
#909 Bug Report
Attributes added to <ebay-menu-button-item> are not being rendered in the resulting link.

## Description
For instance if I use ```<ebay-menu-button type="fake"><ebay-menu-button-item href="http://www.ebay.com" data-track="123"/></ebay-menu-button>```, I'd like the menu item link to have the data-track attribute and value.

ianmcburnie:
Fix added to 9.4.0-1

PR #940, opened 2019-10-28
Changed attributes to passthough to ebay-menu-item

agliga:
## Description
By doing `${item}` this passes through and attributes defined. That way we don't have to define any attributes and then menu-item will process all html attributes. 

## Context
We should change other components to do the same, but this will solve this given issue. 
I also tested the badge-number if it was working as badgeNumber and it worked fine. Also render body is not needed since it will be passes in the $(item)

## References
#939

### DIFF
--- a/src/components/ebay-menu-button/template.marko
+++ b/src/components/ebay-menu-button/template.marko
@@ -60,18 +60,7 @@
         w-on-menu-change="handleMenuChange"
         w-on-menu-select="handleMenuSelect">
         <for(item in data.items)>
-            <ebay-menu-item
-                style=item.style
-                class=item.class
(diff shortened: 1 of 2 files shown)


DylanPiercey:
Probably a good idea to add a test for this. Should be able to copy [this test](https://github.com/eBay/ebayui-core/blob/3a092b475ab99c1b8b595bfb3d166a011bf7e756/src/components/ebay-menu/test/test.server.js#L78) from the menu, to [here](https://github.com/eBay/ebayui-core/blob/3a092b475ab99c1b8b595bfb3d166a011bf7e756/src/components/ebay-menu-button/test/test.server.js#L153).

Issue #918, opened 2019-10-14
ebay-button: dynamic attributes are not being added to the button


BhavinPatel04:
# Bug Report

## Description
**Steps to reproduce:**
````
let commonAttrs = {
    priority: "primary",
    data-href: "some-href"
}
let specificAttrs = {
    data-value: "someValue"
}
<ebayui-button ...commonAttrs ...specificAttrs>Button</ebayui-button>
````
**Expected:**
Button should have `priority`, `data-href` and `data-value` attributes
**Actual:**
Button only takes `priority` attribute

## Workaround
````
<ebayui-button  priority="primary" data-href="some-href" data-value="someValue">Button</ebayui-button>
````

LokeshRishi:
@BhavinPatel04 It seems like you are using incorrect attribute names:
- `data-href` -> `href`
- `data-value` -> `value`

Could you please try with the above mentioned changes?

BhavinPatel04:
@LokeshRishi they are some examples of attributes that are not mentioned in the table in README. For example: `data-template`, `_sp`, `anyOtherAttr`

BhavinPatel04:
This issue is not fixed. I am still facing this issue in v4.4.0-1.

ianmcburnie:
We'll look into it and push another pre-release.

ianmcburnie:
Done in 4.5.0.

PR #944, opened 2019-10-30
918 spread attrs

agliga:
## Description
Changed ebay-button to support spread attributes

## Context
We discussed to add an ignore list in input, that way if that is passed then it will not use any attributes in that list. I also made it so that there are some (like renderBody which will not be added as attributes)

## References
#918 

This needs to be added to every component eventually but we only added button for this case since this was requested to fix (and I also as a proof of concept)

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -30,7 +30,10 @@
   },
   "<ebay-button>": {
     "renderer": "./src/components/ebay-button/index.js",
-    "@*": "expression",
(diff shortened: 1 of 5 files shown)


ianmcburnie:
I feel a little nervous about changing a core utility function at this stage of the release. Might be better left to v4.5.0 and a longer test cycle.

PR #946, opened 2019-11-01
Added support for spread attributes

agliga:
## Description
Expanded processHtmlAttributes helper method to basically parse all given input.
Initially because of target-property, all additional properties were being added into a * object. However, spread attributes `<sometag ...attr />` were not. So since * object is being deprecated we have also added the proper fixes to support the new syntax in marko. 
Basically the input object contains all additional properties. However, there is no way to know which properties are base props and which ones are additional props without reading the marko.json file. So we added a new argument which is an array list of all additional properties which should be ignored (and assuming that the base template will make use of)

## Context
Some notes:
* I basically copied marko.json variables into the new array. 
* All properties need to be camel case because the kabab case converts to camel case in input. (I tried to make sure all of them are camel case but I might have missed some. Maybe I need to use the converter just in case?)
* The only test failure I had was a combobox one. This seems to actually be because the test was not actually running correctly. What happened was the test would inject input.value. However the combobox expects input["*"].value, so the test was not being run. Once I changed it, I noticed that the combobox was only rendering 1 options object which I assume is correct since combobox takes value and filters. So I had updated the test accordingly. 
* I had to add skipAttrs because marko adds those attributes to the input and we don't want them in the HTML.
* I also updated the tests and removed all the * attribute tests. 

## References
#918


src/common/html-attributes/index.js
```diff
@@ -1,5 +1,5 @@
 const assign = require('core-js-pure/features/object/assign');
-
+const skipAttributes = ['*', 'htmlAttributes', 'renderBody', 'widgetState', '__widgetProps', '___widgetProps'];
```
DylanPiercey: For `___widgetProps` we'll a actually need to do a release of Marko 4 to make this non enumerable since the above code would actually not work when running against the production version of Marko since the `___` properties get minified.

src/common/html-attributes/index.js
```diff
@@ -12,21 +12,24 @@ function camelToKebab(s) {
  * Create object of HTML attributes for pass-through to the DOM
  * @param {Object} input
  */
-function processHtmlAttributes(input = {}) {
+function processHtmlAttributes(input = {}, ignore) {
```
DylanPiercey: Since we're passing in the entire input now, there probably isn't a case where it is undefined. Maybe worth removing the default parameter to an empty object?

src/common/html-attributes/index.js
```diff
@@ -12,21 +12,24 @@ function camelToKebab(s) {
...
-        obj = assign(wildcardAttributes, htmlAttributes);
+    let obj = htmlAttributes || {};
+    if (htmlAttributes) {
+        obj = assign({}, htmlAttributes);
     }
-
-    if (obj) {
-        Object.keys(obj).forEach((key) => {
-            attributes[camelToKebab(key)] = obj[key];
+    if (!!ignore) {
```
DylanPiercey: Is there a case where there isn't an ignore?

agliga: That was left in order to validate tests so I could check if it was using the old syntax. I can remove it

src/common/html-attributes/index.js
agliga: That's fine, but I think it should be safe to leave it in there until the release.

DylanPiercey: The thing is it would give a false sense of security. It works in dev, but would break in prod (for Marko 4 apps w/ compatibility layer). We shouldn't release it like this.

DylanPiercey: I can try to get out a release of Marko today along side this. We should make sure to include in the release notes that the latest version of Marko is required if using with Marko 4.

agliga: And marko v3 too correct? Since this should exist there too right? Or just marko 4?

DylanPiercey: The `___` properties are only there in Marko 4, and get renamed/minified in prod.

DylanPiercey: I just realized in Marko 3 this list is missing `widgetBody`

DylanPiercey: Actually `widgetProps` as well was there in Marko 3. Not sure why we made it a `___` prop in Marko 4. I am going to update the compatibility layer in Marko 4 to actually just use the bare `widgetProps` name as well.

agliga: So should I keep the __widgetProps (two underscore) OR rename to widgetProps

agliga: Anyways, I added both widgetProps and widgetBody then. Kept the two underscore __widgetProps

DylanPiercey: The full list should be `['htmlAttributes', 'renderBody', 'widgetState', 'widgetProps', 'widgetConfig', 'widgetBody']`.

We shouldn't need `*` either because we changed everything to `targetProperty: null`.

agliga: 👍 Should be updated

ianmcburnie:
Not sure if this is related, but I'm seeing this error on menu-button example page:

Screen Shot 2019-11-05 at 11 29 26 AM

agliga:
> Not sure if this is related, but I'm seeing this error on menu-button example page:
> 
> Screen Shot 2019-11-05 at 11 29 26 AM
> 
> It's preventing any interaction with the menu buttons.

I was also able to see it, (and on `4.5.0` branch) but after running clean and starting again, I can't see that error anymore.

ianmcburnie:
Ok, everything else seems to be running well for me in local branch.

PR #948, opened 2019-11-05
Added page notice celebration and window notice

agliga:
## Description
Added support for page notice celebration and window notice
Blocked because waiting for #946 to be merged first in order to add spread attr support

## Context
* Both require a new title element (But in different places). I added it in the right context. However, it might be better to have just a title attribute? Or maybe not. 
* There seems to be a lot of conditionals for this, I think we should consider refactoring each of the different parts (content, icon container) to be different components at least. 
* I had to add code in the attr transformer to allow the `ebay-notice-title` tags to be rendered correctly (will remove in marko v4)

## References
#877

## Screenshots
DWEB
image
image
image

MWEB
image
image
image


ianmcburnie:
Should this PR contain the storybook generated updates? Or have we been adding those at a later stage? (i.e. release time)

ianmcburnie:
@agliga Is there a way to opt-in to the `window-notice--full` variant?

ianmcburnie:
@agliga Can you make the window notice example match the skin one, like for like?

Screen Shot 2019-11-05 at 4 06 41 PM

src/components/ebay-notice/examples/11-window-notice/template.marko
```diff
@@ -0,0 +1,10 @@
+<ebay-notice
+    type="window"
+    a11y-heading-text="information">
```
ianmcburnie: I think `a11y-heading-text` becomes irrelevant with window notice, because it just uses the entire contents of the h2 to label the region. We'll clean up these differences when we refactor, but for now just make a note of it in docs and remove from example.

agliga:
> @agliga Can you make the window notice example match the skin one below, like for like?
> 
> Screen Shot 2019-11-05 at 4 06 41 PM

The problem is the CSS, it uses margin instead of padding, so we would need to change the outer container to have a white background
I could do that create a `<div>` wrapper with `background-color: white;`, however shouldn't it have padding instead of margin?

ianmcburnie:
Hmm, I'm not sure I understand. Both page notice and window notice use margin (correctly) on the root element to keep it a minimum of n pixels away from the module above/below.

agliga:
> Hmm, I'm not sure I understand. Both page notice and window notice use margin (correctly) on the root element to keep it a minimum of n pixels away from the module above/below.

I added some examples. I added a container with a border and background which did the trick. 
Also I added height/width constraints on the fill version

agliga:
> @agliga Is there a way to opt-in to the `window-notice--full` variant?

Addressed this

marko.json
```diff
@@ -480,6 +480,7 @@
     "@html-attributes": "expression",
     "@class": "string",
     "@style": "string",
+    "@fill": "boolean",
```
ianmcburnie: Just spotted this. Another potential boolean trap, but I think we all realise the entire component is in need of a refactor/separation, so let's deal with it then. I will rename this to `fill-window` and update the README to make usage a little clearer in the interim.

PR #955, opened 2019-11-12
ebay-combobox-readonly: Upgrade to Marko 4

LokeshRishi:
## Description
Upgraded `ebay-combobox-readonly` component to Marko 4


src/components/ebay-combobox-readonly/index.marko
```diff
@@ -0,0 +1,53 @@
+$ var data = component.getTemplateData(state, input);
```
agliga: I think we should do as before. 
Use input instead of data. 
We prob can move getTemplateData to template itself
For all the classes in template just do
`class=['class1', input.borderless && 'borderlessClass']` etc

LokeshRishi: Updated. Removed `getTemplateData()` altogether.

PR #960, opened 2019-11-15
Marko v4 upgrade (initial migration)

agliga:
## Description
This is the first pass of upgrading to marko v4. The following has been done
1. Updated to marko-widgets 7 (I think it was already there)
2. Removed all marko 3 packages
3. Removed marko v3 hacks
4. Cleaned up code for migration
5. Ran migration on each component individually
6. Cleanup browser.json (and removed some, since marko 4 doesn't need them)
7. Ran migration on whole app (in case anything was missed or examples didn't migrate)
8. Ran all tests and checked on server that everything is working

## Context
Couple of things
* Most tests passed. The migration tool would change some most code to be `onCreate`, which I had to change to `onInput`. If it had both, I tried to clean it up and combine it.
* The main test that had to change was the rerender test. In marko 4 if a component is rerendered with the same input it does not trigger a change. I changed some tests to explicitly change the input in order to force the rerender
* I did a lot of cleanup but some things I left as is from migration. I think we should go further and clean up more from this. We should be only setting state for things we change and use input otherwise.


demo/browser.json
```diff
@@ -1,7 +1,6 @@
 {
     "dependencies": [
         "require: marko",
```
DylanPiercey: `require: marko` shouldn't be needed.

package.json
```diff
@@ -125,8 +121,8 @@
   "peerDependencies": {
     "@ebay/skin": "9.4.0",
     "chromedriver": "^77.0.0",
-    "marko": "^3 || ^4",
-    "marko-widgets": "^6 || ^7"
+    "marko": "^4",
+    "marko-widgets": "^7"
```
DylanPiercey: We shouldn't need marko-widgets @7

src/components/components/ebay-tooltip-overlay/browser.json
```diff
@@ -1,5 +1,5 @@
 {
     "dependencies": [
-        "require: ./index.js"
+        "require: ./index.marko"
```
DylanPiercey: This file shouldn't be needed.

src/components/components/ebay-tooltip-overlay/index.marko
```diff
@@ -0,0 +1,81 @@
+import constants from "./constants"
+
+static var typeRoles = constants.typeRoles;
+
+static var pointerStyles = constants.pointerStyles;
+
+$ {
```
DylanPiercey: Would be nice to merge these two scriptlets.
There is a few places with code like this (also for static blocks that could be merge).

src/components/ebay-carousel/index.marko
```diff
@@ -0,0 +1,146 @@
...
+            key="list"
+            id:scoped="list">
+            <for|item| of=data.items>
+                <li
+                    ...item.htmlAttributes
+                    class=item.class
+                    style=item.style
+                    aria-hidden=(!item.fullyVisible && "true")
+                    no-update-body-if(!!config.preserveItems)>
+                    <if(
```
DylanPiercey: Most places with this renderBody check can be converted to just `<${item.renderBody}/>` instead of the if statement.

src/components/ebay-checkbox/index.marko
```diff
@@ -0,0 +1,29 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+static {
+    var ignoredAttributes = ["class", "style", "disabled"];
```
DylanPiercey: With single line blocks like this I'd ideally like to see it on one line, eg

```
static var ignoredAttributes = ["class", "style", "disabled"];
```

src/components/ebay-infotip/browser.json
```diff
@@ -1,6 +1,6 @@
 {
     "dependencies": [
         "../ebay-icon",
-        "require: ./index.js"
+        "require: ./index.marko"
```
DylanPiercey: Neither of these deps should be needed. (Should be able to remove this browser.json).

PR #961, opened 2019-11-15
Upgrade markov4

agliga:
## Description
This is the first pass of upgrading to marko v4. The following has been done
1. Updated to marko-widgets 7 (I think it was already there)
2. Removed all marko 3 packages
3. Removed marko v3 hacks
4. Cleaned up code for migration
5. Ran migration on each component individually
6. Cleanup browser.json (and removed some, since marko 4 doesn't need them)
7. Ran migration on whole app (in case anything was missed or examples didn't migrate)
8. Ran all tests and checked on server that everything is working

## Context
Couple of things
* Most tests passed. The migration tool would change some most code to be `onCreate`, which I had to change to `onInput`. If it had both, I tried to clean it up and combine it.
* The main test that had to change was the rerender test. In marko 4 if a component is rerendered with the same input it does not trigger a change. I changed some tests to explicitly change the input in order to force the rerender
* I did a lot of cleanup but some things I left as is from migration. I think we should go further and clean up more from this. We should be only setting state for things we change and use input otherwise.


DylanPiercey:
I left some comments in https://github.com/eBay/ebayui-core/pull/960

agliga:
> I left some comments in #960

Sorry, I had closed that one since it was targeted to the wrong branch, and some of those fixes have been fixed since I had a problem pushing upstream (like all the `require:`).
I will check the other issues like merging state, I have done some of them but not all.

PR #971, opened 2019-12-02
Dialog: restore dialog focus on close

### LINKED ISSUE #941: Accessibility: Selling: Focus lost upon dismissing
Focus lost from original element, after closing dialog.

DylanPiercey:
## Description
This PR updates the `ebay-dialog` to track the focused element before trapping the focus within the dialog. When the dialog is closed it then restores focus to the previous element.

The user can still control restoring the focus manually by using the `on-close` event and setting the focus there since the focus is restored internally just before emitting that event.

## References
Fixes #941

### DIFF
--- a/src/components/ebay-dialog/component.js
+++ b/src/components/ebay-dialog/component.js
@@ -53,6 +53,7 @@ module.exports = {
 
         // Ensure focus is set and body scroll prevented on initial render.
         if (isFirstRender && isTrapped) {
+            this._prevFocusEl = document.activeElement;
             focusEl.focus();
             bodyScroll.prevent();
         }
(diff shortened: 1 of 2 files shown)


ianmcburnie:
Is the focus always set on the previous element before the manually chosen element? If so, this can potentially cause the screen reader to start reading the previous element before the manually chosen element. Probably an edge case as it happens so fast, but in general we try and avoid any quick switching of focus to prevent this from happening (some screen readers get confused very easily!).

DylanPiercey:
@ianmcburnie yes it would quickly set focus to two different elements in that case (would be in the same tick). I could update the logic to check if focus was altered via the event to avoid this. Do you think that is worth doing?

ianmcburnie:
I think it's worth doing. I've seen this cause some unexpected issues with screen readers.

DylanPiercey:
Updated the PR to address the above @ianmcburnie

src/components/ebay-dialog/component.js
```diff
@@ -67,8 +68,18 @@ module.exports = {
...
                 } else {
                     bodyScroll.restore();
+                    const activeElement = document.activeElement;
                     this.emit('dialog-close');
 
+                    if (
+                        // Skip restoring focus if the focused element was changed via the dialog-close event
+                        activeElement === document.activeElement &&
+                        // Skip restoring focus if the previously focused element was removed from the DOM.
+                        this._prevFocusEl.parentNode
```
mlrawlings: `isConnected` would be better as if the previously focused element weren't removed directly it would still have a parent.

Problem is browser support.  Edge and IE don't support this property.

DylanPiercey: Good catch. I'll come up with a solution that works in IE and Edge as well.

DylanPiercey: Fixed using `document.documentElement.contains` instead.

ianmcburnie: @mlrawlings is this solution to your satisfaction? :-) If so, we can merge.

PR #973, opened 2019-12-02
Expand-Button: Add click and escape key event handling;

### LINKED ISSUE #970: Expand-button: button doesn't toggle the chevron when clicked and doesn't emit the click event
# Bug Report

## Description
The new `ebay-expand-button` extracted from the `ebay-button` expand variant in v3.0.0 doesn't toggle the chevron when clicked and doesn't emit the click event.

## Workaround
None

seangates:
## Description
- add click event handler
- add escape key event handler
- add widget code
- update template to support being a widget
- add browser tests accordingly

## Context
The expand button component was not sending any click events or allowing the escape key to be captured (as is done in the plain button). These changes make the button capable of being interactive with other Marko components.

**_Note:_** the "expanded" state should be handled by the wrapping component using `aria-expanded="true"` to toggle the chevron. If it is being used to expose a popup (like a menu) the developer would also need to add `aria-haspopup="true"` and make sure that the expand button immediately precedes the menu (or other popup) in the DOM.

## References
Fixes #970

### DIFF
--- a/src/components/ebay-expand-button/index.js
+++ b/src/components/ebay-expand-button/index.js
@@ -1,0 +1,23 @@
+const assign = require('core-js-pure/features/object/assign');
+const eventUtils = require('../../common/event-utils');
+
+module.exports = require('marko-widgets').defineComponent({
+    template: require('./template.marko'),
+    getInitialState(input) {
+        return assign({}, input, {
+            disabled: input.disabled
+        });
+    },
+    handleClick(originalEvent) {
+        if (!this.state.disabled) {
+            this.emit('button-click', { originalEvent });
+        }
+    },
+    handleKeydown(originalEvent) {
+        eventUtils.handleEscapeKeydown(originalEvent, () => {
+            if (!this.state.disabled) {
+                this.emit('button-escape', { originalEvent });
+            }
+        });
+    }
+});
--- a/marko.json
+++ b/marko.json
@@ -159,7 +159,7 @@
     "@a11y-close-text": "string"
   },
   "<ebay-expand-button>": {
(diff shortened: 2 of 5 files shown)


ianmcburnie:
Can this go into 4.5.0 instead? We should be releasing this week. If you prefer it in a patch, that's fine.

seangates:
Sooner the better for the B2C team. 👍

ianmcburnie:
Cool. Patch it is!

src/components/ebay-expand-button/index.js
```diff
@@ -0,0 +1,23 @@
+const assign = require('core-js-pure/features/object/assign');
+const eventUtils = require('../../common/event-utils');
+
+module.exports = require('marko-widgets').defineComponent({
+    template: require('./template.marko'),
+    getInitialState(input) {
```
agliga: Should we add a toggle for the chevron too? IE put it here in initial state (or use input value?) and then toggle to true and change the chevron icon?
I thought that was part of the description as well for this issue?

seangates: No, I don't believe this button should have any toggle state. Specifically, we use `aria-expanded="true"` Skin takes care of the chevron automatically.

(Note my comment in the description.)

seangates: Also, this is prescient given how the `makeup-expander` is used, because of it's DOM manipulation outside of the components we're building. That's a huge part of keeping toggled state outside of this component. Especially when we go to replace `ebay-button` in the `ebay-menu-button` (and elsewhere).

Issue #964, opened 2019-11-18
Menu/Menu-Button: refactor for Marko v4


ianmcburnie:
We have identified this as one of the components most in need of a refactor in v4.

ianmcburnie:
Completed.

PR #975, opened 2019-12-06
Refactored menu and menu button for marko v4

agliga:
## Description
Changed menu and menu button to no longer use state for everything, but input instead. Changed stated to be only the checked items

## Context
* I pass index through the events now, so we don't have to find the pressed item using fancy html manipulation
* I had to add the index on the ebay-menu change event because then the menu-button couldn't sync it's state based on it. A possible better solution (but might be a braking change), is to pass the state from `ebay-menu` in events and have the `ebay-menu-button` sync it's state based on that state

## References
#964

### DIFF
--- a/src/components/ebay-menu-button/component.js
+++ b/src/components/ebay-menu-button/component.js
@@ -1,6 +1,5 @@
 const Expander = require('makeup-expander');
 const assign = require('core-js-pure/features/object/assign');
-const indexOf = require('core-js-pure/features/array/index-of');
 const findIndex = require('core-js-pure/features/array/find-index');
 const eventUtils = require('../../common/event-utils');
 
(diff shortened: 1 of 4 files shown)


src/components/ebay-menu-button/component.js
```diff
@@ -11,26 +10,22 @@ module.exports = {
...
+        const currentIndex = this.state.checkedItems.findIndex(checked => checked);
 
-        if (this.state.type === 'radio' && itemIndex !== currentIndex) {
-            this.state.items.forEach((eachItem, i) => {
-                eachItem.checked = i === itemIndex;
-            });
-
-            this.setStateDirty('items');
+        if (this.type === 'radio' && index !== currentIndex) {
+            this.setState('checkedItems', this.state.checkedItems.map((item, i) => i === index));
```
DylanPiercey: In Marko 4 I'd prefer to see `this.state.checkedItems = ` instead of `setState`.

src/components/ebay-menu-button/component.js
```diff
@@ -11,26 +10,22 @@ module.exports = {
...
             this.emitComponentEvent({
+                index,
                 eventType: 'change',
                 el: itemEl
             });
-        } else if (this.state.type !== 'radio') {
-            item.checked = !item.checked;
-            this.setStateDirty('items');
+        } else if (this.type !== 'radio') {
+            this.state.checkedItems[index] = !this.state.checkedItems[index];
```
DylanPiercey: Thoughts on renaming the state variable to `isChecked`? `checkedItems` seems a little off to me because it is not an array of checked items, it is an array of booleans which map to if an item is checked.

src/components/ebay-menu-button/index.marko
```diff
@@ -90,8 +90,8 @@ $ var isOverflowVariant = state.variant === "overflow";
         onMenu-change("handleMenuChange")
         onMenu-select("handleMenuSelect")
         key="content">
-        <for|item| of=state.items>
-            <ebay-menu-item(item)/>
+        <for|item,index| of=input.items>
+            <ebay-menu-item  ...item checked=state.checkedItems[index] />
```
DylanPiercey: Internally we should change this to `<@item>` to avoid work when we deprecate the nested tag transform.

src/components/ebay-menu-button/component.js
agliga: Done but for type as `radio`, not `type !== radio`

PR #977, opened 2019-12-08
Marko 4 refactoring

DylanPiercey:
DO NOT SQUASH.

## Description
Misc refactoring and improvements to modernize the components for Marko 4 and switching to best practices.

# Major Goals
1. Separate input and state. (Since these were merged in Marko 3 some components had them quite intertwined which was hard to follow).
2. Reduce the number of states, prefer computing values than storing computed data in state.
3. Reduce manual dom manipulation or reading from the dom.
4. Increase consistency around method/variable naming.
5. Some components did not remove event handlers or dom changes added by makeup during the lifecycle. Those are now fixed.
6. Improved tag definitions. Now all of the tag defs reference the `html-attributes` attribute group which pulls in all of the cross functional attributes for html tags, eg `id`, `style`, `aria-x` etc. This vastly improves auto complete. For the most part, where possible, any pass through elements now copy the definitions https://github.com/marko-js/marko/blob/master/src/core-tags/html/marko.json, meaning that when a pass through is to an `<a>` the appropriate autocompletes appear. There is also a `never` type which is used to indicate that although it would be a valid `html-attribute` it is not possible to override (for example `@type` in the `ebay-checkbox` is `never` because it can't be overwritten). The `never` type also prevents autocompletions happening for those attributes. All in all the autocomplete should be much better.
7. Isolate tag definitions. All tag definitions are now moved into their respective folder using `marko-tag.json` and `tags-dir` in the root `marko.json`. I think this makes things much cleaner. All that remains in the root `marko.json` are the transformers which rename nested tags which we intend to migrate from.

### Affected Components

The `ebay-carousel` carousel will see similar changes in a future PR. The `ebay-menu` and `ebay-menu-button` have been tackled by @agliga, although I may in the future create an additional PR implementing some more things from the `Major Goals` section above.

`ebay-pill` has been intentionally skipped since it will be removed in a future version. Although `ebay-combobox-readonly` will also be removed, I had already refactored it before I realized 😋.

Besides those these components everything else has been refactored.

### Notes

This PR comes with a ton of changes. I have made very minimal changes to test (which all should be passing) for a couple components which I believe had bugs, or tests that were testing implementation details. I've also manually tested all of these changes.

That said, obviously with this much change, a good review and QA is necessary.

Looking forward to hearing all and any feedback from you all. 😄


src/components/components/ebay-tooltip-base/component.js
```diff
@@ -10,63 +10,68 @@ module.exports = {
...
 
-    onUpdate() {
-        this.onRenderLegacy({
-            firstRender: false
-        });
-    },
-
-    onRenderLegacy() {
-        const hostClass = `${this.state.type}__host`;
+    _setupMakeup() {
```
ianmcburnie: I like this setup/cleanup convention for Makeup!

DylanPiercey:
@agliga  Regarding the `_` props I did a once over and thought I got them all. I can do another pass but that won’t be until Thursday at the earliest.

Issue #956, opened 2019-11-13
Tracking ticket for divergences between 4.x and 5.x branches


ianmcburnie:
Let's use this ticket to track anything that has gone into 4.x branch *after* the Marko v4 migrations in 5.0.0 branch were started. So far we have:

* Infotip: location fixes - Storybook examples (#952) (possibly undone, intentionally, by #966)
* Change browser.json to require widget properly #958 (not required in Marko4)
* Upgrade to skin 9.5.0
* Infotip: fixed icon alignment issues #966
* Upgraded devDeps
* Docs: added deprecation guidance #352
* Add click and escape key event handling; (#973)
* Combobox: allow custom ID (#972)

ianmcburnie:
This is an issue that Eamon can potentially help with. NOTE: 4.3.3 release from master will need merging into 4.5.0 before beginning this task. @seangates will take care of that as there were a couple of merge conflicts.

ETCaton:
> Change browser.json to require widget properly #958 (not required in Marko4)

Was this resolved by #959 @agliga?

ETCaton:
> Upgrade to skin 9.5.0

This seems to be the case already in 9.5.0 (introduced as part of #966)

agliga:
> > Change browser.json to require widget properly #958 (not required in Marko4)
> 
> Was this resolved by #959 @agliga?

This isn't needed for `5.0.0` since we have removed all `browser.json` dependencies except for external ones.

ETCaton:
That's what I saw, but I just wanted to make sure I was tracing the issue resolutions correctly; "#959 was the resolution to #958?" is what I meant to ask

agliga:
> That's what I saw, but I just wanted to make sure I was tracing the issue resolutions correctly; "#959 was the resolution to #958?" is what I meant to ask

Yes, one is the issue and the other is the pull request 👍

PR #981, opened 2019-12-10
Ported click/escape key event handling for expand-button

ETCaton:
## Description

Ported click/escape key event handling for expand-button (#970) upstream
## Context

Ported exact logic with the exception of using `marko-widgets`, keeping just escape key/click handler in the `component.js`.
## References

#956


ETCaton:
See #983

PR #982, opened 2019-12-10
fix: onClick/Keydown instead of w-*

ETCaton:
## Description

Replaced reference to `w-onclick` and `w-onkeydown` with the `onClick()`/`onKeydown()` functions
## References

#981


ETCaton:
See #983

PR #983, opened 2019-12-10
Ported click/escape key event handling for expand-button

ETCaton:
## Description

Ported click/escape key event handling for expand-button (#970) upstream
## Context

Ported exact logic with the exception of using `marko-widgets`, keeping just escape key/click handler in the `component.js`.
## References

#956


ETCaton:
Was unaware of the protection of `#.#.#` branches so I inadvertendly locked myself out.
#983 -> #982 -> #981 is the merge order, from then on I'll choose better branch names. I was avoiding the ticket number in the branch as it doesn't full address it.

agliga:
> Was unaware of the protection of `#.#.#` branches so I inadvertendly locked myself out.
> #983 -> #982 -> #981 is the merge order, from then on I'll choose better branch names. I was avoiding the ticket number in the branch as it doesn't full address it.

If it's okay, can you put all changes in this 1 PR. We can close the other PRs off. (Leave the branches for now, we can delete them)
This way it's easier to look at the whole set of changes and not have to resolve any merge conflicts after each merge. 

You can merge those branches locally and push all the commits here, We will squash them when we merge.

ETCaton:
This branch is an offshoot of those so it's up-to-date

marko.json
```diff
@@ -198,7 +198,7 @@
     "@a11y-close-text": "string"
   },
   "<ebay-expand-button>": {
-    "template": "./src/components/ebay-expand-button/index.marko",
+    "renderer": "./src/components/ebay-expand-button/index.js",
```
agliga: Revert this line

src/components/ebay-expand-button/component.js
```diff
@@ -0,0 +1,22 @@
+const assign = require('core-js-pure/features/object/assign');
+const eventUtils = require('../../common/event-utils');
+
+module.exports = {
+    onCreate(input) {
+        this.state = assign({}, input, {
```
agliga: Since this is for marko4, we should not assign input to state for that. 
Since we are not changing any state variables like disabled, this should not be needed. Instead we can use `this.input.disabled` in the `handleClick` and `handleKeydown`

src/components/ebay-expand-button/index.marko
```diff
@@ -23,7 +23,10 @@ $ var size = input.size || "medium";
         input.class
     ]
     type="button"
-    aria-disabled=(input.partiallyDisabled && "true")>
+    disabled=data.disabled
```
agliga: Should change to `input.disabled`

PR #984, opened 2019-12-11
Upstream combobox custom id

ETCaton:
## Description

Ported #972 to 5.0.0
#956


src/components/ebay-combobox/component.js
```diff
@@ -84,8 +90,14 @@ module.exports = {
...
-        this.lastValue = input.value;
-        this.state = { currentValue: input.value };
+        let inputId;
+        if (input['*']) {
+            this.lastValue = input['*'].value;
+            inputId = input['*'].id;
+        } else {
+            this.lastValue = input.value;
+        }
+        this.state = { currentValue: this.lastValue, id: inputId };
```
agliga: Where is `currentValue` used?

src/components/ebay-combobox/component.js
```diff
@@ -84,8 +90,14 @@ module.exports = {
     onInput(input) {
         input.autocomplete = input.autocomplete === 'list' ? 'list' : 'none';
         input.options = input.options || [];
-        this.lastValue = input.value;
-        this.state = { currentValue: input.value };
+        let inputId;
+        if (input['*']) {
```
agliga: In marko 4, we should have `input.value` and `input.id` instead of `input['*']` attributes.

src/components/ebay-combobox/index.marko
```diff
@@ -23,7 +23,12 @@ static var itemIgnoredAttributes = ["text"];
         respect the "off" key. Tested and works in all other browsers properly as well.
         -->
         <input
-            ...processHtmlAttributes(input, ignoredAttributes)
+            ...processHtmlAttributes(
+                Object.assign(data, {
+                    id: data.id || component.id + "-input"
```
agliga: Use `state` or `input`. Ideally input.
data should be undefined

src/components/ebay-combobox/component.js
ETCaton: In `_setSelectedText()`, `_emitComboboxEvent()`, and a couple other places.
```
❯ rg currentValue[^a-zA-Z] src/components/ebay-combobox/component.js
55:            this.state.currentValue = this.getEl('combobox').value;
80:        if (this.lastValue !== this.state.currentValue) {
81:            this.lastValue = this.state.currentValue;
100:        this.state = { currentValue: this.lastValue, id: inputId };
161:        if (this.state.currentValue !== text) {
163:            this.state.currentValue = input.value = text;
183:        const currentValueReg = safeRegex(this.state.currentValue);
193:            currentInputValue: this.state.currentValue,
```

src/components/ebay-combobox/component.js
ETCaton: What do you mean? I'm unfamiliar with the `['*']` syntax so I assume it meant something specific, since the referenced PR had `input['*'].id`. What's the difference between `input.id` and `input['*'].id`?

agliga: So earlier marko v3 versions would put any additional parameters in the "*" object (basically anything not defined in `marko.json`). Markov4 basically removes that and puts all additional properties into the base object itself. 
IE taking `<ebay-component something=1/>`
Marko 3:
input["*"].something === 1
Marko 4:
input.something === 1

However when doing spread it always puts it in the base object.
```
$ var attrs = {something: 1}
<ebay-component ...attrs />
```

`input.something === 1` always

(Just for completeness sake and in case you use it in your project.... 
In `marko.json` we have a "@*" object which when set would set the additional parameters to whatever `targetProperty` you define. When setting the `targetProperty: null` then it will put the items into the base object which is what we have defined in our `marko.json`)

ETCaton: OK, got it. So the resolution for these changes would be to not do the `input['*']` check, but just check for the property directly

src/components/ebay-combobox/component.js
agliga: Got it! Looks good

I would remove the id from state and just use the ID directly from input in the template.

src/components/ebay-combobox/index.marko
```diff
@@ -23,7 +23,12 @@ static var itemIgnoredAttributes = ["text"];
         respect the "off" key. Tested and works in all other browsers properly as well.
         -->
         <input
-            ...processHtmlAttributes(input, ignoredAttributes)
+            ...processHtmlAttributes(
+                Object.assign(data, {
+                    id: input.id || component.id + "-input"
```
agliga: Since we are doing marko4 syntax, I would change this to:
`input.id || component.getElId("input")`
See https://github.com/eBay/ebayui-core/blob/5.0.0/src/components/ebay-textbox/index.marko#L17

src/components/ebay-combobox/component.js
ETCaton: After changing the current code to 
```js
onInput(input) {
    input.autocomplete = input.autocomplete === 'list' ? 'list' : 'none';
    input.options = input.options || [];
    let inputId;
    if (input.id) {
        inputId = input.id;
    }
    this.lastValue = input.value;
    this.state = { currentValue: this.value, id: inputId };
}
```
I get a new failing test. My understanding is that this code should be functionally identical to existing code
```
1 failing

  1) src/components/ebay-combobox
       combobox
         renders with second item selected:
     Error: Found multiple elements with the role "option"

(If this is intentional, then use the `*AllBy*` variant of the query (like `queryAllByText`, `getAllByText`, or `findAllByText`)).
```

ETCaton: Resolved with this recommendation: https://github.com/eBay/ebayui-core/pull/984#discussion_r357281806

PR #986, opened 2019-12-13
added focus and blur events to switch component

shanimal:
## Description
Added focus and blur to switch component

## Context
n/a

## References
n/a

## Screenshots
n/a

### DIFF
--- a/src/components/ebay-switch/README.md
+++ b/src/components/ebay-switch/README.md
@@ -19,3 +19,5 @@ Name | Type | Stateful | Required | Description
 Event | Data | Description
 --- | --- | ---
 `switch-change` | `{ originalEvent, value, checked }` | selected value and checked status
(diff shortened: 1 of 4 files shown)


ianmcburnie:
Thanks for the PR @shanimal . Out of curiosity, are these blur and focus hooks something that your app actually needs? Or just something you think that might be needed?

shanimal:
Hi Ian, thanks for the quick response. Good question. Yes, we need this for user and funnel metrics within our product input form. From these we determine pain points for users and better understand where to focus (no pun intended) our energy.

shanimal:
@ianmcburnie After reviewing this for myself, you are right, we probably don't need this after-all. 

I can reopen if I need support for it.

Sorry.

ianmcburnie:
No problem!

PR #989, opened 2019-12-16
Combobox: Fix value change test;

### LINKED ISSUE #988: Combobox: change event is not fired when a new value is entered
# Bug Report

## Description

The current behavior to check for whether the combobox value changed is for the `newValue` (e.g. what is currently in the `value` property of the text input) is tested against ... `this.inputEl.value` ... which is the same thing. Hence, we're simply checking if the value equals itself.

https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-combobox/index.js#L129

## Workaround

None.

seangates:
## Description
- change test for text input and whether the change event should fire on blur

## Context
Presently the test tests whether the entered value equals ... the entered value.

## References
Fixes #988

### DIFF
--- a/src/components/ebay-combobox/index.js
+++ b/src/components/ebay-combobox/index.js
@@ -126,7 +126,7 @@ module.exports = require('marko-widgets').defineComponent({
         const newValue = this.inputEl.value;
 
         eventUtils.handleTextInput(originalEvent, () => {
-            this.valueChanged = this.inputEl.value !== newValue;
+            this.valueChanged = this.state.currentValue !== newValue;
 
             this.activeDescendant.reset();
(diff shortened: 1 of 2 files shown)


DylanPiercey:
I don’t think this issue exists in the v5 branch. Would you be able to put up a PR to that branch which brings these test changes over to verify?

seangates:
@DylanPiercey Sure. I can do the test over there, too.

ianmcburnie:
Let's track anything required in port in #956 Feel free to re-open that ticket.

seangates:
All: https://github.com/eBay/ebayui-core/pull/990

Issue #987, opened 2019-12-14
BUG: Single item carousel paddles sometimes become unresponsive


msendlakowski:
# Bug Report

## Description
In the carousel there appears to be a race condition.  Sometimes when you click on the paddles they become non-responsive.  We've only seen this on the Deals application with the single item carousel.  And the behavior isn't consistent.  Sometimes you have to click many times before it happens.

## Workaround
None

## Screenshots
n/a

PR #991, opened 2019-12-16
987 - Catch race condition in carousel rerender

msendlakowski:
## Description
Add a code branch for the carousel onRender to call emitUpdate so that the paddle click events are re-attached.

## Context
There is a race condition where the user could trigger a rerender by clicking a nav paddle, but the component calculates that the item position is correct by the time `getOffset` is called.  In this case it never calls `emitUpdate` so the click events on the paddles never get re-attached.  This fix handles that scenario.  

I check for `offset !== 0` because this change causes some of the unit tests to fail. In all these cases, the initial offset is 0. I'm not thrilled with this change to address the unit tests, but I couldn't figure out a way to skip this code branch during the unit tests, or why the tests were failing when `emitUpdate` was called.  But manual testing seems to work fine with this change.

## References
Fixes https://github.com/eBay/ebayui-core/issues/987

## Screenshots
no UI change

### DIFF
--- a/src/components/ebay-carousel/component.js
+++ b/src/components/ebay-carousel/component.js
@@ -110,10 +110,10 @@ function onRender() {
                     // Animate to the new scrolling position and emit update events afterward.
                     config.scrollTransitioning = true;
                     this.cancelScrollTransition = scrollTransition(listEl, offset, this.emitUpdate);
-                } else if (offset !== 0) {
+                } else {
                     // Race condition where user clicks the paddles quickly and by the time the offset is calculated
                     // its equal to the element position, so emitUpdate is never called
-                    this.emitUpdate();
+                    this.isMoving = false;
                 }
             }
         }
--- a/src/components/ebay-carousel/component.js
+++ b/src/components/ebay-carousel/component.js
@@ -110,10 +110,10 @@ function onRender() {
                     // Animate to the new scrolling position and emit update events afterward.
                     config.scrollTransitioning = true;
                     this.cancelScrollTransition = scrollTransition(listEl, offset, this.emitUpdate);
-                } else {
-                    // Race condition where user clicks the paddles quickly and by the time the offset is calculated
-                    // its equal to the element position, so emitUpdate is never called
-                    this.isMoving = false;
+                } else if (this.isMoving) {
+                    // Animate to the new scrolling position and emit update events afterward.
+                    config.scrollTransitioning = true;
+                    this.cancelScrollTransition = scrollTransition(listEl, getOffset(state), this.emitUpdate);
                 }
             }
         }
--- a/src/components/ebay-carousel/component.js
+++ b/src/components/ebay-carousel/component.js
@@ -110,6 +110,10 @@ function onRender() {
                     // Animate to the new scrolling position and emit update events afterward.
                     config.scrollTransitioning = true;
                     this.cancelScrollTransition = scrollTransition(listEl, offset, this.emitUpdate);
+                } else if (offset !== 0) {
+                    // Race condition where user clicks the paddles quickly and by the time the offset is calculated
+                    // its equal to the element position, so emitUpdate is never called
+                    this.emitUpdate();
                 }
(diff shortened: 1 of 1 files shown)


agliga:
@msendlakowski Is this needed for marko 3 or only post marko 4. Our `5.0.0` branch is our marko 4 refactor, so this will not be available for marko 3 builds. 
If you need it for marko 3 build, I would add them to the `4.5.0` or `master` branches

msendlakowski:
> @msendlakowski Is this needed for marko 3 or only post marko 4. Our `5.0.0` branch is our marko 4 refactor, so this will not be available for marko 3 builds.
> If you need it for marko 3 build, I would add them to the `4.5.0` or `master` branches

For our case, it's only needed for Marko 4.

msendlakowski:
> I think this change would cause unnecessary update events to fire on rerenders which didn't change anything. For example when the page resizes but the carousel is in the same state. Could you check if that is the case?

@DylanPiercey You're probably right. I'll check and see if I can nail down a more specific fix.

msendlakowski:
@DylanPiercey what do you think about this fix?  It's more specifically handling the problem, and it shouldn't cause any additional events to be called on render.  But if anyone adds something that should be undone in the following lines, they could get missed.

https://github.com/eBay/ebayui-core/blob/987---carousel-bugfix/src/components/ebay-carousel/component.js#L315-L323

DylanPiercey:
Sorry for the delay in review. I thought there was something else going on, but now after having a second look I this PR makes sense. I'm restarting the travis build to get the CI passing, and then this LGTM.

DylanPiercey:
Splitting hairs, but it might be slightly cleaner to replace https://github.com/eBay/ebayui-core/pull/991/files#diff-736301f5ed15565e7953ce5abe74f4aaR107-R117 with

```js
            } else if (this.isMoving) {
                // Animate to the new scrolling position and emit update events afterward.
                config.scrollTransitioning = true;
                this.cancelScrollTransition = scrollTransition(listEl, getOffset(state), this.emitUpdate);
            }
```

My understanding of the issue is that we want to handle the case where the user has clicked the controls but some how the carousel has already moved to that position and so it does not emit en event. That being the case, we can just use the above. In the above it would always try to scroll transition if the controls are clicked (`isMoving` is true then) and so it becomes impossible for `isMoving` to not be reset. It also avoids the issue I mentioned above since it would only try to scroll transition and emit the update once one of the controls has been clicked.

This is somewhat splitting hairs but I think the above more clearly represents the intent. I do think we could get this through as is though.

DylanPiercey:
I have a hunch that we could be hitting this race condition as a result of https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-carousel/index.js#L265. It could be that that carousel is being rerendered mid scroll, which would reset the scroll transition. Then when the scroll transition kicks back up after the render is done, it might have already reached the end and fail to emit an event.

Mind you I haven't been able to reproduce this locally in storybook or the demo page, but that would make sense if the above is true.

ianmcburnie:
@DylanPiercey @msendlakowski What is the status? Can this be merged and included in our pre-release for QA?

msendlakowski:
@ianmcburnie Let me test the recommendation by @DylanPiercey and update this ticket

msendlakowski:
@DylanPiercey Your suggestion fixed the problem too, so I updated.  Ready for review.

ianmcburnie:
As the suggested fix worked, I'm going to merge this one in order to unblock pre-release and QA.

sairam0903:
Hi @ianmcburnie When are we targeting this to release?

msendlakowski:
@sairam0903 It also looks like this will be a major release, that may require some additional work on your part due to breaking changes.
@ianmcburnie Are you expecting breaking changes other than the default being ds6?

ianmcburnie:
Yes, we have removed background icons completely. We have also removed the old "small" buttons (local overrides may be needed to keep those working).

Issue #992, opened 2019-12-16
BUG : <ebay-checkbox/> is not providing given checkbox ID on markoV3, instead it provide generated widget id


udkp:
# Bug Report

## Description

ebay-checkbox is not providing given checkbox ID on markoV3, instead it provide generated widget id (w0-w0-0[0]). However, the label is providing given ID.

Because of this generated id, when we click the label, the checkbox is not getting checked.

version used : 
"@ebay/ebayui-core": "^4.4.3",
"@ebay/skin": "^9.4.0"

## Screenshots

<img width="485" alt="Screen Shot 2019-12-16 at 2 50 09 PM" src="https://user-images.githubusercontent.com/58959234/70950375-37282000-2015-11ea-858a-ef510d438ad0.png">

ianmcburnie:
Patched in 4.5.1. Thanks all.

PR #994, opened 2019-12-17
Added id on checkbox for marko 3

agliga:
## Description
Added support for id on checkbox

## Context
Added similar change to how combobox had. 

## References
#992

### DIFF
--- a/src/components/ebay-checkbox/template.marko
+++ b/src/components/ebay-checkbox/template.marko
@@ -10,6 +10,7 @@
         w-onchange="handleChange"
         w-onfocus="handleFocus"
         disabled=data.disabled
+        id=(data.id || widget.id + "-input")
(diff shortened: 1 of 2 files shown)


src/components/ebay-checkbox/template.marko
```diff
@@ -10,6 +10,7 @@
         w-onchange="handleChange"
         w-onfocus="handleFocus"
         disabled=data.disabled
+        id=(data.id || widget.id + "-input")
```
DylanPiercey: The idiomatic way to do this in Marko 3 would be `data.id || widget.elId("input")`. Otherwise this LGTM.

agliga: I coped this from the combobox PR. Either way, not going to port this to marko 4 so I'll leave it like this for now
But for future ones, I'll do it like this 👍

PR #996, opened 2019-12-18
Adding mousedown handler for menu-button

phoenix727:
## Description
This change causes a mousedown event to be emitted when clicking on ebay-menu-button items.

## Context
While developing a component that requires the use of the mousedown event, the ebay-menu-button was not emitting the appropriate event. This allows completion of the component.


ianmcburnie:
@phoenix727 Can you please shed a little more light on why the mouse down event is needed? Is it for tracking? This kind of usage data would help us, thanks.

phoenix727:
@ianmcburnie Sure. Working on a rich text editor. When text is selected and the `ebay-menu-button` is used, focus is lost on the selection before we are able to perform the correct action via `execCommand`. By using the mousedown instead of the click, we are able to perform actions correctly, since focus is still on the text at the time of the mousedown event. We have other controls that use a similar mechanic, but don't use the `ebay-menu-button`, and since it is via native html rather than `ebayui-core` components, the mousedown event is available in those cases.

ianmcburnie:
Thanks for explaining! Interesting case :-)

You just need to bring you branch up to date with master and we can merge for a patch.

Oh, and do you need to `keydown` too?

phoenix727:
With our usecase, keydown isn't useful since using the keyboard to navigate to the actions removes focus from the selection. That being said, I would fully support all the components in the `ebayui-core` library emitting native html events in addition to the custom ones provided. If that is in the roadmap, I'll offer my help to that upgrade.

ianmcburnie:
@phoenix727 The branch still needs bringing upto date

Also need one more approval from @seangates or @agliga

Issue #1012, opened 2020-01-08
Remove ebay-pill


ianmcburnie:
Pill has been superseded by filter-button.

The pill module was deprecated in v4.  It has been removed from Skin v10. Time for us to remove it from coreui.

Issue #1007, opened 2020-01-06
Upgrade to Skin v10.0.0


ianmcburnie:
Skin v10.0.0 removes all support for background icons.

Issue #897, opened 2019-10-02
ebay-icon: remove background icon support


ianmcburnie:
As per the title, set inline as default (instead of background) and rename to "foreground". This will be a breaking change.

We may even want to remove background icons entirely from coreui (but keep them in skin).

PR #1014, opened 2020-01-13
Update to support skin 10.0.0

agliga:
DO NOT SQUASH.

## Description
Multiple changes in order to fix on ebayui to work with the latest skin 10.0.0 which removed background icons
Also fixed a minor regression issue for tooltip host. 

## Context
* Removed pill (I had to because otherwise I could not push since there were test breakages)
* Removed `type=inline` code from `<ebay-icon>`. Also fixed tests and removed test cases for that.
* Changed all class background icons to use `<ebay-icon>`
* Removed lasso references to foreground icons. Just used normal icons
* Added several new icons using the script (not all, I think we should rerun the script to get all new icons though. Some will be removed). These are mostly the alias ones
* Fixed multiple tests to look for foreground icons instead of background ones.
* Added host to tooltip marko.json. Was missing in a refactor and added it back. 

## References
#1007 
#1012
#897


ianmcburnie:
This too: 

Screen Shot 2020-01-14 at 9 05 44 AM

agliga:
> @agliga The listbox button tick mark doesn't seem to get updated:
> 
> Screen Shot 2020-01-14 at 9 19 18 AM

Done, made tick to always show which works fine with skin 10.

Issue #1013, opened 2020-01-13
Pluralize ebay-breadcrumb & ebay-tab module names


ianmcburnie:
We can hopefully leave in backwards compatible support for the old names and update all references

PR #1022, opened 2020-01-14
Added migrator for ebay-breadcrumbs and ebay-tabs

agliga:
## Description
Added migrators for ebay-breadcrumb and ebay-tab to rename them to ebay-breadcrumbs and ebay-tabs.
This allows it to be backwards compatible but all new code goes through the plural version. 
I also duplicated the marko.json for the non plural versions.

## Context
I left marko.json in the old directory. I could have added a new entry into the marko.json, but this keeps it visible and we can always remove it at a later time.
Added tests to check migration process (I had to add eslint ignore max-len because I could not split it up on multiple lines. It would break if I used backticks.... go figure)

## References
#1013


src/components/ebay-breadcrumb/index.marko
```diff
@@ -1,35 +1 @@
...
-                    <${item.renderBody} />
-                </>
-                <if (!isLast)>
-                    <ebay-icon name="breadcrumb" />
-                </if>
-            </li>
-        </for>
-    </ul>
-</nav>
+<div></div>
```
DylanPiercey: Why is this now just a `<div></div>`?

src/components/ebay-breadcrumb/marko-tag.json
```diff
@@ -5,37 +5,31 @@
     "type": "expression"
   },
   "@html-attributes": "expression",
-  "@a11y-heading-text": "string",
-  "@a11y-heading-tag": "string",
-  "@hijax": "boolean",
-  "@aria-labelledby": "never",
-  "@role": "never",
-  "@items <item>[]": {
+  "@index": "string",
```
DylanPiercey: Why has this tag def changed so much?

src/components/ebay-breadcrumbs/examples/05-hijax/template.marko
```diff
@@ -1,4 +1,4 @@
-<ebay-breadcrumb a11y-heading-text="Page navigation">
+<ebay-breadcrumb on-breadcrumb-select(() => { console.log('oj')}) hijax class="breadcrumb--hijax">
```
DylanPiercey: Not sure this is a great example.

src/components/ebay-tab/index.marko
```diff
@@ -1,93 +1,2 @@
...
-                    <div class="tabs__cell">
-                        <div>
-                            <${panel.renderBody}/>
-                        </div>
-                    </div>
-                </div>
-            </for>
-        </div>
-    </else>
+<div>
```
DylanPiercey: Why is this just a `<div>`? (We should be able to just delete this file and remove the template from the tag def).

src/components/ebay-tab/migrator.js
```diff
@@ -0,0 +1,20 @@
+module.exports = function migrator(el, context) {
+    context.deprecate('<ebay-tab> has been renamed to <ebay-tabs>');
+    const oldAttribute = 'on-tab-select';
+    const newAttribute = 'on-tabs-select';
+    el.setTagName('ebay-tabs');
+    if (el.hasAttribute(oldAttribute)) {
+        const attribute = el.getAttribute(oldAttribute);
+        el.removeAttribute(oldAttribute);
+        attribute.name = newAttribute;
```
DylanPiercey: You shouldn't need to remove then add the attribute, mutating the `.name` should suffice.

src/components/ebay-tab/migrator.js
```diff
@@ -0,0 +1,20 @@
...
+        el.addAttribute(attribute);
+    }
+
+    el.forEachChild(child => {
+        if (child.tagName === 'ebay-tab-heading') {
```
DylanPiercey: This is going to cause issues if either of these are in a `<for>` or `<if>`. We should use traverse here.

src/components/ebay-tabs/component.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
             const keyCode = event.charCode || event.keyCode;
             const direction = keyCode === 37 || keyCode === 38 ? -1 : 1;
             const index = (state.index + len + direction) % len;
-            this.getEl(`tab-${index}`).focus();
+            this.getEl(`tabs-${index}`).focus();
```
DylanPiercey: I feel like renaming the `key` here is not necessary.

src/components/ebay-breadcrumbs/examples/05-hijax/template.marko
agliga: Will remove, was a test.

src/components/ebay-tab/index.marko
agliga: I had issues when there was no index.marko file together with a transformer. I can try again.

src/components/ebay-tabs/component.js
agliga: The key has changed too from tab to tabs. 
As per Ian, we should change names to be consistent.

src/components/ebay-tab/index.marko
agliga: When I remove the empty div I get this error: `Error: Invalid tag file: ebayui-core/src/components/ebay-breadcrumb/marko-tag.json. Neither a renderer or a template was found for tag.`
We do the same for the tags repo. If you have a better suggestion I can change it but I will leave it in for now.

src/components/ebay-breadcrumb/marko-tag.json
agliga: Good catch, I had accidentally copied over the tabs tag instead of the beadcrumb one.

src/components/ebay-breadcrumb/index.marko
agliga: See comment on tab

src/components/ebay-tab/migrator.js
agliga: Done

src/components/ebay-tab/index.marko
agliga: I updated to what you recommended. Now it uses the global marko.json

Issue #1026, opened 2020-01-16
combobox bug: change event not triggered when user types into combobox


fraglord94:
## Description
Type something into the combobox and mouse click outside the combobox. `this.valueChanged` remains `false` and `handleComboboxBlur` does not trigger `emitComboboxEvent` with `change` 

Bug occurs when user types quickly into the combobox

#1026 # Bug Report

## eBayUI Version: 4.5.2

agliga:
Released in `@ebay/ebayui-core@4.5.6`

PR #1032, opened 2020-01-17
Changed logic behind firing change events

agliga:
## Description
Combobox has a keyup event where it basically takes the current input value. If it’s changed it marks it as changed from state and updates it on blur.
Now if I press two keys like: A and then N, but I let go of A first while still holding down N, it sets the change flag to true for “AN”, but once I let go of N, it checks again and says “AN” is the same so it sets the flag to false. Then it won’t trigger a change event. This happens when you type fast and press two in sequence but let go of them at the same time.

## Context
I changed the change event to only fire the following:
A changeValue is set from initial state. If the currentState's value is not equal to change value, OR if state is not equal to input value then trigger the change event. Then set changeValue with the new value. This way the change value will trigger always when there is any change.

## References
#1026


DylanPiercey:
Will a similar change be needed in v5?

agliga:
> Will a similar change be needed in v5?

No it looks like you already did the change in v5 during the refactor (and you used `lastValue` too 👍)
https://github.com/eBay/ebayui-core/blob/5.0.0/src/components/ebay-combobox/component.js#L80

DylanPiercey:
At least I am consistent with myself 😛

Issue #1021, opened 2020-01-14
Carousel: add a peek to all carousels


ianmcburnie:
We recently made changes to the carousel, so that when a single item is presented, there is no peek. In that case we then had to permanently show paddles as affordance. However, after chatting with design, the direction is clear that they want to show a peek, not paddles, in that case.

May require some skin work too.

<img width="408" alt="Screen Shot 2020-01-14 at 2 17 46 PM" src="https://user-images.githubusercontent.com/38065/72387593-b7a37480-36d8-11ea-92d6-1702cd0ffd57.png">

ianmcburnie:
@msendlakowski We are considering making this update in 5.1.0. I know that you have a carousel like this in production that would be affected. Your carousel currently shows 1 item, no peek and permanent paddles. After update it will show 1 item, with peek and no paddles. Are you okay with this going into a minor release?

msendlakowski:
Yeah, we should be good with that.  We are using version 4 for now.  We'll let our PM and design team know that we are frozen there, until they are ready to update the UI to the latest design system guidance.

PR #1041, opened 2020-01-29
Updated carousel with fade, peek, and, no dots 

agliga:
## Description
Added several carousel changes, added fade, removed dots, and added always peek on items-per-slide
This is the ebay-ui portion, see the skin for the fade 

## Context
* Added fade class only when `items-per-slide` is not set or if `autoplay` is not set or if the carousel is not scrolled to end.
* Forced `items-per-slide` peek value to be set at `0.1` if it is not set. 
* Removed all dots and accessitility text associated with it. Also removed all dots tests

## References
#1021
eBay/skin#988
PR with screenshots https://github.com/eBay/skin/pull/1002


ianmcburnie:
Not sure why the a11y status text is now appearing/visible?

Screen Shot 2020-01-30 at 11 25 15 AM

agliga:
@imcburnie it seems to be a skin issue. Its not clipped anymore but the `clipped` class is there

ianmcburnie:
Okay. I believe the `clipped` class comes from the skin utility module. The utility module is a dependency of the core module. The core module is required to be added by all pages.

agliga:
> Okay. I believe the `clipped` class comes from the skin utility module. The utility module is a dependency of the core module. The core module is required to be added by all pages.

Fix for that is in #1043

Issue #1008, opened 2020-01-06
Menu: add support for dividers/groups


ianmcburnie:
More info to follow...

Ignore the weird scrollbar in this mockup:

![Screen Shot 2020-01-06 at 12 14 55 PM](https://user-images.githubusercontent.com/38065/71845922-37e52c80-307e-11ea-85b0-b7489424b088.png)

Here's an example from the MIND patterns:

![Screen Shot 2020-01-06 at 12 16 55 PM](https://user-images.githubusercontent.com/38065/71846060-85fa3000-307e-11ea-9919-29fe9077b0cd.png)

ianmcburnie:
A really quick way of doing this would be to just allow horizontal rule elements as part of the data (with role=separator), but the full, correct way would be to allow separate & distinct ARIA menus.

We have a deadline of early Feb to deliver either one of these solutions.

ianmcburnie:
Merged to 5.1.0

PR #1057, opened 2020-02-11
Added separator for menu

agliga:
DO NOT SQUASH.

## Description
Added a global transform used for both menu and menu-button. 
This allows us to intersperse the separators in the code.
The separator transforms into a menu-item with isSeparator=true attribute .

## Context
* I had to do it this way because separator would lose context if it was a separate tag
* I added some tests to verify that they are working fine.
* Menu button remains unchanged because all the logic is handled in menu3

## References
#1008 

## Screenshots
Screen Shot 2020-02-10 at 4 27 12 PM


agliga:
> Can you update the readme and add a quick example please?

Done

agliga:
Changed base. Some other commits from master are here too though.

ianmcburnie:
Okay, I guess a merge commit should be fine as we need those commits ported over to 5.1.0 branch anyway.

Issue #1059, opened 2020-02-12
Combobox: Typing in the combobox that doesn't have options throws errors


melquan:
# Bug Report

## Description
When typing in the textbox of the combobox that doesn't have options, errors are thrown in the console, due to `this.activeDescendant` being `undefined`.

## Workaround
None

## Screenshots
```
index.js:129 Uncaught TypeError: Cannot read property 'reset' of undefined
    at index.js:129
    at handleNotKeydown (index.js:19)
    at Object.handleTextInput (index.js:64)
    at Widget.handleComboboxKeyUp (index.js:128)
    at event-delegation.js:68
    at Object.batchUpdate (update-manager.js:98)
    at HTMLBodyElement.<anonymous> (event-delegation.js:27)
```

There may be more places that might be missing null checks for `this.activeDescendant` and `this.expander` before calling functions on them.

agliga:
Published `@ebay/ebayui-core@4.5.9` which should have this fix.

PR #760, opened 2019-07-16
Combobox: Fix double selection of options;

seangates:
## Description
- adds usage of `activeDescendant.reset()`
- removes `autoInit` logic and sets it to `-1` permanently

## Context
When users would select an item from the list then open the list again it would select the first item (always) as well as the other item they selected. In certain conditions more than two could be selected.

## References
Fixes #633

PR #1061, opened 2020-02-13
Fixed error when combobox has no options

agliga:
## Description
Added a check to see if `this.activeDescendant`. Also added a test to verify that empty combobox events work. (Test failed without this change)

## References
#1059

### DIFF
--- a/src/components/ebay-combobox/index.js
+++ b/src/components/ebay-combobox/index.js
@@ -130,7 +130,9 @@ module.exports = require('marko-widgets').defineComponent({
         const newValue = this.inputEl.value;
 
         eventUtils.handleTextInput(originalEvent, () => {
-            this.activeDescendant.reset();
+            if (this.activeDescendant) {
+                this.activeDescendant.reset();
+            }
             this.inputEl.value = newValue;
             this.setState('currentValue', newValue);
             this.setSelectedIndex();
--- a/src/components/ebay-combobox/test/test.browser.js
+++ b/src/components/ebay-combobox/test/test.browser.js
@@ -166,6 +166,28 @@ describe('given the combobox starts with zero options', () => {
         });
     });
 
+    describe('when the input receives keyup with no options', () => {
+        beforeEach(async() => {
+            await pressKey(component.getByRole('combobox'), {
+                key: 'A',
+                keyCode: 65
+            });
+        });
+
+        it('then it should emit a input event', () => {
+            expect(component.emitted('combobox-input')).has.length(1);
+        });
+        describe('when blur happens on the combobox', () => {
+            beforeEach(async() => {
+                await fireEvent.blur(component.getByRole('combobox'));
+            });
+
+            it('then it should emit a change event', () => {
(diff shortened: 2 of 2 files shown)


src/components/ebay-combobox/index.js
```diff
@@ -130,7 +130,9 @@ module.exports = require('marko-widgets').defineComponent({
         const newValue = this.inputEl.value;
 
         eventUtils.handleTextInput(originalEvent, () => {
-            this.activeDescendant.reset();
+            if (this.activeDescendant) {
+                this.activeDescendant.reset();
```
ianmcburnie: Hmm, by resetting every time a key is pressed, might this cause the active-descendant to be removed in some cases that we dont want? For example, in the "basic filtering" story:

1. focus on the combobox
1. type "b"
1. press down two times to highlight "Basic Offer 1"
1. type "a"

Does active-descendant still remain highlighted on "Basic Offer 1" or does it dissapear? Sorry I can't test it out myself locally right now.

agliga: Ian, this was added because of this PR https://github.com/eBay/ebayui-core/pull/760
If we remove it, then it will cause the double selection when pressing the down arrow.
At the moment, yes the cursor disappears and then the user has to press down arrow again.

ianmcburnie: I'm not talking about removing the autoReset entirely. Just from doing it every time a key is pressed. Although, having said that, the WCAG example does reset it after every key press, but to me it makes sense to leave the highlighted item in place if a keypress did not trigger any filtering (although how often this scenario is likely to occur is another question - it might be rare especially for long lists). I'll do some testing with screen readers tomorrow to see how this change behaves.

https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html

agliga: Overall that could be a good consideration for another PR or another issue which we can optimize the reset functionality, since this code was introduced in that previous PR to prevent the double highlight. However, this PR was just to fix the no options case, which basically takes the resetting of the active descendant and does it only when it is not null. 

One other thing is the issue seems that the highlight stays if we don't call reset, but when you press down, the first item gets highlighted. So ideally we would have to change makeup to support not resetting the cursor position when the input is changed, and then we can remove the reset. 

Here is an example if I completely remove the reset code. I basically type in the dropdown, select an item with my arrow keys and then type another letter which shouldn't force any  additional filtering (I didn't use the filter one in this example, but it also does the same thing),  and then press the down arrow. 
Here is a gif of that 
combobox-reset

ianmcburnie: Ok, thanks Andew. It sounds like resetting is the best way forward for now then. I also notice a couple of probably related bugs in the MIND pattern example, so it's probably worth me revisiting the entire behaviour and seeing if there's an underlying bug in makeup when I get the chance. For now resetting after every input keeps it simple and predictable.

PR #1064, opened 2020-02-18
Added size modifiers for checkbox and radio

agliga:
## Description
Added new icons for checkbox and radio
Added size modifier to use those icons. Wrapped it all in an if statement since it looks like we can't modify the name directly
Note. This only changes the icons, in mweb/dweb the size is always at 24/18 respectively because of a media query. We would need to also add a large class modifier at a later time.


DylanPiercey:
@agliga is there an actual visual change in the icons or is it literally just the size? If it is just the size why don't we just scale the SVG? As it stands this component now includes 4 icons instead of the two.

agliga:
> @agliga is there an actual visual change in the icons or is it literally just the size? If it is just the size why don't we just scale the SVG? As it stands this component now includes 4 icons instead of the two.

There are different SVGs provided by design that we need to use for regular/large checkbox and radio.

ianmcburnie:
> @agliga is there an actual visual change in the icons or is it literally just the size? If it is just the size why don't we just scale the SVG? As it stands this component now includes 4 icons instead of the two.

There should only ever be two icons rendered to the page though. For large they will have the `-large` postfix. And yeah, we can't just simply scale the SVGs in this case. I know, feels like we have all been fed a lie about SVG! ;-)

DylanPiercey:
@ianmcburnie it's unfortunate that scaling won't work. Why is that?

Also in this case all 4 icons always need to be sent to the page as hidden symbols (only one is event stamped out though). This is because there is no way to statically determine which icons are going to be rendered conditionally within the Marko template.

ianmcburnie:
@DylanPiercey If you look closely, the icons are actually slightly different (i.e chunkier or thinner). It's less pronounced than it was in the previous versions.

DylanPiercey:
@ianmcburnie would it be possible to just override the `stroke-width` or similar with css to achieve that?

ianmcburnie:
Hmm, it could probably be achieved in the particular case of these checkboxes (also assuming that our designers are comfortable doing it that way). However, I don't think it will always be achievable for every icon (I'd have to look through them all and see), so as a concept we probably do need to support different "sizes" of icons. We should mention it to them though, as I agree it would simplify things.

agliga:
@DylanPiercey For the multiple icons we can address them in another PR with designs help. If this looks fine though, I will merge it.

ianmcburnie:
Here's a smattering of different icons, I don't think they could all be differentiated by stroke alone, or any other combination of svg attributes:

Screen Shot 2020-02-18 at 12 12 47 PM

Issue #1058, opened 2020-02-11
ebay-menu-button not collapsing for a single select


sharma46bhawana:
## Description
Using the `<ebay-menu-button>` and `<ebay-menu>` with a single select is not collapsing the menu after an option is selected. It was a feature available in 2.1.0

## Workaround
Tried the workaround of setting the state expanded = true, but it isn't working.

ianmcburnie:
Let's do this for menuitem and radiomenuitem only. I think checboxmenuitem may need to stay open for multi-selectionl

agliga:
Released for on `@ebay/ebayui-core@4.5.11`

Keeping issue open since we need to fix it on `5.1.0` as well

PR #1065, opened 2020-02-18
Added an option collapseOnSelect to close menu on option selection

agliga:
## Description
Added an option to auto collapse on click. I added an option because some teams might not want to automatically collapse since it does not always reflect your changes (and I think some teams update their content after the menu is closed)
This also does not make it a breaking change. 

## References
#1058 

## Screenshots
menu-collapse

### DIFF
--- a/src/components/ebay-menu-button/README.md
+++ b/src/components/ebay-menu-button/README.md
@@ -32,6 +32,7 @@ Name | Type | Stateful | Required | Description
 `checked` (radio) | Number | Yes | No | will set the corresponding index item to `checked` state and use the `aria-checked` attribute in markup
 `disabled` | Boolean | Yes | No | Will disable the entire dropdown (disables the ebay-button label) if set to true
 `variant` | String | No | No | will change the button style, "overflow" / "default"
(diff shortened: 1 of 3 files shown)


src/components/ebay-menu-button/README.md
```diff
@@ -32,6 +32,7 @@ Name | Type | Stateful | Required | Description
 `checked` (radio) | Number | Yes | No | will set the corresponding index item to `checked` state and use the `aria-checked` attribute in markup
 `disabled` | Boolean | Yes | No | Will disable the entire dropdown (disables the ebay-button label) if set to true
 `variant` | String | No | No | will change the button style, "overflow" / "default"
+`collapseOnSelect` | Boolean | Yes | No | Will collapse whole menu when an item is selected in menu. Only works when `type="radio"`
```
ianmcburnie: Nice. I like the fact it's behind a flag and backwards compatible. Given that fact, maybe we should open it up to all menuitem types, and not just radio?

agliga: I added it in two places because I want to maintain not collapsing on radio menu if the selected item is checked.

Issue #660, opened 2019-05-07
New Component: Details


szhou1:
A component version of the Skin details module: https://ebay.github.io/skin/#details

<img width="1106" alt="Screen Shot 2020-02-26 at 3 29 45 PM" src="https://user-images.githubusercontent.com/38065/75397867-d6d31b80-58ac-11ea-8886-73913647758e.png">

PR #1071, opened 2020-02-27
Added details component

agliga:
## Description
Added new details component based on skin markup

## Context
* Added two events for details open/close
* Added a size and type options. I could have done just  center=true and small=true, but who knows if we want to expand those at a later time and they go in line with other components
* Added open as a state variable, although the browser handles the open/close. 

## References
#660 

## Screenshots
Screen Shot 2020-02-27 at 11 01 16 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Each layer does its bit to enforce and enhance accessibility. We consider this l
 * [`ebay-checkbox`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-checkbox)
 * [`ebay-combobox`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-combobox)
 * [`ebay-cta-button`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-cta-button)
+* [`ebay-details`](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-details)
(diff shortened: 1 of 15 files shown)


ianmcburnie:
Here's a quick one, all the examples show "Detals" instead of "Details" :-)

DylanPiercey:
Shouldn't we have some browser tests as well to check the toggle functionality/events?

agliga:
> Shouldn't we have some browser tests as well to check the toggle functionality/events?

I was thinking about that, however that should be handled by marko correct? All it's doing is just emitting an event on open/close which should be handled by marko, so no reason to test that correct?
For open/close, it's actually mostly using native browser implementation. The open/close state is mostly for the initial load (and toggling it is just so it doesn't always stay open when clicked)

DylanPiercey:
@agliga part of the public API for the component is the events it emits and so we still need to test that those are coming through when / how we expect.

Also I think we should not use `onClick` and instead `onToggle` for the details element (https://developer.mozilla.org/en-US/docs/Web/API/HTMLDetailsElement/toggle_event)

DylanPiercey:
`<details>` also has an `open` property with the `open` state. We could avoid using state here all together and just use that for emitting the right event.

agliga:
Added one more change to add client tests and to sync open property on client

@DylanPiercey Can you take another look 👍

agliga:
@DylanPiercey Removed all getByTestId and used closest instead.

PR #1072, opened 2020-02-28
Changed array findIndex to use core-js

agliga:
Changed `findIndex` to use corejs instead of Array.prototype which is undefined in IE11

~~Does `forEach` also need to be changed? I see it's being used as well and not sure if it exists in IE11~~


DylanPiercey:
IE11 supports `Array.prototype.forEach`, so that is fine. It just doesn't support it for `NodeList`'s.

Issue #1075, opened 2020-03-02
Fake Menu button is closed when clicking an item on firefox/safari before event is triggered


agliga:
# Bug Report

## Description
When clicking on a fake menu button using `ebay-menu-button`, if there is a `tabindex=-1` above the `ebay-menu-button` (like on a `role=main`), focus is triggered on the container above forcing the menu to close.
This only happens on safari and firefox

## Workaround
Removing `tabindex=-1` on container above.

PR #1076, opened 2020-03-02
Fixed menu button to not lose focus on firefox and safari

agliga:
## Description
By adding a tabindex to the container above where the buttons/anchor tags are, this will take focus and trigger focus on the element forcing it to not close. 
This is only needed for fake menu.

~~However this introduces another bug which basically will trigger focus on the container instead of the first element. We can fix this in makeup~~
Changed focusManagement on expander to `interactive` for fake menu in order to fix focus on first element. 

## References
#1075

### DIFF
--- a/src/components/ebay-menu-button/index.js
+++ b/src/components/ebay-menu-button/index.js
@@ -16,7 +16,7 @@ module.exports = require('marko-widgets').defineComponent({
         this.expander = new Expander(this.el, {
             hostSelector: '.menu-button__button, .fake-menu-button__button',
             contentSelector: '.menu-button__menu, .fake-menu-button__menu',
-            focusManagement: 'focusable',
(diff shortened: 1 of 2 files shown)


ianmcburnie:
See if you can use the `keyboardOnly` option in makeup, which will ignore programatically focusable elements:

https://github.com/makeup-js/makeup-focusables/pull/1#issuecomment-593511486

agliga:
> See if you can use the `keyboardOnly` option in makeup, which will ignore programatically focusable elements:
> 
> [makeup-js/makeup-focusables#1 (comment)](https://github.com/makeup-js/makeup-focusables/pull/1#issuecomment-593511486)

@ianmcburnie Done, changed focus management on expander. Works fine

agliga:
> Could you please also document this somewhere in the code. A one liner or link to PR/issue is fine.

Done, can you approve and I will push out a patch

Issue #1077, opened 2020-03-02
Pagination component not triggering events in IE11


agliga:
# Bug Report

## Description
In IE 11, the click events are not triggered when pressing the next/prev/number buttons
This is due to the logic here https://github.com/eBay/ebayui-core/blob/4.5.0/src/components/ebay-pagination/index.js#L215
Detail is always 0 because of the implementation in IE https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail

## Workaround
None

PR #1078, opened 2020-03-02
Removed synthetic click and just use browser click

agliga:
## Description
* Removed keyup events since they are being duplicated in click events and not needed
* Removed the isSyntheticClick method
* Removed all tests for keyup as well
This is in line with `5.x` branch which also has the same refactor

## References
For #1077


agliga:
> Not sure why this synthetic click was ever needed, but good to be consistent now with v5. Are you testing link-based pagination and button-based?

Yes I checked that both trigger the events as expected

PR #1093, opened 2020-04-07
Updated dialog for support with new skin changes

agliga:
## Description
Changed dialog to consume the new skin changes

## Context
Most of it is a syntax change. If the new markup is not there it works as before.
* Added slide for full screen dialog animation
* Added `<@header>` and `<@footer>` markup for header and footer

## References
https://github.com/eBay/skin/issues/994

### DIFF
--- a/src/components/ebay-dialog/examples/01-basic/template.marko
+++ b/src/components/ebay-dialog/examples/01-basic/template.marko
@@ -1,5 +1,5 @@
 <ebay-dialog a11y-close-text="Close Dialog" aria-labelledby="dialog-title" open=state.open on-dialog-close('closeDialog')>
-    <h2 id="dialog-title">Heading</h2>
+    <@header><h2 id="dialog-title">Heading</h2></@header>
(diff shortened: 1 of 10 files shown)


ianmcburnie:
@agliga Only thing I notice is this humongous focus indicator in Safari :-) Not sure if it's pre-existing or not.

Screen Shot 2020-04-07 at 9 22 23 PM

agliga:
> @agliga Only thing I notice is this humongous focus indicator in Safari :-) Not sure if it's pre-existing or not.
> 
> Screen Shot 2020-04-07 at 9 22 23 PM

So this happens when you use the old skin version. 
Make sure to link the `10.3.0` skin version locally (especially since the close button should be on the left)
The `10.3.0` version should look fine when linked up with this branch 👍 .

ianmcburnie:
Pretty sure I did link the 10.3.0 version. I'll double check..

agliga:
> Pretty sure I did link the 10.3.0 version. I'll double check..

This is what I get with `10.3.0` linked on Safari
Screen Shot 2020-04-08 at 10 54 34 AM

ianmcburnie:
Did you try the "Fill" example?

agliga:
Oh I think I see it now. Yes I used the fill and see what you mean.
So the case where you have the large focus indicator is the existing case if we don't use the new markup. I did not update the fill with the new markup to show that the old markup works. (and I think fill was an example with the close on right)
This is existing before this change. If I do it on `5.3.0` branch without this change it shows up like this:
Screen Shot 2020-04-08 at 10 58 21 AM

ianmcburnie:
Gotcha. Seems okay to me then. Thanks.

PR #1102, opened 2020-04-21
Split up dialog to use common component 

agliga:
Base changed to `5.4.0`
~~**DO NOT MERGE** Targetting `5.3.0` but will change target after release, doing this because I checked off branch of `5.3.0`. Will update once we release and cut off a new branch.~~

## Description
Split up dialog to use a common component. Added dialog-base, moved most of the dialog tests there. 

## Context
This was done in order for drawer (bottom-contaner) to be able to use the common dialog component. This way it only needs to change the base classes, and pass in the window classes it needs.


src/components/ebay-dialog/index.marko
```diff
@@ -1,48 +1,24 @@
...
 $ var isDocked = input.type === "left" || input.type === "right";
-<div
-    ...processHtmlAttributes(input, ignoredAttributes)
-    aria-modal="true"
-    role="dialog"
+<ebay-dialog-base
+    ...input
+    open=input.open
+    type="dialog"
+    on-modal-show('dialogShow')
```
DylanPiercey, line 10: When re-emitting event's you don't need a `component.js` handler. You can simply (in this case):

```
on-modal-show("emit", "dialog-show")
on-modal-close("emit", "dialog-close")
```

src/components/components/ebay-dialog-base/component.js
```diff
@@ -103,11 +104,11 @@ module.exports = {
 
                 if (isTrapped) {
                     focusEl.focus();
-                    this.emit('dialog-show');
+                    this.emit('modal-show');
```
ianmcburnie, line 107: So these event names are internal ones? Nitpicky, but in future we might have modal and non-modal dialogs. So the name may need to be changed. I guess if it is internal only, then there's no hurry do do that.

agliga: These are internal events only.

Issue #1107, opened 2020-04-22
New Component: Drawer (aka Bottom Container)


ianmcburnie:
Add the Skin Drawer module to CoreUI

ianmcburnie:
Merged into 5.4.0.

PR #1108, opened 2020-04-27
Added drawer component

agliga:
## Description
**NOTE** This uses all the code from https://github.com/eBay/ebayui-core/pull/1102 to make it work. That PR should merged first. 
**ANOTHER NOTE** There are two skin changes that need to be done to make all this work properly.
1. Need to add a fix to flex to allow drawer to be srollable
2. Need to add a change to hide close button if there is no `expanded` class there. 

## Context
Added new component with tests
* 4 events total, `drawer-close`, `drawer-show`, `drawer-expanded`, and `drawer-collapsed`. Added in README how each of these are basically triggered
* All events to drag on handle are touch events and should be tested with an touch emulator at least. 

For tests:
* Added tests to trigger touch drag. Added different tests to check for drag threshold and made sure that the given `drawer` events are not emitted. 
* Added events for scroll to check if expanded is triggered

## References
#1107 

## Screenshots
Screen Shot 2020-04-27 at 1 12 53 PM
Screen Shot 2020-04-27 at 1 12 43 PM
Screen Shot 2020-04-27 at 1 12 37 PM
Screen Shot 2020-04-27 at 1 12 33 PM


ianmcburnie:
Okay, I'll take a look at the toast one and see if we can get both in. Thanks.

Issue #1106, opened 2020-04-22
New Component: Toast


ianmcburnie:
Add the Skin Toast module to CoreUI.

ianmcburnie:
Merged into 5.4.0.

PR #1109, opened 2020-04-29
Added new toast component

agliga:
**Note** this uses changes from #1108 so please merge that one first

## Description
Added ebayui toast component. Added tests and used common component

## Context
Couple of issues 
* since this is a bit different, added some flags to common dialog component to handle different cases, like no body scroll or changing the base tag. 
* Had some issues with animations. Not sure if it was implementation on skin, but changing the dialog hidden attribute on after transition is finished worked and seems better that way. (Also fixed the tests that were failing to call wait first)

## References
#1106

## Screenshots
Screen Shot 2020-04-29 at 11 17 44 AM
Screen Shot 2020-04-29 at 11 17 50 AM


src/components/ebay-drawer/component.js
```diff
@@ -0,0 +1,69 @@
+const findIndex = require('core-js-pure/features/array/find-index');
```
DylanPiercey, line 1: Looks like this PR contains the `drawer` stuff as well? I'm assuming this shouldn't be the case.

agliga: Yes it is. This branch is cut off from the drawer branch because i needed some changes from it and it was not yet merged
Once the drawer is merged it should only have the toast changes.

agliga:
Addressed comments

PR is failing because `toast.browser.json` is not found. Once a skin pre-release is pushed out we can update and it should fix the CI

agliga:
Updated from `5.4.0`, this PR should only have the changes needed for toast now

src/components/components/ebay-dialog-base/component.js
```diff
@@ -12,6 +13,15 @@ module.exports = {
         this.emit('modal-scroll');
     },
 
+    handleKeydown(event) {
+        eventUtils.handleEscapeKeydown(event, () => {
+            // Do not close if the target is an input
```
ianmcburnie, line 18: Nice!

src/components/ebay-toast/index.marko
```diff
@@ -0,0 +1,18 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+<ebay-dialog-base
+    ...input
+    open=input.open
+    baseEl="aside"
+    transitionEl="root"
+    noTrap
```
ianmcburnie, line 8: Can we call this `modeless` instead? A modeless dialog will never trap keyboard or screenreader.

ianmcburnie: Or just `modal=false` (default: true)

src/components/ebay-toast/README.md
```diff
@@ -0,0 +1,22 @@
...
+</ebay-toast>
+```
+
+## Attributes
+
+Name | Type | Stateful | Required | Description
+--- | --- | --- | --- | ---
+`open` | Boolean | Yes | No | Whether toast is open.
+`focus` | String | No | No | An id for an element which will receive focus when the toast opens (defaults to close button).
+`a11y-close-text` | String | No | Yes | A11y text for close button and mask.
```
ianmcburnie, line 15: Remove "and mask". If that was copied over from dialog, it might need removing from there too.

src/components/ebay-toast/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Events
+
+Event | Data | Description
+--- | --- | ---
+`toast-show` |  | toast opened
```
ianmcburnie, line 21: Not sure if we are naming these events consistently elsewhere, but to me it should be show/hide or open/close. One the other and not mixed.

src/components/ebay-toast/examples/01-default/template.marko
```diff
@@ -0,0 +1,22 @@
+<ebay-toast a11y-close-text="Close Toast" aria-labelledby="toast-title" open=state.open on-toast-close('closeToast')>
+    <@header><h2 id="toast-title">Heading</h2></@header>
```
ianmcburnie, line 2: My concern with this syntax is that the developer has to plumb together the accessibility (i.e. the `aria-labelledby` and the `id`). I believe in some other cases we handle those intricacies behind the scenes...?

src/components/ebay-toast/README.md
agliga: On dialog the events are show and close, which is why I labeled them like that. 
We should change that in next refactor

src/components/ebay-toast/examples/01-default/template.marko
agliga: This is no different than the previous versions. In previous versions we had the title as the first element. 
This on `5.2.x` the version before we split up the header
https://github.com/eBay/ebayui-core/blob/v5.2.3/src/components/ebay-dialog/examples/01-basic/template.marko#L1-L2
Also this is in line with the dialog code. 
Again we should address this in refactors (have a title attribute ideally)

src/components/ebay-toast/README.md
ianmcburnie: Ok, I'll make a note of it, thanks.

src/components/ebay-toast/examples/01-default/template.marko
ianmcburnie: Gotcha. Thanks. Will also note this down for a revisit in v6.

src/components/components/ebay-dialog-base/component.js
```diff
@@ -53,10 +61,18 @@ module.exports = {
...
+            this.transitionEls = [this.rootEl];
+        } else {
+            this.transitionEls = [this.windowEl, this.rootEl];
+        }
         // Add an event listener to the dialog to fix an issue with Safari not recognizing it as a touch target.
         this.subscribeTo(this.rootEl).on('click', () => {});
 
+        // Added to document in case the object is not a modal type
+        // This is so we can escape even if focus is not present on popup
+        document.addEventListener('keydown', this.handleKeydown.bind(this));
```
DylanPiercey, line 74: Since we are finally using `marko@4` how do you feel about trying out the `<subscribe>` tag here? (https://github.com/marko-js/tags/tree/master/tags/subscribe).

```marko
<subscribe to=document on-keydown("handleKeydown")/>
```

I bring this up because what is currently in this PR can cause a bug since it is never cleaning up the event listener.

src/components/components/ebay-dialog-base/component.js
```diff
@@ -68,13 +84,31 @@ module.exports = {
         });
     },
 
+    triggerFocus(focusEl) {
```
DylanPiercey, line 87: I'd prefer these new methods use `_` to be consistent with what we've done for the other internal methods.

src/components/components/ebay-dialog-base/index.marko
```diff
@@ -1,20 +1,27 @@
...
     hidden:no-update=!state.open
+    aria-live=(!input.isModal && "polite")
     onClick("handleDialogClick")
     onMousedown("handleStartClick")>
     <div key="window" class=[`${input.classPrefix}__window`, input.windowClass]>
         <if(input.top)>
             <${input.top.renderBody}/>
         </if>
         <${input.header ? "div" : null} class=`${input.classPrefix}__header`>
+            <if(input.header && buttonPosition === 'right')>
```
DylanPiercey, line 22: This could be simplified to:

```marko
<if(buttonPosition === "right")>
  <${input.header}/>
</if>
```

src/components/components/ebay-dialog-base/index.marko
```diff
@@ -23,7 +30,7 @@ static var ignoredAttributes = ["open", "type", "classPrefix", "focus", "a11yClo
                 onClick("handleCloseButtonClick")>
                 <ebay-icon name="close"/>
             </button>
-            <if(input.header)>
+            <if(input.header && buttonPosition === 'left')>
```
DylanPiercey, line 33: Could be simplified to:

```marko
<if(buttonPosition === "left")>
  <${input.header}/>
</if>
```

src/components/components/ebay-dialog-base/index.marko
```diff
@@ -1,20 +1,30 @@
...
+<${baseEl}
     ...processHtmlAttributes(input, ignoredAttributes)
     aria-modal="true"
     role="dialog"
     class=[input.classPrefix, input.class]
     hidden:no-update=!state.open
+    aria-live=(!input.isModal && "polite")
     onClick("handleDialogClick")
     onMousedown("handleStartClick")>
+    <if (state.open)>
```
DylanPiercey, line 17: Can you remove this leading space on the `<if` for consistency?

Issue #1105, opened 2020-04-22
New Component: Wizard Stepper


ianmcburnie:
Add the skin wizard stepper to CoreUI

ianmcburnie:
Merged to 5.5.0. Closing.

PR #1120, opened 2020-05-12
Added wizard stepper component

agliga:
## Description
Added wizard stepper based on skin changes
Created it as a stateless component since the original design did not have any interactive html.

## Context
* Since the marko team doesn't like `component.js` I left it out since there is no client interactions. However, I added a helper since there is a javascript function I don't want to have in a template. 
* Added an interactive example using input to change the current item
* I went with `<@step>` tags since this is the future of dynamic tags. Also I added the current attribute to the `<@step>` tag since it makes more sense there than as a global attribute. I had initially added both but I thought for the first pass the current on `<@step>` works best.

## References
#1105 

## Screenshots
DWEB
Screen Shot 2020-05-12 at 3 29 46 PM
Screen Shot 2020-05-12 at 3 29 48 PM
Screen Shot 2020-05-12 at 3 29 52 PM

Screen Shot 2020-05-12 at 3 30 04 PM
MWEB
Screen Shot 2020-05-12 at 3 30 24 PM
Screen Shot 2020-05-12 at 3 30 29 PM


src/components/ebay-wizard-stepper/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Tag | Required | Description
+--- | --- | ---
+`<@step>` | Yes | Each step to represented. The body will display as text next to the current state
+
+## ebay-wizard-stepper Attributes
+
+Name | Type | Stateful | Required | Description
+--- | --- | --- | --- | ---
+`vertical` | Boolean | No | No | If true, then renders vertical version of stepper
```
ianmcburnie, line 24: `current` makes sense as a boolean, as it's an "off" or "on" state, but I personally would prefer an enum for layout. I'd like us to be more in line with flexbox and grid layout. So instead of `vertical`, I suggest `direction` with values of `column` (default) or `rows`. That keeps us forwards compatible for things like `column-reverse` or `row-reverse` should we ever need it.

src/components/ebay-wizard-stepper/examples/04-interactive/template.marko
```diff
@@ -0,0 +1,29 @@
...
+    }
+    next() {
+        if (this.state.current < this.items.length) {
+            this.state.current++;
+        }
+    }
+}
+
+<div>
+    <ebay-button on-button-click('prev')>Previous</ebay-button>
```
ianmcburnie, line 21: This example is nice. Storybook can also help us achieve something similar with "knobs", right? Mental note for us to look into that again at some point.

src/components/ebay-wizard-stepper/README.md
agliga: Should default be `column`? I assume the column is the top down one, is that considered the default?

ianmcburnie:
@agliga Heads up that I'm going to rename this before we go live with it: https://github.com/eBay/skin/issues/1104

Any preference for `ebay-progress-stepper` or `ebay-stepper`? I could go either way...

agliga:
> @agliga Heads up that I'm going to rename this before we go live with it: https://github.com/eBay/skin/issues/1104
> 
> 
> 
> Any preference for `ebay-progress-stepper` or `ebay-stepper`? I could go either way...
> 
> 

I already renamed it to `ebay-stepper` so i like that the most
For classes its all internal so if we need to refactor its not a big deal

ianmcburnie:
`ebay-stepper` it is :-)

Issue #1115, opened 2020-05-08
`listbox-change` is fired before updated happens in DOM


rajkumarsrk:
# Bug Report

## Description
Component: https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-listbox-button
Listening to `listbox-change` and doing form submit causes the newly selected values not getting updated. Can this event be triggered after it has updated selected filed in `select -> option` HTML

## Workaround
```
listbox.on('listbox-change', () => {
            setTimeout(() => {
                document.getElementById('formid').submit();
            },10);
        });
```

agliga:
After taking a look at this we decided not to do this fix. 

Here is Ians response from the pr

> Now that I've actually sat down and fully read through this, I see that @DylanPiercey is right. We should not be making this change. Also, from an a11y point of view, nine times out of ten, we shouldn't be submitting a form when a listbox has its value changed. A submit button should be used instead. There are some edge cases though where it *might* be appropriate to auto submit.
> 
> 
> 
> Another approach might be to add an 'autoSubmit' option, so that the internals are handled by us? Also happy to go down the `originalEvent` route if that works for @rajkumarsrk?

agliga:
This will be a breaking change to do this the right way. Will fix in next major version

ianmcburnie:
Done in 5.5.0

PR #1127, opened 2020-05-21
Added setTimeout to listbox-change so it will wait for form to be filled

agliga:
## Description
If `listbox-change` is triggered, then will need to wait till stack is cleared in order to allow forms to be submitted. Added a setTimeout to allow that to happen. 

## Context
Tests all passed with this change so it seems safe
Also if there is a better way to do this in more marko syntax, let me know. I couldn't find a better way.  

## References
#1115

### DIFF
--- a/src/components/ebay-listbox-button/component.js
+++ b/src/components/ebay-listbox-button/component.js
@@ -26,16 +26,18 @@ module.exports = {
         elementScroll.scroll(el);
         this.state.selectedIndex = selectedIndex;
 
-        this.emit('listbox-change', {
-            index: selectedIndex,
-            selected: [option.value],
-            el
-        });
-
         if (this.wasClicked) {
             this._expander.collapse();
             this.wasClicked = false;
         }
+
+        setTimeout(() => {
+            this.emit('listbox-change', {
+                index: selectedIndex,
+                selected: [option.value],
+                el
+            });
+        }, 0);
     },
 
     onCreate() {


DylanPiercey:
@agliga this would be a breaking change if teams are relying that the event is fired before the change is completed which is probably what they'd expect considering it's how the native select change event works. I don't think we should make this change, if a user wants to wait for the event to be over they should wait for that using setTimeout or similar. Why are we doing that?

Also for most events we pass through `originalEvent` so that a user can cancel the default browser behavior. I just realized we aren't do that here, do you know why?

agliga:
> @agliga this would be a breaking change if teams are relying that the event is fired before the change is completed which is probably what they'd expect considering it's how the native select change event works. I don't think we should make this change, if a user wants to wait for the event to be over they should wait for that using setTimeout or similar. Why are we doing that?
> 
> Also for most events we pass through `originalEvent` so that a user can cancel the default browser behavior. I just realized we aren't do that here, do you know why?

I agree in a sense, since we are not doing it anywhere else. If Ian is in agreement then we can close that issue. In the original issue that is what the team is doing but they didn't want to do it for some reason. I think part of it is because form is never updated and they have to do it that way.
I think we should be passing `originalEvent`. I will add an issue to add it.

DylanPiercey:
Regarding the original issue, if they were using a native `<select>` and subscribing to the `change` event they'd have the exact same issue. This is by design, `change` is meant to allow you to intercept the change before it happens.

ianmcburnie:
@rajkumarsrk thoughts?

rajkumarsrk:
Considering @DylanPiercey thoughts should we have `listbox-change` as its and have another event `listbox-changed` (post select event) ?

ianmcburnie:
Now that I've actually sat down and fully read through this, I see that @DylanPiercey is right. We should not be making this change. Also, from an a11y point of view, nine times out of ten, we shouldn't be submitting a form when a listbox has its value changed. A submit button should be used instead. There are some edge cases though where it *might* be appropriate to auto submit.

Another approach might be to add an 'autoSubmit' option, so that the internals are handled by us? Also happy to go down the `originalEvent` route if that works for @rajkumarsrk?

agliga:
Will close this pr. Lets move the discussion in the issue itself

Issue #1128, opened 2020-05-21
Add original event to listbox events


agliga:
#909 Bug Report

## Description
There is no originalEvent being passed in listbox events
Was found initially in this issue https://github.com/eBay/ebayui-core/pull/1127

## Workaround
None

agliga:
Wont fix, no reason to add original event to listbox change. It will be a misleading event because its not a change event but rather active descendent change event.

PR #1129, opened 2020-05-22
Added originalEvent object to listbox events

agliga:
## Description
As per the conversation in #1127 listbox has no originalEvents being passed through.
Updated those to be passed through and updated readme

## References
#1128


agliga:
As per discussion with Dylan, we have come up with another solution. The original event is not needed for the change event, since it is misleading. It isn't the select change event but the active-descendent change. Closing this for now

PR #1130, opened 2020-05-22
Added listbox-change even to fire only when marko updates

agliga:
## Description
Added `this.once('update')` to trigger change event on listbox.

## Context
Couple of issues with this new change
* This leads to an issue where if the user presses the currently selected item, it does not trigger a change event. This makes sense but can lead to a possible breaking change since we are changing how `listbox-change` event works.
* I added a check to see if the selected index did change and this way it makes it so the event only is triggered when the non selected item is clicked. It also doesn't collapse the menu since it is already selected. 

Another option would be to always trigger the change event and then maybe have another event on update which only triggers on changing. 

## References
#1115

### DIFF
--- a/src/components/ebay-listbox-button/component.js
+++ b/src/components/ebay-listbox-button/component.js
@@ -23,20 +23,22 @@ module.exports = {
         const el = this.getEls('option')[selectedIndex];
         const option = this.input.options[selectedIndex];
 
-        elementScroll.scroll(el);
-        this.state.selectedIndex = selectedIndex;
+        if (this.state.selectedIndex !== selectedIndex) {
+            elementScroll.scroll(el);
 
-        this.once('update', () => {
-            this.emit('listbox-change', {
-                index: selectedIndex,
-                selected: [option.value],
-                el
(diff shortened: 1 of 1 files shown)


ianmcburnie:
This change makes sense, `listbox-change` should only trigger when selected item changes. But shouldn't we wait until v6 to make this change if it "breaks" existing API?

DylanPiercey:
@ianmcburnie I think this is one of those blurry line breaking changes. Since the team has a work around though maybe it is best to save it for v6 as you say.

ianmcburnie:
Yeah... we could even say it was a bug. I'd be happy to go with that.

agliga:
Closing for now then.

ianmcburnie:
Wait, we don't want to merge this. Sorry if I wasn't clear, but I can see that an argument could be made that nobody should be depending on the change event when the selected item is clicked again. We could argue it was a bug that we are now fixing...

agliga:
Reopening as per Ian, will treat this as a bugfix

PR #1133, opened 2020-05-28
Added skin global and packages to load fonts for storybook

agliga:
## Description
The changes are in `.storybook/config.js` The other changes are running the storybook build and generated files.
Added a require for skin global and marketsans. This is because those are not being loaded by default.


.storybook/config.js
```diff
@@ -10,6 +10,12 @@ configure(() => {
     return result;
   }, {});
 
+  if (process.env.DS !== "4") {
+    require('@ebay/skin/marketsans');
+  }
+
+  require('@ebay/skin/global');
```
DylanPiercey, line 17: If ebayui doesn't ship with this, but the components need it, we should probably document that.

agliga: Will create an issue

PR #1135, opened 2020-06-02
Added migrator for all dialogs to remove header

agliga:
## Description
~~This might be a bigger change than for a patch. However, it should be safe (tests have all passed as before), and it does help going forward~~
~~Either way this should be part of the next release if not for this current one.~~ 
Changed to `5.5.0` branch

## Context
* Added a migrator for all dialog headers. Any `<@header>` that has a `<h1-6>` in it will migrate to add all properties to the `@header` tag. 
* I initially wanted to do this for toast only, but this will make all dialogs consistent. This way this fix won't need to be changed later. 
* This is related to the issue that toast's title is messed up in docs as well as forcing users to add toast__title to the header
* Also made `aria-labelledby` and `id` on header optional, since we will handle those using `:scope`
* I had wanted to put the migrator in `dialog-base` so all dialogs use it, but since we pass the `@header` from the parent we cannot see what the `renderBody` content is. So this way it has to go on each dialog type. 

Only question, I have header in two different places and I was considering using a `<macro>` for that but I heard that we shouln't use macros anymore. I could also use a sub template component if needed so we don't have the code in two different places


src/components/components/ebay-dialog-base/index.marko
```diff
@@ -22,8 +24,13 @@ $ const baseEl = input.baseEl || 'div';
             <${input.top.renderBody}/>
         </if>
         <${input.header ? "div" : null} class=`${input.classPrefix}__header`>
+            $ const header = input.header || {};
```
DylanPiercey, line 27: I'd rather see `<if(input.header && buttonPosition === 'right')>` than defaulting to an empty object here. Also as it stands I believe this would create empty `<h2>`'s if you don't pass a header?

src/components/components/ebay-dialog-base/migrator.js
```diff
@@ -0,0 +1,43 @@
...
+        enter(node) {
+            if (node.tagName === '@header') {
+                // Iterate only once in caes there are multiple tags
+                let foundH2 = false;
+                node.forEachChild((child) => {
+                    const tag = child.tagName;
+                    if (!foundH2 && headerRegExp.test(tag)) {
+                        foundH2 = true;
+                        const isDefault = tag === 'h2';
+                        const message = isDefault ? `<@header>` : `<@header as=${tag}>`;
```
DylanPiercey, line 19: `tag` here needs to be quoted. Currently would output `<@header as=h1>` which is not valid Marko.

DylanPiercey: I guess this is just for the message, but still.

src/components/ebay-toast/index.marko
```diff
@@ -10,9 +11,9 @@ import processHtmlAttributes from "../../common/html-attributes"
     buttonPosition="right"
     on-modal-show("emit", "toast-show")
     on-modal-close("emit", "toast-close")
-    class=[
-        input.class,
-        "toast--transition"
-    ]>
+    class=[input.class, "toast--transition"]>
+    <@header class=[header.class, "toast__title"] ...header>
+        <${header.renderBody}/>
```
DylanPiercey, line 16: This renderBody is not necessary, `renderBody` will be forwarded with `...header`.

`<@header class=[header.class, "toast__title"] ...header/>` should be fine.

PR #1136, opened 2020-06-03
Added changes for dsn support (and dynamic ds support)

agliga:
## Description
WIP

Build failing because we need to generate SVGs from skin dsn


ianmcburnie:
Can we target this one at 5.6.0 (when it's ready)?

agliga:
> So, I'm not really sure about adding all the dsn icons at this point. Especially since they are just copies of DS6. I also need to solve this on the skin side.

I thought they are needed for arc support (as well as the way we get the icons from the server depending on the flag)
@DylanPiercey Can you confirm?

PR #1139, opened 2020-06-08
Added focusables to triger after request Animation

agliga:
## Description
Changed focusables to trigger on callback. If this PR looks good will push it to `4.5.0` as well.

## Context
* For carousel, there are multiple focusable calls. We could optimize to add them all in 1 call, but for now I left it to trigger each one. I also have the return be set in an array and cancel will cancel all the requests
* For tooltip-base I had to refactor since the host retrieved from focusables was used in expander. I split up the setupMakeup to have a setupExpander method too to be used in two places.

### DIFF
--- a/src/components/ebay-carousel/component.js
+++ b/src/components/ebay-carousel/component.js
@@ -91,14 +91,13 @@ function onRender() {
 
         // Ensure only visible items within the carousel are focusable.
         // We don't have access to these items in the template so me must update manually.
-        forEls(listEl, itemEl => {
-            cleanupFocusables.call(this);
-            this.cancelFocusList.push(focusables(itemEl, false, (focusEl) => {
-                focusEl.forEach(itemEl.getAttribute('aria-hidden') !== 'true'
+        this.focusFrame = requestAnimationFrame(() => {
+            forEls(listEl, itemEl => {
+                focusables(itemEl).forEach(itemEl.getAttribute('aria-hidden') !== 'true'
                     ? child => child.removeAttribute('tabindex')
                     : child => child.setAttribute('tabindex', '-1')
                 );
-            }));
+            });
         });
 
         if (config.nativeScrolling) {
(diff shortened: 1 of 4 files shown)


src/components/ebay-carousel/component.js
```diff
@@ -92,10 +92,13 @@ function onRender() {
         // Ensure only visible items within the carousel are focusable.
         // We don't have access to these items in the template so me must update manually.
         forEls(listEl, itemEl => {
-            focusables(itemEl).forEach(itemEl.getAttribute('aria-hidden') !== 'true'
-                ? child => child.removeAttribute('tabindex')
-                : child => child.setAttribute('tabindex', '-1')
-            );
+            cleanupFocusables.call(this);
+            this.cancelFocusList.push(focusables(itemEl, false, (focusEl) => {
```
DylanPiercey, line 96: Hmm... This is a bit weird. I'm almost thinking deferring this to `makeup-focusables` might not have been the right call 😓. We shouldn't need to create a RAF for each item...

agliga: Possibly. I think a cleaner way would be to allow focusables to pass in an array and then return only the focused elements for each or something.

src/components/ebay-carousel/component.js
```diff
@@ -91,14 +91,13 @@ function onRender() {
 
         // Ensure only visible items within the carousel are focusable.
         // We don't have access to these items in the template so me must update manually.
-        forEls(listEl, itemEl => {
```
DylanPiercey, line 94: Is there any reason we couldn't move this into `renderFrame` below?

agliga: This is moved into the render frame. Not sure what you mean by this.

agliga: Basically I took this code and wrapped it in a `requestAnimationFrame`. I thought this is what the fix should be since it will have 1 request animation frame.

agliga: I think I see what you mean. I do not think it's a good idea since this is under an if statement to do preserveItems or not. That boolean is flipped in the renderFrame below. A
lso there is a chance that that renderFrame below is not called in certain cases since there is an early return, such as on autoplay. 

I think these things should be revisited when we refactor, otherwise trying to change it like it is now will cause random bugs to creep up.

Issue #1027, opened 2020-01-16
ebay-textbox: floating label generates wrong HTML


darkwebdev:
# Bug Report

## Description
When using floating label generated HTML is different from what I see in Skin 9/DS6 documentation.

Usage:
```html
<ebay-textbox floating-label="First Name" />
```

Expected:
```html
<span class="floating-label">
    <label class="floating-label__label" for="firstName">First Name</label>
    <span class="textbox">
        <input class="textbox__control textbox__control--underline" id="firstName" type="text" />
    </span>
</span>
```

Generated:
```html
<span class="textbox">
	<label for="dynamic-id" class="floating-label__label floating-label__label--inline">First Name</label>
	<input id="dynamic-id" class="textbox__control textbox__control--underline" type="text">
</span>
```

ianmcburnie:
Seems to be working somehow, but yes floating label class should be a wrapper around a textbox (like in the skin example).

ianmcburnie:
Fixed in 10.6.0 branch.

PR #1141, opened 2020-06-10
Updated textbox to be wrapped into floating label when there is one

biaspro:
## Context
Addresses the issue described here: https://github.com/eBay/ebayui-core/issues/1027


agliga:
@biaspro Can you pull from `5.6.0` one more time? I just updated `5.6.0` with all commits from master. That way we can only see your changes.

src/components/ebay-textbox/index.marko
```diff
@@ -12,54 +12,52 @@ static var ignoredAttributes = [
...
-    </if>
-
-    <if(displayIcon && !isPostfixIcon)>
-        <${input.iconTag}/>
-    </if>
-
-    <${input.multiline ? "textarea" : "input"}
-        ...processHtmlAttributes(input, ignoredAttributes)
-        id=id
+<${input.floatingLabel && "span"} class="floating-label" data-testid="floating-label">
```
agliga, line 18: Don't add `data-testid`
If we do use `data-testid` it should be passed by tests.

src/components/ebay-textbox/test/test.browser.js
```diff
@@ -41,6 +41,10 @@ describe('given an input textbox with floating label and no value', () => {
         component = await render(template, input);
     });
 
+    it('then component is wrapped into floating label element', () => {
+        expect(component.getByTestId('floating-label')).has.class('floating-label');
```
agliga, line 45: I know as a rule we try to not test by TestIds. I ran into the same issue. 
I would try to do `getByText` on ``floatingLabel` and then check the parent if it has floating-label.
You can also probably check the root element has `floatingLabel` as well using `component.container.firstElementChild` or something like that,

biaspro: Cool, done. Out of curiosity, why don't you use test-id's?

src/components/ebay-textbox/index.marko
biaspro: Not sure either, I basically just added a conditional wrapper. I guess it must be because it's a wrapper and all code inside shifted

ianmcburnie:
I'm still seeing the wrong markup when I run this locally:

Screen Shot 2020-06-11 at 11 26 51 AM

ianmcburnie:
Approved, but there's a small conflict for you to take care of before we can merge.

Issue #1086, opened 2020-03-21
Notice: support colored background with no icon


msendlakowski:
The current design system allows for the notice to have a colored background but no icon.  We should support that.

![notice](https://user-images.githubusercontent.com/1562843/77214389-22ec3700-6acc-11ea-813b-6a0c0b1b091d.png)

PR #1146, opened 2020-06-12
add explicit option to hide icon in notice

biaspro:
## Context

Addresses the issue described here: https://github.com/eBay/ebayui-core/issues/1086

### DIFF
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -35,7 +35,7 @@ Name | Type | Stateful | Required | Description
 `status`  | String | No | No | "attention" (default for "page" and "inline"), "confirmation" "information", or "celebration" (page notice only).  The default for "section" type will render with grey background and no icon
 `a11y-heading-text` | String | No | Yes | adding description for the notice for a11y users
 `hidden` | Boolean | Yes | No | whether the widget is hidden or not
-`icon-hidden` | Boolean | Yes | No | whether the icon is hidden or not
+`icon` | String | Yes | No | "default" (matches whatever is specified by the "status") or "none"
 
 ### Additional attributes when type="page"
 
--- a/src/components/ebay-notice/README.md
+++ b/src/components/ebay-notice/README.md
@@ -34,7 +34,8 @@ Name | Type | Stateful | Required | Description
 `type` | String | No | No | "inline", "section", "window", or "page" (default)
 `status`  | String | No | No | "attention" (default for "page" and "inline"), "confirmation" "information", or "celebration" (page notice only).  The default for "section" type will render with grey background and no icon
 `a11y-heading-text` | String | No | Yes | adding description for the notice for a11y users
-`hidden` | Boolean | Yes | No | whether the widget is hidden or not.
+`hidden` | Boolean | Yes | No | whether the widget is hidden or not
+`icon-hidden` | Boolean | Yes | No | whether the icon is hidden or not
 
 ### Additional attributes when type="page"
 
--- a/src/components/ebay-notice/index.marko
+++ b/src/components/ebay-notice/index.marko
@@ -16,6 +16,7 @@ $ var isWindow = type === "window";
 $ var isCelebration = status === "celebration";
 $ var isPage = type === "page" || isSection || isWindow;
 $ var isLightGuidance = isSection && !input.status;
+$ var isIconVisible = !input.iconHidden;
 <${isPage ? "section" : "div"}
     ...processHtmlAttributes(input, ignoredAttributes)
     aria-labelledby:scoped="status"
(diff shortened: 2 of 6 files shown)


src/components/ebay-notice/index.marko
```diff
@@ -26,7 +26,7 @@ $ var isLightGuidance = isSection && !input.status;
         input.fillWindow && `${type}-notice--fill`,
         input.class
     ]>
-    <if(!isLightGuidance)>
+    <if(!isLightGuidance && !input.iconHidden)>
```
agliga, line 29: I would actually move this or wrap the `ebay-icon` with this. As is if the dialog has a header it won't be shown with `icon-hidden` option

ianmcburnie: Was this comment resolved?

biaspro: yes now ebay-icon blocks are wrapped individually rather than the whole header

src/components/ebay-notice/README.md
```diff
@@ -34,7 +34,8 @@ Name | Type | Stateful | Required | Description
 `type` | String | No | No | "inline", "section", "window", or "page" (default)
 `status`  | String | No | No | "attention" (default for "page" and "inline"), "confirmation" "information", or "celebration" (page notice only).  The default for "section" type will render with grey background and no icon
 `a11y-heading-text` | String | No | Yes | adding description for the notice for a11y users
-`hidden` | Boolean | Yes | No | whether the widget is hidden or not.
+`hidden` | Boolean | Yes | No | whether the widget is hidden or not
+`icon-hidden` | Boolean | Yes | No | whether the icon is hidden or not
```
ianmcburnie, line 38: Sorry, one more request. Rather than a boolean (which often lead us into traps), I prefer the idea making this an enumerate type instead:

`icon` | String | Yes | No | "default" (matches whatever specified by the "type") or "none"

This keeps thing a little more open in terms of allowing different kinds of icon in future if we want. @agliga thoughts?

agliga: Agreed, I think it is more fluid that way and we might want to add something like right icon or something else. Better to create it more extensible.

src/components/ebay-notice/index.marko
```diff
@@ -16,6 +16,7 @@ $ var isWindow = type === "window";
 $ var isCelebration = status === "celebration";
 $ var isPage = type === "page" || isSection || isWindow;
 $ var isLightGuidance = isSection && !input.status;
+$ var isIconVisible = input.icon !== "hidden";
```
ianmcburnie, line 19: I believe this should be "none" instead of "hidden".

src/components/ebay-notice/index.marko
```diff
@@ -26,22 +27,24 @@ $ var isLightGuidance = isSection && !input.status;
         input.fillWindow && `${type}-notice--fill`,
         input.class
     ]>
-    <if(!isLightGuidance)>
+    <if(!isLightGuidance && (isIconVisible || input.title))>
```
agliga, line 30: Why is this extra check needed for `isIconVisible` or `input.title`? I see below that there is already a check for `isIconVisible`. 
The question is, is this needed? Is there a better way to simplify it since it is a bit complex?

biaspro: If there is no icon and no title, there is no need to render this whole block, because it will be empty. (and it will create extra space)

biaspro: Will try to simplify it a bit for more clarity

src/components/ebay-notice/index.marko
```diff
@@ -26,15 +28,17 @@ $ var isLightGuidance = isSection && !input.status;
...
+    <if(!isLightGuidance && (isIconVisible || isWindowTitle))>
         <${isPage ? input.a11yHeadingTag || "h2" : "span"}
             class=`${type}-notice__status`
             id:scoped="status">
-            <if(isWindow && input.title)>
-                <ebay-icon
-                    class="window-notice__icon"
-                    name="confirmation-filled"
-                    aria-label=input.a11yHeadingText/>
+            <if(isWindowTitle)>
```
agliga, line 35: Is there an issue now with the icon showing up if windowTitle is false. I think it shows up on the `<else>`

Maybe we should figure out all the different states for this and try to fix it that way. I think it is still overly complex more than it should be. 

My initial guess for this is:
if lightguidance
   if title and window
       if image
          show image
        show title
     else if image
        show image
Something like that. But not sure all the different paths.

ianmcburnie: Notice is going to get some serious refactoring in v6 to make this stuff easier to follow. Just do what you can for now until then.

Issue #1099, opened 2020-04-17
Combobox: add actionable icon option


ianmcburnie:
We added this feature in skin. Let's now expose it in coreui. This will allow teams to toggle the autocomplete property in a predictable way for assistive technology users.

ianmcburnie:
Done in 5.6.0. Closing.

PR #1148, opened 2020-06-15
Added combobox actionable icon

agliga:
## Description
* Added actionable icon to combobox. Also added an example for it. 

## Context
One of the issue was that when clicking the icon, it would trigger a close since the focus left the input. I changed it so it will not close when button is clicked

## References
#1099

## Screenshots
Note, there is a skin issue (which is also present in skin) where the button looks off. Will need a skin change too.

### DIFF
--- a/src/components/ebay-combobox/component.js
+++ b/src/components/ebay-combobox/component.js
@@ -18,9 +18,9 @@ module.exports = {
         return this.expander.collapse();
     },
 
-    handleActionableClick(originalEvent) {
-        this.clickedActionable = true;
-        this.emit('combobox-actionable-click', { originalEvent });
+    handleButtonClick(originalEvent) {
+        this.buttonClicked = true;
+        this.emit('combobox-button-click', { originalEvent });
     },
 
     handleExpand() {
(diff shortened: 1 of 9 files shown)


agliga:
> So I noticed that when I activate the button, focus jumps back to the input box. We probably shouldn't do this. Also, I think as much as possible, when this button is present we should let the app decide on their own logic for the combobox's expanded and autocomplete states.

This actually what is being done. In the example I have the focus triggering on the input. 
Here is the logic for the example: https://github.com/eBay/ebayui-core/pull/1148/files#diff-5626d1eff5dcbc9ef1d67d2a0779e5a1R2-R10
However I can provide another example if needed. Clicking on the button will only trigger an event.

ianmcburnie:
I think just remove the line that sets focus. I wouldn't want teams to copy that. I think maybe just a completely dumb example for now that has no logic. We could maybe add another example that mimics what selling team did: disables autocomplete when the button is clicked, and restores it when user types in box again. But happy to leave that 2nd example for later.

agliga:
> I think just remove the line that sets focus. I wouldn't want teams to copy that. Actually, I think maybe just a completely dumb example for now that has no logic. We could maybe add another example that mimics what selling team did: disables autocomplete when the button is clicked, and restores it when user types in box again. But happy to leave that 2nd example for later.

I initially used that example because that is what skin has as the example for that actionable icon. https://ebay.github.io/skin/#combobox
However, I changed it to toggle autocomplete. What it will do is when clicked it will trigger that autocomplete but if the focus is on the input it will not close the input. However if we need to remove that logic it can be done. 

Again, we also need a skin change to change the alignment

agliga:
> So, I'm still not totally happy with that example. Sorry! Let's just have it simply log the expanded state to console when clicked. I know it's just a rough example, but I'm wary of app teams copying it verbatim (seen this happen before).

Done, I also renamed it to 09 because we have example 08 in the sticky PR.

agliga:
Also changed it to be button instead of actionable to follow the textbox pr

Issue #1116, opened 2020-05-08
Combobox: add sticky item to top of suggestion list


malughanshyam:
# Feature Request

## Description

As mentioned in https://github.com/eBay/ebayui-core/issues/1098#issuecomment-625992217, we have a requirement where we want to show a default sticky option in the `ebay-combobox` to the users always so that they have an easy way to get back to the default option.
 
## Context
We plan to use the `ebay-combobox` to show a list of campaigns to the users. The top option in this list is "All". As per the business logic, this option has a special meaning, i.e., it represents everything in the combo box (all the campaigns). The users can still select a specific campaign from the list of options. But while filtering (typing), as the list of options narrows down (with the auto-complete feature), we want to show the default sticky option "All" always. 

https://github.com/eBay/ebayui-core/issues/1098#issuecomment-625992217

ianmcburnie:
Done in 5.6.0. Closing.

malughanshyam:
Great. Thank you Ian. 
Would it be possible to add this in 4.x as bumping to 5.x is a breaking change for us?

ianmcburnie:
@agliga Did all the work :-) @malughanshyam Typically we only backport critical patches to older versions. Is there no way you can update to v5?

malughanshyam:
Finally, I have upgraded my page to ebayui-core@5.x. 
There were a lot of dependent shared components that weren't up to date with ebayui-core^5.x / ebay-skin^10. 

Thanks, @agliga & @ianmcburnie.

PR #1150, opened 2020-06-17
Added sticky option to combobox

agliga:
## Description
Added option for combobox to have a sticky option.

## Context
Added a modifier to be sticky which is ignored in the filter list. Documented it and added an example

## References
#1116

## Screenshots
image


malughanshyam:
Hi @agliga, 
Would it be possible to add this in 4.x?

Issue #1132, opened 2020-05-28
ebay-textbox: Add actionable stateful post icon and double icon support


gemicci:
# Feature Request
Ebay-textbox: Add actionable stateful post icon and double icon support

Possible new component inputs& actions:

```
<ebay-textbox    
    pre-icon-name="search" 
    post-icon-name="clear" 
    on-textbox-post-icon-click('handlePostIconClick') >
```
If post-icon-name attribute changes from "clear" to null, component should rerender with no icon. "clear" to "arrow-down" should also cause a rerender with the new icon

## Description

There is a product requirement for a complex multiselect dropdown.  We are planning to use ebay-textbox as the base. As per the design, the multiselect box needs to have a pre and post icon. Post icon needs to be actionable and changeable. Notice down arrow will change to X. On click, it will clear the input box content.

## Screenshots
<img width="304" alt="Screen Shot 2020-05-28 at 10 18 05 AM" src="https://user-images.githubusercontent.com/10719382/83153150-9d3cc700-a0cc-11ea-9308-44045ef057ce.png">
<img width="326" alt="Screen Shot 2020-05-28 at 10 16 50 AM" src="https://user-images.githubusercontent.com/10719382/83153148-9ca43080-a0cc-11ea-921a-5a0061d044f9.png">

ianmcburnie:
We'll look into it, thanks. Might also need some work on the skin end to see if we support and/or document the use of two icons.

ianmcburnie:
@gemicci Actually, what is the purpose of the first item being clickable?

gemicci:
@ianmcburnie Thanks for looking into this feature. There is no product requirement for an actionable left icon. Just right one is needed. I just wrote the spec for the sake of consistency. I will update the ticket.

ianmcburnie:
Done in 5.6.0. Closing.

PR #1152, opened 2020-06-17
Added postfix and prefix icon to textbox

agliga:
## Description
Added `postfix-icon` and `prefix-icon` attributes which can show prefix/postfix icons. Can combine both.
In order to render a button I added a `postfix-aria-label` attrbute which is required for the button to render (this way we don't need another attribute to set as button, if aria-label is set it should be a button)

## Context
I added a migrator to take icon to convert to prefix/postfix depdening on the iconPosition attribute. 

## References
#1132

## Screenshots
image

### DIFF
--- a/src/components/ebay-textbox/README.md
+++ b/src/components/ebay-textbox/README.md
@@ -17,7 +17,7 @@ Name | Type | Stateful | Required | Description
 `invalid` | Boolean | No | No | indicates a field-level error with red border if true
 `prefix-icon` | String | No | No | name of the icon from skin to show as the prefix icon
 `postfix-icon` | String | No | No | name of the icon from skin to show as the postfix icon
-`postfix-aria-label` | String | No | No | aria-label for postfix. Required to be set in order to render postfix button and attach a `textbox-postfix-click event`
+`button-aria-label` | String | No | No | aria-label for postfix. Required to be set in order to render postfix button and attach a `textbox-button-click event`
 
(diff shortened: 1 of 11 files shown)


src/components/ebay-textbox/README.md
```diff
@@ -15,8 +15,9 @@ Name | Type | Stateful | Required | Description
 `fluid` | Boolean | No | No |
 `multiline` | Boolean | No | No | renders a multi-line texbox if true
 `invalid` | Boolean | No | No | indicates a field-level error with red border if true
-`icon` | String | No | No | name of the icon from skin
-`icon-position` | String | No | No | Can be "prefix" / "postfix". If not specified or if the value passed is neither "prefix" nor "postfix",  the `icon-position` is set by default to "prefix"
+`prefix-icon` | String | No | No | name of the icon from skin to show as the prefix icon
+`postfix-icon` | String | No | No | name of the icon from skin to show as the postfix icon
+`postfix-aria-label` | String | No | No | aria-label for postfix. Required to be set in order to render postfix button and attach a `textbox-postfix-click event`
```
ianmcburnie, line 20: Can we change it from `postfix-aria-label` to `button-aria-label`? And to `textbox-button-click`? With the idea that the button is always at the end, never at the start, of the textbox. We should make combobox have the same language of 'button' too, for consistency.

agliga: Done

PR #1154, opened 2020-06-18
perf: convert all components to split components where possible

DylanPiercey:
## Description
Recently some components were converted into split components (using `component-browser.js`) to improve performance.

This PR goes through all remaining components and converts them to split components if they are not using state. Also it optimizes the serialized input by adding a `toJSON` for each one to only use `input` which is used within the `component-browser.js` file if it is a top-level component.


ianmcburnie:
Needs targeting at 5.6.0 branch (assuming no breaking changes).

agliga:
> Needs targeting at 5.6.0 branch (assuming no breaking changes).

I think we should release this as a patch for `5.5.x`.
This way it can be used immediately
Also i will update `5.6.0` with master and fix the conflicts.

ianmcburnie:
After chatting offline, the PR is non-urgent and we really need to QA this. So let's bundle it in with upcoming 5.6.0 release. I'll change the target branch... hopefully there'll be no conflicts

ianmcburnie:
...only a couple of conflicts in textbox component.

DylanPiercey:
@ianmcburnie I rebased what I have on top of `5.6.0` which included a few commits from master that haven't made it there. I'm assuming we'll be rebasing `5.6.0` on top of master anyways which would resolve the extra commits and save me some work.

If needed I can rebase again though.

agliga:
> @ianmcburnie I rebased what I have on top of `5.6.0` which included a few commits from master that haven't made it there. I'm assuming we'll be rebasing `5.6.0` on top of master anyways which would resolve the extra commits and save me some work.
> 
> 
> 
> If needed I can rebase again though.
That’s fine. I was planning on doing that anyways.

src/components/components/ebay-tooltip-base/index.marko
```diff
@@ -1,3 +1,14 @@
+static function toJSON() {
+    // Only serialize the following to the browser if we are a top level component.
+    return {
```
agliga, line 3: How does this work?
Are these all the variables needed in client side render? So from now on we would have to specify all client side variables needed from server?

Just wondering so i can understand for future

DylanPiercey: It's only needed for split components. If a split component is loaded (outside of a stateful component) it will not rerender in the browser. However, it still gets all of its `input` serialized so that you can access `this.input` inside of the `component-browser.js` file.

By adding the `toJSON` to the `input` it will make it so that if this component is serialized at the top level then it will only send down the subset of `input` that is actually used in the `component-browser.js`. For some of the components, we have no `input` is used in the `component-browser.js`, that's why many `toJSON`'s are a noop, so nothing gets serialized.

Issue #1151, opened 2020-06-17
Combobox: add aria-roledescription


ianmcburnie:
The `aria-roledescription` attribute now has much better support. I have updated the MIND pattern to make use of it.

For example:

```html
<input aria-roledescription="Use up and down arrow keys to navigate options. Options change based on input." id="combobox-0-input" name="console0" type="text" placeholder="Playstation 4, Xbox One, etc." role="combobox" aria-expanded="true" autocomplete="list" aria-owns="combobox_0-listbox"/>
```

https://ebay.gitbook.io/mindpatterns/input/combobox

agliga:
Trying to understand what to do with this. We already support passthrough attributes so it should be automatic. 
My understanding is we would just change examples to show the `aria-roledescription` here. Should we change all of them or only the first?

ianmcburnie:
Sorry, I wasn't clear. Just hard code `aria-roledescription="Use up and down arrow keys to navigate options." into the template. We don't want apps to have to do this every time. 

If autocomplete is enabled (which is most of the time), the string would be slightly different: "Use up and down arrow keys to navigate options. Options change based on text input".

We can adjust the exact description text as need be, but these strings are a good place to start. Also, for I18N purposes we'll need to allow apps to override those strings with localized strings.

ianmcburnie:
Done in 5.6.0. Closing.

PR #1155, opened 2020-06-19
Added aria-roledescription to combobox

agliga:
## Description
Added aria-roledescription. It will take user input first, and then it will default to english text (and change based on autocomplete value)

## References
#1151


src/components/ebay-combobox/README.md
```diff
@@ -45,6 +45,7 @@ Name | Type | Stateful | Required | Description
 `disabled` | Boolean | Yes | No | sets the disabled attribute of the input
 `expanded` | Boolean | Yes | No | sets whether the listbox is expanded
 `autocomplete` | String | Yes | No | default is `none`; available values are `none` and `list`
+`aria-roledescription` | String | No | No | The role description for accessibility. Default text is set and will be in english. Pass this to override for different locales
```
ianmcburnie, line 48: Can we drop the `aria-` suffix please? We'll come back and revisit all of these attr names in v6, but for now let's make it consistent with what we did with the `autocomplete` attribute.

Issue #690, opened 2019-06-03
Listbox: make independent of button


ianmcburnie:
The MIND patterns have been updated to make listbox independent of any button. This has the following benefits:

* Listbox can be embedded directly in the page without any button (similar to how we made form controls independent of any label)
* Better support for multi-select variant which is often embedded directly in page without button
* Better support for mWeb, which is also often embedded directly on page without button
* Listbox can be nested inside of combobox or other "controlling" widget

Examples:

* https://ebay.github.io/mindpatterns/input/listbox/index.html
* https://ebay.github.io/mindpatterns/input/listbox-button/index.html
* https://ebay.github.io/mindpatterns/input/combobox/index.html

Related tickets:

* https://github.com/eBay/skin/issues/629

ianmcburnie:
Related PR: https://github.com/eBay/ebayui-core/pull/863

ianmcburnie:
This can potentially go in 5.7.0, but also have no problem with it going in 6.0.0 if easier from a refactoring point of view.

ianmcburnie:
Done. Thank you, Andrew.

PR #1166, opened 2020-07-02
Split up listbox to have it's own separate component

agliga:
## Description
Split up listbox and removed it from listbox-button. Have listbox-button call listbox

## Context
* Initially I tired to make is so that `listbox__options` is wrapped by listbox. However there are styles which are requiring the options to be adjacent to the button. 
* Moved active descendent to the listbox and kept only expander in listbox-button. I had to add some scroll changes to the button and the listbox
* Added the keydown event on the listbox__options container to allow going up and down. 
* On listbox-button expand I have to call elementScroll on listbox in order to track it. I can pass attributes and update onInput but seemed like overkill

## References
#690

### DIFF
--- a/marko.json
+++ b/marko.json
@@ -25,6 +25,9 @@
   "<ebay-filter-menu-button-item>": {
     "transformer": "./src/common/transformers/attribute-tags/index.js"
   },
+  "<ebay-listbox-option>": {
+    "transformer": "./src/common/transformers/attribute-tags/index.js"
+  },
   "<ebay-listbox-button-option>": {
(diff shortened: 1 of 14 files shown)


ianmcburnie:
@agliga I'm unable to TAB key to each listbox example. Possibly missing the tabindex from listbox root element. Check Skin for example.

agliga:
> @agliga I'm unable to TAB key to each listbox example. Possibly missing the tabindex from listbox root element. Check Skin for example.

Done, I added tabindex of 0 by default, and I override it in listbox-button

Issue #1159, opened 2020-06-22
Migrations for v6.0.0


ianmcburnie:
Add v6 migrations to 5.x branch when the time is right.

Issue #1158, opened 2020-06-22
Notice: split into separate components


ianmcburnie:
With the additions of window notice and section notice over time, I think we would now benefit from separating ebay-notice out into the following components:

* ebay-window-notice
* ebay-page-notice
* ebay-section-notice
* ebay-inline-notice

I think this will be clearer for end users too.

Related skin issue: https://github.com/eBay/skin/issues/1150

ianmcburnie:
Related Skin PR: https://github.com/eBay/skin/pull/1166

PR #1178, opened 2020-07-17
Split up Notices in various components

agliga:
## Description
Created 4 new components plus 1 common component
* `ebay-page-notice`
* `ebay-inline-notice`
* `ebay-section-notice`
* `ebay-window-notice`

And created a `ebay-notice-base` component which all these components should use. This provides basic classes and icons

## Context
Added a migrator in `ebay-notice` which should do the following:
* Changes each `type` into the given component. If no type is found, defaults to page
* Renames `a11yHeadingText` to `a11yText` (if present)
* Unwraps `<ebay-notice-content>` tag. Should be in body now. (Dylan if I missed something here please let me know a better way to do unwrap)
* Adds either `ebay-button` or `ebay-menu-button` in `<@footer>` tag if it's the immediate child of the parent. 

This should take care of most of the main usecases for notice. 

## References
#1158
#1159

### Page notice
Screen Shot 2020-07-17 at 12 15 54 PM
Screen Shot 2020-07-17 at 12 15 59 PM

### Window notice
Screen Shot 2020-07-17 at 12 16 04 PM
Screen Shot 2020-07-17 at 12 16 37 PM

### Section notice
Screen Shot 2020-07-17 at 12 19 59 PM

### Inline notice (I think the styles are off from skin, not all are coming through This is the issue: https://github.com/eBay/skin/blob/11.0.0/inline-notice.js (should be inline-notice not notice))
Screen Shot 2020-07-17 at 12 19 48 PM


src/components/ebay-window-notice/examples/02-window-notice-fill/template.marko
```diff
@@ -0,0 +1,17 @@
...
+        max-width: 320px;
+        height: 500px;
+        padding: 8px;
+        background: #fff;
+        border: 1px solid #eee;
+    }
+</style>
+
+<div class="demo2-container">
+    <ebay-window-notice type="window" a11y-text="Success!" fill-window>
```
ianmcburnie, line 12: I think I'd prefer an enumerated type here instead of the `fill-window` boolean. I don't see this as something that would get toggled on and off like other booleans (e.g. disabled, expanded, hidden). Perhaps we could make that part of the criteria for a boolean or not? @DylanPiercey @agliga thoughts?

agliga: I would personally not have enums unless they have 2+ options or will possibly have more options in future. The reason is because there will only be 1 option you will specify. Like: `fill="window"` or `"fill="none"` where you'll never want to use `fill="none"`. I think it becomes a waste rather than doing `fill-window`. But that's just me.

agliga:
Note, the build is failing because it requires styles in skin which are not present until we pre-release.

src/components/ebay-window-notice/examples/02-window-notice-fill/template.marko
ianmcburnie: Right now there are only two, but it's not inconceivable that more could be added, e.g. `layout="default"`, `layout="window"`, `layout="split".

Maybe just me, but I don't have a huge problem with only 2 options. There are cases of that in html, css and aria too. 

I think we should try and set some firm rules for when to use a boolean. I guess for me I'm more used to seeing booleans used for things that are semantic rather than presentational.

ianmcburnie: I would be interested to see if we could follow this "rule" with our API in v6, but open to more discussion...

ianmcburnie:
Skin issue fixed: https://github.com/eBay/skin/commit/fa06a7112cafecebab9a61cea2d0b73b26e3c161#diff-8628588396340805b1e50c0de096bbe1L3

src/components/ebay-window-notice/examples/02-window-notice-fill/template.marko
ianmcburnie: @agliga To unblock this, let's move forward with this stricter guideline on use of booleans and change this to be an enum.

The attribute name and values should reflect the two current use cases. Given that window notices appear only in dialog windows, my suggestion is to go with `window=lightbox` (default) and `window=fullscreen` for now. It is not unfeasible that we may add a `window=panel` or `window=drawer` in future, and this would tie in with our new updated dialog terminology.

Screen Shot 2020-07-27 at 9 46 49 PM

src/components/ebay-window-notice/index.marko
```diff
@@ -0,0 +1,9 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+$ var status = "confirmation";
+<ebay-notice-base
+    class=[input.fillWindow && `window-notice--fill`, input.class]
```
ianmcburnie, line 5: Just a heads up that I changed this modifier from `fill` to `screen` in v11 branch.

src/components/ebay-notice/migrator.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * @description
+ * Migrates each notice to different types
+ */
+
+// Types of buttons that if they are direct children of root will be wrapped in footer
+const buttonTypes = ['ebay-button', 'ebay-menu-button'];
+
+function setAttribute(el, oldName, newName) {
```
DylanPiercey, line 9: Looks like this function only sets the attribute if it was already there, is that intentional?

agliga: Yes, it's renaming it from the old name to the new.

src/components/components/ebay-notice-base/index.marko
```diff
@@ -0,0 +1,62 @@
...
+    <${input.headerRoot || "div"} class=`${prefixClass}__header` id:scoped="status"
+        aria-roledescription=input.a11yRoleDescription
+    >
+
+        <if(input.icon !== "none")>
+            <if(status === "confirmation" || status === "celebration")>
+                <ebay-icon
+                    name="confirmation-filled"
+                    aria-label=input.a11yText
+                    class=input.iconClass/>
```
DylanPiercey, line 31: Not a huge fan of this `iconClass` attribute. Let's remember to address this when we look at refactoring the icon apis.

src/components/components/ebay-notice-base/style.js
```diff
@@ -0,0 +1 @@
+require('@ebay/skin/icon');
```
DylanPiercey, line 1: This shouldn't be necessary since we are using the `<ebay-icon>` and it will bring it in.

src/components/ebay-window-notice/examples/02-window-notice-fill/template.marko
DylanPiercey, line 12: Shouldn't need `type` attr here.

DylanPiercey: Also I see a `fill-window` attribute here and in the `marko.json`. It looks like it got renamed or removed?

src/components/ebay-window-notice/index.marko
```diff
@@ -0,0 +1,9 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+$ var status = "confirmation";
```
DylanPiercey, line 3: No point in making this a variable. Should just be in the attribute itself.

src/components/ebay-section-notice/marko-tag.json
```diff
@@ -0,0 +1,21 @@
+{
+  "attribute-groups": ["html-attributes"],
+  "@*": {
```
DylanPiercey, line 3: I think this is missing `a11y-role-description`.

src/components/ebay-section-notice/index.marko
```diff
@@ -0,0 +1,10 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+<ebay-notice-base
+    ...input
```
DylanPiercey, line 4: We're spreading `input` twice here unnecessarily.

src/components/ebay-section-notice/index.marko
```diff
@@ -0,0 +1,10 @@
+import processHtmlAttributes from "../../common/html-attributes"
```
DylanPiercey, line 1: This is imported but not used.

src/components/ebay-notice/index.marko
```diff
@@ -1,77 +1 @@
...
-                <${input.title.renderBody}/>
-            </span>
-        </if>
-        <${input.content || input.renderBody}/>
-    </>
-    <if(input.content)>
-        <${input.renderBody}/>
-    </if>
-</>
+<!-- Removed -->>
```
DylanPiercey, line 1: This file should not be necessary.

src/components/ebay-notice/marko-tag.json
```diff
@@ -1,5 +1,6 @@
 {
```
DylanPiercey, line 1: We should be able to empty out this file besides the `migrator` property.

src/components/ebay-inline-notice/index.marko
```diff
@@ -0,0 +1,14 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+$ var status = input.status || "attention";
+static var ignoredAttributes = ["status", "a11yHeadingText"];
```
DylanPiercey, line 4: Doesn't look like this or the `processHtmlAttributes` are used here.

src/components/ebay-inline-notice/index.marko
```diff
@@ -0,0 +1,14 @@
...
+
+<ebay-notice-base
+    class=[`inline-notice--${status}`, input.class]
+    status=status
+    prefixClass="inline-notice"
+    root="div"
+    header-root="span"
+    main-root="span"
+    ...input
```
DylanPiercey, line 13: Spreading the `input` at the bottom here will override the `class` we've set above.

src/components/ebay-page-notice/index.marko
```diff
@@ -0,0 +1,9 @@
+$ var status = input.status || 'attention';
+
+<ebay-notice-base
+    class=[`page-notice--${status}`, input.class]
+    status=status
+    role="region"
+    prefixClass="page-notice"
+    ...input
```
DylanPiercey, line 8: Spreading the `input` here will override the `class` we have above.

src/components/components/ebay-notice-base/index.marko
```diff
@@ -0,0 +1,62 @@
...
+            </else>
+        </if>
+    </>
+    <${input.mainRoot || "div"} class=`${prefixClass}__main`>
+        <if(input.title)>
+            <${input.title.as || "h2"}
+                class=[`${prefixClass}__title`, input.title.class]
+                ...input.title>
```
DylanPiercey, line 51: Spreading the `input` here will override the `class` we have above.

src/components/ebay-window-notice/index.marko
```diff
@@ -0,0 +1,9 @@
...
+<ebay-notice-base
+    class=[input.window === "fullscreen" && `window-notice--screen`, input.class]
+    role="region"
+    status=status
+    prefixClass="window-notice"
+    ...input/>
```
DylanPiercey, line 9: Spreading the `input` here will override the `class` we have above.

src/components/ebay-notice/index.marko
agliga: If i removed it, it doesn't work properly, says tag is not found.

src/components/ebay-notice/migrator.js
```diff
@@ -0,0 +1,59 @@
...
+    if (el.hasAttribute(oldName)) {
+        const attribute = el.getAttribute(oldName);
+        attribute.name = newName;
+    }
+}
+
+function migrator(el, context) {
+    const type = el.hasAttribute('type') && el.getAttributeValue('type').value;
```
DylanPiercey, line 17: This migrator will only work if the `type` attribute is a string literal.

We should probably error in the case that it's not... Although that would be a breaking change. We can discuss if you like.

src/components/ebay-notice/migrator.js
```diff
@@ -0,0 +1,59 @@
...
+    } else if (type === 'window') {
+        context.deprecate('<ebay-notice type="window"/> has been renamed to <ebay-window-notice>');
+        el.setTagName('ebay-window-notice');
+    } else {
+        // Default is page notice
+        context.deprecate('<ebay-notice/> has been renamed to <ebay-page-notice>');
+        el.setTagName('ebay-page-notice');
+    }
+
+    setAttribute(el, 'a11y-heading-text', 'attyText');
```
DylanPiercey, line 34: Prefer `el.renameAttribute(oldName, newName)`.

agliga: It doesn't exist 👎 
Throws an error

src/components/ebay-notice/index.marko
DylanPiercey: Can you share the full error? That would mean something bad is happening.

src/components/ebay-notice/migrator.js
DylanPiercey: Lame. I thought this was exposed on custom elements but I was wrong.

Either way lets make this more clear and make our helper called `renameAttribute` or something like that, and also use `removeAttribute()` and `setAttribute()`.

agliga: I will create a common method once dialog and this get merged.

Issue #1164, opened 2020-06-29
Dialog: split into separate components


ianmcburnie:
Dialog is going to be split up into multiple skin modules:

* lightbox -> ebay-lightbox
* panel -> ebay-panel
* fullscreen -> ebay-screen (i.e. a fullscreen dialog)
* drawer (already done) -> ebay-drawer
* toast (already done) -> ebay-toast
* confirm (new)
* alert (new)

ianmcburnie:
Changes merged to 6.0.0 branch.

PR #1190, opened 2020-07-31
Split up dialogs into various components

agliga:
PLEASE SQUASH.

## Description
Split up dialogs, created migrators
## Context
* Created fullscreen-dialog
* Created panel-dialog. Used `position:end` for right aligned
* Created lightbox dialog (default for migrator). Also added mini dialog as a different `mode`
I used the variable `mode` for lightbox, which will have different variants. (maybe variant might be better). When we add the alert and confirm dialog, we can use that variable.

## References
#1164 

## Screenshots
Dialogs look the same as in skin.
Only removed fill dialog


src/components/ebay-fullscreen-dialog/index.marko
```diff
@@ -0,0 +1,13 @@
+import processHtmlAttributes from "../../common/html-attributes"
```
DylanPiercey, line 1: `processHtmlAttributes` is not being used here.

src/components/ebay-fullscreen-dialog/marko-tag.json
```diff
@@ -0,0 +1,44 @@
...
+    "html-attributes"
+  ],
+  "@*": {
+    "targetProperty": null,
+    "type": "expression"
+  },
+  "@html-attributes": "expression",
+  "@open": "boolean",
+  "@type": {
+    "enum": [
```
DylanPiercey, line 13: This should be removed now right?

src/components/ebay-lightbox-dialog/README.md
```diff
@@ -0,0 +1,23 @@
...
+</ebay-lightbox-dialog>
+```
+
+## Attributes
+
+Name | Type | Stateful | Required | Description
+--- | --- | --- | --- | ---
+`open` | Boolean | Yes | No | Whether dialog is open.
+`focus` | String | No | No | An id for an element which will receive focus when the dialog opens (defaults to close button).
+`mode` | String | No | No | Either "mini" or "default". Changes the dialog to be normal dialog or mini version (for use in infotip)
```
DylanPiercey, line 15: Not sure how I feel about exposing `mode="mini"` seems like that should be exclusive to the infotip implementation.

src/components/ebay-lightbox-dialog/examples/02-scrolling/template.marko
```diff
@@ -0,0 +1,28 @@
+<ebay-lightbox-dialog
+    a11y-close-text="Close Dialog"
+    aria-labelledby="dialog-title"
```
DylanPiercey, line 3: I thought the `aria-labelledby` was taken care of internally and not an option anymore?

src/components/ebay-lightbox-dialog/index.marko
```diff
@@ -0,0 +1,21 @@
+import processHtmlAttributes from "../../common/html-attributes"
```
DylanPiercey, line 1: `processHtmlAttributes` is not being used.

src/components/ebay-lightbox-dialog/index.marko
```diff
@@ -0,0 +1,21 @@
...
+
+$ var isMini = input.mode === 'mini';
+
+<ebay-dialog-base
+    ...input
+    open=input.open
+    classPrefix="lightbox-dialog"
+    on-modal-open("emit", "lightbox-dialog-open")
```
DylanPiercey, line 9: I am concerned that we are using `open` and `show` interchangeably.

src/components/ebay-panel-dialog/index.marko
```diff
@@ -0,0 +1,15 @@
+import processHtmlAttributes from "../../common/html-attributes"
```
DylanPiercey, line 1: `processHtmlAttributes` is not being used.

src/components/ebay-lightbox-dialog/README.md
agliga: I exposed it because we will add different modes. Since mini was so tied to lightbox at least I have it here. 
I can hide the mode mini, but at least it gives developers a glimpse into what is being done.

src/components/ebay-lightbox-dialog/index.marko
agliga: We should be using `open` now. If you see any more `show` it should be migrated and moved

src/components/ebay-lightbox-dialog/README.md
ianmcburnie: We also have confirm and alert variants to think about. These are very specialized types of lightbox. With just a header, description text, and one or two buttons. Given that they are not open to much customization, they could potentially be their own standalone components, e.g. ebay-confirm-dialog and ebay-alert-dialog, that wrap ebay-lightbox-dialog.

ianmcburnie: Happy for you to leave alert and confirm for a follow up PR, but have a think about it and see how it might affect your thinking (if at all) for this current PR.

Issue #1184, opened 2020-07-28
Icon: support external SVG file


ianmcburnie:
We have updated the Skin website in v11 branch to use an external SVG icon file (and the svg4everybody polyfill for IE11):

https://github.com/eBay/skin/pull/1194

However, we are currently running into cross domain issues preventing the `use` tag from accessing a file on CDN:

https://ir.ebaystatic.com/cr/v/c1/skin/v10.6.1/ds6/icons.svg

An external SVG file solves some issues with our current icon implementation in CoreUI, but if we can't host from a non-relative path I'm not sure if those benefits then become outweighed. Also, I'm not even sure if domains have the ability to host a relative URL (don't all static assets get bundled up and deployed to CDN?)

ianmcburnie:
Moving to superbacklog (i.e. closing it)

PR #1202, opened 2020-08-17
Added icons as separate components

agliga:
## Description
Changed all icons to be components. `<ebay-icon name="settings"/>` becomes `<ebay-settings-icon/>`. All attributes should remain the same, but name has been deprecated (and appropriate message should be shown).
Updated importSVG script to create the icons (symbol for icons and such). Index should just import symbols)
Removed all transformers the icons relied on (and some components which needed them.

## Context
All icons should be migrated automatically. Also added migrators for any `icon="something"` attributes to do `<@icon><ebay-setting-icon/></@icon>` instead. Added a common migrator to do this.

## References
#1184


.nycrc
```diff
@@ -8,7 +8,8 @@
     "demo",
     "**/*.marko.js",
     "src/common/test-utils",
-    "src/components/ebay-icon/symbols"
+    "src/components/ebay-icon/symbols",
```
DylanPiercey, line 11: This directory doesn't exist now right?

marko.json
```diff
@@ -1,6 +1,7 @@
 {
   "taglib-id": "ebayui-core",
   "tags-dir": "./src/components",
+  "taglib-imports": ["./src/components/ebay-icon/marko.json"],
```
DylanPiercey, line 4: Looks like `ebay-icon` is still exposed publically. We should move that into `./src/components/components` (the internal components).

agliga: I still want to expose it publicly because of migrators. Since all the new attributes are internal it should be fine being public. I also need a place to show the examples for it.
Once we get rid of the migrators I will move the code to that components subdirectory.

Issue #976, opened 2019-12-06
Discussion: v6.0.0 refactoring and breaking changes


ianmcburnie:
The primary focus of v6.0.0 will be some cleanup, refactoring and breaking changes. No new features, if possible.

Lots of good background info in this closed ticket too: https://github.com/eBay/ebayui-core/issues/895

This is a list of suggestions/ideas that came from team standup (mostly Dylan!). Feel free to amend and append to this list...

* <strike>P1 Transformers/ issues preventing teams already on Marko v5 upgrading to CoreUI 6
   * Split up icons into individual components
   * All icons dependent on icons</strike> - DONE
* <strike>Reduce strict adherence to HTML tag syntax. Find a middle ground between Marko and HTML
   * Remember: "If you know HTML, CSS, and Javascript, you know Marko"</strike> (Added `<@attribute>` tags)
* <strike>Split up notice components (https://github.com/eBay/ebayui-core/issues/1158)
   * page-notice, inline-notice, section-notice, window-notice</strike> - DONE
* <strike>Split up dialog components (https://github.com/eBay/ebayui-core/issues/1164)
   * lightbox-dialog, panel-dialog, fullscreen-dialog</strike> - DONE
* <strike>The HTML-like API...
   * Revert to pure Marko syntax for nested attribute tag
   * Pass actual types instead of strings</strike>
* <strike>Events
   * Remove event name prefixes (we don’t need this anymore since we removed native DOM events with the imperative API)
   * Revisit event data and make consistent (some of the event data is like a random grab bag of stuff)</strike>
* <strike>Review attr tag usage
   * dialog header
   * drawer header
   * toast header
   * stepper step</strike>

Leave for v7

* Does `ebay-` prefix still make sense?
   * We want to make coreui the North Star for Marko components
   * Maybe ‘core-’ instead
* Button (comments originally from @darkwebdev in #923)
   * Does it make sense to call a property "priority" if it has a delete value? Answer: probably not. Maybe rename to "variant". We could move to even higher level of encapsulation with `ebay-primary-button`, `ebay-delete-button` etc (we already have `ebay-expand-button` and `ebay-cta-button`). See: https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/
   * Do we need a value fake-link for variant property when we can just infer it from the href property?  Long term it might be best to split `ebay-button` out into `ebay-fake-button`.

Leave for v8

* Need a consistent I18N strategy
[...]

PR #1210, opened 2020-08-25
Shortned all events and removed all on-{component} prefixes

agliga:
## Description
* All events should have no more {component prefix} after the on
* Added migrators for all (and improved utilities for it, such as having migrators find onCamelCaseEvents as well)
* Added server tests to check the events are being migrated properly for each component (I had some typos so this helped to catch it, since this change is so large)

## Context
* Some events were not working, such as `update` and `input`. I changed `update`  to `move` (on carousel) and `input` to `input-change`. This is a marko limitation. 
* Ran migrator on all templates

## References
#976

### DIFF
--- a/src/common/migrators/index.js
+++ b/src/common/migrators/index.js
@@ -1,3 +1,13 @@
+function convertToCamel(attribute) {
+    const newAttr = attribute.split('-');
+    newAttr.forEach((attr, index) => {
+        if (index > 0) {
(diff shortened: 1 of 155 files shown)


agliga:
Coverage decreased because of the split marko 5 codepath
Once we get marko 5 and are able to run our test suite against it it should increase coverage.

PR #1215, opened 2020-08-27
Split up menu/menu-button from fake-menu/button

agliga:
## Description
Split up menu to be `fake-menu` and normal menu (with checkbox and radio).
I had considered doing a common base but the code has some differences so I left it as two separate components.

## Context
* I left select and keydown events for fake menu/button. This is for the button type so you can see which item was clicked. 
* Most of the component logic in fake is about events and adding expander. 
* Tried to clean up and remove all instances of fake menu from base class. Only one I wasn't sure was about `aria-current` which in docs it says it's for `fake` only but seems valid for "base" usecase non checked or radio.

## References
#976

## Screenshots
Screen Shot 2020-08-27 at 4 16 07 PM


agliga:
Added a change to fix marko 5 support in this PR as well

src/components/ebay-fake-menu/examples/04-sprites/template.marko
```diff
@@ -0,0 +1,26 @@
...
+        background-repeat: none;
+        display: inline-block;
+        height: 20px;
+        margin-right: 8px;
+        vertical-align: middle;
+        width: 26px;
+    }
+</style>
+
+<ebay-fake-menu text="Country">
```
ianmcburnie, line 13: Nitpick. Remove `text="Country"`

src/components/ebay-fake-menu/examples/03-fake-item-disabled/template.marko
```diff
@@ -0,0 +1,5 @@
+<ebay-fake-menu>
+    <@item href="#" current>item 1</@item>
```
ianmcburnie, line 2: Wondering if we should just remove a layer of abstraction here, and have `<@link>` and `<@button>` instead of `<@item>`?

agliga: I was thinking the same thing actually but ended up not doing it because it might be a lot of changes.
But since you brought it up, I'll go ahead and take care of it.

agliga: And the code would still have to change them to `<@item type="button or link">` so we would still support both formats.

Maybe we can do this later since it won't be a breaking change.

ianmcburnie: Yep, let's save for later

Issue #1219, opened 2020-09-02
ebay-infotip: modal should have label on dialog element


mkchang:
# Bug Report

## Description

The new modal option for infotip does not have a way to apply a label to the element with `role="dialog"`. No input to `ebay-infotip` itself, and no input from `ebay-infotip` to the `ebay-dialog-base` to apply that label.

It looks like a `ariaLabelledby` gets generated on the dialog, but it's not associated with anything in my app at least.

## Screenshots
![image](https://user-images.githubusercontent.com/6880917/91917972-5a176480-ec76-11ea-94b6-6071e166d06d.png)

agliga:
The aria-labelledby is for header (however this doesn't have a header)
You can use `a11yCloseText` for the close button. 
What are you looking for exactly to add? I can add an attribute to pass as aria-label or aria-labelledby to dialog if needed.

ianmcburnie:
When the dialog lacks a visible heading for `aria-labelledby`, I wonder if we can reuse the same `aria-label` text that is used on the infotip button itself as the `aria-label` for the dialog?

agliga:
Yes that is easy to do.
I did add a fix just now to remove the `aria-labelledby` if there is no heading so we can use that too.

ianmcburnie:
Fixed in 5.7.0. Closing.

Issue #1218, opened 2020-09-02
ebay-infotip: modal infotip should not have aria-expanded on the button


mkchang:
# Bug Report

## Description

The new modal option for infotip should not have aria-expanded/collpased being announced, as it is not a dialog.

## Screenshots
![image](https://user-images.githubusercontent.com/6880917/91917972-5a176480-ec76-11ea-94b6-6071e166d06d.png)

ianmcburnie:
Fixed in 5.7.0. Closing.

Issue #1198, opened 2020-08-11
ebay-notice: not recognized as landmark


mkchang:
# Bug Report

## Description
ebay-notice is not recognized as a landmark, at least with VO + Safari on Mac. To test, use Ctrl-Opt-U to turn on macOS VoiceOver Rotor and use arrow keys to select 'Landmarks' mode, and see if the notices show up in the list of landmarks. This is a follow up to the issue that originally caused #1185 to be opened.

Unlike in the [Skin notice docs](https://ebay.github.io/skin/#notice), in my app, `aria-hidden` is applied to the icon, which seems to hide the `aria-label` that `ebay-notice` passes to `ebay-icon`, and prevents it from being detected as a landmark. It works as long as I delete that `aria-hidden`. Not sure how to reconcile that, unless `ebay-notice` should treat the icon as a content image, assuming you are providing an aria-label. 
```html
<section role="region" class="page-notice page-notice--attention" aria-labelledby="s0-0-15-2-2-1[0]-1-status">
    <h2 class="page-notice__status" id="s0-0-15-2-2-1[0]-1-status">
        <svg class="icon icon--attention-filled" focusable="false" aria-label="attention" aria-hidden="true">
            <use xlink:href="#icon-attention-filled"></use>
        </svg>
    </h2>
    <div class="page-notice__content">
        This notice is not recognized as a landmark
    </div>
</section>

<section class="page-notice page-notice--attention" role="region" aria-labelledby="attention-status">
    <h4 class="page-notice__status" id="attention-status">
        <svg class="icon icon--attention-filled" focusable="false" height="24" width="24" aria-label="Attention">
            <use xlink:href="#icon-attention-filled"></use>
        </svg>
    </h4>
    <div class="page-notice__content">
        This notice is recognized as a landmark
    </div>
</section>
```

Also explored from our discussion about `ebay-notice` passing `ebay-icon` `a11Text` instead of setting an `aria-label`. This resulted in this HTML for me, but it wasn't recognized as a landmark either, thinking `title` + `aria-labelledby` does not work for landmark detection.
```html
<section role="region" class="page-notice page-notice--confirmation" aria-labelledby="s0-0-15-2-2-1[1]-1-status">
    <h2 class="page-notice__status" id="s0-0-15-2-2-1[1]-1-status">
        <svg class="icon icon--confirmation-filled"
            focusable="false" aria-labelledby="s0-0-15-2-2-1[1]-1-5-text" role="img">
[...]

PR #1221, opened 2020-09-02
Fixed aria issues with notice and infotip mini

agliga:
## Description
This fixes three bugs
* Fixed notice icon to use `aria-label`. Added `a11y-variant="label"` which will allow this
* Fixed dialog-mini to not use expander (and won't have `aria-expanded`)
* Fixed dialog-mini to pass in `aria-label` from notice.

## References
#1218 
#1219 
#1198


ianmcburnie (approved):
Tested locally with VoiceOver & Safari. All looks good. Thanks.

Issue #1177, opened 2020-07-16
Carousel: append carousel title/heading text to aria-label of prev/next paddles


ianmcburnie:
For screen readers, we can automatically provide additional button context by stitching the title of the carousel to the aria-label (separated by a hyphen/dash).

ianmcburnie:
Looks like  #609 makes this request redundant: 

https://github.com/eBay/ebayui-core/pull/1222#pullrequestreview-482185428

Issue #609, opened 2019-04-02
Carousel: add aria-roledescription


ianmcburnie:
Something I've been wanting to look into for a while, so putting it on backlog. ARIA 1.1 now gives the ability to create custom role descriptions for groups, regions and widgets. Might be something we can leverage for the carousel (e.g. aria-roledescription=slide).

<blockquote>
The aria-roledescription property gives authors the ability to override how assistive technologies localize and express the name of a role. Thus inappropriately using aria-roledescription may inhibit users' ability to understand or interact with an element. Authors SHOULD limit use of aria-roledescription to clarifying the purpose of non-interactive container roles like group or region, or to providing a more specific description of a widget.
</blockquote>

https://www.w3.org/TR/wai-aria-1.1/#aria-roledescription

UPDATE: I have updated the MIND patterns to use `aria-roledescription=carousel`, `role=group` & `aria-labelledby`. Related ticket: https://github.com/eBay/mindpatterns/issues/16

ianmcburnie:
Punting this one to v6.

PR #1222, opened 2020-09-03
Added carousel role description 

agliga:
## Description
Added role description. Will generate an ID used for `aria-labelledby` on the root. This ID is only available when there is a clipped text (`a11yStatusText or `a11yHeadingText`, or is `discrete`).
Added `role="group"` and `aria-roledescription="carousel"`
Also appended title of carousel to buttons. 

## References
#609
#1177

### DIFF
--- a/src/components/ebay-carousel/README.md
+++ b/src/components/ebay-carousel/README.md
@@ -28,7 +28,7 @@ Name | Type | Stateful | Required | Description
 
 Name | Type | Stateful | Required | Description
 --- | --- | --- | --- | ---
-`a11y-status-text` | String | No | Yes | status text (default: "Showing Slide {currentSlide} of {totalSlides} - Carousel")
+`a11y-status-text` | String | No | Yes | status text (default: "Showing Slide {currentSlide} of {totalSlides}")
 `a11y-status-tag` | String | No | Yes | use h1-h6 when there isn't a visible heading before the carousel (default: "span")
 `autoplay` | Boolean or Number | No | No | automatically slides the carousel on an interval. If a number is supplied that is used as the interval in ms, defaults to 4000ms.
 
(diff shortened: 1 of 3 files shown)


ianmcburnie (changes requested):
Good news: the `aria-roledescription` change negates the need for appending the carousel title to each paddle, as the screen reader now automatically announces the roledescription (carousel) and label (top products).

Bad news: you'll need to revert the changes to paddle as it results in carousel and top products being announced twice  :-)

Also, we should remove any mention of the word carousel from our a11y text examples, as this is now handled by the screen reader via the roledescription value.

agliga:
> Good news: the `aria-roledescription` change negates the need for appending the carousel title to each paddle, as the screen reader now automatically announces the roledescription (carousel) and label (top products) when focus first moves inside the carousel group.
> 
> Bad news: you'll need to revert the changes to paddle as it results in carousel and top products being announced twice :-)
> 
> Also, we should remove any mention of the word carousel from our a11y text examples, as this is now handled by the screen reader via the roledescription value.

Done, reverted the change. Also I removed ` - carousel` from the default text

ianmcburnie:
Updates look good, but for some reason the first few examples don't have the ARIA attributes:

Screen Shot 2020-09-03 at 9 18 31 PM

agliga:
> Updates look good, but for some reason the first few examples don't have the ARIA attributes:
> 
> Screen Shot 2020-09-03 at 9 18 31 PM
This is because I added a condition to not add `role=group` and `aria-describeby` if there is not `a11yText` or header text. Otherwise theres nothing to point `aria-labelledby` to and I thought that those two attributes dont work without aria-labelledby

ianmcburnie:
I see what you mean. I think role="group" and aria-roledescription="Carousel" can be ever present, but as you say for continuous case, the aria-labelledby will only be added when dev adds `a11y-heading-text` and `a11y-heading-tag`.

Oops, and yes I should be enjoying the day off, so please ignore this until next week ;-)

agliga:
> I see what you mean. I think role="group" and aria-roledescription="Carousel" can be ever present, but as you say for continuous case, the aria-labelledby will only be added when dev adds `a11y-heading-text` and `a11y-heading-tag`.
> 
> Oops, and yes I should be enjoying the day off, so please ignore this until next week ;-)

Done, and yes you should check this next week :+1:

ianmcburnie (approved):

Issue #1225, opened 2020-09-08
Wrap the <svg> tag stamped into <body> (by ebay-icon) with <div hidden>


dsathyakumar:
# Bug Report

## Description
An `<svg>` tag inherits the SVGElement Interface which does not inherit the HTMLElement interface, thereby not getting the **hidden** property.
Only the HTMLElement interface has the **hidden** property. When this is used by accessibility plugins like makeup-js to add or remove the **hidden** property, it ends up breaking the layouts.

## Workaround
Similar to https://github.com/eBay/skin/blob/master/docs/_includes/ds6/symbols.html#L1 use a div with a **hidden** prop.

ianmcburnie:
Thanks @dsathyakumar. Which version of CoreUI do you need this fix in?

dsathyakumar:
hey @ianmcburnie, i'm guessing this is on all the `ebayui-core` versions thus far.
We do not need this fix as the GH is not on `ebayui-core`.

But all the domain teams that use `ebayui-core` would require this.
The Header relies on the `makeup-screenreader-trap` for accessibility purposes and we are attempting to use the **hidden** prop, provided as part of https://github.com/makeup/makeup-js/pull/9.
But, using this, would end up breaking the domain pages as they use `ebay-icon` that stamps `<svg>` under the `<body>`. 

Also, in the meanwhile, leaving it to you if https://github.com/makeup/makeup-js/issues/10 seems to be a fix that can be made in the interim, while domain teams upgrade.

agliga:
Released for `5.7.1`

PR #1226, opened 2020-09-08
Added hidden div wrapper to Icons

agliga:
## Description
Changed icon wrapper to be a hidden div which has an svg inside it.

## References
#1225


ianmcburnie:
@agliga @dsathyakumar Do we think there are any chances of this being a breaking change for teams? Slightly concerned of any unintended, downstream effects, although I can't see any reason why anybody would go poking about for this in the DOM...

ianmcburnie (approved):

PR #1228, opened 2020-09-09
Fixes to sync up ebayui with skin changes

agliga:
## Description
* Renamed ebay-toast to ebay-toast-dialog
* Renamed ebay-drawer to ebay-drawer-dialog  
* Changed label to floating-label
* Added classes to tooltip for event handlers
* Updated base readme with all the modules that have been changed.

## Context
Kept ebay-toast and ebay-drawer but added migrators for those to change the name. Left the original migrators in the base components since people using the new components will use the new events. 
Also CI will fail since it needs the latest skin to be published


ianmcburnie (approved):
Tested locally. No issues. Great stuff.

Issue #1229, opened 2020-09-09
Fullscreen-Dialog: use `useHiddenProperty ` option in makeup-screenreader-trap


ianmcburnie:
makeup-screenreader-trap was recently updated with an option to allow use of `hidden` property instead of `aria-hidden`. This makes more sense for fullscreen dialogs (and solves some bugs in Talkback) and we should make use of it.

https://github.com/makeup/makeup-js/issues/5

ianmcburnie:
@agliga this one needs to go out in tandem with the SVG fix. I also have another patch lined up for screenreader-trap which should help mitigate this edge case:

https://github.com/makeup/makeup-js/issues/11

PR #1230, opened 2020-09-10
Updated fullscreen dialog to pass in useHiddenProperty argument

agliga:
## Description
Added option of `useHiddenProperty`. (I had to set the default to false if null or undefined because it was failing due to the check being useHiddenProperty === false in makeup)
Added a test to check it is being set.
Set the attribute true for fullscreen dialog only

## References
#1229

## Screenshots
Screen Shot 2020-09-10 at 9 41 46 AM

### DIFF
--- a/src/components/components/ebay-dialog-base/component.js
+++ b/src/components/components/ebay-dialog-base/component.js
@@ -105,6 +105,18 @@ module.exports = {
         }
     },
 
+    _getTrapCallback(restoreTrap, isTrapped, wasTrapped) {
+        const useHiddenProperty = this.input.useHiddenProperty || false;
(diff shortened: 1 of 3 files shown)


ianmcburnie (changes requested):
Looks like we have a timing/ordering issue on our hands. The trap is being executed and the hidden property is being applied to background content before the transition completes and it looks very odd.

agliga:
> Looks like we have a timing/ordering issue on our hands. The trap is being executed and the hidden property is being applied to background content before the transition completes and it looks very odd.

Done. I made the traps trigger after the transition.

ianmcburnie (approved):
Nice! Let's also hold off on the patch release until I've made this patch in makeup too: https://github.com/makeup/makeup-js/issues/11

PR #1241, opened 2020-09-21
Added aria-expanded to drawer handle and default aria-label

agliga:
## Description
Missing `aria-expanded` for drawer. Added it along with default aria-label text


ianmcburnie (changes requested):
As discussed offline, let's go with `a11y-maximize-text` and `a11y-minimize-text` and no ARIA expanded or pressed, and I'll test it out.

agliga:
> As discussed offline, let's go with `a11y-maximize-text` and `a11y-minimize-text` and no ARIA expanded or pressed, and I'll test it out.

Done,
basically when it's minimized will use `a11yMaximizeText` and when maximized will use `a11yMinimizeText`. This makes the most sense since you want to show the text of the action you will do.

ianmcburnie (approved):

ianmcburnie (changes requested):
Oh wait, is this a breaking change....

agliga:
> Oh wait, is this a breaking change....

I was thinking about it... 
We can set the default initially to what `a11yHandleLabel` is. If its not set, then we use the maximize or minimize
However I think that since the default is there we might not need it unless people are using the `a11yHandleText`

ianmcburnie (approved):

Issue #1246, opened 2020-09-28
Combobox: add option to always show all options on focus


ianmcburnie:
We originally had a use case (which we thought might have been a bit of an edge case) whereby the autocomplete combobox is filled with a value on page load. Because the combobox is pre-filled with a value, and the filtering is active, when a user focuses on the combobox they do not see the full list of combobox suggestions, only the filtered list (which can often just be one item). As a kind of workaround, we added an option where the chevron is a separate button element that when clicked always shows the full list of suggestions. However, there is a discoverability issue with this approach and it still feels far from perfect.

Another potential option is to always show the full list of suggestions on focus (apologies to those who have suggested this in the past, but we were a little anxious about it), like is done here:

https://material-ui.com/components/autocomplete/

Let's code it up and see how it behaves, see if it results in other weirdness we might not have anticipated, and take it from there.

NOTE: that material example has a few bugs which we definitely do not want to replicate!

Issue #1201, opened 2020-08-17
Combobox: expanded/collapsed icon doesn't change after pressing "enter" key


joinfelipe:
# Bug Report

## Description
Combobox's  caret icon no longer expands list options after making a selection using keyboard's "enter" key.
1. go to a page with combobox options, e.g: https://ebay.github.io/ebayui-core/ds4/?path=/story/ebay-combobox--lots-of-options
2. type "Basic Offer"
3. navigate with arrow-down key and select first option "Basic Offer" using keyboard's enter key
4. try clicking on the caret expand icon "v"
*Actual Behavior*
5. combobox does not expand
*Expected Behavior*
5. Combobox should expand and display available options

## Workaround
User can still expand options and make selection when using "arrow-down" key, or when leaving the field and clicking on it again.

## Screenshots
![image](https://user-images.githubusercontent.com/69696226/90433233-639eab00-e080-11ea-858a-3e5168287bc5.png)

agliga:
The reason this is happening. Might need Ian's accessibility expertise on what the right way to do this is:
Combobox shows on focus. When selecting an option the focus is placed on the input, however the expander is closed. Clicking again does not trigger focus since you are already focused. 
A solution would be to trigger the expander on click if it the field is focused.

ianmcburnie:
I was tempted to close as wontfix, as the user can click out and click back in again, or use down arrow key, to expand. However, looking at the MIND patterns example, I see I have it that combobox expands if it receives a mouse click:

https://ebay.github.io/mindpatterns/input/combobox/index.html

https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.0pattern/combobox-autocomplete-none.html

This is low priority, but we should make it consistent.

Issue #1234, opened 2020-09-14
Storybook: cleanup, rename and reorganise stories


ianmcburnie:
I think we can do a better job with our stories to make them a little more professional and easier to follow.

ianmcburnie:
Related to this, I think in all cases where an attribute is listed as required, then it should be present in every story.

For example, the pagination readme lists the 3 ally text attributes as required, but not all examples/stories show them, which is a little confusing and leads to a11y bugs.

ianmcburnie:
Closing, as this is a great step forward, but I think we can do another sweep sometime and try and adopt some kind of convention, for example: https://github.com/eBay/ebayui-core/pull/1311#issuecomment-767748098

Issue #1224, opened 2020-09-08
Combobox: add "automatic" keyboard mode


ianmcburnie:
Combobox currently only supports "manual" keyboard mode, where ENTER key must be pressed to fill the combobox with the currently highlighted value. We should support "automatic" mode too, where the comobobox value updates as user cycles through options, as in this example:

https://ebay.github.io/mindpatterns/input/autocomplete/index.html?console=Playstation+4

PR #1247, opened 2020-09-29
Combobox

agliga:
## Description
* Changed combobox to show all suggestions on initial showing
* Changed when clicking combobox to expand
I tried to get the combobox auto mode working, but it might need more changes and might be unsafe for `6.0.0`

## Context
Basically when it shows sets a state variable. That state variable changes when typing to hide all the options.
Had some minor changes to active descendant to make it work properly

## References
#1246 
#1201 

## Screenshots
Screen Shot 2020-09-29 at 4 54 47 PM

### DIFF
--- a/src/components/ebay-combobox/component.js
+++ b/src/components/ebay-combobox/component.js
@@ -23,15 +23,24 @@ module.exports = {
         this.emit('button-click', { originalEvent });
     },
 
-    handleExpand() {
-        const selectedEl = this.getEls('options')[
-            this._getVisibleOptions().indexOf(this._getSelectedOption())
-        ];
-
+    setSelectedView() {
+        const current = this._getVisibleOptions().indexOf(this._getSelectedOption());
+        this.activeDescendant.index = current;
+        const selectedEl = this.getEls('options')[current];
         if (selectedEl) {
             elementScroll.scroll(selectedEl);
         }
+    },
 
+    handleExpand() {
+        if (this.state.viewAllOptions) {
+            this.setSelectedView();
+        } else {
+            this.state.viewAllOptions = true;
+            this.once('update', () => {
+                this.setSelectedView();
+            });
+        }
         this.emit('expand');
     },
 
@@ -78,6 +87,7 @@ module.exports = {
                     this.expander.expand();
                 }
             });
+            this.state.viewAllOptions = false;
 
             this._emitComboboxEvent('input-change');
         });
@@ -110,7 +120,7 @@ module.exports = {
         input.autocomplete = input.autocomplete === 'list' ? 'list' : 'none';
         input.options = input.options || [];
         this.lastValue = input.value;
-        this.state = { currentValue: this.lastValue };
+        this.state = { currentValue: this.lastValue, viewAllOptions: (this.state && this.state.viewAllOptions) || true };
     },
 
     onMount() {
@@ -138,12 +148,12 @@ module.exports = {
                 this.getEl('combobox'),
                 this.getEl('listbox'),
                 '[role="option"]', {
-                    activeDescendantClassName: 'combobox__option--active',
-                    autoInit: -1,
-                    autoReset: -1,
-                    axis: 'y',
-                    autoScroll: true
-                }
+                activeDescendantClassName: 'combobox__option--active',
+                autoInit: -1,
+                autoReset: -1,
+                axis: 'y',
+                autoScroll: true
+            }
             );
 
             this.expander = new Expander(this.el, {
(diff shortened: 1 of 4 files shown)


ianmcburnie:
When you say "auto" mode, do you mean https://github.com/eBay/ebayui-core/issues/1224 ?

agliga:
> When you say "auto" mode, do you mean #1224 ?

Yes, initially since there was some changes, I was trying to sneak that in. I got about 90% of the way but there were some minor edge case bugs so I decided to skip it for `6.0.0`. (I know it's `6.1.0`, but since we're doing some combobox changes I thought it wouldn't hurt since we have to regress anyways)

ianmcburnie:
Fine by me. It will be nice to see how manual and auto modes play out with this "show all" change. Auto mode might be a bit funky, we shall see. We can tell QA to do combobox last to buy us more time.

agliga:
> Fine by me. It will be nice to see how manual and auto modes play out with this "show all" change. Auto mode might be a bit funky, we shall see. We can tell QA to do combobox last to buy us more time.

👍 I'll unstash my changes and try to get it work

src/components/ebay-combobox/README.md
```diff
@@ -44,11 +44,15 @@ Name | Type | Stateful | Required | Description
 `borderless` | Boolean | No | No | whether the input has borders
 `disabled` | Boolean | Yes | No | sets the disabled attribute of the input
 `expanded` | Boolean | Yes | No | sets whether the listbox is expanded
-`autocomplete` | String | Yes | No | default is `none`; available values are `none` and `list`
+`autocomplete` | String | Yes | No | default is `none`; available values are `none`,  `list`, and `automatic`. See below for more info.
```
ianmcburnie, line 47: So it's probably far from obvious, but this `autocomplete` attribute is intended to map to the `aria-autocomplete` attribute. Currently we only support `inline` and `none` from ARIA. We should move `automatic` to be controlled by a different attribute. Maybe call it `selection`?

More info here on "List autocomplete with manual selection" & "List autocomplete with automatic selection":

https://www.w3.org/TR/wai-aria-practices-1.1/#combobox

ianmcburnie: Actually come to think of it, maybe we can still use the `autocomplete` attribute, but with the values as `list-manual` and `list-automatic`...? This would allow us to add `list-inline` in future (see WCAG example of that).

agliga: Should the `roledescription` change for list automatic (like adding the input will be changed?) or fine leaving as is. Basically automatic and manual will be the same, options change based on text input:
```
        (`Use up and down arrow keys to navigate options.${
            input.autocomplete === 'none' ? '': ' Options change based on text input'
        }`);
```

ianmcburnie: I think that's a good idea. Maybe, "Use up and down arrow keys to choose an option"? We can refine the text as we go. We need to keep it as short and sweet as possible.

I also notice a bug with the 1st default example. It shouldn't be announcing "options will change based on input".

agliga: So the first one should be correct since it's `none`. 
I assume `none` should be the default behavior. Let me know if we should change the default?

ianmcburnie: So sorry for the back and forth on this, but I think my instinct was right the first time, list selection type should be a separate attribute. This is because we also might want to configure a non-autocomplete combobox to be manual or automatic. So to summarize:

`autocomplete`: "none", "list" (in future we might also support ARIA values of "inline" and "both"
`list-selection`: "manual" or "automatic"

Where it might get a little quirky, is that for a basic combobox with `autocomplete` set to "none" (which are actually quite rare) we probably want `list-selection` to be manual by default. But for a combobox with `autocomplete` as "list", it probably makes more sense for `list-selection` to be automatic by default.

agliga: Hopefully last one
I split up the `autocomplete` and `list-selection`.
So by default `list-selection` is `automatic` and only when they set it to `manual` then it doesn't select as they key up/down 
And they can mix `autocomplete` and `list-selection`

ianmcburnie: Looks good. Can you please add an example with `list-selection` manual. We can cleanup and reorder these stories later as part of #1234

agliga: There already is one, I moved some of the examples around: https://github.com/eBay/ebayui-core/pull/1247/files#diff-ccae6cf296486645387efb94ef728f77R1-R7

ianmcburnie: Oh sorry, local env problem. I deleted example 03.

ianmcburnie:
The first "default" example now seems to be filtering suggestions as I type.

agliga:
> The first "default" example now seems to be filtering suggestions as I type.

This should be fixed, there was a typo in `autocomplete`.

ianmcburnie (approved):
Things seems to be working great. Nice job!

PR #1250, opened 2020-10-12
Added check-postpublish script to build 

agliga:
## Description
There is a possibility when testing that changes from `prepublish` (which is the changed `marko.json` file) might creep in. 
This adds a check that whenever we build ebayui, we run the `postpublish` script and check that there are no changes in `marko.json`. If there are, then the build fails. 
Added this test to run when pushing code to a branch and as a build script.


ianmcburnie (approved):
Nice.

PR #1268, opened 2020-11-12
Changed node package manager from yarn to npm

agliga:
## Description
Added support for npm and removed yarn support

## Context
* Since we are using yarn resolutions and npm does not support that yet, I added a package to support the resolutions. We need this for `gh-highlight` which we can probably remove if we deprecate and change the old marketplace site (it's used for the code hinting)
* Updated all places to use `npm run` instead of `yarn`


austingardner (approved):
LGTM

ianmcburnie (approved):

Issue #1000, opened 2019-12-20
Menu Button: enhance typeahead functionality to multiple characters (feature request)


ianmcburnie:
The current typeahead is very basic, supporting 1 character only. There has been an ask to enhance this to allow multiple characters (very helpful for menus with long data lists) in a given timeframe (1000ms or so I'd say).

I think a portion of this logic could be extracted out into a small utility method, maybe even a [makeup](https://github.com/makeup) module. The utility would simply return the array of characters entered within a given time range.

PR #1269, opened 2020-11-13
1000 typeahead menu button

austingardner:
## Description 
This PR adds the enhanced multicharacter typeahead functionality that is added to makeup to the menu. 
I also added tests to the menu button to assure that this functionality is working. 
Added makeup-typeahead to deps. 
Updated makeup-floating-label to most recent version. 

## References
https://github.com/eBay/ebayui-core/issues/1000

## Screenshots
Already demo'd -> can add them if needed


src/components/ebay-menu/component.js
```diff
@@ -2,9 +2,11 @@ const assign = require('core-js-pure/features/object/assign');
 const findIndex = require('core-js-pure/features/array/find-index');
 const scrollKeyPreventer = require('makeup-prevent-scroll-keys');
 const rovingTabindex = require('makeup-roving-tabindex');
+const typeahead = require('makeup-typeahead');
 const eventUtils = require('../../common/event-utils');
 const menuUtils = require('../../common/menu-utils');
-const NodeListUtils = require('../../common/nodelist-utils');
+
+const TIMEOUT_LENGTH = 1300;
```
ianmcburnie, line 9: Maybe just rename this variable so that its purpose is clear. Maybe `TYPEAHEAD_TIMEOUT_LENGTH`.

src/components/ebay-menu/component.js
```diff
@@ -48,7 +50,7 @@ module.exports = assign({}, menuUtils, {
     },
 
     handleItemKeypress({ key }) {
-        const itemIndex = NodeListUtils.findNodeWithFirstChar(this.getEl('menu').children, key);
+        const itemIndex = typeahead(this.getEl('menu').children, key, TIMEOUT_LENGTH);
```
ianmcburnie, line 53: Nice clean interface. I like it.

ianmcburnie: Just thought though, do you cover for edge cases like missing menu element or zero children?

austingardner: yes -> added tests in the makeup repo for those cases

ianmcburnie (commented):
I think it would be nice to have a new story/example that better demonstrates the new typeahead.

ianmcburnie (commented):
It seems to be doing a kind of fuzzy-match now. In the new example, when I press 'i' it goes to the first entry that has an 'i' anywhere in it, and not at the start. Is this intentional?

austingardner:
Yes, the fuzzy match is intentional. If there is no match at the start of the string, then the typeahead will look anywhere.

ianmcburnie:
Okay, wasn't totally expecting it, but we can rollout with that and can always make it configurable to offer a more linear option if necessary.

src/components/ebay-menu-button/examples/18-typeahead/template.marko
```diff
@@ -0,0 +1,53 @@
+<style>
```
ianmcburnie, line 1: You can removed this style section from this example as it's not used.

austingardner: done

ianmcburnie (approved):

src/components/ebay-menu/component.js
```diff
@@ -2,9 +2,11 @@ const assign = require('core-js-pure/features/object/assign');
 const findIndex = require('core-js-pure/features/array/find-index');
 const scrollKeyPreventer = require('makeup-prevent-scroll-keys');
 const rovingTabindex = require('makeup-roving-tabindex');
+const typeahead = require('makeup-typeahead');
 const eventUtils = require('../../common/event-utils');
 const menuUtils = require('../../common/menu-utils');
-const NodeListUtils = require('../../common/nodelist-utils');
+
+const TYPEAHEAD_TIMEOUT_LENGTH = 1300;
```
agliga, line 9: @ianmcburnie Would we want to allow this value to be overridden by teams?
If so we can do this where we read the value: `this.input.typeaheadTimeoutLength || TYPEAHEAD_TIMEOUT_LENGTH;`

src/components/ebay-menu/component.js
```diff
@@ -118,12 +120,19 @@ module.exports = assign({}, menuUtils, {
         });
 
         scrollKeyPreventer.add(this.contentEl);
+
+        const { getIndex, destroy } = typeahead();
+        this.getIndex = getIndex;
```
agliga, line 125: Can we call this something like: `getTypeaheadIndex` or something else? `getIndex` seems pretty generic in this case and might be a confusing name.

src/components/ebay-menu/test/mock/index.js
```diff
@@ -30,3 +30,19 @@ exports.Fake_2Items = assign({}, exports.Basic_2Items, {
         renderBody: createRenderBody(`Item text ${i}`)
     }))
 });
+
+function getCountry(num) {
```
agliga, line 34: What's the point of having this as a function?
You can probably use an array instead with various countries. 
Like: 
const countries = ['Albania', 'Alcania', 'Something else'];

`item ${i} ${countries[i]}`;

src/components/ebay-menu/test/test.browser.js
```diff
@@ -11,6 +11,29 @@ afterEach(cleanup);
 /** @type import("@marko/testing-library").RenderResult */
 let component;
 
+describe('typeahead functionality', () => {
```
agliga, line 14: Do we want a test to check the timeout as well, like:
Type a
Wait for 2 seconds
Type l
Should focus on item starting with l

Or is this covered in the makeup portion?

Also do we want to check the functionality if nothing matches? Like if someone types in "a", "l" "x" that nothing is focused or something, or the last one is still focused?

agliga (commented):
Couple of comments

Overall looks really good. I think just needs some minor cleanup (and some of these are pretty nitpicky as well)

src/components/ebay-menu-button/examples/18-typeahead/template.marko
```diff
@@ -0,0 +1,41 @@
+<h3>
+    Click the button, then type abcdefghijk to see focus change.
```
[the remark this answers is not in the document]
austingardner: I guess what I was going for was that they could type and sequentially go down the list. typing ab then abc would move the focus from ab to abc. But yeah if you don't like it we could change it.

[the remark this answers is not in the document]
agliga: That's fine, lets just think of a better example. 
Maybe even Country, State, city, like
`United States`
`United States, California`
`United States, California, Los Angeles`
`United States, California, San Francisco`
`United States, California, San Jose`
Etc
Just something that came off the top of my head. We can do something else, maybe countries or such.

Ideally this would cover users usecases since the abc isn't a real usecase, but we can think of things that can coincide.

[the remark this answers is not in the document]
ianmcburnie: Yeah, I think we should change it for something more real world

src/components/ebay-menu/component.js
ianmcburnie: I don't think we'd want them to override.

src/components/ebay-menu/component.js
ianmcburnie: Agreed.

src/components/ebay-menu/test/test.browser.js
ianmcburnie: I think the nothing matches would be a good test case.

austingardner: done

src/components/ebay-menu/component.js
```diff
@@ -2,9 +2,11 @@ const assign = require('core-js-pure/features/object/assign');
 const findIndex = require('core-js-pure/features/array/find-index');
 const scrollKeyPreventer = require('makeup-prevent-scroll-keys');
 const rovingTabindex = require('makeup-roving-tabindex');
+const typeahead = require('makeup-typeahead');
 const eventUtils = require('../../common/event-utils');
 const menuUtils = require('../../common/menu-utils');
-const NodeListUtils = require('../../common/nodelist-utils');
```
agliga, line 7: Can we remove this whole nodelist-utils file and it's test? If it's not being used anymore no need to keep it around.

src/components/ebay-menu/test/test.browser.js
```diff
@@ -11,6 +11,56 @@ afterEach(cleanup);
...
+
+        expect(component.getByText(firstItemText).parentElement.getAttribute('tabindex')).to.equal('0');
+    });
+
+    it('restarts the search from the beginning after it waits', async() => {
+        await fireEvent.click(component.getByText(firstItemText));
+        await fireEvent.keyPress(component.getByText(firstItemText), { key: 'a', code: 65 });
+        await fireEvent.keyPress(component.getByText(firstItemText), { key: 'l', code: 76 });
+        await fireEvent.keyPress(component.getByText(firstItemText), { key: 'b', code: 66 });
+        setTimeout(() => {}, 100);
```
agliga, line 54: Can we do this instead
`await new Promise(resolve => setTimeout(resolve, 100));`
This seems cleaner and then I don't think you need the `waitFor`. 
It just bugs me seeing the empty function call

src/components/ebay-menu/test/test.browser.js
```diff
@@ -11,6 +11,56 @@ afterEach(cleanup);
...
+    describe('first', () => {
+        beforeEach(async() => {
+            await fireEvent.click(component.getByText(firstItemText));
+            await fireEvent.keyPress(component.getByText(firstItemText), { key: 'a', code: 65 });
+            await fireEvent.keyPress(component.getByText(firstItemText), { key: 'l', code: 76 });
+            await fireEvent.keyPress(component.getByText(firstItemText), { key: 'c', code: 67 });
+        });
+
+        it('shows the correct item in focus when the user types', async() => {
+            await waitFor(() => {
```
agliga, line 32: Did you try to do this without the `waitFor`, not sure if it's needed. 
In the text below you have a `waitFor` and in there there are other `fireEvents` and an assertion without a `waitFor`

agliga (commented):
Couple of final nitpicks, after that it should be good.

One more request, can we add a little blurb about this functionality in the `README.md`, so that teams will know it exists.

agliga (approved):

Issue #1270, opened 2020-11-20
Progress Spinner: new component


ianmcburnie:
A coreui version of Skin's progress spinner: https://ebay.github.io/skin/#spinner

PR #1272, opened 2020-11-25
1270 Progress Spinner

austingardner:
## Description
Adds progress spinner component into ebay-ui, which has an existing skin module. 
Important files start [here](https://github.com/eBay/ebayui-core/pull/1272/files#diff-f47d3060cb5c17ed9f2835f82319cb353dc9cbd01417f6ce3107457428fd09ffR1) and below

## References
https://github.com/eBay/ebayui-core/issues/1270

## Screenshots
progress-spinner


src/components/ebay-progress-spinner/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+```marko
+<ebay-progress-spinner large />
+```
+
+## Attributes
+
+Name | Type | Stateful | Required | Description
+--- | --- | ---- | --- | ---
+`large` | Boolean | No | No | If true, large spinner is displayed; else, normal
```
ianmcburnie, line 11: I vote for making this an enumerated type rather than a boolean. For example, size="large". With "small" being the default size. Reason being we may add other sizes in future (extra-small, extra-large, etc).

ianmcburnie:
I changed the target branch to 7.0.0 and there is now one conflicting file that needs resolving.

austingardner:
Conflict resolved

src/components/ebay-progress-spinner/README.md
```diff
@@ -0,0 +1,11 @@
+# ebay-progress-spinner
+
+```marko
+<ebay-progress-spinner size="large" />
+```
+
+## Attributes
```
ianmcburnie, line 7: @austingardner Sorry, I forgot to mention that we'll also need an `a11y-text` attribute that is used as the value for the aria-label.

ianmcburnie (changes requested):
Hmm, something's not right here. The `a11ytext` attribute should not be showing up in the rendered html. Maybe check to see how another component is doing it.

Screen Shot 2020-12-02 at 6 26 32 PM

austingardner:
This is happening because I saw another component doing `...input` and so I figured I would follow that pattern, so that any additional arguments would be passed into the span... 

What is the best way to do the `const {a11yText, ...rest} = input` syntax I would have used, but in pre-es6? We still want to spread the rest of the input into the span, right?

ianmcburnie:
Actually, I see that badge and infotip are using a direct `aria-label` pass through attribute, rather than `a11y-text`. Let's check with Andrew when he gets back and see if we need to make this consistent everywhere.

src/components/ebay-progress-spinner/index.marko
```diff
@@ -0,0 +1,2 @@
+$ var label = input.a11yText ? input.a11yText : "Busy"
+<span ...input class=[ "spinner", input.size === "large" && "spinner--large"] aria-label=label role="img"></span>
```
agliga: Just so it's here (we chatted about this), use `processHTMLAttributes` instead and set `a11yText` and `size` in the ignored attributes.

agliga: For `class`, attribute, allow the users to override it by adding:
`input.class` in the array.
Also you should

src/components/ebay-progress-spinner/marko-tag.json
```diff
@@ -0,0 +1,7 @@
+{
+    "@size": {
```
agliga, line 2: You will need to add:
```
  "attribute-groups": ["html-attributes"],
  "@*": {
    "targetProperty": null,
    "type": "expression"
  },
  "@html-attributes": "expression"
```
(The only one I know for sure is needed is the `@*` one for now, not sure if the rest are needed, but for constancy, lets add them for now)

src/components/ebay-progress-spinner/marko-tag.json
```diff
@@ -0,0 +1,7 @@
...
+      "enum": ["small", "large"]
+    },
+    "@a11y-text": "string"
```
agliga, line 5: Also add `"@role": "never"`, since you are setting it

src/components/ebay-progress-spinner/index.marko
```diff
@@ -0,0 +1,8 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+static var ignoredAttributes = [
+    "a11yText",
+];
+
+$ var label = input.a11yText ? input.a11yText : "Busy";
+<span ...processHtmlAttributes(input, ignoredAttributes) class=[ "spinner", input.size === "large" && "spinner--large"] aria-label=label role="img"></span>
```
agliga, line 8: Make sure you add `input.class` to the `class` array so users can add their own classes as needed.

src/components/ebay-progress-spinner/index.marko
```diff
@@ -0,0 +1,8 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+static var ignoredAttributes = [
+    "a11yText",
```
agliga, line 4: Add `size` and `class` as well. Basically anything you are using as `input` should be here.

agliga (approved):

ianmcburnie:
I'm still seeing a discrepancy in how aria-label and ally-text are used here, compared to badge and infotip. @agliga Shouldn't i be `aria-label` here as it's a direct pass through?

agliga:
> I'm still seeing a discrepancy in how aria-label and ally-text are used here, compared to badge and infotip. @agliga Shouldn't it be `aria-label` here as it's a direct pass through?

I think so. 
From what I can tell is `a11yText` is used when we need to add accessibility to another element and not root and when we force users to use pass it in. IE on notice we use `a11yText` to put `aria-label` on the notice icon and we require it because the icon needs to announce what state it is.
However in cases where it is not needed and optional, like on button we allow `aria-label` as passthrough.
Possibly we might need to add it to badge for consistency since badge should require `aria-label`.

ianmcburnie:
Thanks, Andrew. It's pretty much going to always be required in this case, as the spinner graphic conveys information/state/

src/components/ebay-progress-spinner/index.marko
```diff
@@ -0,0 +1,8 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+static var ignoredAttributes = [
+    "class",
+    "size"
+];
+
+<span ...processHtmlAttributes(input, ignoredAttributes) class=[ "spinner", input.size === "large" && "spinner--large", input.class] role="img"></span>
```
agliga, line 8: Fix thr space after class=[
Additionally run prettyprint https://github.com/marko-js/cli/blob/master/packages/prettyprint/README.md

agliga (commented):
Lgtm overall just a minor formatting issue

src/components/ebay-progress-spinner/index.marko
austingardner: done

ianmcburnie (approved):
Nice! Congrats on your first ebayui component! :-)

Issue #1275, opened 2020-12-09
ebay-tourtip: pass the event to be able to stop propogation when close tourtip


AnnZhao:
# Feature Request

## Description
pass the event to be able to stop propogation when close tourtip

PR #1288, opened 2020-12-11
Updated original event to be passed from the tooltip-base

agliga:
## Description
Had to pass original event from `overlay-close` in order to pass it up
Also updated chromedriver

## References
#1275


ianmcburnie (approved):

austingardner (commented):
LGTM

Issue #1285, opened 2020-12-10
Panel dialog: 'position="end"' rendered on the root 'div' tag


melquan:
# Bug Report

## Description
When the `ebay-panel-dialog` is specified with `position="end"`, this attribute is rendered on the root `div` tag.

## Workaround
None

## Screenshots

![image](https://user-images.githubusercontent.com/19419138/101706790-bd1b5300-3a3e-11eb-99bf-465492160795.png)

Issue #1281, opened 2020-12-10
Notice(docs): update examples with the new 'a11y-text' attribute


melquan:
# Documentation

## Description
1. `ebay-page-notice`:
   a. 03-page-notice-celebration example has previous `a11y-heading-text` attribute instead of `a11y-text`
   b. 04-custom-heading example example has `a11yText` instead of `a11y-text` (kebab-case for consistency)
2. `ebay-inline-notice` examples have previous `a11y-heading-text` attribute instead of `ally-text`

Issue #1280, opened 2020-12-10
Notice: 'variant' being passed to 'ebay-icon' instead of 'a11yVariant'


melquan:
# Bug Report

## Description
`variant` is being passed to the `ebay-*-icon`'s instead of `a11yVariant` which is causing a couple of issues:
1. `variant` attribute is being rendered on the `svg` tag
2. `aria-label` is missing

## Workaround
None

## Screenshots
![image](https://user-images.githubusercontent.com/19419138/101703177-33b45280-3a37-11eb-8dbc-269e809e0fb7.png)

Issue #1279, opened 2020-12-09
Notice: some attributes are not being ignored and rendered in the HTML


melquan:
# Bug Report

## Description
1. All notice types: the `type` attribute is rendered on the root tag (`section` / `div`)
2. Inline/section notices: `main-root` attribute is rendered on the root tag (`section` / `div`)

## Workaround
None

## Screenshots
1. All notices: `type` attribute
![image](https://user-images.githubusercontent.com/19419138/101701364-a3284300-3a33-11eb-8d79-f40c5a14022e.png)

2. Inline/section notices: `main-root` attribute
![image](https://user-images.githubusercontent.com/19419138/101701539-f4d0cd80-3a33-11eb-8a2f-38049fd2dbff.png)

PR #1290, opened 2020-12-15
Added attributes to ignored list and fixed a11y-variant in notice

austingardner:
## Description
Fixes issues Melissa raised that were assigned to me.

## References
https://github.com/eBay/ebayui-core/issues/1285
https://github.com/eBay/ebayui-core/issues/1281
https://github.com/eBay/ebayui-core/issues/1280
https://github.com/eBay/ebayui-core/issues/1279


src/components/ebay-section-notice/test/test.server.js
```diff
@@ -9,7 +9,7 @@ describe('section-notice', () => {
     it('renders with status', async() => {
         const input = mock.Section_Info;
         const { getAllByLabelText, getByText } = await render(template, input);
-        const status = getAllByLabelText(input.a11yText)[1].parentElement;
+        const status = getAllByLabelText(input.a11yText)[0].parentElement;
```
agliga, line 12: Can we make sure now that `getAllByLabelText` returns a list and not just one element?
If it returns a list this is fine, but if it returns only one, we can use `getByLabelText` instead

src/components/ebay-inline-notice/index.marko
```diff
@@ -8,4 +8,5 @@ $ var status = input.status || "attention";
     root="div"
     header-root="span"
     main-root="span"
+    a11y-text=input.a11yText
```
agliga, line 11: This is not needed, `a11yText` should get passed through from the `...input`. 
If it's not, then we need to add `a11y-text` in marko.json
Check this again

agliga (commented):
Two things
* Change the title for the pull request
* We shouldn't have any `yarn.lock` changes
Could of other minor issues to check too

Also don't have to change this now, but usually for combined PRs try to make each commit to resolve and link to each issue. Once we merge, we can merge without squashing since there is a commit associated for each fix. 
However for now leave it as is, but for future we can do that.

src/components/ebay-page-notice/test/test.server.js
```diff
@@ -10,7 +10,7 @@ describe('page-notice', () => {
     it('renders with defaults', async() => {
         const input = mock.Page;
         const { getAllByLabelText, getByText } = await render(template, input);
-        const status = getAllByLabelText(input.a11yText)[1].parentElement;
+        const status = getAllByLabelText(input.a11yText)[0].parentElement;
```
agliga, line 13: This one didn't work with `getByLabelText`?

austingardner: ah dang missed that one. no it should work. one sec

agliga (approved):
Lgtm
Just change the title to describe the changes. I wouldnt keep 6.0 bugfixes but rather like
Added attributes to ignored list and fixed a11y-variant in notice

PR #1292, opened 2020-12-22
Update DS version numbers

### LINKED ISSUE #974: Docs: add design system version numbers for each component
To make it clearer which version of the playbook a component has been built to, let's add version numbers for each component. These version numbers should match up with what's in skin: https://github.com/eBay/skin/blob/master/docs/index.html#L6

The version numbers should probably go in each section of our Storybook, and perhaps on our README too.

austingardner:
Closes #974

PLEASE SQUASH

## Description
This PR adds a versions json file that will allow us to easily keep track of the versions for each component.

I also added a script that takes the version from the json file and adds it to the readme of the corresponding component.

## Screenshots
Screen Shot 2021-01-04 at 1 24 02 PM


ianmcburnie (commented):
Should we add an entry point to the new script into our package.json script? Or perhaps incorporate it into an existing script.

src/components/ebay-badge/README.md
```diff
@@ -1,4 +1,4 @@
-# ebay-badge
+# ebay-badge v1.2.0
```
ianmcburnie, line 1: As discussed offline, let's prefix the version number with "DS". It would be great if we could reduce it's visual prominence somehow - perhaps by reducing the font-size and/or aligning it over to the right, if possible.

austingardner: Yeah I was thinking we could do something like `<div style="text-align: right; font-size: small;"> DS v1.2.0 </div>`

ianmcburnie: Does Storybook/Markdown support classes instead? e.g. `<div class="ds-version">DS v1.2.0 </div>`

austingardner: I don't think they do unless you add some third party libraries. I think in this case the inline styles would be alright because it's just in the markdown file.

ianmcburnie (approved):

agliga (approved):

Issue #1029, opened 2020-01-17
Storybook: expose the source code for each example


ianmcburnie:
For developer conveniance, it would be nice to show the source code / markup for each example in Storybook:

https://ebay.github.io/ebayui-core

ianmcburnie:
@biaspro I'm not super familiar with what Storybook has to support this, but maybe something for you to pick up and look into.

ianmcburnie:
Done in 7.1.0

Issue #1018, opened 2020-01-14
Storybook: center examples


ianmcburnie:
Would be nice to dynamically center our storybook examples on the canvas.

This plugin might help:

https://github.com/storybookjs/storybook/tree/master/addons/centered

Then we wouldn't see overlays disappearing off the top of the screen like this (notice the infotip overlay):

<img width="786" alt="Screen Shot 2020-01-13 at 8 07 21 PM" src="https://user-images.githubusercontent.com/38065/72313528-7bbcd080-3640-11ea-898c-850570eace30.png">

ianmcburnie:
This is low priority, but perhaps @DylanPiercey could take a quick look and see what it would take to add Marko support for this addon:

https://github.com/storybookjs/storybook/blob/master/ADDONS_SUPPORT.md

mlrawlings:
While I'm not opposed to centering the examples, I see this as a limitation of the component itself.  The tooltip should open where there is available space.

It could argued that the `pointer` attribute should be used, but there are going to be cases where it's not possible to hard-code the optimal position due to differing screen sizes or the current scroll position.

ianmcburnie:
@mlrawlings Yep, I agree, and we have a ticket already for that: https://github.com/eBay/ebayui-core/issues/535

ianmcburnie:
Maybe another one for @biaspro to investigate as part of the other storybook work...

ianmcburnie:
Done in 7.1.0.

PR #1308, opened 2021-01-20
Updated storybook, added code viewer, and directory support

agliga:
## Description
* Updated storybook to the latest version, added support for it to work
* Added code syntax viewer
* Made center examples
* Added support for groups items
* Put button into it's own group (did this to test the new group items)
* Misc cleanup for storybook code
* Reran `storytbook:build` to make sure things are working
* Also upgraded all devDeps (part of upgrading storybook)

## Context
For code syntax, I had to basically compile all raw code using webpack. Then it gets loaded and read when each code is loaded. 
I had to do the lookup with the "ID" instead of filename since the ID was available in each module while filename was not.
I used the old highlights plugin for now, but we can switch

__Note__ I saw a couple of bugs, carousel does not seem to work and breadcrumbs are not showing up. I will address those in a future PR since this one is getting too large.

## References
#1018
#1234 (infastructure work for this)
#1029 

## Screenshots
Screen Shot 2021-01-20 at 10 16 57 AM
Screen Shot 2021-01-20 at 10 17 05 AM
Screen Shot 2021-01-20 at 10 17 13 AM
Screen Shot 2021-01-20 at 10 17 46 AM


ianmcburnie:
Very cool set of changes. Love the code syntax viewer.

Couple of nitpicks:

1. Switch: each story currently shows two instances of the switch. On and off. Please separate those states out into separate stories (this is better for visual regressing testing)
2. Badge: I think we can move the menu and icon stories to their respective components (menu/menu-button and button)
3. Filter: same as above. Some stories are showing more than one instance. If you see this anywhere else, please also fix so that we always have one instance of component per story.

agliga:
> Very cool set of changes. Love the code syntax viewer.
> 
> 
> 
> Couple of nitpicks:
> 
> 
> 
> 1. Switch: each story currently shows two instances of the switch. On and off. Please separate those states out into separate stories (this is better for visual regressing testing)
> 
> 2. Badge: I think we can move the menu and icon stories to their respective components (menu/menu-button and button)
> 
> 3. Filter: same as above. Some stories are showing more than one instance. If you see this anywhere else, please also fix so that we always have one instance of component per story.

Sounds good. I will plan to do them in a follow up pr. I have a couple of fixes to get in but wanted to separate them out so this pr doesn't get too big

ianmcburnie:
No problem. It would probably good if you walk us through some of the plumbing/configuration/plugin aspects.

ianmcburnie (approved):
Happy for this to go in now, if it helps you juggling branches.

Issue #1301, opened 2021-01-12
"ebay-listbox-button" value doesn't truncate 


nikitajain7007:
Facing this issue in ebay-listbox-button  where the value doesn't truncate when the text length width is greater than the container width. ebay-button provides a truncate param https://ebay.github.io/ebayui-core/?path=/story/ebay-button--truncated can ebay-listbox-button support truncate param

<img width="386" alt="Screen Shot 2021-01-12 at 3 17 16 PM" src="https://user-images.githubusercontent.com/22912869/104297825-4d8afe80-54e9-11eb-9227-dc488b482cce.png">

ianmcburnie:
Done in 7.1.0

PR #1310, opened 2021-01-20
Added truncated listbox button option

agliga:
## Description
Added option to truncate. This is the same as the `ebay-button` one.
Also added example and test.
I had to made the listbox button to be a `div` instead of a `span` otherwise it would not truncate.

## References
#1301 

## Screenshots
Screen Shot 2021-01-20 at 2 23 00 PM


ianmcburnie:
In the example, the width of the button changes when a new value is selected. Should we add a min width too? Or just replace max-width and min-width with a fixed width for this scenario?

agliga:
> In the example, the width of the button changes when a new value is selected. This feels odd. Should we add a min width too? Or just replace max-width and min-width with a fixed button width for this scenario?

Screen Shot 2021-01-26 at 2 08 23 PM
Done

ianmcburnie (approved):

PR #1311, opened 2021-01-25
Added new progress-stepper component with chages from 3.1.1

agliga:
## Description
Added progress-stepper component. Changed it to support changes from skin

## References
https://github.com/eBay/skin/pull/1336

## Screenshots
Screen Shot 2021-01-25 at 3 42 46 PM

Screen Shot 2021-01-25 at 3 42 53 PM
Screen Shot 2021-01-25 at 3 43 01 PM


ianmcburnie:
Kind of related to the storybook cleanup, but I think it would be good if we start with some kind of naming convention for examples/stories. For example, a name that highlights exactly which part/attriibute of the API the story relates to:

Direction: vertical
Step-type: attention
Step-type: information
Default-State: upcoming
Default-State: complete

I guess with storybook knobs we would be able to reduce the amount of stories to:

Direction
Step-type
Default-State

ianmcburnie:
Nit pick, but could you put the interactive example last? We should layer-in the complexity of examples, starting with the most common/basic and ending with the more complex/uncommon.

agliga:
> Kind of related to the storybook cleanup, but I think it would be good if we start with some kind of naming convention for examples/stories. For example, a name that highlights exactly which part/attriibute of the API the story relates to:
> 
> Direction: vertical
> Step-type: attention
> Step-type: information
> Default-State: upcoming
> Default-State: complete
> 
> We'd have to see if this convention scales across components.
> 
> I guess with storybook knobs we would be able to reduce the amount of stories to:
> 
> Direction
> Step-type
> Default-State

Screen Shot 2021-01-26 at 1 17 23 PM
Updated

ianmcburnie (approved):

Issue #1317, opened 2021-01-28
Pagination: add ellipses/truncation 


agliga:
Let's scope out the work involved to add a story for ellipses/truncation

![Screen Shot 2021-01-22 at 1 12 08 PM](https://user-images.githubusercontent.com/38065/105547623-82e6d780-5cb3-11eb-9e23-4f0b6e203254.png)

This one might need backporting to v10 or v11.

https://github.com/eBay/skin/issues/1337

ianmcburnie:
Done in 7.2.0

PR #1318, opened 2021-01-28
Added option to show dots on pagination module

agliga:
## Description
* Added `variant` option to allow pagination with dots. 
* Made dots show up and force 2 less items (so that `...` will take up and item and last page)
* Added <strike>`a11yDotsText`</strike> and `onDotsClick` in case those events and labels are needed
* Added tests for dots

## References
#1317 

## Screenshots
pagination

### DIFF
--- a/src/components/ebay-pagination/index.marko
+++ b/src/components/ebay-pagination/index.marko
@@ -55,13 +55,12 @@ $ var range = component._getVisibleRange(items);
         <for|item, i| of=items>
             <if(range.dotsIndex === i)>
                 <li hidden=(range.hideDots)>
-                    <button
+                    <span
                         aria-label=input.a11yDotsText
-                        onClick("handleDotsClick")
                         class=["pagination__item", item.class]
                         >
                         …
-                    </>
+                    </span>
                 </li>
             </if>
             <li hidden=((i < range.start || i > range.end) && range.dotsIndex !== i)>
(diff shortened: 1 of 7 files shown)


src/components/ebay-pagination/test/mock/index.js
```diff
@@ -79,6 +105,78 @@ exports.Links_9Items_7Selected = assign({}, exports.Base_0Items, {
     )
 });
 
+exports.Links_16ItemsDots_1Selected = assign({}, exports.Base_0Items_Dots, {
```
austingardner, line 108: Is there a reason we're not using ES6 object spread instead of assign? I thought we could use ES6 everywhere besides the .marko files. 

Maybe there's some background info I'm misssing?

austingardner (approved):
LGTM!

src/components/ebay-pagination/index.marko
```diff
@@ -53,7 +53,17 @@ $ var range = component._getVisibleRange(items);
         key="items"
         class="pagination__items">
         <for|item, i| of=items>
-            <li hidden=(i < range.start || i > range.end)>
+            <if(range.dotsIndex === i)>
+                <li hidden=(range.hideDots)>
+                    <span
```
ianmcburnie, line 58: An aria-label doesn't do anything without a role (implicit or explicit). A role of `img` is needed in this case. However, I think we should just remove the `a11yDotsText` attribute for now, as the text/ellipses by itself should be adequate. I'm wary of apps populating a label with something that does more harm than good!

agliga: Done

ianmcburnie (approved):

PR #1322, opened 2021-02-03
Added Progress Bar Component to Ebay-UI

### LINKED ISSUE #1267: Progress-Bar: new component
Related Skin issue: https://github.com/eBay/skin/issues/1289

austingardner:
## Description
This PR adds the Progress Bar component to ebay-ui. I included a several examples. 

## References
closes #1267 

## Screenshots
Screen Shot 2021-02-02 at 11 26 05 PM

### DIFF
--- a/src/components/ebay-progress-bar/component.js
+++ b/src/components/ebay-progress-bar/component.js
@@ -1,10 +1,8 @@
 module.exports = {
-    getValue(value, min, max) {
-        const parsedValue = parseInt(value);
+    getValue(value, min) {
+        const parsedValue = parseInt(value, 10);
         if (parsedValue <= min) {
             return min + 1;
-        } else if (parsedValue >= max) {
-            return max - 1;
         }
         return parsedValue;
     }
--- a/src/components/ebay-progress-bar/component.js
+++ b/src/components/ebay-progress-bar/component.js
@@ -1,6 +1,5 @@
 module.exports = {
     getValue(value, min, max) {
-        console.log(value, min, max);
         const parsedValue = parseInt(value);
         if (parsedValue <= min) {
             return min + 1;
--- a/src/components/ebay-progress-bar/component.js
+++ b/src/components/ebay-progress-bar/component.js
@@ -1,0 +1,12 @@
+module.exports = {
+    getValue(value, min, max) {
+        console.log(value, min, max);
+        const parsedValue = parseInt(value);
+        if (parsedValue <= min) {
+            return min + 1;
+        } else if (parsedValue >= max) {
+            return max - 1;
+        }
+        return parsedValue;
+    }
+};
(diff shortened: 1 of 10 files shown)


src/components/ebay-progress-bar/examples/01-empty/template.marko
```diff
@@ -0,0 +1,3 @@
+<div style="background-color: white; height: 35px; display: flex; align-items: center;">
```
ianmcburnie, line 1: Let's remove these inline styles if we can.

src/components/ebay-progress-bar/index.marko
```diff
@@ -0,0 +1,7 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+static var ignoredAttributes = ["class"];
+
+<progress
+    ...processHtmlAttributes(input, ignoredAttributes)
```
agliga, line 6: Can we default `min` and `max` to 0 and 100?
You can do something like:
`min = (input.min || 0)`
`max = (input.max || 100)`
And add that info in the readme and update the examples, but keep 1 with a custom min/max

src/components/ebay-progress-bar/index.marko
```diff
@@ -0,0 +1,7 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+static var ignoredAttributes = ["class"];
+
+<progress
```
agliga, line 5: Are any `aria-labels` required for this to announce the progress?
If so can you add `a11yText` input.

agliga (commented):
Also can you add a link in the root `README.md` to this module.
Just copy one and rename it.

src/components/ebay-progress-bar/index.marko
austingardner: I don't think so, because we're using HTML progress.
Looking [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role) they just have `aria-labels` for things that the progress bar already has.

agliga:
Also make sure you address the issue we discussed with design here, to make sure value is at least min + 1

src/components/ebay-progress-bar/component.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+    getValue(value, min, max) {
+        const parsedValue = parseInt(value);
```
agliga, line 3: Make sure you do `parseInt(value, 10);` (should always have the `radix` param, in case it accidentally parses as a hex or something else)

agliga (commented):
Just fix the radix issue.
Also do we not want to fill it to 100% when it's full? I think for that case we would but not sure. Only when it's not full right?

agliga (approved):

PR #1323, opened 2021-02-04
Adding Github Actions to ebayui-core

austingardner:
PLEASE SQUASH

## Description
Adding github actions to ebayui-core. Removed .travis.yml file that is no longer needed.


ianmcburnie:
This is a great start! There is also an integration with Browserstack (to run our tests in real browsers), which requires a private key. Let's chat about that offline.

Makefile
```diff
@@ -0,0 +1,7 @@
+REPORTER = spec
```
ianmcburnie, line 1: Where does this makefile get used?

ianmcburnie:
Do you want to do the browserstack piece as a separate PR?

austingardner:
Yes let's do the browserstack stuff separately

ianmcburnie (approved):

src/components/ebay-tooltip/test/test.browser.js
```diff
@@ -26,7 +26,7 @@ describe('given the default tooltip', () => {
             expect(component.emitted('expand')).has.length(1);
         });
 
-        describe('when the host element loses hover', () => {
+        describe.skip('when the host element loses hover', () => {
```
agliga, line 29: Why is this test skipped?

austingardner: It kept failing randomly. One run through it would pass, and then another it would fail.

agliga: I think we should leave it as is, and then revisit it if it's failing in another PR. 
If we keep seeing it fail and retest it's fine since we will want to fix it, but removing it will make us lose coverage and we might forget about it

austingardner: okay I unskipped it 👍

Issue #1321, opened 2021-02-02
Listbox-Button: allow ability to construct label from external element plus button value


ianmcburnie:
# Feature Request

## Description

Allow ability to construct label from external text element plus button value

Example:

```html
<span id="el1">Reason</span>
<button aria-labelledby="el1 el2"><span id="el2">Cannot contact seller</span></button>
```

## Context

We often see teams trying to use listbox-button like a `<select>`, by providing a `<label>` tag. However, a `<label>` tag does not work for buttons. We would need to allow the accessible label to be constructed via a space separated list of ids using `aria-labelledby`.

## Screenshot

![Screen Shot 2021-02-03 at 10 44 03 AM](https://user-images.githubusercontent.com/38065/106793843-d4288c80-660c-11eb-86e6-0431f55dd3d8.png)

agliga:
This has been published in `6.1.0` and in 7.2.0` beta for now (will go out in `7.2.0`)

Issue #1320, opened 2021-02-02
Listbox-Button: allow ability to add prefix text to the button.


ianmcburnie:
# Feature Request

## Description

Allow ability to add prefix text to the button. This prefix would remain in place every time a new value is chosen from the listbox.

For example:

`<button>Reason: Cannot contact seller</button>`

`prefixText="Reason"`

## Context

We often see teams trying to use listbox-button like a `<select>`, by providing a `<label>` tag. However, this is not how buttons are labelled. A buttons label is derived from its inner text/value (although we *could* also allow it to be constructed from an external text element plus the value, using aria-labelledby, see: #1321).

Let's clear up this confusion and make it easier for devs to correctly label this component and thus cut down a11y defects.

agliga:
This has been published in `6.1.0` and in 7.2.0` beta for now (will go out in `7.2.0`)

PR #1328, opened 2021-02-09
Added prefix-id and prefix-label to listbox

agliga:
## Description
Added both `prefix-label` and `prefix-id` to listbox button

## References
https://github.com/eBay/ebayui-core/issues/1320
https://github.com/eBay/ebayui-core/issues/1321

## Screenshots
Screen Shot 2021-02-09 at 1 54 17 PM
Screen Shot 2021-02-09 at 1 54 21 PM

### DIFF
--- a/src/components/ebay-listbox-button/index.marko
+++ b/src/components/ebay-listbox-button/index.marko
@@ -1,4 +1,4 @@
-import processHtmlAttributes from "../../common/html-attributes";
+import processHtmlAttributes from "../../common/html-attributes"
 
 static var ignoredAttributes = [
     "name",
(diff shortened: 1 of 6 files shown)


src/components/ebay-listbox-button/index.marko
```diff
@@ -33,9 +35,15 @@ $ var selectedText = selectedOption && selectedOption.text;
         disabled=input.disabled
         name=input.buttonName
         aria-haspopup="listbox"
+        aria-labelledby=(labelId && `${input.prefixId} ${labelId}`)
```
DylanPiercey, line 38: Is this intentional? I'm assuming this should just be `aria-labelledby=labelId`.

austingardner (approved):

src/components/ebay-listbox-button/index.marko
agliga: Yes it is intentional, the `labelledby` needs to be:
`externalLabelID internalLabelId`
So if the `input.prefixId` is present, then we will add a `labelId` to our internal label and then combine both. 
See: https://github.com/eBay/ebayui-core/issues/1321

PR #1329, opened 2021-02-10
Optimized video component

agliga:
## Description
* Switched to shaka player
* Added play button as loading, when clicked will switch to loading
* Added check to use `requestIdleCallback` and use that if available to load from CDN
* Removed all checks for IOS since shaka handles that automatically
* Added error event in case there is an error when loading the source


src/components/ebay-video/README.md
```diff
@@ -17,13 +17,19 @@ For resizing, `ebay-video` supports fixed width or variable width. If no width i
...
 Name | Type | Stateful | Required | Description
 --- | --- | --- | --- | ---
-`src` | String | No | Yes | The url path for the playlist MPD or M3U8 to use. If format is set to auto, will detect format based on file extension
-`format` | String | No | No | Either `auto`, `mpd` or `m3u8`. Default is auto. The format to use. If auto, will try to detect based on file name. If mpd will trigger downloading dashjs from given CDN.
-`cdn-url` | String | No | No | The full url to point to where to download dashjs. This is only used to override the default CDN path.
-`cdn-version` | String | No | No | If using the default CDN path, you can override what version of dashjs to use.
+`src` | String | No | Yes | The url path for the playlist MPD or M3U8 to use.
+`thumbnail` | String | No | No | The url path for the video thumbnail
+`cdn-url` | String | No | No | The full url to point to where to download shaka. This is only used to override the default CDN path.
+`cdn-version` | String | No | No | If using the default CDN path, you can override what version of shaka to use.
```
austingardner, line 23: maybe we could be more specific with this name - it could be 
`shaka-version-cdn` or something like that. 

Just a suggestion though

austingardner (approved):
LGTM

src/components/ebay-video/README.md
agliga: Good comment, I think I want to make an option to change possibly, but not sure.
We can revisit it

PR #1330, opened 2021-02-11
Add Prettier to Ebay-UI

austingardner:
## Description
This PR adds Prettier to eBay-UI, allowing automatic formatting. 
It also adds lint-staged and husky to run prettier on any changed files as a pre-commit hook.

Additionally, the existing pre-push hook was interfering with husky, so I added a pre-push hook in husky that runs `npm run lint && npm test`, which is what the previous pre-push hook did. I deleted the previous pre-push hook and added the ones that husky automatically generates. Husky now runs well.  

## Context
Ian mentioned adding lint-staged and husky in a previous conversation. I don't think there will be any issues with it. 

One thing we may have to think about is if our IDEs allow prettier to run on save for js, css, less, json and other files, while still allowing marko-pretty-print to run on marko files.


demo/style.less
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable */
```
austingardner, line 1: disabled this for now. We can run a different PR to run prettier on all of the files if we want.

src/components/ebay-progress-bar/component.js
```diff
@@ -1,9 +1,9 @@
 module.exports = {
-    getValue(value, min) {
-        const parsedValue = parseInt(value, 10);
-        if (parsedValue <= min) {
-            return min + 1;
-        }
-        return parsedValue;
+  getValue(value, min) {
```
ianmcburnie, line 2: Not sure why this file got changed from 4 spaces to 2...?

ianmcburnie (approved):

.prettierignore
```diff
@@ -0,0 +1,7 @@
+*.marko.js
+static/
+dist/
+docs/
+_cdn/
+demo/style.less
+src/components/ebay-video/style.less
```
agliga, line 7: Put a newline at the end of the file.
Just curious why is ebay-video style ignored?

agliga (commented):
Is this running on all files, or will it run as we change? Only a couple of files changed.
Also is marko working with this or not?

austingardner:
Right now Marko is not running with this. Pretty-print will have to run separately. We could add it in to lint-staged.

This will run as we change, although maybe we should run it on the src files now so that there aren't so many changes later.

.prettierignore
austingardner: I was trying not to change the files for this PR, because there weren't any changes in that file related to this PR, other than styling. For some reason that file kept getting picked up and prettier was running on it

agliga: Well there is a prettier ignore thing you added in the top of the file, so maybe thats why?
Regardless either we have ignore on those files or in here but no reason to have it in both places. 

And I think it's fine to have those files changed

.prettierignore
```diff
@@ -0,0 +1,6 @@
+*.marko.js
+static/
+dist/
+docs/
+_cdn/
+demo/style.less
```
agliga, line 6: This too? I see that there is some ignore put in demo/style.less but it shows up here.

austingardner: good catch, got that one too

agliga (approved):
LGTM, I think the only outstanding issue is fixing progress-bar indentation, otherwise I'm fine with it

src/components/ebay-progress-bar/component.js
ianmcburnie: @austingardner I think this is the only thing that needs addressing before merge

ianmcburnie (changes requested):
Can we configure prettier to 2 spaces of indentation for JSON? That's how we have it in all other projects.

austingardner:
Yes I can do that. Let me add that in

austingardner:
Okay I think this is ready - JSON width is set to 2

src/components/ebay-progress-bar/component.js
austingardner: Done. What happened was that prettier's default is 2 and prettier was run on this file. Now that we have it set to 4 I ran it again and it's back to where it was.

ianmcburnie (approved):

PR #1333, opened 2021-02-19
Fake menu now uses list instead of div

### LINKED ISSUE #1316: Fake-Menu: should use list based markup (like on Skin)
The fake menu and fake menu button should use list based markup instead of div based markup.

See skin for reference: https://ebay.github.io/skin/#menu-fake

austingardner:
## Description
Changes the structure of the ebay-fake-menu (which fake-menu-button uses) to use a list based structure instead of a div structure. 

## Context
Accessibility 

## References
closes #1316 

## Screenshots
Screen Shot 2021-02-18 at 9 32 02 PM

Screen Shot 2021-02-18 at 9 32 20 PM

Screen Shot 2021-02-18 at 9 32 46 PM

### DIFF
--- a/src/components/ebay-fake-menu/index.marko
+++ b/src/components/ebay-fake-menu/index.marko
@@ -34,37 +34,41 @@ $ var baseClass = input.classPrefix || "fake-menu";
         input.class
     ]
     style=input.style>
-    <div
-        class=`${baseClass}__items`
+    <ul
+        class=`fake-menu__items`
         key="menu"
         tabindex="-1"
         id:scoped="menu">
         <for|item, index| of=input.items>
             $ var isDisabled = !item.href || item.disabled;
             <if (item._isSeparator)>
-                <hr class=`${baseClass}__separator` role="separator" />
+                <li>
+                    <hr class=`${baseClass}__separator` role="separator" />
+                </li>
             </if>
             <else>
-            <${item.type === "button" ? "button" : "a"}
-                ...processHtmlAttributes(item, itemIgnoredAttributes)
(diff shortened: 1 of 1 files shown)


agliga (approved):

src/components/ebay-fake-menu/index.marko
```diff
@@ -34,37 +34,41 @@ $ var baseClass = input.classPrefix || "fake-menu";
         input.class
     ]
     style=input.style>
-    <div
-        class=`${baseClass}__items`
+    <ul
+        class=`fake-menu__items`
```
ianmcburnie, line 38: Is this change intentional?

austingardner: It is intentional, the styling looked good after using it. The issue is that with the fake-menu-button__items class, the bullet points were showing and the margins were off. 

Shifting to the fake-menu__items class fixed everything, so I figured we could just hard code it.

austingardner: Is that a bad idea? I wasn’t 100% sure but it looked great so I thought it was a good direction to go

ianmcburnie: Hmm. This suggest that we might need a skin change also. Let me check...

ianmcburnie: I don't see any issue with skin.

ianmcburnie (approved):

ianmcburnie:
Argh, this was pointed at 7.2.0 instead of 7.3.0. Nevermind, should be fine as it's a fairly minor change.

Issue #967, opened 2019-11-20
Browserstack: failing on Firefox for Windows


ianmcburnie:
Browserstack has been having problems with Firefox on Windows. We should probably temporarily disable this browser until we can investigate the root of the issue.

![Screen Shot 2019-11-20 at 10 40 30 AM](https://user-images.githubusercontent.com/38065/69268139-6ddc5900-0b83-11ea-9a4c-f8e70314d6a7.png)

ianmcburnie:
@DylanPiercey Can maybe have a quick look at this. Otherwise, we can disable it.

ianmcburnie:
Related ticket: #1307

ianmcburnie:
Moving to super back log.

PR #1336, opened 2021-02-24
ci: add browserstack

ianmcburnie:
Adds browserstack support to our github actions. Marko-test and WebDriverIO do all the heavy lifting here. We just have to set the correct env vars.

Also fixed broken config in `marko-cli.js`. After much head scratching the offending line of code was the `localIdentifier` param. It was causing issues with the browserstack local connection. I just removed it and a bunch of other config which seems superfluous now.

Browserstack is now again working from CLI and CI. I updated the CI to only trigger on pull requests. We can fine tune this later if desired.

I cleaned up the list of browsers. There still seems to be an issue with carousel in Firefox (#967) so it remains omitted.

Screen Shot 2021-02-23 at 5 02 42 PM

Screen Shot 2021-02-23 at 5 05 18 PM

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,6 +1,6 @@
 name: eBay UI CI
 
-on: [pull_request, push]
+on: [pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


agliga (approved):
Nice! I like how we reenabled other browsers

austingardner (approved):
👍

PR #1338, opened 2021-02-24
1298 Snackbar Dialog Component

### LINKED ISSUE #1298: Snackbar-Dialog: new component
# Feature Request

## Description

Acknowledgment messaging looks like it could be a variant of toast-dialog component. We will know for sure when the skin work (https://github.com/eBay/skin/issues/1290) has been completed. UPDATE: nope, we are going with snackbar-dialog. 

## Screenshots

Screen Shot 2020-11-09 at 10 14 55 AM

austingardner:
## Description
Added snackbar dialog component to ebay-UI.

More tests will be added in separate PR

## References
closes #1298 

### Basic Snackbar
basic-snackbar

### Action Snackbar
action-snackbar

### Stacked Snackbar
stacked-snackbar


src/components/ebay-snackbar-dialog/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+A snackbar is a non-modal dialog that appears in response to a lightweight user action. It disappears automatically after a minimum of 6 seconds.
+
+In the case of type="action", the user should provide the open state as a boolean and a function that will be called with an on-close event to close the snackbar.
+
+## Attributes
+
+| Name   | Type    | Stateful | Required | Description                                                                         |
+| ------ | ------- | -------- | -------- | ----------------------------------------------------------------------------------- |
+| `type` | String  | No       | No       | Options are default or action. Action types include a button to close the snackbar. |
```
ianmcburnie, line 26: Do we  need the `type` attribute? There isn't one in toast-dialog.

src/components/ebay-snackbar-dialog/examples/03-stacked/template.marko
```diff
@@ -0,0 +1,20 @@
+<ebay-snackbar-dialog open=state.open windowClass="snackbar-dialog__window--column" type="action" on-close("closeSnackbar")>
```
ianmcburnie, line 1: Rather than `windowClass`, can we add a `layout` attribute? Values should be `row` (default) or `column`.

src/components/ebay-snackbar-dialog/component.js
```diff
@@ -0,0 +1,22 @@
...
+module.exports = {
+    onInput(input) {
+        if (input.open === true || input.open === false) {
+            this.state.open = input.open;
+            clearTimeout(this.timeout);
+        }
+        if (input.open !== false) {
+            clearTimeout(this.timeout);
+            const timeout = setTimeout(() => {
+                this.emit('close');
```
ianmcburnie, line 11: One enhancement to make is that the dialog should never close while keyboard focus or mouse is inside.

src/components/ebay-snackbar-dialog/examples/02-action/template.marko
```diff
@@ -0,0 +1,20 @@
+<ebay-snackbar-dialog type="action" open=state.open on-close('closeSnackbar')>
+    This is the action snackbar
+</ebay-snackbar-dialog>
+<ebay-button on-click("openSnackbar")>
+    Open Action Snackbar
```
ianmcburnie, line 5: The action button should be passed in by the app developer as part of the body. The same way that toast-dialog does it.

src/components/ebay-snackbar-dialog/examples/03-stacked/template.marko
agliga: Window class should be a private method. This should be in the snackbar template.

src/components/ebay-snackbar-dialog/component.js
```diff
@@ -0,0 +1,22 @@
+const DEFAULT_TIMEOUT_LENGTH = 6000;
+module.exports = {
+    onInput(input) {
+        if (input.open === true || input.open === false) {
+            this.state.open = input.open;
```
agliga, line 5: Why do we need this?
It doesn't look like state is ever changed

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,18 @@
...
+    isModal=false
+    classPrefix="snackbar-dialog"
+    open=state.open
+    class=[input.class, "snackbar-dialog--transition"]
+    buttonPosition="hidden">
+    <${input.renderBody}/>
+    <if(input.type === 'action')>
+    <@actions>
+        <span class="snackbar-dialog__actions">
+            <button on-click('emit', 'close') accesskey="u" class="fake-link snackbar-dialog__cta">Undo<span class="clipped"> - Access Key: U</span></button>
```
agliga, line 14: This should be adjustable by user imo, like the text and accesskey.
I would add a `<@undo>` tag where the user can add what they need.
`<@undo accesskey="u" a11y-text="- access key: U">Undo</@undo>`
(Or split it in multiple attributes on the snackbar, like `undo-text` and `a11yUndoText`)
And then
```
<button on-click() ...processHTMLAttributes(input.undo, ['list of ignored stuff like a11yText'] >
  <${input.undo}/>
  <span class="clipped">${input.undo.a11yText}</span>
</button>
```

Remember there are non english usecases too!

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,18 @@
...
+
+<ebay-dialog-base
+    ...input
+    isModal=false
+    classPrefix="snackbar-dialog"
+    open=state.open
+    class=[input.class, "snackbar-dialog--transition"]
+    buttonPosition="hidden">
+    <${input.renderBody}/>
+    <if(input.type === 'action')>
```
agliga, line 11: I would remove this and add the `<@undo>` check to show that part

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,18 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+<ebay-dialog-base
+    ...input
+    isModal=false
+    classPrefix="snackbar-dialog"
+    open=state.open
```
agliga, line 7: Do `input.open`
Or you don't need it because you should be passing it through from `...input`

src/components/ebay-snackbar-dialog/index.marko
ianmcburnie: It's not always an undo. Just bear that in mind.

src/components/ebay-snackbar-dialog/README.md
austingardner: That's right, we don't because they can pass in an action if they need in the marko file.

src/components/ebay-snackbar-dialog/component.js
austingardner: good catch

src/components/ebay-snackbar-dialog/examples/02-action/template.marko
austingardner: that's right. fix in progress

src/components/ebay-snackbar-dialog/index.marko
austingardner: Okay so I know you suggested the @undo tag, but I'm wondering whether just giving the user the `@action` tag and letting them put in the button as well as the text inside of it would be an adequate solution. 

Can you take a look at the change I just made and see whether it works?

agliga: Yes action is probably better if it can be any action

src/components/ebay-snackbar-dialog/README.md
```diff
@@ -0,0 +1,27 @@
...
+<ebay-snackbar-dialog>
+    Basic Snackbar
+</ebay-snackbar-dialog>
+```
+
+## Description
+
+A snackbar is a non-modal dialog that appears in response to a lightweight user action. It disappears automatically after a minimum of 6 seconds.
+
+In the case of type="action", the user should provide the open state as a boolean and a function that will be called with an on-close event to close the snackbar.
```
ianmcburnie, line 20: You can delete this line now.

src/components/ebay-snackbar-dialog/examples/02-action/template.marko
```diff
@@ -0,0 +1,23 @@
+<ebay-snackbar-dialog open=state.open on-close('closeSnackbar')>
+    This is the action snackbar
+    <@actions>
+            <button on-click('closeSnackbar') accesskey="u" class="fake-link snackbar-dialog__cta">Undo<span class="clipped"> - Access Key: U</span></button>
```
ianmcburnie, line 4: This should probably use ebay-button. Right @agliga?

We would also want to handle the addition of those classes ourselves.

I think we could also take care of appending the "Access Key: U" text ourselves. We would need a new `a11y-text-access-key` (default: "Access Key").

src/components/ebay-snackbar-dialog/README.md
austingardner: done

src/components/ebay-snackbar-dialog/examples/02-action/template.marko
austingardner: hmm what would that look like exactly? Because if the user gives the button or whatever else they want there, how would we add the accesskey into the button the user is adding?

ianmcburnie: @agliga If using ebay-button can we derive the accesskey char from the input? Or can we not do that for an attribute that is a straight pass through?

ianmcburnie: Also, @austingardner look at toast-dialog for an example of what I mean for using ebay-button here.

austingardner: wait so we don't want the user to be able to pass whatever they want into the @action section? Like for example if they wanted an icon or something instead of the button

ianmcburnie: It's always going to be an ebay-button

austingardner: okay so in this case we could just have the user pass in the button's text and then we could take care of everything else? we wouldn't need an @action in that case, just an @undo like @agliga said? and then that text between the @undo would be placed in the text the ebay-button

ianmcburnie: Yes, we could do that. We could even do without the `@undo` and just use attributes of `action-text` and `action-access-key`. However I was hoping to keep the API somewhat consistent with toast-dialog. That would also allow more than one action button, if design team ever wanted.

ianmcburnie: What do you guys think? Keep it similar to toast-dialog? Or go for something different and perhaps more streamlined?

austingardner: Yeah I like the `action-text` and `action-access-key` idea! I don't think we should let the user pass in something like `<button/>` if we're always going to use `<ebay-button/>`

austingardner: but at the same time yeah that's true that they might need more than one action button... hmmm...

agliga: You can also have multiple attribute tags￼
they are passed in as an array. Look at my video changes with the source tag and you can see how it's done￼

austingardner: okay so if I have an action tag it could look something like this: 
```
<for|action| of=input.actions>
        <ebay-button class="fake-link snackbar-dialog__cta" on-click(action.on-click) >
            <${action.actionText}/>
            <span class="clipped"> - Access Key: ${action.action-access-key}</span>
        </ebay-button>
    </for>
```

But how would the on-click work? I don't know syntactically how to pass functions like that in Marko

austingardner: nvm about the function -> I just need to think Marko and not React

src/components/ebay-snackbar-dialog/component.js
austingardner: Okay so I made progress on this, but I need a new selector. Right now I am using the CSS one and that doesn't seem very stable. See the component.js in the snackbar dir [commit here](https://github.com/eBay/ebayui-core/pull/1338/commits/ed58d0bf87fbb3168b7ba0c51c189eb53f42187c)

austingardner:
I think I've settled on the API... let me know what you think @agliga @ianmcburnie.

src/components/ebay-snackbar-dialog/examples/02-action/template.marko
```diff
@@ -0,0 +1,23 @@
+<ebay-snackbar-dialog open=state.open on-close('closeSnackbar') on-action("closeSnackbar")>
+    This is the action snackbar
+    <@action access-key="U">
+        Undo
+    </@action>
+</ebay-snackbar-dialog>
+<ebay-button on-click("openSnackbar")>
```
ianmcburnie, line 7: In all examples, can you put the button that launches the snackbar before the dialog. This will ensure correct reading & keyboard order.

austingardner: done

src/components/ebay-snackbar-dialog/test/mock/action-snackbar-template.marko
```diff
@@ -0,0 +1,23 @@
+<ebay-button on-click("openSnackbar")>
+    Open Action Snackbar
+</ebay-button>
+<ebay-snackbar-dialog open=state.open on-close('closeSnackbar') on-action("closeSnackbar")>
```
ianmcburnie, line 4: We shouldn't need the developer to listen for close event and then set open state. We should handle this for them.

src/components/ebay-snackbar-dialog/README.md
```diff
@@ -0,0 +1,27 @@
...
+<ebay-snackbar-dialog>
+    Basic Snackbar
+</ebay-snackbar-dialog>
+```
+
+## Description
+
+A snackbar is a non-modal dialog that appears in response to a lightweight user action. It disappears automatically after a minimum of 6 seconds.
+
+The user manages the state of the snackbar, and so should provide the open state as a boolean as well as a function to close the snackbar when the on-close event occurs.
```
ianmcburnie, line 20: Second part of this sentence needs removing (regarding the close function requirement).

src/components/ebay-snackbar-dialog/test/mock/action-snackbar-template.marko
austingardner: Hmm I mean I did have it so the user didn't have to pass a close function, but then we talked about it more and I got rid of it because the user will basically always need to manage the state of the snackbar. 

In the action case, we will always need the developer to manage the state, because there are so many ways they might want to handle closing it. 

I agree though that with the no action basic example, we *could* have it so that the snackbar just needs to be passed the open boolean. But then if the user is already passing the boolean for open / close, it just makes sense to let them manage the state. 

The user would definitely need to at least pass the open / close boolean though, even in the basic state. Would they want it to start open and then close, or would they want it to trigger based on something and then auto close? It just seems much easier to have them manage the state. 

I know with material UI components when I was working with them, the user managed open / close states in a fair number of components. This is a case where that makes sense, and I know we have other components like that as well.

ianmcburnie: By default, when user clicks the action button, toast should always close and set open to false. 
By default, after timeout, toast should always close and set open to false.
If the developer wants to do something else in *addition*, then yes they can tap into the on-close event to do that.

Can you compare with toast dialog? It doesn't seem to be requiring user to explicitly handle a close event.

austingardner: In the toast-dialog exmaple, it does require the user to have a close event. 

[example from toast dialog](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-toast-dialog/examples/01-default/template.marko)

```marko
<ebay-toast-dialog a11y-close-text="Close Toast" open=state.open on-close("closeToast")>
    <@header>Heading</@header>
    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit</p>
    <p>
        <a href="http://www.ebay.com">www.ebay.com</a>
    </p>
    <@footer>
        <ebay-button accesskey="i" on-click("closeToast")>Close</ebay-button>
    </@footer>
</ebay-toast-dialog>
<ebay-button on-click("openToast")>Open Toast</ebay-button>
class {
    onCreate() {
        this.state = { open: false };
    }
    openToast() {
        this.state.open = true;
    }
    closeToast() {
        this.state.open = false;
    }
}
```

ianmcburnie: Hmm, yes. It's done differently though. It is closed via a click event on the ebay-button, which makes more sense to me than a close event on the snackbar itself. Calling a close function on the close event of the snackbar is what struck me as odd initially. Can we add event listeners to attribute tags?

ianmcburnie: Hmm, this doesn't seem right: `on-close("closeToast")>`. Let's check in tomorrow with Andrew.

ianmcburnie: So after some offline discussion it might be better to rename `on-close("closeToast")` to `on-close("handleClose")` to make it clearer that this is something that an application _can_ do to synchronize its state with the components current state, but is not necessary to actually close the toast. Likewise for snackbar.

src/components/ebay-snackbar-dialog/README.md
ianmcburnie: After today's discussion, my interpretation is that if this component were a fully controlled component, then this sentence and instruction would be needed. However, given that it is not a fully controlled component, the sentence needs modifying to be something like:

`The application manages the initial state of the snackbar, and so should provide the open state as a boolean. The on-close event handler can be used to synchronize application state with the component's internal state`.

austingardner: okay I adjusted the README... let me know what you think

ianmcburnie (approved):
Thanks for your patience, Austin! The API LGTM!

src/components/components/ebay-dialog-base/index.marko
```diff
@@ -98,6 +101,9 @@ $ var header = input.header;
         <div id=input.mainId class=`${input.classPrefix}__main`  key="body" onScroll("handleScroll")>
             <${input.renderBody}/>
         </div>
+        <span class=`${input.classPrefix}__actions`>
+            <${input.action && input.action.renderBody}/>
+        </span>
```
agliga, line 106: If no action is present then this section should not get rendered.
Can we add it as a part of main or does it have to be it's own section? 
I'm fine, but generally we had header, main, footer

src/components/components/ebay-dialog-base/index.marko
```diff
@@ -29,6 +29,7 @@ static var ignoredHeaderAttributes = ["id", "as"];
 $ var buttonPosition = input.buttonPosition || "left";
 $ var baseEl = input.baseEl || "div";
 $ var header = input.header;
+$ var actions = input.actions;
```
agliga, line 32: Unused?

src/components/ebay-snackbar-dialog/examples/02-action/template.marko
```diff
@@ -0,0 +1,27 @@
...
+    Open Action Snackbar
+</ebay-button>
+<ebay-snackbar-dialog open=state.open on-close('handleClose') on-action("handleAction")>
+    This is the action snackbar
+    <@action access-key="U">
+        Undo
+    </@action>
+</ebay-snackbar-dialog>
+
+class {
```
agliga, line 11: Put class before html

src/components/ebay-snackbar-dialog/examples/03-stacked/template.marko
```diff
@@ -0,0 +1,27 @@
...
+    Open Stacked Snackbar
+</ebay-button>
+<ebay-snackbar-dialog open=state.open layout="column" on-close("handleClose") on-action("handleAction")>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+    <@action action-access-key="U">
+        Undo
+    </@action>
+</ebay-snackbar-dialog>
+
+class {
```
agliga, line 11: Put class before html

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,105 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+static var ignoredAttributes = [
+    "open"
```
agliga, line 4: Should also add layout and any other params?

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,105 @@
...
+    }
+    onDestroy() {
+        this.clearAllTimeouts();
+    }
+
+    onCreate() {
+        this.DEFAULT_TIMEOUT_LENGTH = 6000;
+        this.eventSet = new Set();
+        this.state = {open: false};
+        this.timeouts = [];
```
agliga, line 32: I'm not sure why is this array needed like this?

IMO, there should be 1 timeout to close the snackbar. 
Once you leave it should start to trigger the timeout. 
If you enter the snackbar, it clears the timeout. 
Why do we have multiple ones?

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,105 @@
+import processHtmlAttributes from "../../common/html-attributes"
+
+static var ignoredAttributes = [
+    "open"
+];
+
+class {
```
agliga, line 7: Our convention is to have `component.js` if the class is pretty large. Please move this into that file, since this is quite a bit of logic.

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    onInput(input) {
+        this.clearAllTimeouts();
+        this.state = {open: input.open || this.state.open || false}
+        if (input.open !== false) {
+            this.timeouts.push(setTimeout(() => {
+                this.emit("close");
+                this.state.open = false;
+            }, input.timeout || this.DEFAULT_TIMEOUT_LENGTH));
```
agliga, line 21: We can put this whole logic into another method since we are using it in multiple places.

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+    onDestroy() {
+        this.clearAllTimeouts();
+    }
+
+    onCreate() {
+        this.DEFAULT_TIMEOUT_LENGTH = 6000;
```
agliga, line 29: I would put this as a global variable outside the class scope (it works better in `component.js`)

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,105 @@
...
+        if (this.eventSet.size === 0 && this.state.open === true) {
+            this.timeouts.push(setTimeout(() => {
+                this.emit("close");
+                this.state.open = false;
+            }, this.DEFAULT_TIMEOUT_LENGTH));
+        }
+    }
+
+    handleBaseClose() {
+        console.log("handlebaseclose in snackbar")
```
agliga, line 73: Remove console.log

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,105 @@
...
+        this.clearAllTimeouts();
+        this.state.open = false;
+        this.emit("close");
```
agliga, line 76: Dialog base should emit close when dialog is closed
Try removing this and adding: `on-close('emit', 'close')` to `<ebay-dialog-base>` in the template and see if that works

src/components/components/ebay-dialog-base/component.js
```diff
@@ -36,6 +36,7 @@ module.exports = {
     handleKeydown(event) {
         eventUtils.handleEscapeKeydown(event, () => {
             this.state.open = false;
+            this.emit('baseClose');
```
agliga, line 39: Remove, `close` should be called when dialog finishes transition

src/components/ebay-snackbar-dialog/test/test.server.js
```diff
@@ -0,0 +1,34 @@
...
+        expect(dialog).has.class('snackbar-dialog--transition');
+        expect(getByText(input.renderBody.text)).has.class('snackbar-dialog__main');
+        expect(getByText(input.header.renderBody.text).parentElement).has.class(
+            'snackbar-dialog__header'
+        );
+        expect(getByText(input.action.renderBody.text).parentElement).has.class(
+            'snackbar-dialog__actions'
+        );
+        // debug(getByText(input.action.renderBody.text));
+        // expect(getByText(input.action.renderBody.text)).has.class('fake-link fake-link--secondary');
```
agliga, line 25: Comments remove?

src/components/ebay-snackbar-dialog/test/mock/mock-basic.marko
```diff
@@ -0,0 +1,17 @@
+<ebay-snackbar-dialog open=state.open on-close('closeSnackbar')>
```
agliga, line 1: Is this needed anymore?

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+
+<ebay-dialog-base
+    onBaseClose("handleBaseClose")
```
agliga, line 81: Change this to `onClose('emit', 'close')`

agliga (commented):
Lots of comments, feel free to chat with me if you need more info

src/components/ebay-snackbar-dialog/test/mock/mock-basic.marko
austingardner: yes, if they want to re-open the snackbar later. 

I will add an example with only the open state for clarity.

src/components/ebay-snackbar-dialog/index.marko
austingardner: So what was happening is in some cases multiple timeouts were created before they were cleared. Certain combinations of mouseLeave and Blur events were happening so that a couple timeouts were created before the first was cleared, and then the snackbar was timing out and closing too early later on. Having the array ensures that the timeouts are always cleared together.

src/components/ebay-snackbar-dialog/test/mock/mock-basic.marko
agliga: This is in mocks, not in examples; I don't think this is needed.

src/components/ebay-snackbar-dialog/index.marko
austingardner: Maybe we could consider switching up that convention? I definitely have a preference for putting things in one file. It tends to make things easier to follow because you don't have to switch between files. 

Also trying to switch to a `component.js` is breaking and suddenly it is saying that `this.emit is not a function` which doesn't make a lot of sense.

src/components/ebay-snackbar-dialog/index.marko
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    handleMouseEnter() {
+        this.clearAllTimeouts();
+        this.eventSet.add("mouseOver");
+    }
+
+    handleMouseLeave() {
+        this.eventSet.delete("mouseOver");
```
agliga, line 63: Should we not clear all timeoutes here and `handleBlur` as well?
This is related to the comment about the timeout array.

IMO this is what should happen.
* Anytime we add a new timeout, we clear the old one.
* Anytime we want to delay the timeout, we clear the old one
* On destroy clears the old one
This way we probably don't need multiple timeouts.

src/components/ebay-snackbar-dialog/index.marko
agliga: I did this change locally and it seems to be working (feel free to use it and let me know if there is a problem)
Only change I would do is remove the `clearAllTimeouts` and just do `clearTimeout(this.timeout)`
```
diff --git a/src/components/ebay-snackbar-dialog/index.marko b/src/components/ebay-snackbar-dialog/index.marko
index 6521528f..62bd96d3 100644
--- a/src/components/ebay-snackbar-dialog/index.marko
+++ b/src/components/ebay-snackbar-dialog/index.marko
@@ -6,19 +6,17 @@ static var ignoredAttributes = [
 
 class {
     clearAllTimeouts() {
-        while(this.timeouts.length) {
-            clearTimeout(this.timeouts.pop());
-        }
+        clearTimeout(this.timeout);
     }
 
     onInput(input) {
         this.clearAllTimeouts();
         this.state = {open: input.open || this.state.open || false}
         if (input.open !== false) {
-            this.timeouts.push(setTimeout(() => {
+            this.timeout = setTimeout(() => {
                 this.emit("close");
                 this.state.open = false;
-            }, input.timeout || this.DEFAULT_TIMEOUT_LENGTH));
+            }, input.timeout || this.DEFAULT_TIMEOUT_LENGTH);
         }
     }
     onDestroy() {
@@ -29,7 +27,6 @@ class {
         this.DEFAULT_TIMEOUT_LENGTH = 6000;
         this.eventSet = new Set();
         this.state = {open: false};
-        this.timeouts = [];
     }
 
     handleAction() {
@@ -47,10 +44,11 @@ class {
     handleBlur() {
         this.eventSet.delete("focus");
         if (this.eventSet.size === 0) {
-            this.timeouts.push(setTimeout(() => {
+            this.clearAllTimeouts();
+            this.timeout = setTimeout(() => {
                 this.emit("close");
                 this.state.open = false;
-            }, this.DEFAULT_TIMEOUT_LENGTH));
+            }, this.DEFAULT_TIMEOUT_LENGTH);
         }
     }
 
@@ -62,10 +60,11 @@ class {
     handleMouseLeave() {
         this.eventSet.delete("mouseOver");
         if (this.eventSet.size === 0 && this.state.open === true) {
-            this.timeouts.push(setTimeout(() => {
+            this.clearAllTimeouts();
+            this.timeout = setTimeout(() => {
                 this.emit("close");
                 this.state.open = false;
-            }, this.DEFAULT_TIMEOUT_LENGTH));
+            }, this.DEFAULT_TIMEOUT_LENGTH);
         }
     }
 ```

src/components/components/ebay-dialog-base/index.marko
austingardner: Okay putting it in the main section causes CSS weirdness. But I added an if tag so it doesn't render if there's no action.

agliga (commented):
I was going to let these slide, but you need to update the examples one. So while you are at it, these should be quick.

agliga (approved):

agliga:
Just change the title and remove the number before merging, the PR number will show up

PR #1339, opened 2021-02-24
video features for multiple sources, noscript, and fullscreen

agliga:
## Description
Several changes from the videeo review
* Added support for multiple sources. Can be a single file or array. (I left it like that because they might want one file so we shouldn't force them to use array syntax). Will try each video and then go to the next one. 
* Added `source` initially to allow noscript version to work (had to make loading show on initial rerender)
* Added fullscreen option which triggers fullscreen on player play
* Changed retry to error (for now). Will have a followup to differentiate between error and retry.


ianmcburnie:
Is there a reason for not doing nested `<source>` tags like the HTML video tag does?

agliga:
> Is there a reason for not doing nested `<source>` tags like the HTML video tag does?

We are doing that, it's iterating through all the sources. However the mpd and m3u8 ones cause some other issues (and those are added by the video player)

austingardner (approved):
LGTM!

ianmcburnie (approved):

Issue #1341, opened 2021-02-24
Dialog: Cannot focus back to the button opening the dialog when the dialog is closed


cynthiahhc:
The issue happens in Safari and Firefox (using OX), and it's in both 6.x.x and 5.x.x versions.

Usually, we open the dialog by clicking a button, but in Safari and Firefox, clicking the button will not focus on the button. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus

When `ebay-dialog-base` tries to cache the previously focused element before it opens by using `document.activeElement`, it caches the incorrect element. The button is never focus and when the dialog close, it will restore the focus to some other element, causing the page to scroll/bounce.

https://github.com/eBay/ebayui-core/blob/5.7.0/src/components/components/ebay-dialog-base/component.js#L126

ianmcburnie:
Thank you for reporting this @cynthiahhc

agliga:
Released in `@ebay/ebayui-core@5.7.13` and `@ebay/ebayui-core@7.1.2`

PR #1342, opened 2021-02-25
Updated dialog for adding safari activeElement support

agliga:
## Description
Added a click listener on window to track click changes in order to support safari's activeElement issues. 
Used the code from https://github.com/DylanPiercey/parse-form/blob/master/src/index.ts#L85-L120 (Thanks Dylan) 

## References
#1341

### DIFF
--- a/src/components/components/ebay-dialog-base/component.js
+++ b/src/components/components/ebay-dialog-base/component.js
@@ -5,6 +5,34 @@ const eventUtils = require('../../../common/event-utils');
 const transition = require('../../../common/transition');
 
 module.exports = {
+    trackLastClick(e) {
+        if (
+            e.defaultPrevented ||
+            e.metaKey ||
+            e.ctrlKey ||
+            e.shiftKey ||
+            e.button !== 0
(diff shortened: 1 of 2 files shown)


src/components/components/ebay-dialog-base/component.js
```diff
@@ -155,13 +184,13 @@ module.exports = {
                     this.emit('open');
                 } else {
                     this._triggerBodyScroll(false);
-                    const activeElement = document.activeElement;
+                    const activeElement = this.getActiveElement();
```
DylanPiercey, line 187: Since calling the function clears the click target, shouldn't we cache the call in this function call to avoid calling it more than once and getting different `activeElement`s?

agliga: I removed that clearing of the activeElement for now. 
I don't think there's a need to clear it when we call it. We are removing the event when the dialog opens.

ianmcburnie (approved):

DylanPiercey (approved):

PR #1357, opened 2021-03-24
Added fix for video player runtime and with amd loading

agliga:
## Description
* ~~With the way we are doing babel bundling and we are using promises, we have to add a `regenrator-runtime` module to make it work with `commonjs`. Since video won't work without this, we need to add it as a dependency.~~ Removed not needed `async` function from client code. 
* There is an issue with loading with `define`, so in the minifed shaka code, I removed that check to always fail


austingardner (approved):
LGTM

PR #1359, opened 2021-03-29
Menu Button Labeling

### LINKED ISSUE #1350: Menu-Button: support other labelling scenarios
Basically repeat what we did for listbox-button, by adding `prefix-id` and `prefix-label` attributes to ebay-menu-button.

https://github.com/eBay/ebayui-core/issues/1321

austingardner:
## Description
Adds prefixLabel and prefix-id to the menu-button. 

## Context
These were added previously to listbox-button and are now being added to menu-button.

## References
closes #1350 

## Screenshots
Screen Shot 2021-04-02 at 12 56 59 PM
Screen Shot 2021-04-02 at 12 57 09 PM

### DIFF
--- a/src/components/ebay-menu-button/examples/19-prefix-label/template.marko
+++ b/src/components/ebay-menu-button/examples/19-prefix-label/template.marko
@@ -1,8 +1,20 @@
-<ebay-menu-button type="radio" prefix-label="Select From Menu">
+class {
+    onCreate() {
+        this.state = { labelText: 'label text', index: -1 };
+    }
+
+    handleChange(e) {
+        console.log(e);
+        this.state.labelText = e.el.textContent;
+        this.state.index = e.index;
+    }
+}
+
+<ebay-menu-button type="radio" prefix-label="Sort By:" on-change('handleChange')>
     <@label>
-        label text
+        ${state.labelText}
     </@label>
-    <@item checked>item 1 that has very long text</@item>
-    <@item>item 2</@item>
-    <@item>item 3</@item>
+    <@item checked=state.index==0>Color</@item>
+    <@item checked=state.index==1>Year</@item>
+    <@item checked=state.index==2>Mileage</@item>
 </ebay-menu-button>
--- a/src/components/ebay-menu-button/examples/19-prefix-label/template.marko
+++ b/src/components/ebay-menu-button/examples/19-prefix-label/template.marko
@@ -4,7 +4,6 @@ class {
     }
 
     handleChange(e) {
-        console.log(e);
         this.state.labelText = e.el.textContent;
         this.state.index = e.index;
     }
--- a/src/components/ebay-menu-button/examples/19-prefix-label/template.marko
+++ b/src/components/ebay-menu-button/examples/19-prefix-label/template.marko
@@ -1,6 +1,6 @@
 class {
     onCreate() {
-        this.state = { labelText: 'label text', index: -1 };
+        this.state = { labelText: 'Color', index: 0 };
     }
 
     handleChange(e) {
--- a/src/components/ebay-menu-button/examples/19-prefix-label/template.marko
+++ b/src/components/ebay-menu-button/examples/19-prefix-label/template.marko
@@ -1,0 +1,8 @@
+<ebay-menu-button type="radio" prefix-label="Select From Menu">
+    <@label>
+        label text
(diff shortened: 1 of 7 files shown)


src/components/ebay-menu-button/examples/19-prefix-label/template.marko
```diff
@@ -0,0 +1,8 @@
+<ebay-menu-button type="radio" prefix-label="Select From Menu">
+    <@label>
+        label text
```
ianmcburnie, line 3: Like in the listbox button example, can you make sure the label text is updated to reflect the current selection.

src/components/ebay-menu-button/examples/19-prefix-label/template.marko
```diff
@@ -0,0 +1,8 @@
+<ebay-menu-button type="radio" prefix-label="Select From Menu">
```
ianmcburnie, line 1: Let's go with a more realistic example here, such as `"Sort: "` or `"Filter: "`

src/components/ebay-menu-button/examples/20-prefix-id/template.marko
```diff
@@ -0,0 +1,8 @@
+<div id="custom-label">
+    Menu Label
+</div>
+<ebay-menu-button prefix-id="custom-label" text="eBay Menu">
```
ianmcburnie, line 4: Same as previous example, can you make sure the label text is updated to reflect the current selection. If the button label text never updates, then an external label shouldn't be used, so I want to make sure the examples are good.

src/components/ebay-menu-button/examples/19-prefix-label/template.marko
```diff
@@ -0,0 +1,8 @@
...
+    <@label>
+        label text
+    </@label>
+    <@item checked>item 1 that has very long text</@item>
```
ianmcburnie, line 5: Let's remove this very long text option in both examples and use something more real-life for each option.

ianmcburnie (changes requested):
For prefix-id, the inner text of the button is not being included in the screen reader readout. Seems like the span inside the button is missing an id:

Screen Shot 2021-04-03 at 8 49 16 AM

src/components/ebay-menu-button/examples/20-prefix-id/template.marko
```diff
@@ -0,0 +1,18 @@
+class {
+    onCreate() {
+        this.state = { buttonText: 'Select An Option' };
```
ianmcburnie, line 3: Very last thing I promise, but can we have the first value (i.e. Name) instead of 'Select an Option' as the initial text.

src/components/ebay-menu-button/examples/19-prefix-label/template.marko
```diff
@@ -0,0 +1,19 @@
+class {
+    onCreate() {
+        this.state = { labelText: 'label text', index: -1 };
```
ianmcburnie, line 3: Likewise, can we have the first value (i.e. Color) instead of 'label text' as the initial text.

src/components/ebay-menu-button/index.marko
```diff
@@ -55,17 +60,26 @@ $ var isOverflowVariant = input.variant === "overflow";
...
-                    <${input.label.renderBody}/>
+                    <if(input.prefixLabel)>
+                        <span class="menu-button-prefix-label">${input.prefixLabel}</span>
+                        <${input.label.renderBody}/>
+                    </if>
+                    <else >
+                        <span id=labelId>
+                            <${input.label.renderBody}/>
+                        </span>
+                    </else>
```
agliga, line 71: I don't think this is all needed.
You can probably wrap the if statement with a span like this:
```
<${labelId ? "span" : null} id=labelId>
   <if>
   <else-if>
</>
```
That way, if labelId is set, then it gets rendered and wraps all the possible text with the span.

ianmcburnie (approved):

src/components/ebay-menu-button/index.marko
ianmcburnie: @austingardner I'm okay with what we have for now. I find it readable. If you think Andrews suggestion would be equally as readable, then go ahead and make the change. Otherwise, I'm fine with it as is.

Issue #1340, opened 2021-02-24
NEW: EEK Badge Component


ianmcburnie:
Related skin ticket: https://github.com/eBay/skin/issues/1374

![Screen Shot 2021-02-24 at 12 27 08 PM](https://user-images.githubusercontent.com/38065/109062576-f5751980-769c-11eb-97ff-ce7c93778bb8.png)
![Screen Shot 2021-02-24 at 12 26 56 PM](https://user-images.githubusercontent.com/38065/109062584-f73edd00-769c-11eb-8bb8-33fac3febe5e.png)

ianmcburnie:
Done in 7.4.0

PR #1365, opened 2021-04-05
feat(eek): added new eek component

agliga:
## Description
Added new eek component. Imported icons from skin

## References
https://github.com/eBay/ebayui-core/issues/1340

## Screenshots
Screen Shot 2021-04-05 at 2 54 06 PM


ianmcburnie:
Trying to avoid direction specific language here, I think "high" and "low" might be better than "top" and "bottom". Hopefully that aligns too with EEK language...

ianmcburnie (approved):

PR #1366, opened 2021-04-06
Signal

### LINKED ISSUE #1363: NEW: signal component
A simple component for the "Hotness Signal" in Skin.

https://github.com/eBay/skin/issues/1367

austingardner:
## Description
Signals are data-backed recommendations to help customers make more informed decisions.
There are four signal types, each corresponding to a specific color: trustworthy, recent, time-sensitive & neutral.

## Context
Finishes up the Signal component already merged in Skin. 

## References
Closes #1363 

## Screenshots
Screen Shot 2021-04-09 at 7 48 31 AM

### DIFF
--- a/src/components/ebay-signal/index.marko
+++ b/src/components/ebay-signal/index.marko
@@ -1,0 +1,5 @@
+$ var type = input.type;
+
+<span class=`signal signal--${input.type}`>
+    <${input.renderBody}/>
+</span>
--- a/src/components/ebay-signal/index.marko
+++ b/src/components/ebay-signal/index.marko
@@ -1,5 +1,11 @@
-$ var type = input.type;
+import processHtmlAttributes from "../../common/html-attributes";
 
-<span class=`signal signal--${input.type}`>
+static var ignoredAttributes = [
+    "status",
+];
+
+$ var defaultStatus = "neutral";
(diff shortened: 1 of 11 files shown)


src/components/ebay-signal/index.marko
```diff
@@ -0,0 +1,5 @@
+$ var type = input.type;
```
agliga, line 1: We should have a default value
Also I think we generally don't like `type`. Maybe `status` is better?

src/components/ebay-signal/index.marko
```diff
@@ -0,0 +1,5 @@
...
+
+<span class=`signal signal--${input.type}`>
```
agliga, line 3: Pass in the `...input`

Also do we need any `role` or aria attributes here? At least `role` would probably be needed.

agliga (commented):
Can you also add some server tests, just to check that the type is getting set correctly?

src/components/ebay-signal/index.marko
ianmcburnie: Good point. I'll have a think if we need anything and add it to the Skin docs if so.

ianmcburnie (approved):

Issue #1354, opened 2021-03-18
fake-tabs: allow aria-current="true" 


ianmcburnie:
Scope: allow configuration/enumeration of aria-current.

Currently, fake-tabs always sets `aria-current=page` for the current tab. However, this is not always valid in cases when there is another level of site navigation under the tabs.

In the screenshot below, both the first tab link  "Activity") and the first link of secondary navigation ("Summary') both reflect the current page url, and so it is valid for them to both have aria-current=page.

<img width="893" alt="Screen Shot 2021-03-18 at 1 40 54 PM" src="https://user-images.githubusercontent.com/38065/111696138-2c989f80-87f1-11eb-8474-d56b7c651a04.png">

However, in the following case, only the "Recently viewed" link should have aria-current=page. The "Activity" link should have `aria-current=true`.

<img width="963" alt="Screen Shot 2021-03-18 at 1 41 17 PM" src="https://user-images.githubusercontent.com/38065/111696159-33271700-87f1-11eb-8d5e-f759e11b71fd.png">

ianmcburnie:
Done in 7.4.0

PR #1369, opened 2021-04-09
Allow Aria Current for Fake Tabs

austingardner:
PLEASE SQUASH & MERGE

## Description
This allows the user to set aria-current to `true` as well as `"page"`, which remains the default. 

I wasn't quite sure what to name the attribute - `tab-aria-current` doesn't seem quite right. Definitely could use some suggestions on that one. 

## References

#1354 

## Screenshots
Screen Shot 2021-04-08 at 6 17 46 PM
Screen Shot 2021-04-08 at 6 17 37 PM


src/components/ebay-fake-tabs/marko-tag.json
```diff
@@ -6,6 +6,9 @@
   },
   "@html-attributes": "expression",
   "@selected-index": "number",
+  "@tab-aria-current": {
+    "enum": ["page", true]
```
ianmcburnie, line 10: Value needs to be a string. See how `aria-current` is not being rendered properly in example:

Screen Shot 2021-04-09 at 9 56 27 AM

ianmcburnie (changes requested):

src/components/ebay-fake-tabs/marko-tag.json
ianmcburnie: Also, I think `aria-current-value` would be better than `tab-aria-current`.

However, given that "current" and "true" are the only possible values, ideally we might want to use an abstraction. Something like a boolean, `tab-matches-current-url` (can't think of a good name right now) which would be `true by default and use `aria-current="page"`; if set to `false` would use `aria-current="true"`.

ianmcburnie:
Seems to be a lint-spaces error in CI build

src/components/ebay-fake-tabs/README.md
```diff
@@ -20,10 +20,11 @@
...
-| `selectedIndex`  | String | Yes      | No       | 0-based index of selected tab tab and panel                             |
-| `tabAriaCurrent` | Enum   | No       | No       | If provided, will set aria-current to provided value. Default is "page" |
+| Name                                                | Type    | Stateful | Required | Description                                                                                                       |
+| --------------------------------------------------- | ------- | -------- | -------- | ----------------------------------------------------------------------------------------------------------------- |
+| `selectedIndex`                                     | String  | Yes      | No       | 0-based index of selected tab tab and panel                                                                       |
+| `tabMatchesCurrentUrl`                              | Boolean | No       | No       | If provided, will set aria-current to "page" if true, or will set aria-current "true" if value provided is false. |
+| default is true and will set aria-current to "page" |
```
ianmcburnie, line 27: Markdown syntax error here? Let's go with:

"Specify whether the href of the currently active fake tab matches the current window url. Default is true. This property is used to configure the underlying aria-current attribute (i.e. a value of "page" (default) or "true").

ianmcburnie (approved):

PR #1374, opened 2021-04-13
fix(menu-button): default priority was incorrect #1372

### LINKED ISSUE #1372: fake-menu-button: various issues with stories
# Bug Report

## Description

Firstly, I noticed the disabled story does not look correct, and it is not actually disabled. This is a simple fix. The `disabled` attribute is missing from the example.

Screen Shot 2021-04-12 at 4 18 16 PM

Then I notice that all the stories are being rendered with `expand-btn--secondary` (blue) modifier. By default they should have no modifier (black).

This is due to button using secondary as default. This should be set to priority=none for expand button.

Looks like the default priority was never updated from secondary to none:

`button priority, "primary" / "secondary" (default) / "none`

UPDATE: This applies to ebay-menu-button too.

Screen Shot 2021-04-12 at 4 21 18 PM

Careful comparison with what we have in Skin and playbook will be needed to ensure we get these buttons properly in sync.

Related ticket: https://github.com/eBay/skin/issues/1405

ianmcburnie:
Fixes #1372 

## Description

The default menu button priority is changed from "secondary" to "none". This could *technically* be considered a breaking change.... however, I'm going to argue that it is a **bug** since v7.0.0 was supposed to have made this change (which mirrors what we have in Skin v12). I will put a big notice in the release notes to make this very clear. 

## Screenshots

Disabled story is fixed:

Screen Shot 2021-04-13 at 2 30 30 PM

Default button priority has been corrected from "secondary" to "none".

Screen Shot 2021-04-13 at 3 45 00 PM


austingardner (approved):

austingardner:
What leaves me a little confused about this is the button priority's options - is the only valid option now "none", because primary and secondary are deprecated?

PR #1375, opened 2021-04-21
fix(ebay-video): fixed some async reloading problems

agliga:
## Description
There is an issue on the examples when the component is rerendered it won't load the video anymore. This fixes it by retrying on error code 11 which is video not ready yet. 
Also added some other changes to check for retrying if the component is not destroyed yet since it uses an async way to do destroy.

### DIFF
--- a/src/components/ebay-video/component.js
+++ b/src/components/ebay-video/component.js
@@ -3,6 +3,13 @@ const versions = require('./versions.json');
 const MAX_RETIRES = 3;
 
 module.exports = {
+    reattach(callback) {
+        if (this.isDetaching) {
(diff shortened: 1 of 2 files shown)


ianmcburnie (dismissed):
Seems to be working better now!

austingardner (approved):
LGTM!

src/components/ebay-video/component.js
```diff
@@ -3,6 +3,13 @@ const versions = require('./versions.json');
 const MAX_RETIRES = 3;
```
austingardner, line 3: typo here

ianmcburnie (approved):

PR #1376, opened 2021-04-21
Storybook Fix

austingardner:
## Description
There were issues running storybook, something related to webpack. 

I was unable to run storybook locally - it would crash. 

This adds a dependency, html-webpack-plugin, that was throwing out a warning, and now things work. 

Also checked - this works on safari and edge, code examples are working 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "build:ci": "npm run lint && node scripts/check-postpublish && npm test && npm run build",
     "build:readme": "node scripts/add-ds-version",
     "storybook:build": "DS=6 build-storybook -o ./docs && DS=4 build-storybook -o ./docs/ds4",
-    "storybook:start": "start-storybook -s demo/assets -p 6006",
+    "storybook:start": "start-storybook -p 6006",
     "importSVG": "node scripts/import-svg",
     "ts": "npm run test:server",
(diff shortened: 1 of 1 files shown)


ianmcburnie (approved):
Nice fix!

ianmcburnie:
i thought https://github.com/eBay/ebayui-core/commit/6b15c8df2dcfcaee062fc034c99f5447859e14c8 had already been reverted...?

package-lock.json
```diff
@@ -4022,6 +4022,23 @@
           "integrity": "sha512-EykJT/Q1KjTWctppgIAgfSO0tKVuZUjhgMr17kqTumMl6Afv3EISleU7qZUzoXDFTAHTDC4NOoG/ZxU3EvlMPQ==",
           "dev": true
         },
+        "html-webpack-plugin": {
```
austingardner, line 4025: this isn't in package.json, it is here because it is a dependency of one of our dependencies.

agliga (commented):
Can you do a `git checkout origin/7.4.0 package-lock.json`? I don't want to push this change with those new versions.

austingardner:
Storybook doesn't run for me without those changes. 😬

agliga (approved):

PR #1380, opened 2021-04-26
Emit Keydown Event for Radio and Checkbox

### LINKED ISSUE #1377: Checkbox & radio: emit the 'keydown' event
# Feature Request

## Description
Emit the `keydown` event for `ebay-checkbox` and `ebay-radio`.

## Context
There are usecases in our flow where we need to perform actions based on certain keys pressed when focused on checkboxes and radios.

austingardner:
## Description
This PR forwards the keydown even from the Radio and Checkbox components. 

## References
closes #1377


src/components/ebay-combobox/component.js
```diff
@@ -175,12 +183,13 @@ module.exports = {
...
                 this.getEl('listbox'),
-                '[role="option"]', {
+                '[role="option"]',
+                {
                     activeDescendantClassName: 'combobox__option--active',
                     autoInit: -1,
                     autoReset: -1,
                     axis: 'y',
-                    autoScroll: true
+                    autoScroll: true,
```
ianmcburnie, line 192: Is this a default prettier setting or something we enabled? Not a huge fan of these trailing commas. Looks so odd to me.

austingardner: Yeah it's a prettier setting. We can change it to 'trailingCommas: none' if we want to, but I think having trailing commas is the default. I feel like it's easier because then when you add an item to an array or object you don't have to go back to the previous one and add a comma

src/components/ebay-radio/index.marko
```diff
@@ -11,7 +11,8 @@ $ input.toJSON = noop;
         type="radio"
         class="radio__control"
         onClick("handleClick")
-        onFocus("handleFocus")/>
+        onFocus("handleFocus")
+        on-keydown("handleKeyDown")/>
```
agliga, line 15: Follow the other syntax
`onKeydown` 
Also for the methodname, do `handleKeydown` to match the text. (Don't worry about changing it in the 5.0 branch)

src/components/ebay-combobox/component.js
ianmcburnie: Okay, despite my cognitive dissonance, let's keep the default setting to keep my aging brain on its toes (a brain with toes... weird imagery there).

ianmcburnie (approved):

src/components/ebay-combobox/component.js
```diff
@@ -115,14 +115,21 @@ module.exports = {
             this.focus();
         }
 
-        if (this.expander && this.expander.isExpanded() && !wasClickedOption && !this.buttonClicked) {
+        if (
```
austingardner, line 118: this file just had some leftover prettier changes... I can drop those from the PR if needed

ianmcburnie: Yeah, let's drop any unrelated files.

agliga (approved):

Issue #1379, opened 2021-04-26
Program Badge: new component


ianmcburnie:
Simple wrapper component for program badges that will be added in Skin 12.5.0.

ianmcburnie:
Done. Merged to 7.5.0

PR #1387, opened 2021-05-07
feat: added program badges and imported new icons

agliga:
PLEASE SQUASH

## Description
Added program badges to ebayui

## Context
So since program badges used another file, I had to change the script. Also the icon syntax was a bit different so I had to use some changes to our existing pattern in icons.
I exposed a new variable `_type` which is internal to detonate between `icons` or `program-badges`.

## References
#1379

## Screenshots
Screen Shot 2021-05-07 at 2 00 45 PM


agliga:
> Running into this error:
> 
> Screen Shot 2021-05-10 at 8 54 20 AM

Gotta link skin with the `program-badges` module 👍

ianmcburnie (approved):

PR #1397, opened 2021-05-19
feat: Added itemMatchesUrl to fake-menu, allows aria-current=true

### LINKED ISSUE #1395: fake-menu-button: allow aria-current="true"
Similar to https://github.com/eBay/ebayui-core/issues/1354, let's support `aria-current=true` instead of only the hardcoded `aria-current=page`

austingardner:
## Description
Added itemMatchesUrl to the fake-menu @item, which allows the user to set aria-current to `true` instead of just page.

Right now it doesn't show the check mark when aria-current is set to `true` - @ianmcburnie is that the intention?

## References
closes #1395 

## Screenshots
Screen Shot 2021-05-19 at 4 08 46 PM
Screen Shot 2021-05-19 at 4 09 05 PM

### DIFF
--- a/src/components/ebay-fake-menu/README.md
+++ b/src/components/ebay-fake-menu/README.md
@@ -45,15 +45,16 @@
 
 ### @item Attributes
 
-| Name                          | Type    | Stateful | Required                               | Description                                                                 |
-| ----------------------------- | ------- | -------- | -------------------------------------- | --------------------------------------------------------------------------- |
-| `href` (fake menu)            | String  | No       | No                                     | for link that looks like a menu-item. If set to null then will disable item |
-| `type` (fake menu)            | String  | No       | No                                     | Set to "button" for fake menu-item `<button>`                               |
-| `value` (radio or checkbox)   | String  | No       | No                                     | the value to use with event responses for for the `checked` array           |
-| `checked` (radio or checkbox) | Boolean | No       | No                                     | whether or not the item is checked                                          |
-| `current` (fake menu)         | Boolean | No       | No                                     | whether or not the href is the current href of the page                     |
-| `badge-number`                | Number  | No       | No                                     | used as the number to be placed in the badge                                |
-| `badge-aria-label`            | String  | No       | Yes (only if badge number is provided) | passed as the `aria-label` directly to the badge                            |
+| Name                          | Type    | Stateful | Required                               | Description                                                                                                                                                                |
+| ----------------------------- | ------- | -------- | -------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `href`                        | String  | No       | No                                     | for link that looks like a menu-item. If set to null then will disable item                                                                                                |
+| `type`                        | String  | No       | No                                     | Set to "button" for fake menu-item `<button>`                                                                                                                              |
+| `value` (radio or checkbox)   | String  | No       | No                                     | the value to use with event responses for for the `checked` array                                                                                                          |
+| `checked` (radio or checkbox) | Boolean | No       | No                                     | whether or not the item is checked                                                                                                                                         |
+| `current`                     | Boolean | No       | No                                     | whether or not the href is the current href of the page                                                                                                                    |
+| `itemMatchesUrl`              | Boolean | No       | No                                     | used in conjunction with `current` -- This determines whether aria-current will be `'page'` or `'true'` -- Defaults to `true` which gives aria-current a value of `'page'` |
+| `badge-number`                | Number  | No       | No                                     | used as the number to be placed in the badge                                                                                                                               |
+| `badge-aria-label`            | String  | No       | Yes (only if badge number is provided) | passed as the `aria-label` directly to the badge                                                                                                                           |
 
 ## ebay-menu-separator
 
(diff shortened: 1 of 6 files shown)


src/components/ebay-fake-menu/README.md
```diff
@@ -45,15 +45,16 @@
...
+| `value` (radio or checkbox)   | String  | No       | No                                     | the value to use with event responses for for the `checked` array                                                                                                    |
+| `checked` (radio or checkbox) | Boolean | No       | No                                     | whether or not the item is checked                                                                                                                                   |
+| `current` (fake menu)         | Boolean | No       | No                                     | whether or not the href is the current href of the page                                                                                                              |
+| `itemMatchesUrl` (fake menu)  | Boolean | No       | No                                     | used in conjunction with `current` -- This determines whether aria-current will be `page` or `true` -- Defaults to `true` which gives aria-current a value of `page` |
```
agliga, line 55: We can remove `(fake menu)` since they're all fake menu

agliga, line 55: I'm not sure I really like this name and how it works. It seems a bit off having to name it as `itemMatchesUrl=false`

agliga (changes requested):
Couple of things:

* Make sure you have a `marko.json`
* Make sure we add some sanity tests

Maybe Ian can answer, we have two types only right? "page" or "true"?
Maybe we should have an argument saying: `a11yCurrentValue` or something, and then either set `page` or `true`? We can document that so that people know to set either those two values. We can default to `page` if not set.
We can also make it so that anything other than `true` does not get set, so like:
$ var defaultAriaCurrent = item.a11yCurrentValue === 'true' ? 'true' : 'page';
I think that might be cleaner

src/components/ebay-fake-menu/README.md
ianmcburnie: We are trying to abstract away the aria internals here, so that developers don't need knowledge of what is the correct ARIA attribute to use.

ianmcburnie: We could also just go with `currentPage` or `currentItem` (and deprecate `current`). Might be cleaner.

ianmcburnie: Also, for tabs I know we were trying to do this in a non-breaking way. Same applies here. For next major version we could look into making `current` an enumerated type.

ianmcburnie: EDIT: actually, scrap that, I like Andrews approach better:

```
We can also make it so that anything other than true does not get set, so like:
$ var defaultAriaCurrent = item.a11yCurrentValue === 'true' ? 'true' : 'page';
```

austingardner:
Added a test as well 👍

src/components/ebay-fake-menu/examples/06-a11y-current-value/template.marko
```diff
@@ -0,0 +1,5 @@
+<ebay-fake-menu>
+    <@item href="#" current a11y-current-value='true'>item 1</@item>
```
ianmcburnie, line 2: I thought we were going with enhancing `current` to be an enumerated type, and falling back to a boolean for backwards compatibility? If we are sticking with two attributes, my preference would be `itemMatchesUrl` to be consistent with what we did in tabs.

agliga: I think `current` should stay as a boolean. We have booleans in other places where it makes sense and there can only be 1 possible value. HTML has the same. 
I do think that either:
`a11y-current-value` should be on the parent though if we are going this route.
If we do `current='true|page'` that can work too, but it might be confusing what the default would be as well as overloading both a boolean/string. 

My personal preference is keep current as a boolean, set the default on the `<ebay-fake-menu>`. I don't care about the variable name as much, but I think an enum makes sense because we can potentially have other values in the future.

ianmcburnie: A problem with `a11y-current-value` is that it implies that any [aria-current](https://www.w3.org/TR/wai-aria-1.1/#aria-current) value can be used. Really, it's only every going to be "true" or "page" in this case. Probably not a huge deal, and we could clear it up in our docs.

A potential problem of putting this on the root element is that for some fake menus there might be a mix of items where some should have `page` and some should have `true`. Definitely getting into edge cases here, but it is a possible scenario.

austingardner: Hmm I guess I don't quite understand what you're saying about the root element. The `a11y-current-value` is applied on each item, not on the `fake-menu`. So that should allow the user to configure each item individually

ianmcburnie:
This will go into 7.6.0 once the branch is ready (after we release 7.5.3)

src/components/ebay-fake-menu/examples/06-a11y-current-value/template.marko
ianmcburnie: Andrew was proposing moving it from each item to the item parent, but I foresee a problem with that.

I think for the sake of moving things forward, I recommend just going with:

`<@item href="#" current item-matches-url='false'>item 1</@item>`

This unblocks us and aligns with what we currently have in fake-tabs. We can come back to this and make some breaking changes and cleanup in v8.

ianmcburnie (dismissed):

ianmcburnie (approved):

PR #1398, opened 2021-05-20
fix(dialog): added a check to remove traps only on modals

agliga:
## Description
There's an issue with snackbar dialog, when it's on the page with another dialog and it gets rerendered, it will remove all the keyboard traps for other dialogs.
Added a fix for snackbar to NOT remove any traps since it is not a modal and does not add any traps


austingardner (approved):
LGTM!

ianmcburnie (approved):

PR #1400, opened 2021-05-24
Removed aria-labelledby from inline-notice

### LINKED ISSUE #1394: Inline-Notice: invalid/redundant aria-labelledby attribute
The `aria-labelledby` attribute should not be showing up on inline notice. In general `aria-labelledby` and `aria-label` should never be on elements that do not have an explicit or implicit role (i.e. spans & divs). Good catch by axe here!

image (2)

I also notice that an `icon` attribute is showing up in the rendered HTML:

Screen Shot 2021-05-19 at 8 36 08 AM

austingardner:
## Description
This PR removes aria-labelledby from inline-notice

## References
closes #1394

### DIFF
--- a/src/components/components/ebay-notice-base/index.marko
+++ b/src/components/components/ebay-notice-base/index.marko
@@ -12,13 +12,14 @@ static var ignoredAttributes = [
     "title",
     "footer",
     "type",
-    "mainRoot"
+    "mainRoot",
+    "noA11yLabel"
 ];
 
 $ var status = input.status;
 $ var prefixClass = input.prefixClass;
 <${input.root || "section"}
-    aria-labelledby:scoped="status"
+    aria-labelledby=(!input.noA11yLabel && component.elId('status'))
     class=[prefixClass, input.class]
     ...processHtmlAttributes(input, ignoredAttributes)>
     <${input.headerRoot || "div"} class=`${prefixClass}__header` id:scoped="status"
--- a/src/components/components/ebay-notice-base/index.marko
+++ b/src/components/components/ebay-notice-base/index.marko
@@ -3,6 +3,7 @@ import processHtmlAttributes from "../../../common/html-attributes"
 static var ignoredAttributes = [
     "status",
     "a11yText",
+    "icon",
     "iconClass",
     "class",
     "root",
--- a/src/components/ebay-inline-notice/index.marko
+++ b/src/components/ebay-inline-notice/index.marko
@@ -1,5 +1,6 @@
 $ var status = input.status || "attention";
 
+
 <ebay-notice-base
     ...input
     class=[`inline-notice--${status}`, input.class]
@@ -8,4 +9,5 @@ $ var status = input.status || "attention";
     root="div"
     header-root="span"
     main-root="span"
+    noA11yLabel=true
 />
--- a/src/components/ebay-inline-notice/index.marko
+++ b/src/components/ebay-inline-notice/index.marko
@@ -8,5 +8,5 @@ $ var status = input.status || "attention";
     root="div"
     header-root="span"
     main-root="span"
-    noA11yLabel=true
+    noA11yLabel
 />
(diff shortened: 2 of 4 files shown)


src/components/components/ebay-notice-base/index.marko
```diff
@@ -15,11 +15,12 @@ static var ignoredAttributes = [
     "mainRoot"
 ];
 
+$ var ariaLabelledBy = input.prefixClass.includes("inline-notice") ? false : "status";
```
ianmcburnie, line 18: Not sure it makes sense for inline notice to be part of notice base. Maybe something we can revisit in v8 refactoring.

agliga: I think it should be, there is a lot of stuff that is being used like the icon and such.

That said, I think we should not do it this way but add a new argument or something.

Let's make a new input like `noA11yLabel`.
Also I would just do: 
`aria-labelledby=(!input.noA11yLabel && component.elId('status')`

src/components/components/ebay-notice-base/index.marko
```diff
@@ -15,11 +15,12 @@ static var ignoredAttributes = [
...
 $ var status = input.status;
 $ var prefixClass = input.prefixClass;
 <${input.root || "section"}
-    aria-labelledby:scoped="status"
-    class=[prefixClass, input.class]
+    aria-labelledby=ariaLabelledBy && component.elId(ariaLabelledBy),
+    class=[prefixClass, input.class],
```
agliga, line 23: Why is there a comma here?

agliga (changes requested):

src/components/components/ebay-notice-base/index.marko
austingardner: the linter told me I needed a comma

agliga: I think this is a syntax error.
You need to surround the `ariaLabelledBy && component.elId(ariaLabelledBy)` by parentheses.

austingardner: OK I can do that. However I'm pretty sure it actually is legitimate Marko to have the comma there. See 
[Marko issue](https://github.com/marko-js/marko/issues/424)

ianmcburnie:
While you are digging around in this code, can you also see if you can prevent `icon` from being rendered in the output:

Screen Shot 2021-05-19 at 8 36 08 AM

ianmcburnie (dismissed):

src/components/ebay-inline-notice/index.marko
```diff
@@ -8,4 +8,5 @@ $ var status = input.status || "attention";
     root="div"
     header-root="span"
     main-root="span"
+    noA11yLabel=true
```
agliga, line 11: Remove true, just `noA11yLabel`

src/components/components/ebay-notice-base/index.marko
```diff
@@ -12,13 +13,14 @@ static var ignoredAttributes = [
...
 ];
 
 $ var status = input.status;
 $ var prefixClass = input.prefixClass;
 <${input.root || "section"}
-    aria-labelledby:scoped="status"
+    aria-labelledby=(!input.noA11yLabel && component.elId('status'))
     class=[prefixClass, input.class]
     ...processHtmlAttributes(input, ignoredAttributes)>
     <${input.headerRoot || "div"} class=`${prefixClass}__header` id:scoped="status"
```
agliga, line 26: Can you check and make sure that this value is the same in the normal case. From what I remember it might be different.
If it's different, have it be set by `component.elId('status')`

austingardner: Page notice and Section Notice still look good and have the aria-labelledby

Screen Shot 2021-05-28 at 3 38 23 PM
Screen Shot 2021-05-28 at 3 38 50 PM

agliga (approved):

Issue #1402, opened 2021-05-28
Dialog: some marko attributes being rendered in HTML output


ianmcburnie:
I see the following attrs that shouldn't be there:

*  variant
* reject-text
* confirm-text

We should check alert dialog too.

https://ebay.github.io/ebayui-core/?path=/story/ebay-confirm-dialog--default

![Screen Shot 2021-05-28 at 12 00 29 AM](https://user-images.githubusercontent.com/38065/120012832-de441180-bf94-11eb-904f-157d286ea8d3.png)

ianmcburnie:
Fixed in 7.6.0

PR #1405, opened 2021-06-01
fix(dialog-base): added attributes to ignoreAttributes list

agliga:
## Description
Added some attributes to ignore list for dialog-base.
Initially I had added them in individual components, but they were causing issues with tests.
Also rearranged the input so that they get replaced correctly.

## References
#1402


austingardner (approved):
nice! LGTM

PR #1406, opened 2021-06-02
fix(video): added focus when clicking report, and text input

agliga:
## Description
* Changed report button to focus on overflow menu after cliking
* Added text override for report button


ianmcburnie (approved):
Kb focus is working great!

Issue #1413, opened 2021-06-18
ebay-textbox: missing size attribute


ianmcburnie:
Seems like the size attribute is missing in order to add Skin large modifier 

https://ebay.github.io/skin/#textbox-large

ianmcburnie:
Fixed in 12.x

PR #1415, opened 2021-06-21
fix(textbox): added input-size attribute to control textbox size

agliga:
## Description
Added `input-size` attribute. 
`size` is a reserved textbox attribute, used for the length of the input https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/size
So I called it `input-size` instead.

## References
https://github.com/eBay/ebayui-core/issues/1413

## Screenshots
Screen Shot 2021-06-21 at 10 12 18 AM


BhavinPatel04:
Will this attribute affect the postfix-icon height as well? Because right now, if I manually add large classes, the postfix icon's height is still small and so it looks weird

image

ianmcburnie (approved):

PR #1424, opened 2021-06-23
fix(button): removed span from icon-button

agliga:
## Description
We had a span wrapper around the icon, which should not be there. Removed it since it serves no purpose.
Updated documentation for button

## References
https://github.com/eBay/skin/issues/1480

## Screenshots
Screen Shot 2021-06-23 at 10 37 38 AM

### DIFF
--- a/src/components/ebay-button/README.md
+++ b/src/components/ebay-button/README.md
@@ -13,6 +13,12 @@
 <ebay-button>text</ebay-button>
 ```
 
+## ebay-button (Icon variant) Usage
+
+```marko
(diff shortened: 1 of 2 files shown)


austingardner (approved):

ianmcburnie:
We might have to leave this for 8.0.0 as technically it could cause a break if the app code is depending on this element (shouldn't be.. but we never know).

ianmcburnie:
Actually let's go with this for 12.x. I think we'll be fine.

PR #1430, opened 2021-07-01
Added programatic volume, added timer, some cc support, and ds4 icon

agliga:
## Description
* Added programatic volume controls (muted and volume)
* Added timer to video
* Added cc. There needs more work on the icon
* Added ds4 report icon
* Misc fixes like adding guards and removing unused code


src/components/ebay-video/README.md
```diff
@@ -15,18 +15,21 @@ For resizing, `ebay-video` supports fixed width or variable width. If no width i
...
+| `css-url`          | String  | No       | No       | The full url to point to where to download shaka css. This is only used to override the default CDN path.                   |
+| `cdn-version`      | String  | No       | No       | If using the default CDN path, you can override what version of shaka to use.                                               |
+| `a11y-load-text`   | String  | No       | Yes      | The accessibility text for the loading spinner. Default is "Loading"                                                        |
+| `a11y-play-text`   | String  | No       | Yes      | The accessibility text for the play button. Default is "Click to play"                                                      |
+| `error-text`       | String  | No       | Yes      | The content for error when an either the library or video cannot load. Default is "An error has occurred"                   |
+| `a11y-report-menu` | String  | No       | Yes      | The text for screen readers for the report menu popup. Default is "Report"                                                  |
```
ianmcburnie, line 31: For consistency, we should have 'text' as part of the name.

src/components/ebay-video/README.md
```diff
@@ -52,8 +55,9 @@ If multiple sources are supplied, it will go in the order they are supplied and
...
-| Event        | Data                        | Description           |
-| ------------ | --------------------------- | --------------------- |
-| `load-error` | `{ originalEvent }`         | error occurred        |
-| `play`       | `{ originalEvent, player }` | play start            |
-| `report`     | `{ }`                       | report button clicked |
+| Event          | Data                               | Description                                                                          |
+| -------------- | ---------------------------------- | ------------------------------------------------------------------------------------ |
+| `load-error`   | `{ originalEvent }`                | error occurred                                                                       |
+| `play`         | `{ originalEvent, player }`        | play start                                                                           |
+| `volumechange` | `{ originalEvent, volume, muted }` | volume is changed. Will return the values of the volume decimal and muted true/false |
```
ianmcburnie, line 62: Name should be separated with a dash?

agliga: This is one that I'm not sure about. 
The html5 event is without a dash and non camel case. (Which is what this is and how i kept it)
Which is why i kept it as is. I can rename it to be with a dash though which is more consistent with our codebase

ianmcburnie: Yeah, let's keep it consistent with our side.

ianmcburnie:
In Chrome I see what appears to be the default play button/icon appear for a split second underneath our custom icon. Is there any way to suppress this?

agliga:
> In Chrome I see what appears to be the default play button/icon appear for a split second underneath our custom icon. Is there any way to suppress this?

Unfortunately the only way to override these icons is using dom selectors. They do not have options to do so otherwise
One way we can try to do it is to override the styles to show the material icons.

ianmcburnie:
I think we'll have to figure something out, as it kinda looks bad as it currently is.

agliga:
> I think we'll have to figure something out, as it kinda looks bad as it currently is.

Done, it was my fault. This was the old play icon that we were using. Replaced it with the new one. 
Will need these css changes to make it look good and not jump https://github.com/eBay/skin/pull/1503

src/components/ebay-video/marko-tag.json
```diff
@@ -12,8 +12,11 @@
   "@cdn-version": "string",
   "@a11y-play-text": "string",
   "@a11y-load-text": "string",
+  "@a11y-report-menu": "string",
```
ianmcburnie, line 15: What's the difference between `a11y-report-menu` and `report-text`?

agliga: I missed changing this, good catch!
Changed it to `a11y-report-text`

ianmcburnie (approved):

PR #1465, opened 2021-07-17
feat(video): added option to remove volume slider

### LINKED ISSUE #1463: video: add option to configure volume controls
Let's enable the ability to show a simple mute on/off button only, instead of slider.

This will solve/mitigate: #1435 & #1436

austingardner:
## Description
adds option to `<ebay-video/>` to remove the volume slider. 

## References
closes #1463 

## Screenshots
Screen Shot 2021-07-16 at 6 34 45 PM

Screen Shot 2021-07-16 at 6 35 10 PM

Screen Shot 2021-07-16 at 6 35 19 PM

### DIFF
--- a/src/components/ebay-video/README.md
+++ b/src/components/ebay-video/README.md
@@ -30,6 +30,7 @@ For resizing, `ebay-video` supports fixed width or variable width. If no width i
 | `error-text`       | String  | No       | Yes      | The content for error when an either the library or video cannot load. Default is "An error has occurred"                   |
 | `a11y-report-text` | String  | No       | Yes      | The text for screen readers for the report menu popup. Default is "Report"                                                  |
 | `report-text`      | String  | No       | Yes      | The text for report button. Default is "Report"                                                                             |
+| `volume-slider`    | Boolean | Yes      | No       | True/False to keep or remove volume slider. Default is true                                                                 |
 
 ## @source Tag (required)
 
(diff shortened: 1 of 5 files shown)


src/components/ebay-video/README.md
```diff
@@ -30,6 +30,7 @@ For resizing, `ebay-video` supports fixed width or variable width. If no width i
 | `error-text`       | String  | No       | Yes      | The content for error when an either the library or video cannot load. Default is "An error has occurred"                   |
 | `a11y-report-text` | String  | No       | Yes      | The text for screen readers for the report menu popup. Default is "Report"                                                  |
 | `report-text`      | String  | No       | Yes      | The text for report button. Default is "Report"                                                                             |
+| `volumeSlider`     | Boolean | Yes      | No       | True/False to keep or remove volume slider. Default is true                                                                 |
```
ianmcburnie, line 33: Typo. Should be hyphenated instead of camelCase, I believe.

src/components/ebay-video/component.js
```diff
@@ -115,10 +121,14 @@ module.exports = {
             this.state.action = input.action;
             this.takeAction();
         }
+        if (input.volumeSlider === false) {
```
ianmcburnie, line 124: Can this be reduced to just `this.state.volumeSlider = input.volumeSlider`?

austingardner: We could, but then we still have to check if input.volumeSlider is undefined or if setting it to undefined puts the component in a weird state.... I think that would work though

ianmcburnie: Ok, just leave as is for now then, to be safe.

ianmcburnie (approved):

PR #1471, opened 2021-07-20
fix(tooltip): tooltip closes on escape keydown

### LINKED ISSUE #1444: Tooltip: Add ESC key functionality to dismiss
Tooltips should be dismissible with ESC key while focus is on interactive host.

austingardner:
## Description
This PR adds a keydown listener to the tooltip, and then closes the tooltip when the `Escape` key is pressed. 

## References
closes #1444 

## Screenshots
tooltip-escape


src/components/components/ebay-tooltip-base/index.marko
```diff
@@ -7,12 +7,16 @@ static function toJSON() {
...
 
+static const isBrowser = typeof window !== "undefined";
 $ input.toJSON = toJSON;
 
 <span
     key="container"
     onExpander-expand("handleExpand")
     onExpander-collapse("handleCollapse")
     overlay-style=input.overlayStyle>
+    <!-- <if(isBrowser)>
```
ianmcburnie, line 18: Are the changes in this file still needed?

austingardner: nope I will drop those...

src/components/components/ebay-tooltip-base/component-browser.js
```diff
@@ -10,10 +10,29 @@ module.exports = {
...
+            }
+        }
+    },
+
     onMount() {
         if (this.input.type !== 'dialog--mini') {
             this._setupMakeup();
         }
+
+        this.subscribeTo(window).on('keydown', (e) => {
```
ianmcburnie, line 33: You can/should probably use this instead: https://github.com/eBay/ebayui-core/blob/master/src/common/event-utils/index.js#L32

Example usage here: https://github.com/eBay/ebayui-core/blob/master/src/components/components/ebay-dialog-base/component.js#L37

ianmcburnie: I think the event delegate should be `this.rootEl` instead of `window`.

austingardner: So I tried that and it breaks the tooltip... the tooltip doesn't show up at all.

agliga:
Should be ready to go, tooltip dismisses when button is focused and escape key is pressed.

austingardner:
So I don't quite understand why it is better to only close when the tooltip is focused. Isn't it better that the tooltip closes on escape when the user is hovering, as well?

agliga:
> So I don't quite understand why it is better to only close when the tooltip is focused. Isn't it better that the tooltip closes on escape when the user is hovering, as well?

So escape should only be used when the user is using keyboard correct? If they're using mouse, then I don't think escape is needed because they can just mouse out. 
For keyboard use, the only time the tooltip will be shown is on focus.

austingardner:
> > So I don't quite understand why it is better to only close when the tooltip is focused. Isn't it better that the tooltip closes on escape when the user is hovering, as well?
> 
> So escape should only be used when the user is using keyboard correct? If they're using mouse, then I don't think escape is needed because they can just mouse out.
> For keyboard use, the only time the tooltip will be shown is on focus.

Okay that makes sense! For some reason hitting escape just seemed more satisfying and instant than the mouse out. But I agree it isn't needed except on focus.

ianmcburnie:
So, technically you are both right. When I originally filed the issue, the desire was to only close it on ESC when the host had keyboard focus (what Andrew is saying). But given some recent changes to WCAG I can now foresee that we might have to have it close on ESC even if user initiated with mouse hover (what Austin is saying). Please bear with me while I confirm.

agliga:
> So, technically you are both right. When I originally filed the issue, the desire was to only close it on ESC when the host had keyboard focus (what Andrew is saying). But given some recent changes to WCAG I can now foresee that we might have to have it close on ESC even if user initiated with mouse hover (what Austin is saying). Please bear with me while I confirm.

That's fine.
However if we need to have escape on `mouse hover` we need to attach an event to the `window` because we do not know what element is "focused". Another element could be focused while hovering over the tooltip element.

ianmcburnie:
So, as mentioned, this is one of the more recent additions to WCAG: https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html

This one always feels a little odd to me, but basically we need to be able to dismiss the content that appears on mouse hover *without* moving the mouse again. So yep, let's go with the window based event handler that Austin had initially.

austingardner:
> So, as mentioned, this is one of the more recent additions to WCAG: https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html
> 
> This one always feels a little odd to me, but basically we need to be able to dismiss the content that appears on mouse hover _without_ moving the mouse again. So yep, let's go with the window based event handler that Austin had initially.

So why does WCAG concern themselves with content that appears with a mouse? I thought visually impaired users would use the keyboard? Isn't that out of scope for them?

austingardner:
LGTM!

ianmcburnie (approved):

PR #1472, opened 2021-07-21
feat(video): place report text and icon on main controls bar

austingardner:
## Description
This PR puts the report text and icon on the main controls bar. 

This works for all cases that view item needs, except that the developer must now detect mobile devices and put `volume-slider=false` on `ebay-video` for mobile, because with the volume slider, a mobile device viewing a video in vertical mode won't be able to see all of the controls. I have talked with view item and this works for them. 

## Context
The VI team wants the video on the main controls bar. 

## Screenshots
Screen Shot 2021-07-20 at 3 26 11 PM
Screen Shot 2021-07-20 at 3 27 34 PM
Screen Shot 2021-07-20 at 3 28 06 PM


ianmcburnie:
How about inverting the logic and having `volume-slider=false` be the default? That way we don't have a broken layout in mobile. If desktop app overlooks this setting they don't get a slider, but the layout wont be broken.

ianmcburnie (approved):

Issue #1431, opened 2021-07-02
Carousel: broken 2-finger scrolling when saving and setting index


benc3467:
# Bug Report

## Description

This issue occurs when saving and setting the carousel index:

```
handleScroll({ index }) {
      this.state.index = index;
}
<ebay-carousel index=state.index on-scroll('handleScroll')>...</ebay-carousel>
```

This appears to occur when using the carousel-update event as well. 

When using 2-finger scrolling on mac, the carousel slide appears to "shake" slightly before resting into position. In addition, scrolling quickly sometimes results in the carousel rapidly shifting (see end of demonstration video).

## Screenshots
https://user-images.githubusercontent.com/43257212/124317513-4910db80-db45-11eb-81e1-8f12d4389cfa.mp4

ianmcburnie:
Fixed in master. Will go out in 7.6.12

PR #1482, opened 2021-08-04
fix(carousel): smooth scrolling with state change

benc3467:
## Description

This PR fixes the broken carousel scroll that occurs when manually setting the carousel index.

The skipScrolling variable was used to avoid calling scrollTo when the user used horizontal scrolling (it should only be called when the user clicks the carousel paddles). This worked for most cases, but when the state is changed, a rerender occurs and automatically calls the scrollTo again, skipping the check to make sure the user isn't already performing the scrolling. Here I made sure that the skipScrolling variable is persistent across renders and resets in the appropriate place in order avoid the scrollTo call after the rerender.

## Context

The issue occurred when saving and setting the carousel index:

```
handleScroll({ index }) {
      this.state.index = index;
}
<ebay-carousel index=state.index on-scroll('handleScroll')>...</ebay-carousel>
```
It occurred when using the carousel-update event as well.

When using 2-finger (or any horizontal) scrolling, the carousel slide appeared to "shake" slightly before resting into position. In addition, scrolling quickly sometimes resulted in the carousel slide rapidly shifting.

## References

Original bug report: https://github.com/eBay/ebayui-core/issues/1431

## Screenshots

Previous scroll behavior when setting the index manually:
https://user-images.githubusercontent.com/43257212/128260739-59e82ef0-006b-4c65-9027-b72262f1e6c3.mov
scroll_broken

Behavior with these changes implemented:
https://user-images.githubusercontent.com/43257212/128260796-84535c27-423a-44cc-ac1e-c6ba1e58bbff.mov
scroll_fixed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11825,9 +11825,9 @@
       "dev": true
     },
     "chromedriver": {
-      "version": "90.0.1",
-      "resolved": "https://registry.npmjs.org/chromedriver/-/chromedriver-90.0.1.tgz",
-      "integrity": "sha512-jvyhin0I/Bacxfet7eg29B1j+5mKR35XwWGbgcCOUALeE3mqcCKJY8xUW9cVrqVqTK9/iUOq8/kar7qrTVshPA==",
+      "version": "92.0.1",
+      "resolved": "https://registry.npmjs.org/chromedriver/-/chromedriver-92.0.1.tgz",
+      "integrity": "sha512-LptlDVCs1GgyFNVbRoHzzy948JDVzTgGiVPXjNj385qXKQP3hjAVBIgyvb/Hco0xSEW8fjwJfsm1eQRmu6t4pQ==",
       "dev": true,
       "requires": {
         "@testim/chrome-version": "^1.0.7",
@@ -11849,9 +11849,9 @@
           }
         },
         "debug": {
-          "version": "4.3.1",
-          "resolved": "https://registry.npmjs.org/debug/-/debug-4.3.1.tgz",
-          "integrity": "sha512-doEwdvm4PCeK4K3RQN2ZC2BYUBaxwLARCqZmMjtF8a51J2Rb0xpVloFRnCODwqjpwnAoao4pelN8l3RJdv3gRQ==",
+          "version": "4.3.2",
+          "resolved": "https://registry.npmjs.org/debug/-/debug-4.3.2.tgz",
+          "integrity": "sha512-mOp8wKcvj7XxC78zLgw/ZA+6TSgkoE2C/ienthhRD298T7UNwAg9diBpLRxC0mOezLl4B0xV7M0cCO6P/O0Xhw==",
           "dev": true,
           "requires": {
             "ms": "2.1.2"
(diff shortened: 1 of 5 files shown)


ianmcburnie:
Thank you @benc3467! :-)

austingardner:
Hey @benc3467 could you replace the `.mov` files with `.gif`? Github allows `.gif` files to play inline, so that reviewers of this PR don't have to download them.

benc3467:
Yep, will do. I did record some gifs, but I thought the low frame rate made it hard to see the shaking behavior. I'll just attach them in addition to the .mov files

austingardner:
@benc3467 By running `npx marko-test src/components/ebay-carousel/test/test.browser.js` you can see the relevant tests for the carousel. Looking at the build output I think there are a couple that are breaking. See if you can fix them - in this case the tests probably just need a couple of minor changes; I can see that your fix works, so in this case we just need to refactor the tests to accommodate your changes.

benc3467:
@austingardner Yep, I see the issue. It is an actual problem caused by my changes, so I don't think updating the tests is the right call. I'll try to come up with a fix.

DylanPiercey:
@benc3467 my understanding of the issue is that when you ultimately tell the carousel element to scroll to the same position twice that it does this kind of jumping. Is that accurate?

If so then wouldn't it make more sense to keep track of the current target position for the carousel, and only trigger the scrolling if the target position has changed? Rather than tracking if the interaction was performed via the prev/next buttons?

benc3467:
@DylanPiercey Yes, that does seem to make more sense. That would also likely fix the issue that was causing the test cases to fail, which was that it would not automatically scroll to the correct position when manually changing the index.

benc3467:
@austingardner @DylanPiercey I looked more into it at the end of last week and saw that there was already a system built in to avoid the double scroll behavior, but this didn't work with a state change and subsequent rerender. I simplified my changes to take advantage of this, so could you take a look when you have time?

DylanPiercey:
I'm not sure the changes you've made have the desired impact. Also it looks like things are now using a mix of `this.skipScrolling` and `config.skipScrolling` which are two different things.

I think the best solution would be to just have the scroll transition utility (https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-carousel/utils/scroll-transition/index.js#L14) just keep track of what the current target scroll transition was and clear it once it has reached that point.

Then if a new scroll transition comes in, and it is ultimately toward the same target, we'll just ignore it.

---

Not related to the issue, but it might simplify a fix: Since we've dropped ie11 support we can actually drop a huge portion of the code from this scroll transition utility. Scroll behaviour and the `scrollTo` API is supported well enough in the browsers we currently target. This means we could delete all of this code: 

https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-carousel/utils/scroll-transition/index.js#L2-L4
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-carousel/utils/scroll-transition/index.js#L15
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-carousel/utils/scroll-transition/index.js#L18
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-carousel/utils/scroll-transition/index.js#L20-L75
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-carousel/utils/scroll-transition/index.js#L78-L87

---

An approximate implementation of what I've described might look like this:

```js

const onScrollEnd = require('../on-scroll-end');
const targetPositions = new WeakMap();

/**
 * Utility to animate scroll position of an element using an `ease-out` curve over 250ms.
 * Cancels the animation if the user touches back down.
 *
 * @param {HTMLElement} el The element to scroll.
 * @param {number} to The offset to animate to.
 * @param {function} fn A function that will be called after the transition completes.
 * @return {function} A function that cancels the transition.
 */
module.exports = function scrollTransition(el, pos, fn) {
  if (targetPositions.get(el) !== pos) {
    el.scrollTo({ left: to });
    targetPositions.set(el, to);
  }

  return onScrollEnd(el, () => {
    targetPositions.set(el, undefined);
    fn();
  });
}
```

This could be optimized further if `onScrollEnd` was converted to be a promises that we could store. The code I have above would needlessly register multiple scroll end handlers, but it's not a big deal.

DylanPiercey (approved):

agliga (approved):

PR #1483, opened 2021-08-05
feat(storybook): added controls to components

austingardner:
## Description
This PR adds controls to the majority of the components. There were a few that Andrew had added already. I added a few functions that Dylan helped me with in order to render source code better for many components, as well as rendering the render body in the sub tags. 

There are a few minor issues (especially the menu button is displaying two chevron icons) which I'll resolve in upcoming PRs, but this is at the point where we need to review and merge it. 

## Context
I tried to add controls to as many of the fields in each component where it made sense to do so. In most cases the sub tags are editable. We can add in more examples to the components if we need.


agliga:
I think there's some other changes we can make, but I think this to me is good to merge as is. We can always change stuff in upcoming PRs (like how we handle render bodys and such)

agliga (approved):

ianmcburnie:
Thanks, Andrew. Good to have you back! And thanks Austin for a huge amount of work here. Let's merge it!

PR #1492, opened 2021-08-17
fix(menu-button): passed noToggleIcon down to expand button

### LINKED ISSUE #1491: Menu-Button: dropdown icon shown twice
Screen Shot 2021-08-17 at 2 11 33 PM

austingardner:
## Description
noToggleIcon property now gets passed down to the expandButton, as Andrew mentioned in our conversation earlier.

## References
closes #1491 

## Screenshots (apply to menu-button as well as fake-menu-button)
Screen Shot 2021-08-17 at 4 34 18 PM

### noToggleIcon = true
Screen Shot 2021-08-17 at 4 34 33 PM

### DIFF
--- a/src/components/ebay-expand-button/index.marko
+++ b/src/components/ebay-expand-button/index.marko
@@ -11,6 +11,7 @@ static var ignoredAttributes = [
     "truncate",
     "fixedHeight",
     "partiallyDisabled",
+    "noToggleIcon",
     "toJSON"
 ];
 
@@ -38,6 +39,6 @@ $ var isIconOnly = input.iconOnly || !input.renderBody
                 <${input.renderBody}/>
             </span>
         </if>
-        <ebay-dropdown-icon/>
+        <if(!input.noToggleIcon)><ebay-dropdown-icon/></if>
     </span>
 </button>
--- a/src/components/ebay-menu-button/index.marko
+++ b/src/components/ebay-menu-button/index.marko
@@ -52,6 +52,7 @@ $ debugger;
         aria-label=input.a11yText
         aria-labelledby=(labelId && `${input.prefixId} ${labelId}`)
         disabled=input.disabled
+        no-toggle-icon=input.noToggleIcon
         on-escape("handleButtonEscape")
         key="button">
         <if(isOverflowVariant)>
@@ -82,9 +83,6 @@ $ debugger;
                         <span id=labelId>${input.text}</span>
                     </if>
                 </else>
-                <if(!input.noToggleIcon)>
-                    <ebay-dropdown-icon/>
-                </if>
             </span>
         </else>
     </>
--- a/src/components/ebay-expand-button/marko-tag.json
+++ b/src/components/ebay-expand-button/marko-tag.json
@@ -15,5 +15,6 @@
   "@autofocus": "#html-autofocus",
   "@disabled": "#html-disabled",
   "@type": "never",
-  "@aria-disabled": "never"
+  "@aria-disabled": "never",
+  "@no-toggle-icon": "boolean"
 }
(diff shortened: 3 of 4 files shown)


agliga (commented):
I think there's a better way to achieve this.

There is no reason to add a noToggleIcon on expand button because by removing that it becomes a button.
What I would do is this:

On line 41 we choose what element to render. We should add a third one which is just Button if it's noToggleIcon
So something like this:
$ {
  let buttonElmenet = Expandbutton;
  if (input.isOverflowVariant) {
     buttonElement = IconButton;
  } else if (input.noToggleIcon) {
    buttonElement = Button;
 }
}
Something like this.

austingardner:
> I think there's a better way to achieve this.
> 
> There is no reason to add a noToggleIcon on expand button because by removing that it becomes a button.
> What I would do is this:
> 
> On line 41 we choose what element to render. We should add a third one which is just Button if it's noToggleIcon
> So something like this:
> 
> ```
> $ {
>   let buttonElmenet = Expandbutton;
>   if (input.isOverflowVariant) {
>      buttonElement = IconButton;
>   } else if (input.noToggleIcon) {
>     buttonElement = Button;
>  }
> }
> ```
> 
> And then on line 41 do
> `<${buttonElement}`
> Something like this.

I think we need to think about things more if we're going to put a normal button there... that allows some strange scenarios such as 

Screen Shot 2021-08-18 at 3 14 51 PM

agliga:
> > I think there's a better way to achieve this.
> > There is no reason to add a noToggleIcon on expand button because by removing that it becomes a button.
> > What I would do is this:
> > On line 41 we choose what element to render. We should add a third one which is just Button if it's noToggleIcon
> > So something like this:
> > ```
> > $ {
> >   let buttonElmenet = Expandbutton;
> >   if (input.isOverflowVariant) {
> >      buttonElement = IconButton;
> >   } else if (input.noToggleIcon) {
> >     buttonElement = Button;
> >  }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And then on line 41 do
> > `<${buttonElement}`
> > Something like this.
> 
> I think we need to think about things more if we're going to put a normal button there... that allows some strange scenarios such as
> 
> Screen Shot 2021-08-18 at 3 14 51 PM

Okay, I see the issue. 
Lets keep it as is then. Let me take another look.

src/components/ebay-menu-button/index.marko
```diff
@@ -52,6 +52,7 @@ $ debugger;
         aria-label=input.a11yText
         aria-labelledby=(labelId && `${input.prefixId} ${labelId}`)
         disabled=input.disabled
+        noToggleIcon=input.noToggleIcon
```
agliga, line 55: Maybe call this `no-icon` or `text-only`? (We have icon-only).
The main change I want is to have it support dashes. (You need to update the marko.json too)

src/components/ebay-expand-button/index.marko
```diff
@@ -11,7 +11,8 @@ static var ignoredAttributes = [
     "truncate",
     "fixedHeight",
     "partiallyDisabled",
-    "toJSON"
+    "toJSON",
```
agliga, line 14: Can you swap this, have `toJSON` be the last one since that's an internal variable.

agliga (commented):
Couple of minor nits, but overall I'm fine with it

src/components/ebay-menu-button/index.marko
austingardner: okay if we change the name do we need to add it to the migrations? I think `noToggleIcon` already existed on the menu-button

agliga: Okay, leave it then. I think it's fine. 
Just make dashes supported.

austingardner: `  "@no-toggle-icon": "boolean",` is what is in marko-tag.json. Is that adequate?

agliga: Yep!

agliga (approved):

PR #1501, opened 2021-08-25
1477 breadcrumbs current item tab order

### LINKED ISSUE #1477: breadcrumbs : "current" item should not be in tab order
https://github.com/eBay/ebayui-core/blob/4f33bf56babecc4db9c0483175095a5a3f485afb/src/components/ebay-breadcrumbs/index.marko#L26

https://ebay.github.io/skin/#breadcrumbs
logic results in last element button that is tab-able
`            <a aria-current="location">Smart Watch Bands</a> vs    <button aria-current="location">Smart Watch Bands</button>`

Screen Shot 2021-07-26 at 6 07 38 PM

austingardner:
## Description
The last breadcrumb can no longer be tabbed if it doesn't have an href. Just had to get rid of the button - Ian said we should be consistent and only have buttons or anchor tags in the breadcrumbs, and not mix them.

## Context
had to delete some tests to make this work

## References
closes #1477 

## anchor with no href on last item
breadcrumbs-tab

## buttons

breadcrumbs-tab-buttons

### DIFF
--- a/src/components/ebay-breadcrumbs/index.marko
+++ b/src/components/ebay-breadcrumbs/index.marko
@@ -10,6 +10,7 @@ static var ignoredAttributes = [
 ];
 
 $ input.toJSON = noop;
+$ var anyHref = (input.items || []).some(element => element.href != null);
 <nav
     ...processHtmlAttributes(input, ignoredAttributes)
     aria-labelledby:scoped="breadcrumbs-heading"
@@ -23,7 +24,7 @@ $ input.toJSON = noop;
             <li>
                 $ var isLast = i === input.items.length - 1;
                 $ var current = !item.href && isLast;
-                <${item.href ? "a" : "button"}
+                <${anyHref ? "a" : "button"}
                     ...processHtmlAttributes(item)
                     aria-current=(current && "location")
                     onClick(!current && "handleClick")>
--- a/src/components/ebay-breadcrumbs/test/test.server.js
+++ b/src/components/ebay-breadcrumbs/test/test.server.js
@@ -16,37 +16,25 @@ describe('breadcrumbs', () => {
 
         expect(getByLabelText(input.a11yHeadingText)).has.attr('role', 'navigation');
 
-        input.items
-            .slice(0, -1)
-            .forEach((item) =>
-                expect(getByText(item.renderBody.text)).has.property('tagName', 'A')
-            );
-
-        expect(getByText(input.items[input.items.length - 1].renderBody.text)).has.property(
-            'tagName',
-            'BUTTON'
-        );
-    });
-
-    it('renders buttons when hrefs are missing', async () => {
-        const input = mock.Buttons;
-        const { getByText } = await render(template, input);
         input.items.forEach((item) =>
-            expect(getByText(item.renderBody.text)).has.property('tagName', 'BUTTON')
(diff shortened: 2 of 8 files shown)


src/components/ebay-breadcrumbs/index.marko
```diff
@@ -10,6 +10,7 @@ static var ignoredAttributes = [
 ];
 
 $ input.toJSON = noop;
+$ let anyHref = input && input.items && input.items.some(element => element.href != null);
```
agliga, line 13: Use `var` here. 
There's an issue with `let/const` in older marko versions in the template.

src/components/ebay-breadcrumbs/index.marko
```diff
@@ -23,7 +24,7 @@ $ input.toJSON = noop;
             <li>
                 $ var isLast = i === input.items.length - 1;
                 $ var current = !item.href && isLast;
-                <${item.href ? "a" : "button"}
+                <${anyHref && "a" || "button"}
```
agliga, line 27: Why not use ternary?
`${anyHref ? "a" : "button"}`

src/components/ebay-breadcrumbs/test/test.server.js
```diff
@@ -16,37 +16,23 @@ describe('breadcrumbs', () => {
...
 
-        input.items.slice(0, -1).forEach((item) => {
+        input.items.forEach((item) => {
             const itemEl = getByText(item.renderBody.text);
             if (item.href) {
                 expect(itemEl).has.property('tagName', 'A');
                 expect(itemEl).has.attr('href', item.href);
             } else {
-                expect(itemEl).has.property('tagName', 'BUTTON');
+                expect(itemEl).has.property('tagName', 'A');
```
agliga, line 34: You have duplicate expects. Combine both and move them out of the the conditionals.

src/components/ebay-breadcrumbs/test/test.server.js
```diff
@@ -16,37 +16,23 @@ describe('breadcrumbs', () => {
...
+                expect(itemEl).attr('aria-current', 'location');
```
agliga, line 35: Actually this test is somewhat invalid. I would expect that `aria-current` to be ONLY on the last one, and not on any items without a `href`

agliga (commented):
Left a couple of code comments.

@ianmcburnie Isn't there an issue where there will be an anchor tag rendered without a `href`? In case the users don't pass a `href` on any.

ianmcburnie:
It's fine (i.e. valid HTML) if anchor tag doesn't receive an href. Really, it's up to the app developer to ensure this is set (just like with an actual anchor tag).

austingardner:
@agliga addressed your comments 👍

src/components/ebay-breadcrumbs/index.marko
```diff
@@ -10,6 +10,7 @@ static var ignoredAttributes = [
 ];
 
 $ input.toJSON = noop;
+$ var anyHref = input && input.items && input.items.some(element => element.href != null);
```
agliga, line 13: Sorry, just noticed this. No need to check for `input`.
You can do something like: 
`(input.items || []).some(...etc)`

agliga (commented):
Just one last one, then it should be good to go. Sorry I didn't notice it the first time.

agliga (approved):

Issue #1337, opened 2021-02-24
ebayui 8.0.0 migrate breaking changes


agliga:
# Feature Request
I'm adding this in order to document any possibly breaking changes between 7 and 8. 
The reason  add these as migrations in ebayui 7. This way users can upgrade to the lastest 7, fix their migrations, and then upgrade to 8 without any issues.

austingardner:
Should we migrate any "type" attributes to "variant"?

ianmcburnie:
Tracking relevant PRs here:

https://github.com/eBay/ebayui-core/pull/1429

PR #1502, opened 2021-08-25
fix(listbox): added listboxSelection option with auto/manual

agliga:
## Description
Changed listbox to support `listboxSelection` either `auto` or `manual`

## Context
* For auto will use `activeDescendantChange` as it was before
* For manual will hook into `keydown` and `click`. 
* Added checks to not listen to the events when not in use.  

## References
https://github.com/eBay/ebayui-core/issues/1337

### DIFF
--- a/src/components/ebay-listbox-button/README.md
+++ b/src/components/ebay-listbox-button/README.md
@@ -35,16 +35,17 @@ When a selected option is specified:
 
 ### ebay-listbox-button Attributes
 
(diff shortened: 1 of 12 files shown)


agliga:
> Tested it locally. Very cool!
> 
> 
> 
> If combobox is using listbox under the hood (not sure it is or not) then we need to make sure combobox is using automatic list selection.

No its not
Combobox is also using the activeDescendant but the pattern used is the same as in this pr.

Issue #1503, opened 2021-08-26
ebay-infotip: Add support for "open" attribute


okannivadikubendran:
# Feature Request

## Description
`ebay-infotip` component doesn't have support for "open" attribute which can be used by the component consumers to dynamically open/close the infotip. e.g, Closing the infotip when user click anywhere outside the infotip

Similar to `eba-fullscreen-dialog`, `ebay-details` components, can  we have an `open` attribute for `ebay-infotip` component

ianmcburnie:
Released in 7.x

PR #1506, opened 2021-09-01
feat(infotip): added programatic show/hide for infotip

agliga:
## Description
Exposed `expander` methods in infotip. Change infotip base to support those.
Also added two examples (for expander and for modal versions). 
I think in the future we should add passing `open` as a param. The reason I didn't do it now is because it would be a much large change and potential regression

## References
https://github.com/eBay/ebayui-core/issues/1503


austingardner (approved):
LGTM; seems like we're moving more towards the controlled components with this one

Issue #1489, opened 2021-08-17
confirm and alert dialogs: skin classname changes


ianmcburnie:
Confirm-dialog and alert-dialog were promoted to first class modules in Skin. We need to update CoreUI markup accordingly.

Relevant skin ticket: https://github.com/eBay/skin/issues/1453

ianmcburnie:
Done in 8.0.0 branch

PR #1510, opened 2021-09-01
fix: fixed broken tests

agliga:
## Description
* Note, this requires https://github.com/eBay/skin/pull/1548 and https://github.com/eBay/skin/pull/1547
* Added alert dialog and confirm dialog css changes to fix test cases.
* Fixed pagination breakpoints (since the items are smaller now)
* Updated some packages which also helped fix some other test issues

## References
https://github.com/eBay/ebayui-core/issues/1489


src/components/ebay-alert-dialog/index.marko
```diff
@@ -4,24 +4,23 @@ $ var mainId = component.getElId("alert-dialog-main");
...
+    window-class=["alert-dialog__window alert-dialog__window--fade"]>
     <@footer>
         <ebay-button
             priority="primary"
             aria-describedby=mainId
             on-click("emit", "confirm")
             key="confirm"
             id=confirmId
-            class="lightbox-dialog__confirm">
+            class="alert-dialog__confirm">
```
austingardner, line 23: I think the class here was alert-dialog__acknowledge -> Ian changed it in his last PR

agliga: I don't see the class being set in css/less. I only see it shown in the docs. So does it matter since the class isn't really being used?

austingardner: You're right it probably doesn't matter then. If that's the case we should drop it from the skin docs to avoid confusion

agliga: Just updated it in ebayui for consistency

austingardner (approved):
LGTM! nice job resolving all of the issues with the tests. I was able to run them locally - looks very good!

ianmcburnie (approved):
Great work, Andrew!

PR #1512, opened 2021-09-03
feat: added open attribute to tourtip, tooltip, and infotip

### LINKED ISSUE #1508: Infotip/Tourtip/Tooltip: Change expanded/collpased state based on open input

austingardner:
## Description
Added open attribute to tooltip, infotip, and tourtip. This gives the developer greater control over how they are used. 

## References
closes #1508 

## Infotip
infotip-open

## Tourtip
tourtip

## Tooltip
tooltip-open

### DIFF
--- a/src/components/ebay-tooltip/component.js
+++ b/src/components/ebay-tooltip/component.js
@@ -7,11 +7,8 @@ module.exports = {
         };
     },
     onInput(input) {
-        if (input.open === true) {
-            this.getComponent('base').expand();
-        }
-        if (input.open === false) {
-            this.getComponent('base').collapse();
+        if (input.open === true || input.open === false) {
+            this.state.open = input.open;
         }
     },
     handleExpand() {
@@ -24,7 +21,7 @@ module.exports = {
     },
     handleKeydown(e) {
         eventUtils.handleEscapeKeydown(e, () => {
-            this.getComponent('base').collapse();
+            this.handleCollapse();
         });
     },
 };
--- a/src/components/ebay-tooltip/component.js
+++ b/src/components/ebay-tooltip/component.js
@@ -21,7 +21,7 @@ module.exports = {
     },
     handleKeydown(e) {
         eventUtils.handleEscapeKeydown(e, () => {
-            this.handleCollapse();
+            this.state.open = false;
         });
     },
 };
--- a/src/components/ebay-tooltip/component.js
+++ b/src/components/ebay-tooltip/component.js
@@ -1,10 +1,18 @@
 const eventUtils = require('../../common/event-utils');
 
 module.exports = {
-    onInput() {
+    onCreate() {
         this.state = {
             open: false,
         };
+    },
+    onInput(input) {
+        if (input.open === true) {
+            this.getComponent('base').expand();
+        }
+        if (input.open === false) {
+            this.getComponent('base').collapse();
+        }
     },
     handleExpand() {
         this.state.open = true;
--- a/src/components/ebay-tooltip/index.marko
+++ b/src/components/ebay-tooltip/index.marko
@@ -21,6 +21,7 @@ $ var pointer = input.pointer || "bottom";
 
 <span>
     <ebay-tooltip-base
+        open=state.open
         key="base"
         type="tooltip"
         pointer=pointer
--- a/src/components/components/ebay-tooltip-base/component-browser.js
+++ b/src/components/components/ebay-tooltip-base/component-browser.js
@@ -20,17 +20,21 @@ module.exports = {
         if (this.input.type !== 'dialog--mini') {
             this._setupMakeup();
         }
+        if (this.action && this._expander) {
+            if (this.action === 'expand') {
+                this.expand();
+            } else if (this.action === 'collapse') {
+                this.collapse();
+            }
+            this.action = null;
+        }
     },
 
     onInput(input) {
-        if (this._expander) {
-            if (input.open === true) {
-                this.handleExpand();
-                this.expand();
-            } else if (input.open === false) {
-                this.handleCollapse();
-                this.collapse();
-            }
+        if (input.open === true) {
+            this.action = 'expand';
+        } else if (input.open === false) {
+            this.action = 'collapse';
         }
     },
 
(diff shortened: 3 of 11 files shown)


agliga:
Infotip and tourtip work great! Good job

Tooltip has an issue where if open is false then then events do not trigger when hovering on the item. It opens for a second and closed. If open is true then it never opens.

src/components/ebay-tooltip/index.marko
```diff
@@ -21,6 +21,7 @@ $ var pointer = input.pointer || "bottom";
 
 <span>
     <ebay-tooltip-base
+        open=input.open
```
agliga, line 24: Should be `state.open`

agliga (commented):
So the tooltip not working properly is because of the `toJson` in the tooltip-base
Since we are rendering it based on open, the `toJSON` prevents the rerender from happening properly.
We have to add all the variables we need in that, or remove the `toJSON` method in the index.marko.
Try to optimize it first off

src/components/ebay-tooltip/component.js
```diff
@@ -1,11 +1,19 @@
 const eventUtils = require('../../common/event-utils');
 
 module.exports = {
-    onInput() {
+    onCreate() {
         this.state = {
             open: false,
         };
     },
```
agliga, line 8: Why are we even handling the state?
It's all based on input.

agliga (commented):
One comment, everything else looks great!

src/components/ebay-tooltip/component.js
austingardner: I couldn't get it to work without initializing the state in the `onCreate` method. It just seemed natural to add the state there; at the very least, managing the state the way we previously were in the `onInput` method wasn't working out very well.

agliga: So then we should have `input` set the state correctly
And then use `state` instead of input

austingardner: Well here's the thing - we do want the state initially set to false. If there are no inputs, the base component emitting events will change the state of `tooltip`. Then in the case that there are inputs, we just want to call `expand` and `collapse` on the base component, which will then emit events that `tooltip` listens to and then changes its own state.

So right now what we are doing is indeed what we want to happen.

agliga:
```
diff --git a/src/components/ebay-tooltip/component.js b/src/components/ebay-tooltip/component.js
index 1e513e38..ed5bb99c 100644
--- a/src/components/ebay-tooltip/component.js
+++ b/src/components/ebay-tooltip/component.js
@@ -7,12 +7,7 @@ module.exports = {
         };
     },
     onInput(input) {
-        if (input.open === true) {
-            this.getComponent('base').expand();
-        }
-        if (input.open === false) {
-            this.getComponent('base').collapse();
-        }
+        this.state.open = input.open;
     },
     handleExpand() {
         this.state.open = true;
```

```
diff --git a/src/components/components/ebay-tooltip-base/component-browser.js b/src/components/components/ebay-tooltip-base/component-browser.js
index 68b986b4..4b4986a9 100644
--- a/src/components/components/ebay-tooltip-base/component-browser.js
+++ b/src/components/components/ebay-tooltip-base/component-browser.js
@@ -23,15 +23,13 @@ module.exports = {
     },

     onInput(input) {
-        if (input.type !== 'tooltip') {
-            if (this._expander) {
-                if (input.open === true) {
-                    this.expand();
-                    this.handleExpand();
-                } else if (input.open === false) {
-                    this.collapse();
-                    this.handleCollapse();
-                }
+        if (this._expander) {
+            if (input.open === true) {
+                this.expand();
+                this.handleExpand();
+            } else if (input.open === false) {
+                this.collapse();
+                this.handleCollapse();
             }
         }
     },
```
This seems to be working locally.

agliga: I think it is much cleaner and easier to understand that what was before. 
It gets very confusing when there is this `input.type !== "tooltip"`, this component-base should handle all the cases.

I don't see any issues with this, if there is let me know.

austingardner: oh now I see. That makes sense

austingardner: So I tried this and it works mostly... except that the proper events aren't being emitted from the tooltip-base to the tooltip.

agliga: Are they emitting properly on the other ones or just not tooltip?

austingardner: just not tooltip

agliga: Then we should take a look why that is, there's something probably missing in tooltip

austingardner: I think the hard thing is that the tooltip is activated on hover, and the hover events aren't changing the state quite correctly.

agliga: So I pulled in my `8.0.0` event changes to test it. I added `onCollapse` and `onExpand` events to storybook. I see the events are getting fired correctly on tooltip when open toggles and when I hover over. 
I think the only issue I see is it's firing twice? But not a big deal imo.

austingardner: oh okay. What about the escape key press?

agliga: Works just fine. I would actually change that to do `this.state.open = false` so it rerenders and calls into tooltip base.

Make sure you're focused on the `document`. The trigger is on `document`, which in storybook is in an iframe

agliga: To test the `open` attribute, what I did is I triggered it to open. Clicked and held on the storybook iframe. This focuses but does not close the tooltip (releasing the mouse button triggers the close). Then I pressed escape and it closed just fine.

agliga (approved):
Austin, I updated this to fix the tests, can you review it too, especially the part I added

ianmcburnie (commented):
We have a small discrepancy with the name being `open`, but the events being `on-expand/on-collapse` (which ties in with `aria-expanded`). Was there a reason for going with `open` instead of `expanded`?

austingardner (commented):
LGTM! Thanks for the help on that tooltip Andrew

austingardner:
> We have a small discrepancy with the name being `open`, but the events being `on-expand/on-collapse` (which ties in with `aria-expanded`). Was there a reason for going with `open` instead of `expanded`?

We have other components that use `open`... I think it makes sense to keep as is

Issue #1494, opened 2021-08-18
Add events from README.md into Controls; Add all attributes into Controls for all components


austingardner:
Because we will no longer be showing the docs page in storybook, we need to add the events into the controls for all components; additionally, we need to double check that all attributes have been added into the controls. We can disable the attribute in controls if needed.

PR #1514, opened 2021-09-09
feat: addeed storybook actions

agliga:
## Description
Added actions tabs. Added actions for most stories. 

## Referecnes
https://github.com/eBay/ebayui-core/issues/1494

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,13 +1,5 @@
 module.exports = {
     stories: ['../src/**/*.stories.js'],
 
-    addons: [
-        './plugins/theme-switcher/register.jsx',
-        {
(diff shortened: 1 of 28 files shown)


austingardner:
Looks like the confirm dialog is emitting the `on-reject` even when the user clicks on `Okay`.

austingardner:
Also the toast-dialog emits the 'on-close' event twice when the "Close" button is clicked, but only one when the "X" button is clicked. Not a big issue, might not be worth fixing.

austingardner (approved):
Looks great! Nice job here Andrew! I think this will definitely help people understand how our components are working.

PR #1515, opened 2021-09-10
fix(fake-menu): anchor tags display aria-disabled when item is disabled

### LINKED ISSUE #1480: ebay-fake-menu: Disabled link is not clear
# Bug Report

## Description

On [ebay-fake-menu readme](https://github.com/eBay/ebayui-core/tree/v7.6.11/src/components/ebay-fake-menu#item-attributes), you can disable a link if you set `@item` `href` to null.

When I check [ebay-fake-menu storybook](https://ebay.github.io/ebayui-core/?path=/story/ebay-fake-menu--fake-item-disabled), the link does not output `aria-disabled="true"` even if the code seems 'might' have it.

Screenshot 2021-07-30 at 17 12 14

[From the Marko template](https://github.com/eBay/ebayui-core/blob/v7.6.11/src/components/ebay-fake-menu/index.marko#L52-L61), the `aria-disabled="true"` is only rendered if the `@item` prop `disabled` is true.

Screenshot 2021-07-30 at 17 56 02

### Questions

1. What is the expected output for a disabled fake menu item? If an `<a>` without `href` is sufficient, then this issue is invalid and you can close it immediately 🙂
2. From the current implementation, it seems the only way to render `aria-disabled` is to set `@item.disabled` to `true` instead of setting `href` to null. If so, then we only need to update the README.

If you want, I can make a PR once it's clear what the issue is.

## Workaround

None

## Screenshots

See on section Description above.

austingardner:
## Description
The fake menu now shows aria-disabled=true on anchor tags when the corresponding item is disabled. 

## Context
This is part two of the fake-menu fixes; part one was [here](https://github.com/eBay/ebayui-core/pull/1511)

## References
closes #1480 

## Screenshots
fake-menu-aria-disabled

### DIFF
--- a/src/components/ebay-fake-menu/index.marko
+++ b/src/components/ebay-fake-menu/index.marko
@@ -18,7 +18,8 @@ static {
         "checked",
         "current",
         "badgeNumber",
-        "badgeAriaLabel"
+        "badgeAriaLabel",
+        "disabled"
     ];
 }
 
@@ -53,6 +54,8 @@ $ var baseClass = input.classPrefix || "fake-menu";
                         ...processHtmlAttributes(item, itemIgnoredAttributes)
                         class=[`${baseClass}__item`, item.class]
                         style=item.style
+                        disabled=(item.type === "button" && item.disabled)
+                        aria-disabled=(item.type !== "button" && item.disabled && 'true')
                         aria-current=(item.current ? defaultAriaCurrent : false)
                         href=(item.disabled ? null : item.href)
                         onKeydown("handleItemKeydown", index)
--- a/src/components/ebay-fake-menu/README.md
+++ b/src/components/ebay-fake-menu/README.md
@@ -1,25 +1,25 @@
 <h1 style='display: flex; justify-content: space-between; align-items: center;'>
     <span>
-        ebay-menu
+        ebay-fake-menu
     </span>
     <span style='font-weight: normal; font-size: medium; margin-bottom: -15px;'>
         DS v1.1.0
     </span>
 </h1>
 
-## ebay-menu Tag
+## ebay-fake-menu Tag
 
-### ebay-menu Usage
+### ebay-fake-menu Usage
 
 ```marko
-<ebay-menu text="text">
+<ebay-fake-menu text="text">
     <@item>item 1</@item>
     <@item>item 2</@item>
     <@item>item 3</@item>
-</ebay-menu>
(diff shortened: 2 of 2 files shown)


src/components/ebay-fake-menu/index.marko
```diff
@@ -53,6 +54,8 @@ $ var baseClass = input.classPrefix || "fake-menu";
                         ...processHtmlAttributes(item, itemIgnoredAttributes)
                         class=[`${baseClass}__item`, item.class]
                         style=item.style
+                        disabled=(item.type === "button" && item.disabled)
+                        aria-disabled=(item.type !== "button" && item.disabled ? true : null)
```
ianmcburnie, line 58: I think you need quotes around the `true` as it is only rendering `aria-disabled`. It should render `aria-disabled="true"`

austingardner: that's right; the `true` or `'true'` always gets me mixed up

src/components/ebay-fake-menu/index.marko
```diff
@@ -53,6 +54,8 @@ $ var baseClass = input.classPrefix || "fake-menu";
                         ...processHtmlAttributes(item, itemIgnoredAttributes)
                         class=[`${baseClass}__item`, item.class]
                         style=item.style
+                        disabled=(item.type === "button" && item.disabled)
+                        aria-disabled=(item.type !== "button" && item.disabled ? 'true' : null)
```
agliga, line 58: Just for consistency, doesn't this work?:
`aria-disabled=(item.type !== "button" && item.disabled && 'true')`
No readon to default to null I'm assuming

agliga:
One small nit
I think it's fine overall, just need to update documentation to tell them to use `disabled` even for links.

agliga (approved):

ianmcburnie (commented):
I don't see a way to set an item to disabled in the storybook.

Screen Shot 2021-09-16 at 9 48 48 AM

austingardner:
> I don't see a way to set an item to disabled in the storybook.
> 
> Screen Shot 2021-09-16 at 9 48 48 AM

You can add disabled to an item in the json in the renderBody on Storybook

ianmcburnie:
Thanks Austin! I probably need a tour of all the new features :-)

ianmcburnie (approved):

PR #1518, opened 2021-09-15
fea(storybook): documentation additions

agliga:
## Description
* Finished adding all events
* Added tagsToString fixes for non reprated attribute tags
* Added tagsToString for some most components
* Added `text` doc
* Added `spread` attribute to button (just for a sample of how it should be, if it's good then will add to all components.)

## References
https://github.com/eBay/ebayui-core/issues/1494

### DIFF
--- a/.storybook/storybook-code-source/index.js
+++ b/.storybook/storybook-code-source/index.js
@@ -1,27 +1,38 @@
-function tagToString(tagName, args, fromArray = false) {
-    const { body, attrs } = argsToString(args);
+function tagToString(tagName, args, plurarls = {}, indent = '', fromAttribute = false) {
+    const nextIndex = indent + '\t';
+    const { body, attrs } = argsToString(args, plurarls, nextIndex);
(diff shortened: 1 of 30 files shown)


austingardner (approved):
LGTM!

.storybook/storybook-code-source/index.js
```diff
@@ -1,28 +1,39 @@
-function tagToString(tagName, args, fromArray = false) {
-    const { body, attrs } = argsToString(args);
+function tagToString(tagName, args, plurarls = {}, indent = '', fromAttribute = false) {
```
austingardner, line 1: is `plurarls` for args to subcomponents? Maybe that name could be a little clearer

agliga: Yes, it's basically to turn plural argument tags to singular
So `@items` becomes `@item`

ianmcburnie (approved):
Nice! Storybook is MUCH improved from our v7 branch :-)

PR #1521, opened 2021-09-19
feat(video): added closed captions and language to overflow menu

austingardner:
## Description
adds closed captions and language options as well as overflow menu to ebay-video. 

Note: closed captions are only available when the video has them. 

## Screenshots
captions

### DIFF
--- a/src/components/ebay-video/component.js
+++ b/src/components/ebay-video/component.js
@@ -14,7 +14,9 @@ const videoConfig = {
         'mute',
         'report',
         'fullscreen',
+        'overflow_menu',
     ],
+    overflowMenuButtons: ['captions'],
 };
 
 module.exports = {


agliga:
I think we only want captions, not language. Can you verify?

austingardner:
Yeah that had been the plan but I think in this case it makes sense to include language just in case they need it.

agliga:
> Yeah that had been the plan but I think in this case it makes sense to include language just in case they need it.

I think as long as the button is not visible if there's only 1 language, then it's fine to include. 
But since we are not supporting multiple languages yet, having this button might be misleading. (again just check if it gets hidden if there's only 1 language).

austingardner:
> > Yeah that had been the plan but I think in this case it makes sense to include language just in case they need it.
> 
> I think as long as the button is not visible if there's only 1 language, then it's fine to include.
> But since we are not supporting multiple languages yet, having this button might be misleading. (again just check if it gets hidden if there's only 1 language).

Yeah it was still displaying when there was only one language - I dropped the language option

PR #1523, opened 2021-09-22
feat(dialog): changed useHiddenProperty logic and set default for fullscreen

agliga:
## Description
* Set fullscreen dialog to `useHiddenProperty` by default
* Changed dialog-base to not run `keyboard-traps` if `useHiddenProperty` is true.

### DIFF
--- a/src/components/components/ebay-dialog-base/test/test.browser.js
+++ b/src/components/components/ebay-dialog-base/test/test.browser.js
@@ -365,10 +365,10 @@ describe('given a closed dialog with useHiddenProperty', () => {
             await component.rerender(Object.assign({}, input, { open: true }));
         });
 
-        thenItIsOpen(true);
+        useHiddenPropertyIsOpened(true);
     });
 
-    function thenItIsOpen(wasToggled) {
+    function useHiddenPropertyIsOpened(wasToggled) {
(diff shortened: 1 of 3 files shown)


src/components/components/ebay-dialog-base/test/test.browser.js
```diff
@@ -365,10 +365,10 @@ describe('given a closed dialog with useHiddenProperty', () => {
             await component.rerender(Object.assign({}, input, { open: true }));
         });
 
-        thenItIsOpen(true);
+        useHiddenPropertyIsOpened(true);
     });
 
-    function thenItIsOpen(wasToggled) {
+    function useHiddenPropertyIsOpened(wasToggled) {
```
ianmcburnie, line 371: Do we still have the existing test for when hidden prop is not being used?

agliga: Yes, this one is changed since this method was copied from that method initially but not renamed. I just renamed it so it makes more sense.
https://github.com/eBay/ebayui-core/pull/1523/files#diff-75e8adbaaf3f958943e9bb789744ffc500ec020dbcba951ea7dbd946ac457f80R319-R326
Here is the original method above it

ianmcburnie (approved):

PR #1532, opened 2021-10-06
fix(video): added reportText, dropped a11yReportText

### LINKED ISSUE #1533: Report text for video is not passed correctly
# Bug Report
## Description
Video doesn't pass report text to the report button correctly 

## Screenshots
Screen Shot 2021-09-29 at 12 00 42 PM

austingardner:
## Description
Fixed Report Text for video and dropped a11yReportText

## References
closes #1533 

## Screenshots
Screen Shot 2021-10-05 at 10 33 39 PM

### DIFF
--- a/src/components/ebay-video/component.js
+++ b/src/components/ebay-video/component.js
@@ -224,7 +224,7 @@ module.exports = {
         );
 
         // eslint-disable-next-line no-undef,new-cap
-        shaka.ui.Controls.registerElement('report', new Report.Factory());
+        shaka.ui.Controls.registerElement('report', new Report.Factory(this.input.reportText));
 
         // eslint-disable-next-line no-undef,new-cap
         shaka.ui.Controls.registerElement('captions', new TextSelection.Factory());
--- a/src/components/ebay-video/elements.js
+++ b/src/components/ebay-video/elements.js
@@ -26,8 +26,12 @@ function getElements(self) {
         }
     };
     Report.Factory = class {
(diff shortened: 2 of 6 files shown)


src/components/ebay-video/component.js
```diff
@@ -114,6 +114,10 @@ module.exports = {
             this.video.muted = input.muted;
         }
 
+        if (input.reportText) {
```
agliga, line 117: Why set it in state? 
Just call it directly from input

agliga (changes requested):
Just change the state to input
Also makes sure you change marko.json too. 
Otherwise it looks good

src/components/ebay-video/component.js
```diff
@@ -224,7 +224,10 @@ module.exports = {
         );
 
         // eslint-disable-next-line no-undef,new-cap
-        shaka.ui.Controls.registerElement('report', new Report.Factory());
+        shaka.ui.Controls.registerElement(
+            'report',
+            new Report.Factory(this.input.reportText || null)
```
agliga, line 229: One more nit, sorry. I don't think there's a need to default to null?

You have the same thing here. I would just do `this.input.reportText` here.
https://github.com/eBay/ebayui-core/pull/1532/files#diff-ec1aeb41a76f173410fc83de66c8305a2e27f80a2053cb23a4b638495b445783R34

And same with it on the next one. If it's undefined or blank it should be fine I would assume right?

agliga (approved):

PR #1552, opened 2021-11-04
feat(video): added ability to specify mimeType

austingardner:
## Description
Added mimeType option to video component

## Screenshots
ralph-time-video

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -136,38 +136,38 @@
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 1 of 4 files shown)


ianmcburnie:
Are we doing this one...? Or should we close?

agliga:
> Are we doing this one...? Or should we close?

I moved this to draft.
I want to keep it around until we figure out if it's absolutely needed or not. 
For now it's not, but still waiting on response if there is no other workaround

Issue #1555, opened 2021-11-11
Menu-button label is not showing


agliga:
# Bug Report

## Description
Seems like when label and postfix label are there it doesn't show up.

Issue #1554, opened 2021-11-11
Alert Dialog & Confirm Dialog: h2 styling not correct


ianmcburnie:
Seems like the h2 heading is missing the class, `alert-dialog__title` `confirm-dialog__title` for alert and confirm respectively.
![Screen Shot 2021-11-11 at 9 49 50 AM](https://user-images.githubusercontent.com/38065/141345559-79dc89d1-d882-4f1c-ba9a-afa017800dc4.png)
![Screen Shot 2021-11-11 at 9 49 56 AM](https://user-images.githubusercontent.com/38065/141345560-52b5aba4-f33d-42c5-8ae7-b4d119236d4c.png)

Skin:

![Screen Shot 2021-11-11 at 9 56 12 AM](https://user-images.githubusercontent.com/38065/141346151-45293103-3b08-434f-ba10-6cec96bfb50d.png)

PR #1556, opened 2021-11-11
fix: dialog title and menu-button label/prefix label

agliga:
## Description
* Fixed dialog to have `__title` on `h2` element
* Fixed menu button by removing `icon-only` attribute. It is no longer needed. 

## References
https://github.com/eBay/ebayui-core/issues/1555
https://github.com/eBay/ebayui-core/issues/1554

## Screenshots
Screen Shot 2021-11-11 at 10 17 11 AM

Screen Shot 2021-11-11 at 10 58 53 AM

### DIFF
--- a/src/components/ebay-menu-button/test/test.server.js
+++ b/src/components/ebay-menu-button/test/test.server.js
@@ -61,12 +61,6 @@ describe('menu-button', () => {
         expect(getByRole('button')).has.class('expand-btn--small');
     });
 
-    it('renders without text', async () => {
-        const input = Object.assign({}, mock.Basic_2Items, { text: '' });
-        const { getByRole } = await render(template, input);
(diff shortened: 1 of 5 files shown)


src/components/ebay-menu-button/test/test.server.js
```diff
@@ -61,12 +61,6 @@ describe('menu-button', () => {
         expect(getByRole('button')).has.class('expand-btn--small');
     });
```
PrashantAshok, line 63: Should there be a test case for `prefix-label` as well?

agliga: We do have one i think. 
Regardless we will update with snapshots at some point.

austingardner (approved):

Issue #1561, opened 2021-11-19
ebay-infotip: passing 'open=true' does not open the infotip by default


melquan:
# Bug Report

## Description
Passing 'open=true' to have the infotip open by default only works for the modal variant, when it should apply for all variants.

## Workaround
None.

PR #1562, opened 2021-11-19
fix(tooltip-base): added checking for open on onMount

agliga:
## Description
Added check for opening tooltips in `onMount`

Basically changed both checks to be in a similar function and just changed the way they are called

## Context
Added a storybook example for it (we can remove it if needed, but it's the only way to simulate a first render open)

## References
https://github.com/eBay/ebayui-core/issues/1561


ianmcburnie (approved):
Looks good. Hopefully this wont get abused too much and we wont see lots of tooltips open on initial page load :-|

Issue #1567, opened 2021-11-30
ebay-section-title: adjust for Skin Evo updates 


ianmcburnie:
CTA (see all) variant has been deprecated in Skin 13.2.0. We'll keep it around until v9. 

A new simpler version of a hyperlinked section title has been added, which basically just adds an SVG icon directly after the link:

```html
<div class="section-title">
    <div class="section-title__title-container">
        <h2 class="section-title__title">
            <a href="https://www.ebay.com">Today’s Deals – All With Free Shipping</a>
            <svg class="icon icon--arrow-right-extra-small" focusable="false" height="10" width="10" aria-hidden="true">
                <use xlink:href="icons.svg#icon-arrow-right-extra-small"></use>
            </svg>
        </h2>
    </div>
</div>
```

Related PR: https://github.com/eBay/skin/pull/1592

I also notice we are missing the `icon-btn--transparent` class on infotip in coreui:

![Screen Shot 2021-11-30 at 1 31 17 PM](https://user-images.githubusercontent.com/38065/144131301-9fbc4d45-7d82-4959-9545-8c8d6064e7f3.png)

PR #1574, opened 2021-12-06
feat(section-title): updated section-title to match EVO

agliga:
## Description
Updated section-title to match EVO changes

## References
#1567

## Screenshots
Screen Shot 2021-12-06 at 12 50 13 PM

### DIFF
--- a/src/components/ebay-section-title/index.marko
+++ b/src/components/ebay-section-title/index.marko
@@ -1,6 +1,7 @@
 import processHtmlAttributes from "../../common/html-attributes"
 
 static var ignoredAttributes = [
+    "ctaText",
     "href",
     "size",
     "title",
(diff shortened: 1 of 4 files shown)


src/components/ebay-section-title/index.marko
```diff
@@ -41,22 +44,6 @@ static var ignoredAttributes = [
             </span>
         </if>
     </div>
-    <!-- CTA -->
```
ianmcburnie, line 44: Do we need to retain this for backwards compatibility?

agliga: Are these classes still available? What would be the right structure for this
I personally don't think so because if this is the new design this is what we should want teams to do

ianmcburnie: There are two pre-existing "CTA" stories which have a different HTML structure. We need to make sure these don't break:

https://ebay.github.io/skin/storybook/?path=/story/section-title-section-title--arrow-cta-see-all
https://ebay.github.io/skin/storybook/?path=/story/section-title-section-title--arrow-cta-no-text

agliga: I readded it back, however the no-text variant does not exist (and we can't get it unless we add a new argument)

ianmcburnie (approved):

Issue #1586, opened 2021-12-16
ebay-menu @item is still clickable when disabled=true


jmdiego:
# Bug Report

## Description
for the `ebay-menu` component with `type="radio"`, when the disabled attribute is added to an `@item` it's still clickable, selects the disabled item and triggers the on-change event.

The expected result is that the click/touch/keyboard enter events are ignored for the disabled items.

## Workaround
Disabling pointer-events for the disabled items is almost a perfect workaround, the keyboard events are still enabled and the disabled option can be still selected that way:
```css
.menu-button__item[disabled] {
  pointer-events: none;
}
```

agliga:
Will be released in `8.3.0`

PR #1587, opened 2021-12-16
fix(menu): don't trigger events when item is disabled

agliga:
## Description
Added a check to remove events if the item is disabled (added the same in fake-menu too)

## References
https://github.com/eBay/ebayui-core/issues/1586


austingardner (approved):
LGTM!

Issue #1598, opened 2022-01-18
menu-button: add support for regular button


agliga:
# Feature Request
In skin we added support for btn in menu-button. We have a form variant which is expand-btn and standard which will be button.

## Context
Original issue is: https://github.com/eBay/skin/issues/1631

Issue #1592, opened 2022-01-05
ebay-fake-menu-button: "priority" attribute rendered in the HTML


melquan:
# Bug Report

## Description
`priority` is rendered in the `ebay-fake-menu-button` HTML.

## Workaround
None.

## Screenshots
![image](https://user-images.githubusercontent.com/19419138/148147990-b8a47eae-02b3-43f7-8399-be432611c7b0.png)

PR #1601, opened 2022-01-21
feat(menu-button): added support for normal button

agliga:
## Description
Added support for normal button (using variant). 
Should also fix priority being passed through (will only be passed to button now, should be null for other cases)

## References
#1598
#1592

## Screenshots
Screen Shot 2022-01-21 at 8 53 29 AM
Screen Shot 2022-01-21 at 8 53 32 AM


ArtBlue (approved):

ianmcburnie (approved):
Works great! I guess we already knew that `priority` being a no-op on certain combinations of `variant` is a code-smell, but we will make a note of it and perhaps address it during our next round of refactoring.

Issue #1599, opened 2022-01-20
listbox: allow non selected state


PrashantAshok:
# Bug Report

## Description

We are using ebay-listbox in our app. I have a requirement to show the listbox without any option selected. However, if none of the option is set as selected, the component set first option as selected by default. Any reason behind this or it's a bug?
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-listbox/component.js#L66

As per skin, Listbox should allow a non-selected state. https://ebay.github.io/skin/#listbox

## Workaround

None

PR #1602, opened 2022-01-21
fix(listbox): changed default selected value to be blank

agliga:
## Description
Fixed listbox to not selected a default option. 
Added an `unselectedText` attribute to default the initial text when no option is selected. (Should default to - like in skin).
Fixed tests and made sure they work with the new missing attribute. Added a new mock to check for first item selected (As most tests are)

## References
https://github.com/eBay/ebayui-core/issues/1599

## Screenshots
Screen Shot 2022-01-21 at 11 53 54 AM
Screen Shot 2022-01-21 at 11 53 56 AM
Screen Shot 2022-01-21 at 11 54 00 AM

### DIFF
--- a/src/components/ebay-listbox/test/mock/index.js
+++ b/src/components/ebay-listbox/test/mock/index.js
@@ -12,6 +12,15 @@ exports.Basic_3Options = {
     })),
 };
 
+exports.Basic_3Options_FirstSelected = {
+    name: 'listbox-name',
+    options: getNItems(3, (i) => ({
+        selected: i === 0,
+        value: String(i),
+        text: `option ${i}`,
+    })),
+};
+
 exports.Basic_3Options_1Selected = {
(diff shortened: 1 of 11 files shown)


ArtBlue (approved):

src/components/ebay-listbox/test/mock/index.js
```diff
@@ -12,6 +12,15 @@ exports.Basic_3Options = {
     })),
 };
 
+exports.Basic_3Options_FirstSelected = {
```
PrashantAshok, line 15: would it be helpful to add a test case where none of the options are selected?

agliga: I have one on server. It checks if no items are selected.

ianmcburnie (approved):
Nice! :-)

PR #1608, opened 2022-01-26
fix(progress-spinner): added override for progress spinner

agliga:
## Description
So because of all the issues with CSS progress spinner and double encoding, I added the spinner background-icon url directly in ebayui. 
This solves the issue (verified it on a sample page)
Added logic to switch between dark-mode as well (@DylanPiercey can you offer a solution if there's a better way to toggle the icon).
I had tried to go the route where we generate the icon, but the icon was having issues importing (need to be normalized I believe)

## References
https://github.com/eBay/skin/issues/1648

## Screenshots
Screen Shot 2022-01-25 at 6 39 19 PM
Screen Shot 2022-01-25 at 6 39 25 PM

### DIFF
--- a/src/components/ebay-filter-menu/examples/01-basic/template.marko
+++ b/src/components/ebay-filter-menu/examples/01-basic/template.marko
@@ -1,4 +1,11 @@
-<ebay-filter-menu>
+class {
+    handleChange({el}) {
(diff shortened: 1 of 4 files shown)


ArtBlue (approved):

agliga:
Closing pr. I think i got a better solution.

PR #1614, opened 2022-02-03
updated package that have an unment dependency i.e nanoid

WaliHassanKhan:
## Description

ebayui-core has an unmet dependency on nanoid.
nanoid has been removed as dependency from the makeup-next-id NPM package.
Updating these three packages fixes the issue.
## Screenshots

Screen Shot 2022-02-02 at 6 08 35 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,14 +151,14 @@
   },
   "dependencies": {
     "@marko-tags/subscribe": "^0.4.2",
-    "makeup-active-descendant": "0.4.0",
-    "makeup-expander": "~0.9.0",
+    "makeup-active-descendant": "0.5.0",
+    "makeup-expander": "~0.10.0",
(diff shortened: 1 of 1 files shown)


ianmcburnie:
makeup-next-id v0.3.0 was published without dependency on nanoid. We just need to update our dependency.

ianmcburnie (changes requested):
The correct versions are in this related skin commit: https://github.com/eBay/skin/pull/1658/commits/36f8c362b45ec99f0a388515503013f2e9a4dc3e

agliga:
Already commited directly to master (and updated yarn.lock)
https://github.com/eBay/ebayui-core/commit/8f710bd1b32db2a2ac4a5a4bc4e6b18f9efa9091

Issue #1612, opened 2022-01-27
Focus is lost when a ebay-menu-button item open a dialog


cc-ebay:
# Bug Report

## Description
I have a problem with focus on a `ebay-menu-button` - One of the item `on-select` opens a dialog.
When the dialog closes, the focus is on a element at the bottom of the page instead of the `ebay-menu-button`.

## Workaround
As a work around I added `el.closest('.menu-button').querySelector('.menu-button__button').focus();`  in the `select` event but  the component should do something about it.

agliga:
So technically there isn't anything we can do about this
The reason is becasue when you click an item on `menu-button` that element you click on will become the focus element, not the menu-button items. When the dialog closes, it will try to focus that element but because it is hidden it will not. 
One possible solution is to have an argument like we have `focusEl` to have `closeFocusEl`, which is the element which the dialog will focus when it's closed. 
@ianmcburnie any thoughts?

ianmcburnie:
I agree with `closeFocusEl`

cc-ebay:
Are you saying to add `closeFocusEl` as an attribute of the dialog? 
Did you think how the code will look like if you do this? 
Would not be make more sense to have that attribute in the `.open` function of the dialog?

agliga:
> Are you saying to add `closeFocusEl` as an attribute of the dialog? Did you think how the code will look like if you do this? Would not be make more sense to have that attribute in the `.open` function of the dialog?

Similarly to how we have `focus` attribute on dialog to force the initial focus element, we can add `closeFocusEl` which you can pass the menu-button selector. When the dialog closes it will focus that.

Issue #1597, opened 2022-01-13
Unused variable in ebay-confirm-dialog


montoya332:
I believe this variable is not used in dialog base therefore it can most likely be removed 
https://github.com/eBay/ebayui-core/blob/b8a608b51258cd1a78540d6d5355fd3e175c709c/src/components/ebay-confirm-dialog/index.marko#L5

Note:
https://github.com/eBay/ebayui-core/blob/master/src/components/components/ebay-dialog-base/component.js
https://github.com/eBay/ebayui-core/blob/master/src/components/components/ebay-dialog-base/index.marko

Issue #1540, opened 2021-10-14
Toast-dialog close button transparency (code smell)


agliga:
Initial PR:
https://github.com/eBay/ebayui-core/pull/1538#pullrequestreview-779193835

We should change the toast dialog close button transparency modifier using an argument or something different than checking toast-dialog base class
If we add more dialogs with that close button support we will need that
Maybe even a way to modify close button itself

PR #1615, opened 2022-02-07
feat(dialog): added close focus attribute (and other cleanup)

agliga:
## Description
* Added close-focus attribute to focus an element on close. This will normally default to the button clicked to open the dialog. 
* Removed unused variable
* Cleaned up a bit of code smell (added `close-button-class` attribute to control the close buttons class)

## References
[#1597](https://github.com/eBay/ebayui-core/issues/1597)
[#1540](https://github.com/eBay/ebayui-core/issues/1540)
[#1612](https://github.com/eBay/ebayui-core/issues/1612)

## Screenshots
Tested this by changing `src/components/ebay-lightbox-dialog/examples/01-basic/template.marko` to add `closeFocus="testid"` to dialog and added `<ebay-textbox id="testid"/>` to the template
dialog

### DIFF
--- a/src/components/components/ebay-dialog-base/component.js
+++ b/src/components/components/ebay-dialog-base/component.js
@@ -23,10 +23,14 @@ module.exports = {
         this.clickTarget = el;
     },
 
-    getActiveElement() {
+    getActiveElement(input) {
+        let closeFocusEl;
+        if (input && input.closeFocus) {
+            closeFocusEl = document.getElementById(input.closeFocus);
+        }
         const el =
             document.activeElement === document.body ? this.clickTarget : document.activeElement;
-        return el;
+        return closeFocusEl || el;
     },
 
     handleStartClick({ target }) {
@@ -165,7 +169,7 @@ module.exports = {
 
         // Ensure focus is set and body scroll prevented on initial render.
         if (isFirstRender && this.input.isModal && isTrapped) {
-            this._prevFocusEl = this.getActiveElement();
+            this._prevFocusEl = this.getActiveElement(this.input);
             this._triggerFocus(focusEl);
             this._triggerBodyScroll(true);
         }
@@ -205,7 +209,7 @@ module.exports = {
 
             if (isTrapped) {
                 if (!isFirstRender) {
-                    this._prevFocusEl = this.getActiveElement();
+                    this._prevFocusEl = this.getActiveElement(this.input);
                     this._triggerBodyScroll(true);
                     this.cancelTransition = transition(
                         {
--- a/src/components/components/ebay-dialog-base/index.marko
+++ b/src/components/components/ebay-dialog-base/index.marko
@@ -7,6 +7,7 @@ static var ignoredAttributes = [
     "type",
     "classPrefix",
     "focus",
+    "closeFocus",
     "a11yCloseText",
     "windowClass",
     "baseEl",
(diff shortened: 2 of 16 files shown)


src/components/components/ebay-dialog-base/component.js
```diff
@@ -23,10 +23,14 @@ module.exports = {
...
-    getActiveElement() {
+    getActiveElement(input) {
+        let closeFocusEl;
+        if (input && input.closeFocus) {
+            closeFocusEl = document.getElementById(input.closeFocus);
+        }
         const el =
             document.activeElement === document.body ? this.clickTarget : document.activeElement;
-        return el;
+        return closeFocusEl || el;
```
cc-ebay, line 33: If the `closeFocus` exists you do not have to get `el` so oyu can put that in an `else`

agliga: No we want to default it to whatever was clicked.
It only defaults to whatever was clicked when closeFocus is not set or doesn't resolve to an element.

cc-ebay: I got it. I suggest:
```
if (closeFocusEl) return closeFocusEl;
return document.activeElement === document.body ? this.clickTarget : document.activeElement;
```
Instead of:
```
const el =
            document.activeElement === document.body ? this.clickTarget : document.activeElement;
        return closeFocusEl || el;
```

cc-ebay: By the way I am not sure what is the style for dev warn/error but it would be definitely useful if you have: 
```
        if (input && input.closeFocus) {
            closeFocusEl = document.getElementById(input.closeFocus);
        } else {
            console.warn('closeFocus id not found');
        }

File: src/components/components/ebay-dialog-base/component.js [hunk 2]
```diff
@@ -23,10 +23,14 @@ module.exports = {
         this.clickTarget = el;
     },
 
-    getActiveElement() {
+    getActiveElement(input) {
+        let closeFocusEl;
+        if (input && input.closeFocus) {
+            closeFocusEl = document.getElementById(input.closeFocus);
```
cc-ebay, line 29: `document.querySelector` would be more flexible than `document.getElementById`

src/components/components/ebay-dialog-base/component.js
agliga: So the reason is this:
We want to make sure this goes to a happy path. If someone accidentally messes up, or an element isn't there, we want to default to the last click target. This is expected for dialog. We do not want this to lose focus. 
If the closeFocus is null, or if it cannot resolve to an element, it should default to the last click target. This is why the code is written like this because if the `document.getElementById` does not match, it returns null. So it will default to the click target.

I don't really want to warn, because users should be testing this before going out. I don't think this is necessary...

src/components/components/ebay-dialog-base/component.js
agliga: Yes, possibly.
However this is in line with `focus` which uses `document.getElementById`. For consistency this is why it is done that way. 
Plus we don't want to deal with conflicts, if there are multiple matches on the page that can lead to other errors. Forcing ID makes it much cleaner and easier.

However, in the next major version we can revisit this and make both support `querySelector` if necessary. But we definitely want to keep this in sync with `focus`

Issue #1607, opened 2022-01-25
ebay-filter-menu: Pass current (un)checked item data in change event


Vaishnavi-Thiyagaraj:
# Feature Request

## Description

Currently `ebay-filter-menu` change event is not passing (un)checked item data which triggered the `change` event.
On change event, we need to know what item was changed and whether it got checked/unchecked. This is because our item list is dynamic based on search result and we need to persist/know the user (un)selected items.

As suggested workaround for now, we are getting the (un)checked state and value of the current checked item like below
```
<@option data-myvalue=value>${value}</@option>
  handleChange({el}) {
        console.log(el.ariaChecked));
        console.log(dataset.myvalue);
        console.log(el.innerText)
    }

```

PR #1621, opened 2022-02-14
fix(filter-menu): added index and currentChecked arguments

agliga:
## Description
* Added extra arguments to change event for filter-menu and filter-menu-button. Added `index` which is the index of the clicked item and `checkedValue` which will be true/false depending if that item is checked (on radio it will always be checked. I kept it for radio to be consistent) 

## References
https://github.com/eBay/ebayui-core/issues/1607

## Screenshots
Screen Shot 2022-02-14 at 10 47 15 AM
Screen Shot 2022-02-14 at 10 47 31 AM

### DIFF
--- a/src/components/ebay-filter-menu-button/component.js
+++ b/src/components/ebay-filter-menu-button/component.js
@@ -11,7 +11,7 @@ module.exports = Object.assign({}, menuUtils, {
         // TODO: the event data from the filter-menu should probably
         // change to include which items are checked not just the values.
         this.toggleChecked(checkedIndex);
-        this._emitComponentEvent('change', el, originalEvent, index, currentChecked);
+        this._emitComponentEvent('change', originalEvent, { el, index, currentChecked });
     },
 
     handleFooterButtonClick() {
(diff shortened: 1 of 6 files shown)


src/components/ebay-filter-menu-button/component.js
```diff
@@ -7,11 +7,11 @@ module.exports = Object.assign({}, menuUtils, {
         eventUtils.handleEscapeKeydown(originalEvent, () => (this._expander.expanded = false));
     },
 
-    handleMenuChange({ checkedIndex, el, originalEvent }) {
+    handleMenuChange({ checkedIndex, el, originalEvent, index, currentChecked }) {
         // TODO: the event data from the filter-menu should probably
         // change to include which items are checked not just the values.
         this.toggleChecked(checkedIndex);
-        this._emitComponentEvent('change', el, originalEvent);
+        this._emitComponentEvent('change', el, originalEvent, index, currentChecked);
```
ianmcburnie, line 14: Maybe anything beyond the 3 "always present" params (i.e. `'change', el, originalEvent`) should be rolled into a single object parameter?  Otherwise this signature could get unwieldy if we keep adding more params.

agliga: Done, good call

ianmcburnie (approved):

Issue #1593, opened 2022-01-05
ebay-button: add support for destructive button


agliga:
# Feature Request

## Description
In skin, we are adding a new split button. Need to add support in ebayui for this button.
Also there is a destructive button added, need to make sure this gets documented.

https://github.com/eBay/skin/issues/1545
https://github.com/eBay/skin/issues/1567

PR #1628, opened 2022-02-18
feat(destructive-buttons): implementing destructive button variant

ArtBlue:
## Description
Added support for destructive button variant.

## Context
Added destructive option for creating that variant for button.

## References
#1593 

## Screenshots
sbook-destructive-p
sbook-destructive-s
sbook-destructive-t

sbook-destructive-p-loading
sbook-destructive-s-loading
sbook-destructive-t-loading

### DIFF
--- a/src/components/ebay-button/button.stories.js
+++ b/src/components/ebay-button/button.stories.js
@@ -91,6 +91,17 @@ export default {
                 },
             },
         },
+        variant: {
+            options: ['destructive'],
+            description:
+                'transforms to a specific variant that styles in conjunction with priority',
+            table: {
+                defaultValue: {
+                    summary: 'none',
+                },
+            },
+            type: { category: 'Options' },
+        },
         'partially-disabled': {
             description: 'programmatically disabled, but remains keyboard focusable',
             table: {
@@ -192,6 +203,7 @@ Standard.args = {
     transparent: false,
     'fixed-height': false,
     truncate: false,
+    destructive: false,
 };
 
 Standard.parameters = {
--- a/src/components/ebay-button/button.stories.js
+++ b/src/components/ebay-button/button.stories.js
@@ -92,12 +92,12 @@ export default {
             },
         },
         variant: {
-            options: ['destructive'],
+            options: ['standard', 'destructive'],
             description:
                 'transforms to a specific variant that styles in conjunction with priority',
             table: {
                 defaultValue: {
-                    summary: 'none',
+                    summary: 'standard',
                 },
             },
             type: { category: 'Options' },
@@ -203,7 +203,6 @@ Standard.args = {
     transparent: false,
     'fixed-height': false,
     truncate: false,
-    destructive: false,
 };
 
 Standard.parameters = {
--- a/src/components/ebay-button/index.marko
+++ b/src/components/ebay-button/index.marko
@@ -40,6 +40,7 @@ $ {
     var fixedHeightClass =
         input.fixedHeight &&
         (sizeClass ? sizeClass + "-fixed-height" : baseClass + "--fixed-height");
+    var variantClass = input.variant ? baseClass + "--" + input.variant : "";
     var tag = input.href ? "a" : "button";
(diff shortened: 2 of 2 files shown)


src/components/ebay-button/index.marko
```diff
@@ -40,6 +40,7 @@ $ {
     var fixedHeightClass =
         input.fixedHeight &&
         (sizeClass ? sizeClass + "-fixed-height" : baseClass + "--fixed-height");
+    var destructiveClass = input.destructive ? baseClass + "--destructive" : "";
```
agliga, line 43: We tend to prefer to to have a "enum" type value for things that might have multiple values. 
So maybe 
`variant="destructive"` or `variant="standard"` (default value)
Because we might add different type.

ArtBlue: So change to this?:

`var variantClass = input.variant ? baseClass + "--" + variant : "";`

If so, I imagine we would need to also change the destructive toggle (bool) in `button.stories.js` from toggle to possibly a combobox with (for now) dropdown option of "destructive?"

agliga: Yes, that's correct. 
See how priority is. 

Usually we have a default like `none` or `standard`.

ArtBlue: @agliga , gotcha. Ok, checked in the tweaks with commit verification.

src/components/ebay-button/button.stories.js
```diff
@@ -91,6 +91,17 @@ export default {
                 },
             },
         },
+        variant: {
+            options: ['destructive'],
```
agliga, line 95: This is close to what we were thinking.
We should have `options: ['destructive', 'standard']`

src/components/ebay-button/index.marko
```diff
@@ -40,6 +40,7 @@ $ {
     var fixedHeightClass =
         input.fixedHeight &&
         (sizeClass ? sizeClass + "-fixed-height" : baseClass + "--fixed-height");
+    var variantClass = input.variant ? baseClass + "--" + input.variant : "";
```
agliga, line 43: Here we should do something like:
```
var variant = input.variant || 'standard';
var variantClass = variant !== 'standard' ? baseClass etc
```
Also we should use backticks (Yes it's not in the previous example but that's better to do)
So 
```
`${baseClass}--${variant}`
```

src/components/ebay-button/button.stories.js
ArtBlue: Gotcha! Yea, being explicit here makes more sense. Ok, adding...

agliga:
One more comment, need to add `variant` to ignored attributes list otherwise it gets spread through. https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-button/index.marko#L9

src/components/ebay-button/index.marko
ArtBlue: Ah, it looks like you already answered that in a comment of your own. Cool.

agliga: Screen Shot 2022-02-22 at 1 22 15 PM

Yes, so just for completeness, this is what happens when it doesn't get added to `ignoredAttribute`

ArtBlue: Yep, saw that happening...the reason I asked.

src/components/ebay-button/index.marko
```diff
@@ -40,6 +41,8 @@ $ {
     var fixedHeightClass =
         input.fixedHeight &&
         (sizeClass ? sizeClass + "-fixed-height" : baseClass + "--fixed-height");
+    var variant = input.variant || 'standard';
+    var variantClass = variant !== 'standard' ? `${baseClass}--${input.variant}` : "";
```
agliga, line 45: Sorry one more nit. 
We can just do: 
```
`var variantClass = variant !== 'standard' && `${baseclass}--etc`;
```
If variantClass is undefined then it does not get added as a class in marko.

agliga (approved):

Issue #1673, opened 2022-03-22
combobox: add expanded attribute


agliga:
# Bug Report
combobox used to support an expanded attribute. We should set it. `onSetupMakeup` if combobox has expanded attribute then we should expand it. Something like this:
```
if(input.expanded && !this.expander.expanded) { this.expander.expanded = true } else if (!input.expanded && this.expander.expanded) { this.expander.expanded = false }
``` 
(Would need to check if the second part of the if statement is needed, since it might rerender closed on rerenders)

ianmcburnie:
Merged to 8.6.0 branch.

PR #1676, opened 2022-03-23
feat(combobox): added expanded attribute

agliga:
## Description
* Added expanded attribute.
* Used the expanded to always force expanded state. Only collapses when expanded is false (and then regular behavior happens.)

## References
https://github.com/eBay/ebayui-core/issues/1673


ianmcburnie (approved):
Tested locally in storybook. Looks great.

Issue #1639, opened 2022-03-09
esm: migrate ebayui to use ESM syntax


agliga:
# Feature Request
In order to be able to import stories into tests to use as mock data, we need to convert all the modules to ESM. This should be doable. Since we use babel to output our components there shouldn't be an issue in terms of compatibility.

PR #1679, opened 2022-04-01
chore: updated to esm

agliga:
## Description
Changed all components to use ESM syntax (__NOTE__ some server tests, like transforms had to remain using non esm. There are a couple server tests which use non esm just based on time constraints)
## Context
* Had to add a package.json with `type: module` in scripts to use ESM there.
* By default this package is non type: module, and it's because storybook breaks with that. 
* I forked the babel config for tests and prod (there was a regenerator runtime error in the deployment with tests. Might have to revisit to have one babel config)
* Removed demo folder
* ~~I switched to c8 for coverage from nyc. There are some issues with ESM on nyc~~ Got nyc working again
* Using marko 5 for tests now

## References
https://github.com/eBay/ebayui-core/issues/1639


ianmcburnie:
Should we leave this for v9 to give us a bit more time to test things out? A little nervous at putting this out in a minor...

agliga:
> Should we leave this for v9 to give us a bit more time to test things out? A little nervous at putting this out in a minor...

We can
However i feel safe because:
- We are already using babel config which is unchanged. So the compiled output will be the same
- all tests are passing. No tests structures were changed except for transforms/migrators which only 1 of them is being really used.

agliga:
> > Should we leave this for v9 to give us a bit more time to test things out? A little nervous at putting this out in a minor...
> 
> We can However i feel safe because:
> 
> * We are already using babel config which is unchanged. So the compiled output will be the same
> * all tests are passing. No tests structures were changed except for transforms/migrators which only 1 of them is being really used.

One more thing I can add is we also have the test page that our QA used to use where we can see if there are any errors or issues that come up. That should be pretty safe to check and verify that the components are working on publish.

ianmcburnie (approved):
A monumental effort! Great set of changes!

Issue #1678, opened 2022-03-31
button: remove aria live


agliga:
As per https://github.com/eBay/skin/issues/1701 need to remove live region from loading button

PR #1680, opened 2022-04-04
fix: removed aria-live from loading button

agliga:
## Description
Removed aria-live from button
Also removed static variable which is not used anymore.

## Context
* Split button snapshots had aria-live in them. Made sure to update snapshots.

## References
https://github.com/eBay/ebayui-core/issues/1678

### DIFF
--- a/src/components/ebay-button/index.marko
+++ b/src/components/ebay-button/index.marko
@@ -28,7 +28,6 @@ static var validPriorities = [
     'delete'
 ];
 
-static var hasLoaded;
 $ {
     input.toJSON = toJSON;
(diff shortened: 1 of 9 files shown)


src/components/ebay-button/index.marko
```diff
@@ -74,7 +70,6 @@ $ {
     type=(tag === "button" && (input.type || "button"))
     aria-disabled=((input.partiallyDisabled || input.loading) && "true")
```
ianmcburnie, line 71: we need to remove aria-disabled when in loading state too

ianmcburnie (approved):

Issue #1594, opened 2022-01-05
star-rating-select: new module


agliga:
# Feature Request
Based on the skin requirement, need to add the ebayui comoponent for interactive star rating.
It should have a way to show the highlighted stars, but also trigger events for clicking a star. 
https://github.com/eBay/skin/issues/1610

agliga:
We should plan on doing:
`<ebay-star-rating value=1.5>` 
`<ebay-star-rating-select value=1.5>`

`star-rating` will be static and `star-rating-select` will be interactive for users and have it's own state.

Value will use the correct SVG icon (might need a map of ratings: component)
```
import rating1 from "<ebay-star-rating-1/>;
import rating1-5 from "<ebay-star-rating-1.5/>;
etc

$ const ratingMap = {
   1: rating1,
   1.5: rating1-5,
};

<${ratingMap[input.value]} /> 
```
Something like this
OR we might want to revisit how we import star-rating's svg's.

PR #1690, opened 2022-04-20
feat(star-rating-select): new component

agliga:
## Description
* New component based on https://github.com/eBay/skin/pull/1725

## Context
* Most of the component is similar to radio. 
* __NOTE__ in draft mode because still waiting on skin part to be merged. But wanted to verify this current approach was okay.
* ~Questions: Should we add an example with fieldset? Also do we want fieldset to wrap the star-rating?~ Added fieldset example and removed `role` if there is no `a11yText`.

## References
https://github.com/eBay/ebayui-core/issues/1594

## Screenshots
Screen Shot 2022-04-20 at 3 48 48 PM

Screen Shot 2022-04-20 at 3 48 59 PM

### DIFF
--- a/src/components/ebay-star-rating-select/README.md
+++ b/src/components/ebay-star-rating-select/README.md
@@ -1,6 +1,6 @@
 <h1 style='display: flex; justify-content: space-between; align-items: center;'>
     <span>
-        ebay-star-ratingselect-
+        ebay-star-rating-select
     </span>
     <span style='font-weight: normal; font-size: medium; margin-bottom: -15px;'>
         DS v1.2.0
--- a/src/components/ebay-star-rating-select/README.md
+++ b/src/components/ebay-star-rating-select/README.md
@@ -1,0 +1,20 @@
+<h1 style='display: flex; justify-content: space-between; align-items: center;'>
+    <span>
+        ebay-star-ratingselect-
+    </span>
+    <span style='font-weight: normal; font-size: medium; margin-bottom: -15px;'>
+        DS v1.2.0
+    </span>
+</h1>
+
+## ebay-star-rating-select Usage
+
+```marko
+<ebay-star-rating-select/>
+```
+
+## Links
+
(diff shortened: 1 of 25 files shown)


src/components/ebay-star-rating-select/README.md
```diff
@@ -0,0 +1,20 @@
+<h1 style='display: flex; justify-content: space-between; align-items: center;'>
+    <span>
+        ebay-star-ratingselect-
```
ArtBlue, line 3: typo `ebay-star-rating-select`

ArtBlue (approved):
Very nice! Great job!

src/components/ebay-star-rating-select/star-rating-select.stories.js
```diff
@@ -0,0 +1,113 @@
...
+    },
+
+    argTypes: {
+        disabled: {
+            control: { type: 'boolean' },
+        },
+        value: {
+            control: { type: 'number' },
+            description:
+                '1 - 5, depending on how many starts are selected. If 0 or null defaults to no stars selected',
```
ianmcburnie, line 36: Small typo. Should be stars not starts

src/components/ebay-star-rating-select/examples/02-fieldset/template.marko
```diff
@@ -0,0 +1,12 @@
+class {}
+
+<fieldset>
+    <legend>Select rating</legend>
```
ianmcburnie, line 4: A small thing, but legend and aria-label strings are different. Let's consolidate on "Rate product" for both

ianmcburnie:
I think we'll definitely need to review the size of these stars with @rbascue

agliga:
> I think we'll definitely need to review the size of these stars with @rbascue

Yes, since in coreui we are relying only on the CSS sizes and not setting them we can probably go ahead with this PR.

agliga:
Also verified that `aria-hidden` is set. In coreui we set it automatically when an icon does not have an `aria-label`

ianmcburnie (approved):

Issue #1640, opened 2022-03-14
star-rating: static module


agliga:
# Feature Request
Related issues: https://github.com/eBay/ebayui-core/issues/1594
Skin issue: https://github.com/eBay/skin/issues/1695

Basically at the moment teams using star-rating have to do `<ebay-star-rating-1/>`. But it would be better to do `<ebay-star-rating rating=1/>` so that the stars can be dynamic. This could also be the basis for the interactive star rating module.

PR #1691, opened 2022-04-21
feat(star-rating): added dynamic star rating (static)

agliga:
## Description
Static star rating, using the `data-stars` value to show stars. It allows teams to do something like`<ebay-star-rating value=input.stars>`

## Context
Still kept icon version of stars around. 

## References
https://github.com/eBay/ebayui-core/issues/1640

## Screenshots
Screen Shot 2022-04-20 at 9 24 13 PM

### DIFF
--- a/src/components/ebay-star-rating/README.md
+++ b/src/components/ebay-star-rating/README.md
@@ -14,12 +14,15 @@
 
 ```marko
 <ebay-star-rating-1/>
+<ebay-star-rating value="1"/>
 <ebay-star-rating-1-5/>
+<ebay-star-rating value="1-5"/>
 ```
 
 ## ebay-star-rating-{rating} Attributes
 
-| Name              | Type    | Stateful | Required | Description                                                                                 |
-| ----------------- | ------- | -------- | -------- | ------------------------------------------------------------------------------------------- |
-| `no-skin-classes` | Boolean | No       | No       | Used for special cases where `icon` classes from Skin should not be applied                 |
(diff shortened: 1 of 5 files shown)


src/components/ebay-star-rating/star-rating.stories.js
```diff
@@ -24,12 +26,37 @@ export default {
         },
     },
 
-    argTypes: {},
+    argTypes: {
+        value: {
+            control: { type: 'text' },
```
ArtBlue, line 31: Maybe the value should be controlled with a dropdown so only valid options can be used. 
image

ArtBlue (commented):
Looks good with one suggestion (see inline)

src/components/ebay-star-rating/star-rating.stories.js
agliga: Good idea

src/components/ebay-star-rating/README.md
```diff
@@ -14,12 +14,15 @@
 
 ```marko
 <ebay-star-rating-1/>
```
ianmcburnie, line 16: so we’ll be phasing out the old syntax <ebay-star-rating-1/> or not?

agliga: Yes we have both
A so its not breaking 
B so its in line with skin

I think the new one is dynamic and the other one is fixed or static in storybook

ianmcburnie (approved):

PR #1696, opened 2022-05-17
PIEACC-587: added optional label for the listbox-button el

ashrdy567:
## Description
Added optional label to announce to know the context

## Context
In feedback-ratings-filter component, added the label(shouldn't be visible in the UI), provides an additional context for the dropdown.

## References
https://jirap.corp.ebay.com/browse/PIEACC-587
Related PR: https://github.corp.ebay.com/ebay-toronto/vi-component-feedback/pull/122

## Screenshots

PIEACC-587-after
PIEACC-587-before

### DIFF
--- a/src/components/ebay-listbox-button/index.marko
+++ b/src/components/ebay-listbox-button/index.marko
@@ -58,7 +58,12 @@ $ var displayText = selectedText || unselectedText;
                 <span class="expand-btn__text">${input.prefixLabel} ${displayText}</span>
             </else-if>
             <else>
-                <span id=labelId class="expand-btn__text">${displayText}</span>
+                <span>
+                    <if(input.label)>
+                        <span class="clipped">${input.label}</span>
+                    </if>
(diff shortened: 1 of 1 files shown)


ianmcburnie (commented):
Adding a hidden prefix to a button in this manner breaks accessibility for voice recognition software (because the start of the programmatic label does not match the visible label). That fact aside, we already have a variety of ways to label a listbox button, or add additional context, in an accessible way. Did you check the docs or storybook?

ashrdy567:
@ianmcburnie I checked the storybook, couldn't find a prop which can set the hidden label.
https://ebay.github.io/ebayui-core/?path=/story/ebay-listbox-button--standard

ianmcburnie:
It shouldn't be hidden. The purpose of a button should be explicit to all users (not just screen reader users) at all times by its text.

More info here: https://ebay.github.io/skin/#menu-button-label
https://ebay.github.io/skin/#listbox-button

Listbox button supports fixed, floating, compact and compound types of labels.

PR #1707, opened 2022-06-30
Issue #1698 - Ignore values in `ebay-tourtip`

### LINKED ISSUE #1698: ebay-tourtip & ebay-snackbar-dialog: attributes that should ignored are being rendered on the span/div tags
# Bug Report

## Description
1. **`ebay-tourtip` - `heading` & `content` attributes**
It appears that `heading` & `content` are not being ignored when processing the html attributes on the span in `ebay-tourtip`?  Looks like it's rendering `heading="{}"` & `content="{}"`.  While upgrading to Marko5, looks like it changes to `heading="[object Object]"` & `content="[object Object]"` (at least that's what I'm seeing in the fixture snapshot tests).

2. **`ebay-snackbar-dialog` - `action` attribute**
Similarly, the `action` attribute is being rendered on the div in `ebay-snackbar-dialog`.

## Workaround
None

## Screenshots
`ebay-tourtip`
image

`ebay-snackbar-dialog`
In the snapshot test:
```
<div
  action="{}"
  aria-live="polite"
  aria-modal="true"
  class="snackbar-dialog snackbar-dialog--transition"
  role="dialog"
>
```

LuLaValva:
- Resolves #1698 

## Description
Added `heading` and `content` to the ignore list for `ebay-tourtip`.

## Context
`action` is already ignored in `ebay-snackbar-dialog`, so I did not have to remove that as well.

### DIFF
--- a/src/components/ebay-tourtip/index.marko
+++ b/src/components/ebay-tourtip/index.marko
@@ -10,7 +10,9 @@ static var ignoredAttributes = [
     "a11yCloseText",
     "host",
     "toJSON",
-    "open"
+    "open",
+    "heading",
+    "content"
 ];
 
 $ input.toJSON = noop;


agliga (approved):
I'm fine with this going in as is. No need for tests

PR #1708, opened 2022-06-30
Issues #1636 and #1703 - Changes to `ebay-listbox-button`

### LINKED ISSUE #1636: ebay-listbox-button: add collapseOnSelect attribute
Add `collapseOnSelect` option to listbox-button so that listbox can auto collapse after selection is made (we did something similar for menu button)

### LINKED ISSUE #1703: ebay-listbix-button: focus lost after option selected
# Bug Report
https://jirap.corp.ebay.com/browse/PIEACC-588

## Description
On ebay-listbox-button component, when option is selected from the dropdown, focus stays on the button. when the previous option is selected, focus shits up to random element.

## Workaround
I tried in my app, onChange of option, I set the focus to button. But, somehow it's not working.
 [https://github.corp.ebay.com/ebay-toronto/vi-component-feedback/compare/master...amaggidi:amaggidi-deploy-branch?expand=1#diff-a3a021[…]f6fe9507R107](https://github.corp.ebay.com/ebay-toronto/vi-component-feedback/compare/master...amaggidi:amaggidi-deploy-branch?expand=1#diff-a3a021025c6ecd274791f1a51fb17da90c59bcca821f4f7772aca630f6fe9507R107)

LuLaValva:
- Resolves #1636
- Resolves #1703 

## Description
- Added boolean argument `collapseOnSelect` to `ebay-listbox-button`.
- Modified `ebay-listbox` to emit a `select` event in addition to the `change` event in the case that the same option was selected again.
- When an option is selected with the keyboard and `collapseOnSelect` is on, focus returns to the original button

## Context
Moving focus back to the button _only_ on button press and not on click was an opinion-based decision, I would like to hear if you all agree.

### DIFF
--- a/src/components/ebay-listbox-button/component.js
+++ b/src/components/ebay-listbox-button/component.js
@@ -12,12 +12,18 @@ export default {
     },
 
     handleListboxChange(event) {
-        if (event.wasClicked) {
-            this._expander.expanded = false;
-        }
         const selectedIndex = event.index;
         this.state.selectedIndex = selectedIndex;
         this.emit('change', event);
+    },
+
+    handleListboxSelect(event) {
+        if (this.input.collapseOnSelect) {
+            this._expander.expanded = false;
+            if (!event.wasClicked) this.el.firstChild.focus();
+        } else if (event.wasClicked) {
+            document.activeElement.blur();
+        }
     },
 
     onCreate() {
--- a/src/components/ebay-listbox-button/component.js
+++ b/src/components/ebay-listbox-button/component.js
@@ -8,22 +8,17 @@ export default {
     },
 
     handleCollapse() {
+        this.getEl('button').focus();
         this.emit('collapse');
     },
 
     handleListboxChange(event) {
+        if (this.input.collapseOnSelect) {
+            this._expander.expanded = false;
+        }
         const selectedIndex = event.index;
         this.state.selectedIndex = selectedIndex;
         this.emit('change', event);
-    },
-
-    handleListboxSelect(event) {
-        if (this.input.collapseOnSelect) {
-            this._expander.expanded = false;
-            if (!event.wasClicked) setTimeout(() => this.getEl('button').focus(), 0);
-        } else if (event.wasClicked) {
-            document.activeElement.blur();
-        }
     },
 
     onCreate() {
--- a/src/components/ebay-listbox-button/component.js
+++ b/src/components/ebay-listbox-button/component.js
@@ -20,7 +20,7 @@ export default {
     handleListboxSelect(event) {
         if (this.input.collapseOnSelect) {
             this._expander.expanded = false;
-            if (!event.wasClicked) this.el.firstChild.focus();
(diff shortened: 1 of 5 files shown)


LuLaValva:
I would like to add some tests for these changes as well, after the first round of review. First I'd like to ensure that I'm moving in the right direction.

src/components/ebay-listbox-button/component.js
```diff
@@ -12,14 +12,20 @@ export default {
...
-        }
         const selectedIndex = event.index;
         this.state.selectedIndex = selectedIndex;
         this.emit('change', event);
     },
 
+    handleListboxSelect(event) {
+        if (this.input.collapseOnSelect) {
+            this._expander.expanded = false;
+            if (!event.wasClicked) this.el.firstChild.focus();
```
ArtBlue, line 23: Does `<ENTER>` and/or `<SPACE>` get treated the same as a click? Does this logic tree work when interacting with a keyboard?

LuLaValva: Good call for additional testing. `<ENTER>` and `<SPACE>` are treated differently than clicks, but I have discovered that selecting an option with `<ENTER>` triggers both an `on-collapse` _and_ `on-expand` at the same time, which leads to some strange behavior. This fixes the problem, but it is very hacky and I would like to find a better solution. Do you have any suggestions?
```suggestion
            if (!event.wasClicked) setTimeout(() => this.el.firstChild.focus(), 0);
```

ArtBlue: Hmmm...very odd, indeed. This is a bit troubling. Before we try to fix this, especially in a hacky way, it might be worth delving into it a bit to get at the root cause.

> selecting an option with <ENTER> triggers both an on-collapse and on-expand at the same time, which leads to some strange behavior

That shouldn't be happening. I have a concern that this might be a bug in Marko itself. Maybe Marko is not capturing/trapping the keyboard event or double-firing it somehow? Dunno.  @ianmcburnie , @agliga , thoughts?

agliga (commented):
LGTM! I will approve, but we need to add a `collapseOnSelect` to `marko-tag.json` 

`"@collapseOnSelect": "boolean",` just add this and I will approve.

src/components/ebay-listbox/component.js
```diff
@@ -12,19 +12,17 @@ export default class {
...
-                this.emit('change', {
-                    index,
-                    selected: [option.value],
-                    el,
-                    wasClicked,
-                });
+                this.emit('change', details);
             });
         }
+        this.emit('select', details);
```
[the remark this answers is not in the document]
LuLaValva: This is needed, because `change` only fires when a *new* option is selected. If this is not included, the dropdown will not close after selecting the item that has already been picked.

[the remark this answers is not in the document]
agliga: That should be fine. That is the existing behavior right?
We shouldn't need to fire any events if they select the same item.

agliga (approved):

Issue #1709, opened 2022-07-14
ebay-page-notice: Add support for dismiss icon


ArtBlue:
Add support for dismiss icon `x` implemented in Skin 14.1.0

This is the updated page notice UI:
<img width="1130" alt="image" src="https://user-images.githubusercontent.com/1675667/179005991-81174378-4803-4a6b-bd9a-c1e9c3fade3c.png">

## Open Questions
1. Should the dismiss icon `x` be present all the time or can it be absent in some use cases?
2. If the dismiss icon is present only in some use cases, is it enabled by default and removable by implementer or does the implementer have to opt in to see it?

PR #1711, opened 2022-07-19
feat(page-notice): added support for dismiss button

ArtBlue:
## Description
Adds support for the dismiss button/icon and CTA link(s) for page notice.

## Context
We needed to maintain some semblance of consistency with regard to markup to align with styling and also keeping things backwards compatible to avoid breaking changes. As such...

1. I introduced a CTA option for users to add main call to action links.
2. I added support for a dismiss button/icon that hides the notice when clicked.

## References
#1709 

## Screenshots
image

### DIFF
--- a/src/components/components/ebay-notice-base/component.js
+++ b/src/components/components/ebay-notice-base/component.js
@@ -1,0 +1,19 @@
+module.exports = {
+    onCreate() {
+        this.state = {
+            dismissed: false,
+        };
+    },
+    handleDismissClick: forwardEvent('on-dismiss'),
+    handleDismissKeydown: forwardEvent('on-dismiss'),
+};
+
+function forwardEvent(eventName) {
+    return function (value, originalEvent) {
+        this.state.dismissed = true;
+        this.emit(eventName, {
+            originalEvent,
+            value: value,
+        });
+    };
+}
--- a/src/components/components/ebay-notice-base/index.marko
+++ b/src/components/components/ebay-notice-base/index.marko
@@ -15,11 +15,13 @@ static var ignoredAttributes = [
     "footer",
     "type",
     "mainRoot",
-    "noA11yLabel"
+    "noA11yLabel",
+    "a11yDismissText"
 ];
 
 $ var status = input.status;
 $ var prefixClass = input.prefixClass;
+<if(!state.dismissed)>
 <${input.root || "section"}
     aria-labelledby=(!input.noA11yLabel && component.elId('status'))
     aria-roledescription=input.a11yRoleDescription
@@ -48,6 +50,11 @@ $ var prefixClass = input.prefixClass;
             </else-if>
         </>
     </if>
+    <if(input.cta)>
+        <p class="page-notice__cta">
+            $!{input.cta}
+        </p>
+    </if>
     <${input.mainRoot || "div"} class=`${prefixClass}__main`>
         <if(input.title)>
             <${input.title.as || "h2"}
@@ -58,9 +65,21 @@ $ var prefixClass = input.prefixClass;
         </if>
         <${input.renderBody}/>
     </>
-    <if(input.footer)>
+    <if(input.footer && !input.a11yDismissText)>
         <div class=`${prefixClass}__footer`>
             <${input.footer}/>
         </div>
     </if>
+    <if(input.a11yDismissText)>
+        <div class=`${prefixClass}__footer`>
+            <button
+                aria-label=`${input.a11yDismissText}`
+                class="fake-link page-notice__dismiss"
+                onClick("handleDismissClick")
+                onKeydown("handleDismissKeydown")>
+                <ebay-close-icon class="icon icon--close-small"/>
+            </button>
+        </div>
+    </if>
 </>
+</if>
--- a/src/components/ebay-page-notice/README.md
+++ b/src/components/ebay-page-notice/README.md
@@ -34,8 +34,10 @@ The `<ebay-page-notice>` is a tag used to create a custom-designed notice elemen
 
 ## ebay-page-notice Attributes
 
-| Name             | Type   | Stateful | Required | Description                                                                                    |
-| ---------------- | ------ | -------- | -------- | ---------------------------------------------------------------------------------------------- |
-| `status`         | String | No       | No       | "attention" (default for "page" and "inline"), "confirmation" "information", or "celebration". |
-| `a11y-icon-text` | String | No       | Yes      | adding description for the notice for a11y users                                               |
-| `icon`           | String | Yes      | No       | "default" (matches whatever is specified by the "status") or "none"                            |
+| Name              | Type   | Stateful | Required | Description                                                                                    |
+| ----------------- | ------ | -------- | -------- | ---------------------------------------------------------------------------------------------- |
+| `status`          | String | No       | No       | "attention" (default for "page" and "inline"), "confirmation" "information", or "celebration". |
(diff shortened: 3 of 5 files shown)


src/components/components/ebay-notice-base/component.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+    onCreate() {
+        this.state = {
+            dismissed: false,
+        };
+    },
+    handleDismissClick: forwardEvent('on-dismiss'),
+    handleDismissKeydown: forwardEvent('on-dismiss'),
```
agliga, line 8: This is the problem. It should just be `dismiss` and not `on-dismiss`. The event it emits should be without the `on-`

src/components/components/ebay-notice-base/component.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+    onCreate() {
+        this.state = {
+            dismissed: false,
+        };
+    },
```
agliga, line 6: Not sure how I feel about this. 

We used to have a `hidden` property, which when it was set to true it would hide the notice. 
I think we can go back to that, but we will need to make sure it syncs up with the user input. 
The problem is once it is dismissed, there's no way the user can show the notice again. 

So maybe we can do this:
```
    onInput(input) {
        this.state = { dismissed: input.dismissed || false };
    },
```
And then add a new input.dimissed to page notice. 

This way, the user can toggle the dismissed attribute, in case they want to reset it or not.

src/components/ebay-page-notice/README.md
```diff
@@ -34,8 +34,10 @@ The `<ebay-page-notice>` is a tag used to create a custom-designed notice elemen
...
+| Name              | Type   | Stateful | Required | Description                                                                                    |
+| ----------------- | ------ | -------- | -------- | ---------------------------------------------------------------------------------------------- |
+| `status`          | String | No       | No       | "attention" (default for "page" and "inline"), "confirmation" "information", or "celebration". |
+| `a11y-icon-text`  | String | No       | Yes      | adding description for the notice for a11y users                                               |
+| `icon`            | String | Yes      | No       | "default" (matches whatever is specified by the "status") or "none"                            |
+| `cta`             | String | No       | No       | adding a dedicated CTA link in notice with full anchor tag in HTML                             |
+| `a11yDismissText` | String | No       | No       | adding a dismiss button to dismiss the notice                                                  |
```
agliga, line 43: Lets remove the page-notice-attributes. We should only have documentation on storybook

Look at `split-button` docs, lets copy the readme from there. Have it have links to point to storybook.

src/components/ebay-page-notice/index.marko
```diff
@@ -6,4 +8,5 @@ $ var status = input.status || 'attention';
     status=status
     role="region"
     prefixClass="page-notice"
+    on-dismiss('emit', 'on-dismiss')
```
agliga, line 11: Again. should be just `('emit', 'dismiss')`

src/components/components/ebay-notice-base/index.marko
```diff
@@ -48,6 +50,11 @@ $ var prefixClass = input.prefixClass;
             </else-if>
         </>
     </if>
+    <if(input.cta)>
+        <p class="page-notice__cta">
+            $!{input.cta}
+        </p>
```
agliga, line 56: Oh we should use attribute tags instead of this. This is unsafe to do. 

You will need to add the definition in `marko-tag.json` to get it to work. 

By doing an attribute tag, we could actually just do:

`<@cta href="link">Body text</@cta>`

Which is pretty nice. 

We can do:
```
<if(input.cta)>
        <p class="page-notice__cta">
          <a ...input.cta><${input.cta}/></a>
       </p>
```

We could also override the `<a>` if needed, but I think it should be fine for now.

src/components/ebay-page-notice/page-notice.stories.js
```diff
@@ -59,19 +62,36 @@ export default {
...
-            description: 'The footer content to be displayed. Used to show a CTA button generally',
+            description:
+                'The footer content to be displayed. Used to show the dismiss button generally',
             table: {
                 category: '@attribute tags',
             },
         },
+        cta: {
+            description: 'This adds a CTA link using full HTML anchor tag',
+        },
```
agliga, line 73: Should be an attribute tag, like `@cta`.

src/components/components/ebay-notice-base/component.js
ArtBlue: Gotcha. The `on-` is implied. Good to know. I fixed this.

src/components/components/ebay-notice-base/component.js
ArtBlue: This was intentional actually. I was basing my design decisions here on a true "dismiss" and those don't typically need to be re-displayed. When users dismiss something, they're usually completely getting rid of it from the screen and it's usually implied that this is going away for good. If you think it's worth creating this additional level of abstraction, I suppose we can. Just not sure about its utility.

src/components/ebay-page-notice/README.md
ArtBlue: That sounds much better - single source. Will do. What's odd is that I don't recall writing the docs for the first few. I think I just added the last 2. Weird.

agliga: Yes. 
And don't worry about it. I have started doing that method in a couple of components but not for all of them. So as we touch code, we should update them to use the new method (like I did in my dialog PR)
I believe the docs for all the others are just format/whitespace changes. Our prettier changes the tables to align correctly.

src/components/components/ebay-notice-base/component.js
agliga: So I agree in theory with what you have. 
The concern I have is on app teams. I know a lot of app teams tend to reuse our components in different situations. It might not be ideal, but I would hate it for them to dismiss a component and then expect it to be rendered later. Having a way to basically show the notice that was dismissed again is helpful imo. 

@ianmcburnie any input? I'm fine to go either way.

ArtBlue: Oh, I see. You were thinking of implementers (devs), not actual end users. I was just doing a page refresh to get it back when building 😄 .

If it's an enhancement for devs, I guess I can go either way on this too but I still wonder how many devs would just refresh the page anyway.

src/components/components/ebay-notice-base/index.marko
ArtBlue: Yea, I was wondering about this. I realized unescaped HTML may not be the best option. Ok, let me work on changing this.

ArtBlue: Ok, I made the switch. I wasn't sure about the changes in `marko-tag.json` since I don't know what those options mean; was just a copy/paste/modification.

agliga (approved):
I'm going to approve it for now. 
I think the dismiss part might still be an issue, but we can always change it if needed. 

I would like to wait for more input on that part before merging.

src/components/components/ebay-notice-base/component.js
ianmcburnie: Agree with Andrew, that we've seen that quite often devs want to re-use/re-show the same notice component. I think we should allow that flexibility, otherwise we should remove the notice from the DOM to make it very clear once it's gone it's gone (and they would need to instantiate a new ebay-notice).

agliga (approved):

PR #1713, opened 2022-07-19
feat: added borderless to button (sync from skin)

agliga:
## Description
We missed adding borderless support for button last version. 
Added it here and updated menu button to support the new button borderless argument

### DIFF
--- a/src/components/ebay-button/button.stories.js
+++ b/src/components/ebay-button/button.stories.js
@@ -60,6 +60,15 @@ export default {
                 },
             },
         },
+        borderless: {
+            description: 'Shows button without border.',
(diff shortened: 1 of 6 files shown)


LuLaValva (approved):
Looks good!

ianmcburnie (approved):
Nice!

ArtBlue (approved):
Everything looks good with this specific PR but there is a minor issue with the styling.

image

It looks like there is an additional `20px` padding on the right. It looks like `btn--borderless` overrides the left padding with 0 but doesn't do it for the right; the right padding is not overridden and creates an imbalance.

PR #1727, opened 2022-08-12
fix: reset ebay-select on form reset

### LINKED ISSUE #1714: ebay-select: not reset to default option on form reset
# Bug Report

## Description

If a `<select>` is used inside a `<form>`, clicking on `<button>` or `<input>` with `type="reset"` will reset the `<select>` to its default option (usually the first option if `selected` is not set on options) without any js.
This is expected to work with `<ebay-select>` as well.

LuLaValva:
- Fixes #1714 

## Description
The `ebay-select` component did not reset when the form it was wrapped in received a "reset" event. This has been fixed.

### DIFF
--- a/src/components/ebay-select/examples/08-in-form/template.marko
+++ b/src/components/ebay-select/examples/08-in-form/template.marko
@@ -10,7 +10,6 @@
     <ebay-button type="reset">
       Reset
     </ebay-button>
-    &nbsp
     <ebay-button type="submit">
       Submit
     </ebay-button>
--- a/src/components/ebay-select/examples/08-in-form/template.marko
+++ b/src/components/ebay-select/examples/08-in-form/template.marko
@@ -1,0 +1,18 @@
+<form style="text-align: center;">
+  <div>
+      <ebay-select name='formFieldName'>
+          <@option value='1' text='Option 1'/>
+          <@option value='2' text='Option 2'/>
+          <@option value='3' text='Option 3'/>
+      </ebay-select>
+  </div>
+  <div style="padding: 1em;">
+    <ebay-button type="reset">
+      Reset
+    </ebay-button>
+    &nbsp
+    <ebay-button type="submit">
+      Submit
+    </ebay-button>
+  </div>
+</form>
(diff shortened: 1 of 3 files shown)


agliga (approved):
Good job, simple and works well.

ArtBlue (approved):
Very nice! The simplicity is eloquent.

src/components/ebay-select/examples/08-in-form/template.marko
```diff
@@ -0,0 +1,18 @@
...
+          <@option value='1' text='Option 1'/>
+          <@option value='2' text='Option 2'/>
+          <@option value='3' text='Option 3'/>
+      </ebay-select>
+  </div>
+  <div style="padding: 1em;">
+    <ebay-button type="reset">
+      Reset
+    </ebay-button>
+    &nbsp
```
ianmcburnie, line 13: Remove this line. Will cause issues in actual code.

ianmcburnie:
This is setting selected back to the first value. What happens when there is a `selected` value? Both for server rendered and client rendered.

agliga (changes requested):
We should first verify how to get the default value set. Usually when `selected` is set it forces  that to be the default value. 
What we can do:
We should have a default value saved `onMount`. (Uses should also be able to pass that default value in case they want to change it). That should be the reset value.

So we can do:
```
<ebay-select>
   <@option selected default-value>Option1</>
   <@option>Option2</>
</ebay-select>
```
And then either find what the default-value is onMount or use the currently selected one as the default value. Then when the reset is triggered it should select that element.

LuLaValva:
I've pushed a fix that chooses the first option selected as the default value. Including a `default-value` attribute seems to be outside the scope of this issue since `ebay-select` doesn't currently have that as a feature. Would you like me to add it anyway?

agliga:
> I've pushed a fix that chooses the first option selected as the default value. Including a `default-value` attribute seems to be outside the scope of this issue since `ebay-select` doesn't currently have that as a feature. Would you like me to add it anyway?

I think the issue is if the component gets unmounted and remounted then it might change the default selected value. 
It should be pretty easy to add that

agliga (approved):

Issue #1716, opened 2022-07-25
tooltip: design updates sync with skin


agliga:
Initial issue https://github.com/eBay/skin/issues/1745
Need to change the modal version of infotip to show drawer instead of lightbox.

Issue #1688, opened 2022-04-18
Accessibility issue `ARIA attribute is not allowed: aria-expanded="true"`


joinfelipe:
# Bug Report

## Description
Component has accessibility issue. The following error is given when running accessibility validation tool: 
```
ARIA attribute is not allowed: aria-expanded="true"
Element source:
<span class="tourtip__host" aria-expanded="true" aria-controls="s0-0-0-24-17-4-1-3-18[0]-overlay"></span>
```

## Screenshots
<img width="1003" alt="Screen Shot 2022-04-18 at 11 53 19 AM" src="https://user-images.githubusercontent.com/69696226/163860719-3004020b-7508-4057-aeaf-576b1ab437d2.png">

agliga:
Will either need to:
Fix tourtip to not use expander
OR 
Fix expander to not add aria-expanded

ianmcburnie:
Same issue for `aria-controls` too, I'd imagine. It shouldn't be showing up here.

PR #1728, opened 2022-08-16
feat: tooltip, infotip, tourtip sync with skin and fixes

agliga:
## Description
* Synced with tooltip changes from skin (icon mainly and slight positioning)
* Changed tourtip to not rely on expander and manually show/hide
* Changed infotip to use drawer dialog on modal variant. 

## Context
* Changed tests to use storybook and not rely on mocks

## References
#1688 #1716

## Screenshots
Screen Shot 2022-08-16 at 11 26 42 AM
Screen Shot 2022-08-16 at 11 26 48 AM
Screen Shot 2022-08-16 at 11 27 13 AM
Screen Shot 2022-08-16 at 11 33 04 AM


LuLaValva (commented):
- I think infotip should probably be moved from `buttons` to `notices & tips` in the docs, is that update appropriate to add to this PR while you update `ebay-infotip`?
- The text inside the infotip modal looks like a function displayed as a string, this may be a quirk of storybook or a bug with the modal

As far as I have discovered, everything else looks great.

agliga:
> - I think infotip should probably be moved from `buttons` to `notices & tips` in the docs, is that update appropriate to add to this PR while you update `ebay-infotip`?
> 
> - The text inside the infotip modal looks like a function displayed as a string, this may be a quirk of storybook or a bug with the modal
> 
> 
> 
> As far as I have discovered, everything else looks great.

1. I'm fine with it but i think Ian wants it there. @ianmcburnie 
2. I'll check. It might be due to marko storybook.

agliga:
> * I think infotip should probably be moved from `buttons` to `notices & tips` in the docs, is that update appropriate to add to this PR while you update `ebay-infotip`?
> * The text inside the infotip modal looks like a function displayed as a string, this may be a quirk of storybook or a bug with the modal
> 
> As far as I have discovered, everything else looks great.

Good catch, I checked the heading and there was a bug. I pushed up the fix. Thanks!

LuLaValva (approved):
Ok, I'll approve for now but I'd like to see an approval from @ArtBlue or @ianmcburnie as well before it is merged.

ianmcburnie:
I think the initial idea was it for to live under notices & tips for now and if we ever rename it to infotip-button, then it could be moved under buttons. I'm fine with it staying under buttons as is though. If we are to continue trying to align these based on ARIA roles (which I like) then there would be a section called tooltips, with a single component under it: ebay-tooltip (as that is the only one that is an actual ARIA tooltip role). Infotip would be under buttons and tour tip would be under....hmm, "Tours" maybe?

ArtBlue (approved):
Looks good to me.

PR #1742, opened 2022-08-24
chore: removed unused dependencies

agliga:
## Description
__Note__ Targeting master so we remove any dependabot alerts
* Removed unused dependencies. Over time we have not removed and deps we are no longer using. This PR resolves that
* There was an issue with `mdx` using a new version of react. I assume one of the unused depdencies was calling it, so I loaded mdx/react for that case.


ianmcburnie (approved):
Nice cleanup!

ArtBlue (approved):

Issue #1701, opened 2022-06-01
Discussion: v10 refactoring & breaking changes 


agliga:
Original Issue: #1531

Maybes:

* ebay-menu & ebay-menu-button
   * does not support groups (see: https://makeup.github.io/makeup-js/makeup-menu/index.html)
* Button (comments originally from @darkwebdev in #923)
   * Does it make sense to call a property "priority" if it has a delete value? Answer: probably not. Maybe rename to "variant". We could move to even higher level of encapsulation with ebay-primary-button, ebay-delete-button etc (we already have ebay-expand-button and ebay-cta-button). See: https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/
   * Do we need a value fake-link for variant property when we can just infer it from the href property? Long term it might be best to split ebay-button out into ebay-fake-button (see above)
* Does page-notice-base still make sense? Not sure inline notice has much in common with page notice and section notice
   * https://github.com/eBay/ebayui-core/pull/1400/files#diff-5140941efa2fb8731469d23631806f4bce10e377aca91833034b18c57aa4938dR18

Other

* ebayui & vite
   * support commonjs & esm 
* Marko 5 features? Anything we can take advantage of?
* Does ebay- prefix still make sense?
   * We want to make coreui the North Star for Marko components
   * Ties in with our headless ui / unbranded vision
   * Maybe ‘core-’ instead? Or 'marko-ui'? Can we think of a cool name?
* Need a consistent I18N strategy
* Remove hardcoded English defaults
* Perhaps make all strings required at compile time?
* Align with standard L10N bundle support (.properties file)
* rename "fake" things for better findability?
   * ebay-fake-menu => ebay-menu-fake
   * ebay-fake-tabs => ebay-tabs-fake
   * ebay-fake-menu-button => ebay-menu-button-fake

agliga:
Fix default for fake-menu-button to be type=button

PR #1743, opened 2022-08-26
chore: removed expand button, progress, and window notice

agliga:
## Description
* Removed ebay-progresss. There was already a migrator for it, so now component has been removed
* Removed expand-button. Added a migrator to auto migrate teams to use ebay-button
* Removed window-notice. No migrator since this is no longer being supported
* Fixed all components which used expand button to use button. 

## References
#1701


LuLaValva (approved):
Looks good! A few small documentation critiques that shouldn't hold back from merging but I think are worth pointing out:

- In the `ebay-button--expand-button` storybook page, `bodyState` still has a default label of `none` and `variant` is labeled as `standard`, even though it probably makes sense to label them as `expand` and `form`. This seems like it is not a trivial fix, so since it's not that big of a deal there's probably no need to figure out how to modify it.
- `src/components/ebay-expand-button/README.md:12` doesn't include `variant="form"`. This may be intentional since you can have expand buttons in all variants, but I wanted to point it out in case it wasn't.

agliga:
I'm going to merge this to move forward. 
I will fix those readme issues in a further PR.
Overall readmes should be deprecated and we should remove all attributes and info from there and only link out to storybook.

Issue #1744, opened 2022-08-29
ebay-infotip: rendering issues on safari


agliga:
Safari has a bug with filter which causes rerendering issues when using filter with transform: translate. This is mainly seen in ebayui.
Should add a scale3d to all transforms to fix this.

PR #1745, opened 2022-08-29
fix: added scale3d for safari rerender bug

agliga:
## Description
* Added scale3d to all transitions for all tooltips. This fixes a bug on safari

## References
https://github.com/eBay/ebayui-core/issues/1744

## Screenshots
Screen Shot 2022-08-29 at 12 35 09 PM


LuLaValva (approved):

ArtBlue (approved):
Looks good. I think we may need a similar fix in skin storybook.

Issue #1748, opened 2022-09-02
Deprecated `ebay-stepper` references non-existent skin file


DylanPiercey:
# Bug Report

https://github.com/eBay/ebayui-core/blob/7ba7f4e994b85da5fea069a6cd1dd30ece1e5c93/src/components/ebay-stepper/style.js References a no longer existent module from `@ebay/skin`. If this module is loaded it will cause a build error.

This is especially an issue when the module is used with Vite where it is unconditionally prebundled.

## Description
If you try to load `ebay-stepper` it should error. Also if you load `@ebay/ebayui-core` at all within a Vite project it will also error.

## Workaround
N/A

## Screenshots
N/A

agliga:
Stepper has been removed in 10.0.0. It was used for migration only.

Issue #1738, opened 2022-08-23
deprecations: remove deprecated skin classes and components


agliga:
See https://github.com/eBay/skin/issues/1650 and https://github.com/eBay/skin/pull/1841

Some code has been removed in skin 15. We need to remove it from ebayui. 
* expand-button (component and any use. Make sure we use button instead)
* window-notice (remove)
* borderless, make sure we use button instead
* wide button, remove
* panel dialog center title, removed
* reliance on split icon packs. Should only pull from icons

Issue #1735, opened 2022-08-23
progress-stepper: Improved state styles and guidelines


agliga:
original issue: https://github.com/eBay/skin/issues/1826

PR #1749, opened 2022-09-07
feat(progress-stepper): updated to new design

agliga:
## Description
We had several changes in this PR which are needed for progress stepper
* Updated icon import. We need to import the new progress stepper icons
* Reimported icons to get new progress stepper icons
* Removed stepper module, it should be unused.
* Updated progress stepper with new code

## Context
* Removed type from `@step`. With this new change there would only be two types, attention and whatever the default is. If someone wanted to override all steps, they should use `defaultState`. So because of that, I rolled in `attention` into `defaultState`. 
* Cleaned up some unused classes and structure.

## References
https://github.com/eBay/ebayui-core/issues/1748
https://github.com/eBay/ebayui-core/issues/1738
#1735

## Screenshots
Screen Shot 2022-09-07 at 10 32 39 AM
Screen Shot 2022-09-07 at 10 32 33 AM
Screen Shot 2022-09-07 at 10 34 03 AM
Screen Shot 2022-09-07 at 10 34 10 AM


ArtBlue (approved):
Nice work! Looks great!

LuLaValva (approved):
Looks good to me

PR #1752, opened 2022-09-12
chore: updated readmes for components

agliga:
## Description
* Changed readme for about half of the components
* Added links to all storybook components and common examples
* Removed most of the text from readmes, since they should be looking up in storybook on how to's

I will do the rest for the other components, but wanted to get any feedback on this currently


LuLaValva:
I like the direction this is going, as it will naturally keep documentation in sync with the components as they are developed. Creating an "Examples and Documentation" section for each README is definitely a good idea.

It seems you are suggesting that we remove all code examples from documentation and instead provide a link to the `examples` folder in GitHub. However, for some components such as `ebay-alert-dialog` and `ebay-badge` you kept the README example. Is there something special about these examples that means they should be kept, or will they be removed when you complete the documentation update?

agliga:
> I like the direction this is going, as it will naturally keep documentation in sync with the components as they are developed. Creating an "Examples and Documentation" section for each README is definitely a good idea.
> 
> It seems you are suggesting that we remove all code examples from documentation and instead provide a link to the `examples` folder in GitHub. However, for some components such as `ebay-alert-dialog` and `ebay-badge` you kept the README example. Is there something special about these examples that means they should be kept, or will they be removed when you complete the documentation update?

I haven’t done all of them. I’m going to in a followup. 
I mainly wanted to get some early feedback. I started to do a couple of them and then lost track and started to do even more before stopping and realizing i want to make sure we are fine with this direction. So that’s why there’s more than i initially had planned

LuLaValva (approved):
Approving the current changes under the assumption that merge conflicts will be resolved and this will have a follow-up PR that completes the transition for the rest of the components

PR #1754, opened 2022-09-12
docs: added class to labels of disabled fields

LuLaValva:
- addresses eBay/skin#1843

## Description

Set an example by assigning the `field__label--disabled` class to disabled fields in the `ebayui-core` storybook.

### DIFF
--- a/src/components/ebay-checkbox/examples/WithLabel.marko
+++ b/src/components/ebay-checkbox/examples/WithLabel.marko
@@ -9,7 +9,7 @@ class {}
         on-change("emit", "change")
         on-focus("emit", "focus")
         on-keydown("emit", "keydown") />
-    <label class=["field__label", "field__label--end", input.disabled && "field__label--disabled"] for="checkbox">
+    <label class="field__label field__label--end" for="checkbox">
         Option
     </label>
 </span>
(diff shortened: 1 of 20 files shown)


src/components/ebay-checkbox/examples/WithLabel.marko
```diff
@@ -9,7 +9,7 @@ class {}
         on-change("emit", "change")
         on-focus("emit", "focus")
         on-keydown("emit", "keydown") />
-    <label class="field__label field__label--end" for="checkbox">
+    <label class=["field__label", "field__label--end", input.disabled && "field__label--disabled"] for="checkbox">
```
agliga, line 12: I think what might be better is to actually have another story with disabled example, rather than using the toggle.
The toggles should only be used for arguments that get passed to the actual component.

LuLaValva:
Alright, I've updated these changes so that each form component that needs it has a "Disabled With Form" example. Some quick notes of storybook issues I ran into and fixed along the way:

- Some examples lacked `...input`, which prevented them from being controlled by the storybook Controls tab
- I changed some of the "text" controls which only had limited options to "Options" controls (`checkbox`, `radio`, `textbox`, `tri-state-checkbox`)

src/components/ebay-checkbox/examples/WithLabel.marko
LuLaValva: Should be moved now, thanks Andrew.

Issue #1736, opened 2022-08-23
section-title: updated "see all" variant


agliga:
Original issue: https://github.com/eBay/skin/issues/1728

PR #1758, opened 2022-09-15
feat(section-title): synced styles with skin and updated docs

agliga:
## Description
* Synced section title with skin
* Removed the CTA without text
* Restructured some code
* Added snapshot tests.
* Updated readme for section title

Arthur, make sure I got all the cases and removed all the unused usecases

## References
#1736

## Screenshots
Screen Shot 2022-09-15 at 4 38 15 PM

Screen Shot 2022-09-15 at 4 38 18 PM
Screen Shot 2022-09-15 at 4 38 23 PM


ArtBlue (commented):
Overall looks good. I think you got rid of all the obsolete variations and the new ones appear to cover the current ones. A few minor things...

1. The "size" option (large or regular) is not needed anymore since the size is determined by the viewport width.

2. The ctaText option control doesn't appear to be doing anything. Maybe there's a DOM selector mismatch somewhere since I flattened the DOM (the cta link text now lives under `.section-title__cta`.

3. Under "Standard" changing the `@info` and `@overflow` causes an error. I'm guessing those options may not be available under that variation?

agliga:
> Overall looks good. I think you got rid of all the obsolete variations and the new ones appear to cover the current ones. A few minor things...
> 
> 1. The "size" option (large or regular) is not needed anymore since the size is determined by the viewport width.
> 2. The ctaText option control doesn't appear to be doing anything. Maybe there's a DOM selector mismatch somewhere since I flattened the DOM (the cta link text now lives under `.section-title__cta`.
> 3. Under "Standard" changing the `@info` and `@overflow` causes an error. I'm guessing those options may not be available under that variation?
Thanks for this!
The only one that isn't an issue is #2. We have ctaText which should control the text of the cta link (should be "see all" by default). You need to have the HREF in order to see it. I will update the documentation for this to make it more clear.

ArtBlue (approved):
Looks good! I'm still not able to get the UI to change when adding options, but we can deal with that later. 
image

agliga:
> Looks good! I'm still not able to get the UI to change when adding options, but we can deal with that later.
> image

Oh i see. We need to just pass in the input for that case. 
it should work in the first case since that’s loading the index directly.

PR #1761, opened 2022-09-19
10.0.0

ArtBlue:
Release pull request for 10.0.0 release.


src/components/ebay-drawer-dialog/component.js
```diff
@@ -20,15 +20,20 @@ export default {
 
     handleTouchStart(event) {
```
ianmcburnie, line 21: Might be worth briefly documenting at some point what is actually happening here with these touch events, at a high level.

src/components/ebay-icon/examples/all.marko
```diff
@@ -1554,6 +1554,12 @@ div.icon-examples
         span.text
             -- ebay-stepper-confirmation-icon
 
+   div
```
ianmcburnie, line 1563: Good use of Marko concise syntax here.... from somebody who isn't a fan of concise syntax ;-)

agliga: I’m not a fan either but generating this file in concise syntax was much easier than doing html

src/components/ebay-menu-button/test/__snapshots__/renders-with-overflow-variant.expected.html
```diff
@@ -12,21 +12,21 @@
...
         [36m<defs>[39m
           [36m<symbol[39m
-            [33mid[39m=[32m"icon-overflow"[39m
-            [33mviewBox[39m=[32m"0 0 4 18"[39m
+            [33mid[39m=[32m"icon-overflow-small"[39m
+            [33mviewBox[39m=[32m"0 0 3 13"[39m
           [36m>[39m
             [36m<path[39m
-              [33md[39m=[32m"M2 4A2 2 0 102.001.001 2 2 0 002 4zm0 7a2 2 0 10.001-3.999A2 2 0 002 11zm2 5a2 2 0 11-3.999.001A2 2 0 014 16z"[39m
+              [33md[39m=[32m"M1.5 10a1.5 1.5 0 110 3 1.5 1.5 0 010-3zm0-5a1.5 1.5 0 110 3 1.5 1.5 0 010-3zm0-5a1.5 1.5 0 110 3 1.5 1.5 0 010-3z"[39m
```
ianmcburnie, line 24: Might be worth documenting how these snapshots are generated, and the general process around it, in our contributing guide if we haven't done so already.

src/components/ebay-drawer-dialog/component.js
agliga: Yes. Overall its basically detecting if the area was tapped and dragged up/down by a certain amount. 
the change here was that native events don’t support forEach. I could have used them in a prototype but i kinda like just using the base for loop

src/components/ebay-menu-button/test/__snapshots__/renders-with-overflow-variant.expected.html
agliga: I will add a task to do for post release. Or i can add a new command to also update snapshots automatically

ianmcburnie (approved):

src/components/ebay-drawer-dialog/component.js
ianmcburnie: > its basically detecting if the area was tapped and dragged up/down by a certain amount.

Cool. Please create a task or note for that.

agliga: Will do

LuLaValva (approved):

Issue #1762, opened 2022-09-19
ebay-carousel: existing tabindex value is being clobbered


ianmcburnie:
# Bug Report

## Description

The ebay-carousel adds `tabindex=-1` to any interactive element that moves outside of the viewport (to prevent kb access to it).The `tabindex` attribute is then removed when the item comes into view.  The issue is that this brute-force removal is too aggressive, and any _pre-existing_ tabindex is not restored back to its initial state.

To summarize: if an element was rendered with ` tabindex=-1`, our code is not restoring that value when it comes back into view.

The same issue applies to `aria-hidden`.

https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-carousel/component.js#L93-L103

An example of this behavior can be found here on the refurbished items carousel: https://www.ebay.com/sch/i.html?_from=R40&_trksid=p2380057.m570.l1313&_nkw=iphone&_sacat=0#mainContent

## Workaround

None

## Dev Notes

![Screen Shot 2022-09-19 at 12 34 25 PM](https://user-images.githubusercontent.com/38065/191100719-30c59cdc-a095-4688-807c-f687e6d03e6e.png)

agliga:
The problem is we don't know what the initial tabindex is. This is run on render. We could do a check if it's first render to preserve the tabindexes, but if they change from the parent, then we have no idea if it was first render, or if we are rerendering. If we do the first render, then it should be pretty easy to remember the tabindex. But if any of the structure changes, we could lose that or cause mismatches. 
There are several possible ways to get around this. We could ask teams to have something like a `data-tabindex` on each item they want to be restored, so we know what the initial value is. Or we can add that `data-tabindex` and say it only works on initial render and then restore it as needed.

agliga:
Based on my discussions with Dylan, we will add support for users to pass a `default-tabIndex`, which if set we will default to when we remove the tabindex. The users can add this to any element inside the carousel and we will look for it when we reset the tabindex, otherwise we will remove the tabindex.

PR #1773, opened 2022-09-29
feat(carousel): added support for default-tabindex

agliga:
## Description
Added `default-tabindex` option to any focusable element to basically allow a default to be set when carousel changes the tabindex on an item
Added an example of that to the page. (I'm happy to change it as needed but it should be pretty basic)
Also updated readme for carousel to use new pattern.

## References
https://github.com/eBay/ebayui-core/issues/1762


ianmcburnie:
Is `default-tabindex` going on a native HTML element? If so, shouldn't we use a data attribute? Something like `data-carousel-default-tabindex`?

agliga:
> Is `default-tabindex` going on a native HTML element? If so, shouldn't we use a data attribute? Something like `data-carousel-default-tabindex`?

Yes, you're right.

LuLaValva (approved):
I'm not sure if this is the cleanest way to allow tab index and it's a little strange to be using the `dataset` directly, but it looks like it is working on my end so I will approve.

agliga:
> I'm not sure if this is the cleanest way to allow tab index and it's a little strange to be using the `dataset` directly, but it looks like it is working on my end so I will approve.

So here's why:

Basically we are scanning each element inside the carousel and seeing if it's a focusable element (button/link etc). And for all those focusable elements, we either remove or set a tabindex to -1 if it's either in view or obscured by the carousel. 

The problem is we don't know what the initial tabindex was. We can try to save the first render tabindex, but if it changes in anyway, it could cause issues or not render correctly. 

This change is supposed to solve that by allowing users to specify the tabindex they want it restored to by using a data-attribute. This way we won't be guessing what the right tabindex is.

agliga:
I'm going to merge, if there are any issues I can fix them in a followup PR

Issue #1777, opened 2022-10-03
floating-label: add support for opaque floating label


agliga:
Original issue: https://github.com/eBay/skin/issues/1698

Should be able to add the opaque modifier to textarea to allow floating label to not obscure text.
Might add it as default for textarea

PR #1784, opened 2022-10-06
feat(textbox): added transparent modifier and cleanup

agliga:
## Description
* Added opaqueLabel option to textbox

## Context
Since I touched textbox code, I cleaned up the stories, tests, and readme to match the new syntax.
Tests use the new snapshots and storybook mocks
Storybook has been cleaned up (And added missing examples)
Example directory has been cleaned up

## References
#1777

## Screenshots
Screen Shot 2022-10-06 at 9 57 46 AM


ArtBlue (commented):
The code looks good overall, but it doesn't look like I can set the floating label. 
image

agliga:
> 

Hmm, that's what it's like currently in production. 
I will add a change to the docs

agliga:
Okay, looks like just needed to add the type to floating label since we're not setting it by default. Should be fixed.

ArtBlue (approved):

Issue #1782, opened 2022-10-06
Combobox multiple options selected on keyboard input


cc-ebay:
# Bug Report

## Description

I changed the story on https://github.com/eBay/ebayui-core/blob/397db2c44b40d9e329ca764f1ffbc735417db7bf/src/components/ebay-combobox/combobox.stories.js with: 
```javascript
FloatingLabel.args = {
...
    floatingLabel: 'Default Label',
    'onSelect': function(...args) {console.log('SELECT', args);}
};

for(let i = 5000; i > 4; i--){
    FloatingLabel.args.options.push({text: i % 2 ? `Basic Offer ${i}`: `August Campaign ${i}`} )
}
```
Beside taking forever to load and being quite slow in responding to input, I get this weird multi-selection. You can see around the ned of the video.

## Screenshots

![combobox-bug-small](https://user-images.githubusercontent.com/64445558/194189522-b47a6296-602e-4349-97bb-9cb9d680fb28.gif)

cc-ebay:
In general, the logic of the `ebay-combobox` seems overly complex and optimizable. 

For example: `_getVisibleOptions()` seems to be called a lot when it could be called only when the text in the combobox change.

agliga:
I believe it's coming from there being too many options and it's probably lagging behind. 
I have seen that multi highlight thing before. We had a fix where we were clearing the selected item, but it caused other issues. This has been a known issue for a while now. We can take a look.

agliga:
> In general, the logic of the `ebay-combobox` seems overly complex and optimizable.
> 
> For example: `_getVisibleOptions()` seems to be called a lot when it could be called only when the text in the combobox change.

Unfortunately, from what I remember, this is because of the filtering logic.

cc-ebay:
> I believe it's coming from there being too many options and it's probably lagging behind. I have seen that multi highlight thing before. 

Well, it is clearly a problem that should be fixed. 

>We had a fix where we were clearing the selected item, but it caused other issues.

In my option this is a hint that the logic is "overly complex"

> This has been a known issue for a while now. We can take a look. This bug is actually a bug with makeup-active descendant not with combobox itself.

Well, regardless, it is affecting the combobox and should be fixed.

cc-ebay:
> > In general, the logic of the `ebay-combobox` seems overly complex and optimizable.
> > For example: `_getVisibleOptions()` seems to be called a lot when it could be called only when the text in the combobox change.
> 
[...]

Issue #1771, opened 2022-09-28
JS Error on ebay-combobox


cc-ebay:
# Bug Report

## Description
If we press `Esc` or `Return` in the input there is a JS error:
```
component.js:81 Uncaught TypeError: Cannot read properties of null (reading 'expanded')
    at component.js:81:1
    at handleKeydown (index.js:10:1)
    at Module.handleEnterKeydown (index.js:24:1)
    at EbayCombobox.handleComboboxKeyDown (component.js:80:1)
    at delegateEvent (event-delegation.js:65:1)
    at addEventListener.listeners.<computed> (event-delegation.js:117:1)
```

cc-ebay:
`this.expander` may be `null` (closed) in `this.expander.expanded`

PR #1785, opened 2022-10-07
fix(combobox): #1771 JS error because expander did not exist and #1782 multiple options selected on keyboard input

cc-ebay:
## Description
Fixes: 
- #1771 JS error because expander did not exist
- #1782 multiple options selected on keyboard input

### DIFF
--- a/src/components/ebay-combobox/component.js
+++ b/src/components/ebay-combobox/component.js
@@ -193,7 +193,7 @@ export default {
         this._cleanupMakeup();
     },
 
-    _setupFloatingLabel() { 
+    _setupFloatingLabel() {
         // TODO: makeup-floating-label should be updated so that we can remove the event listeners.
         // It probably makes more sense to just move this functionality into Marko though.
         if (this._floatingLabel) {
--- a/src/components/ebay-combobox/component.js
+++ b/src/components/ebay-combobox/component.js
@@ -247,7 +247,9 @@ export default {
             }
         }
 
-        this._setupFloatingLabel();
+        if (this.input.floatingLabel) {
+            this._setupFloatingLabel();
+        }
     },
 
     _cleanupMakeup() {
--- a/src/components/ebay-combobox/component.js
+++ b/src/components/ebay-combobox/component.js
@@ -198,9 +198,27 @@ export default {
         this._cleanupMakeup();
     },
 
+    _setupFloatingLabel() { 
+        // TODO: makeup-floating-label should be updated so that we can remove the event listeners.
+        // It probably makes more sense to just move this functionality into Marko though.
+        if (this._floatingLabel) {
+            this._floatingLabel.refresh();
+            this.handleFloatingLabelInit();
+        } else if (document.readyState === 'complete') {
+            if (this.el) {
+                this._floatingLabel = new FloatingLabel(this.el);
+                this.handleFloatingLabelInit();
+            }
+        } else {
+            this.subscribeTo(window).once('load', () => {
+                this._setupFloatingLabel();
+            });
+        }
+    },
+
     _setupMakeup() {
         if (this._hasVisibleOptions()) {
-            this.activeDescendant = createLinear(
+            if (this.activeDescendant) this.activeDescendant = createLinear(
                 this.el,
                 this.getEl('combobox'),
                 this.getEl('listbox'),
@@ -214,7 +232,7 @@ export default {
                 }
             );
 
-            this.expander = new Expander(this.el, {
+            if (this.expander) this.expander = new Expander(this.el, {
                 autoCollapse: !this.expanded,
                 expandOnFocus: true,
                 collapseOnFocusOut: !this.expanded && !this.input.button,
@@ -229,21 +247,8 @@ export default {
                 this.expandedChange = false;
             }
         }
-        // TODO: makeup-floating-label should be updated so that we can remove the event listeners.
-        // It probably makes more sense to just move this functionality into Marko though.
-        if (this.input.floatingLabel) {
-            if (this._floatingLabel) {
-                this._floatingLabel.refresh();
-                this.handleFloatingLabelInit();
-            } else if (document.readyState === 'complete') {
-                if (this.el) {
-                    this._floatingLabel = new FloatingLabel(this.el);
-                    this.handleFloatingLabelInit();
-                }
-            } else {
-                this.subscribeTo(window).once('load', this._setupMakeup.bind(this));
-            }
-        }
+
+        this._setupFloatingLabel();
     },
 
     _cleanupMakeup() {
--- a/src/components/ebay-combobox/component.js
+++ b/src/components/ebay-combobox/component.js
@@ -82,11 +82,7 @@ export default {
     handleComboboxKeyDown(originalEvent) {
         eventUtils.handleUpDownArrowsKeydown(originalEvent, () => {
             originalEvent.preventDefault();
-
-            if (!this.isExpanded()) {
-                this.activeDescendant.reset();
-                this.expand();
-            }
+            this.expand();
         });
 
         eventUtils.handleEnterKeydown(originalEvent, () => {
@@ -131,7 +127,6 @@ export default {
         }
 
         if (
-            this.isExpanded() &&
             !wasClickedOption &&
             !this.buttonClicked &&
             this.input.expanded !== true
@@ -218,29 +213,33 @@ export default {
 
     _setupMakeup() {
         if (this._hasVisibleOptions()) {
-            if (this.activeDescendant) this.activeDescendant = createLinear(
-                this.el,
-                this.getEl('combobox'),
-                this.getEl('listbox'),
-                '[role="option"]',
-                {
-                    activeDescendantClassName: 'combobox__option--active',
-                    autoInit: -1,
-                    autoReset: -1,
-                    axis: 'y',
-                    autoScroll: true,
-                }
-            );
+            if (!this.activeDescendant) {
+                this.activeDescendant = createLinear(
+                    this.el,
+                    this.getEl('combobox'),
+                    this.getEl('listbox'),
+                    '[role="option"]',
+                    {
+                        activeDescendantClassName: 'combobox__option--active',
+                        autoInit: -1,
+                        autoReset: -1,
+                        axis: 'y',
+                        autoScroll: true,
+                    }
+                );
+            }
 
-            if (this.expander) this.expander = new Expander(this.el, {
-                autoCollapse: !this.expanded,
-                expandOnFocus: true,
-                collapseOnFocusOut: !this.expanded && !this.input.button,
-                contentSelector: '[role="listbox"]',
-                hostSelector: '[role="combobox"]',
-                expandedClass: 'combobox--expanded',
-                simulateSpacebarClick: true,
-            });
+            if (!this.expander) {
+                this.expander = new Expander(this.el, {
+                    autoCollapse: !this.expanded,
+                    expandOnFocus: true,
+                    collapseOnFocusOut: !this.expanded && !this.input.button,
+                    contentSelector: '[role="listbox"]',
+                    hostSelector: '[role="combobox"]',
+                    expandedClass: 'combobox--expanded',
+                    simulateSpacebarClick: true,
+                });
+            }
 
             if (this.expandedChange) {
                 this.expander.expanded = this.expanded;
--- a/src/components/ebay-combobox/component.js
+++ b/src/components/ebay-combobox/component.js
@@ -15,11 +15,23 @@ export default {
     },
 
     isExpanded() {
-        return this.expander.expanded;
+        return this.expander && this.expander.expanded;
+    },
+
+    isCollapsed() {
+        return this.expander && !this.expander.expanded;
+    },
+
+    expand() {
+        if (this.isCollapsed()) {
+            this.expander.expanded = true;
+        }
     },
 
     collapse() {
-        return (this.expander.expanded = false);
+        if (this.isExpanded()) {
+            this.expander.expanded = false;
+        }
     },
 
     handleButtonClick(originalEvent) {
@@ -62,8 +74,8 @@ export default {
     },
 
     handleComboboxClick(e) {
-        if (e.target === document.activeElement && this.expander && !this.isExpanded()) {
-            this.expander.expanded = true;
+        if (e.target === document.activeElement) {
+            this.expand();
         }
     },
 
@@ -71,14 +83,14 @@ export default {
         eventUtils.handleUpDownArrowsKeydown(originalEvent, () => {
             originalEvent.preventDefault();
 
-            if (this.expander && !this.expander.expanded) {
+            if (!this.isExpanded()) {
                 this.activeDescendant.reset();
-                this.expander.expanded = true;
+                this.expand();
             }
         });
 
         eventUtils.handleEnterKeydown(originalEvent, () => {
-            if (this.expander.expanded) {
+            if (this.isExpanded()) {
                 const selectedIndex = this.activeDescendant.index;
 
                 if (selectedIndex !== -1) {
@@ -86,13 +98,13 @@ export default {
                 }
 
                 if (this.input.expanded !== true) {
-                    this.expander.expanded = false;
+                    this.collapse();
                 }
             }
         });
 
         eventUtils.handleEscapeKeydown(originalEvent, () => {
-            this.expander.expanded = false;
+            this.collapse();
         });
     },
 
@@ -103,9 +115,7 @@ export default {
                 // If we have an expander after the update
                 // that could mean that new content was made visible.
                 // We force the expander open just in case.
-                if (this.expander) {
-                    this.expander.expanded = true;
-                }
+                this.expand();
             });
             this.state.viewAllOptions = false;
 
@@ -121,13 +131,12 @@ export default {
         }
 
         if (
-            this.expander &&
-            this.expander.expanded &&
+            this.isExpanded() &&
             !wasClickedOption &&
             !this.buttonClicked &&
             this.input.expanded !== true
         ) {
-            this.expander.expanded = false;
+            this.collapse();
         }
 
         this.buttonClicked = false;


src/components/ebay-combobox/component.js
```diff
@@ -62,37 +70,37 @@ export default {
...
                 const selectedIndex = this.activeDescendant.index;
 
                 if (selectedIndex !== -1) {
                     this._setSelectedText(this._getVisibleOptions()[selectedIndex].text);
                 }
 
                 if (this.input.expanded !== true) {
-                    this.expander.expanded = false;
+                    this.collapse();
                 }
```
cc-ebay, line 98: A little tangential to this PR: I think that the parts that touch `input.expanded` should have adequate comments why they are necessary.

src/components/ebay-combobox/component.js
```diff
@@ -121,13 +127,12 @@ export default {
...
-            this.expander &&
-            this.expander.expanded &&
+            this.isExpanded() &&
             !wasClickedOption &&
             !this.buttonClicked &&
             this.input.expanded !== true
         ) {
-            this.expander.expanded = false;
+            this.collapse();
         }
```
cc-ebay, line 136: A little tangential to this PR: I think that the parts that touch `input.expanded` should have adequate comments why they are necessary.

src/components/ebay-combobox/component.js
```diff
@@ -179,17 +184,12 @@ export default {
         this._setupMakeup();
     },
 
-    onRender() {
-        if (typeof window !== 'undefined') {
-            this._cleanupMakeup();
-        }
-    },
-
```
agliga, line 187: This will not correctly clean up when the component rerenders when this triggers. 
We do not need to touch the lifecycle methods. These were implemented by the marko team and can cause issues in certain server side rendering/rerendering cases.
We should call active descendant reset on click, like I recommended earlier.

agliga (changes requested):
Couple of issues, I added comments. 

I would also add, let's only check expander in places that its causing an issue. I don't think we need to check for expander to be defined in every single case, only in the one that's breaking., which is enter and escape keydown. The rest should be left as it is.

src/components/ebay-combobox/component.js
```diff
@@ -15,11 +15,19 @@ export default {
     },
 
     isExpanded() {
-        return this.expander.expanded;
+        return this.expander && this.expander.expanded;
+    },
+
+    expand() {
+        if (this.expander && !this.expander.expanded) {
```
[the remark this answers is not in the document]
cc-ebay: Ok.

src/components/ebay-combobox/component.js
```diff
@@ -15,11 +15,19 @@ export default {
...
+            return (this.expander.expanded = true);
```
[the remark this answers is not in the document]
cc-ebay: I do not think it is necessary but I followed your previous code: https://github.com/eBay/ebayui-core/pull/1785/files#diff-3989921c7990e6038ef0b82eff0f698c230a1c327258db78bdf1c986e72614bcL22

src/components/ebay-combobox/component.js
```diff
@@ -15,11 +15,19 @@ export default {
     },
 
     isExpanded() {
-        return this.expander.expanded;
+        return this.expander && this.expander.expanded;
```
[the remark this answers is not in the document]
cc-ebay: Not sure what you mean. It does not have any collateral effects on state

src/components/ebay-combobox/component.js
cc-ebay: Can you tell me when it will not work? 
It seems it works fine. 

`_setupMakeup()` is called `onUpdate` and `onMount` and `_cleanupMakeup()` is called in it to clean before set things up again.

src/components/ebay-combobox/component.js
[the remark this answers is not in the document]
cc-ebay: If you mean change it to `get isExpanded(){ ...}`.  The use of a function makes things more explicit. 

Getter have a place but should be used everywhere. My personal feeling about getters and setters is that they make code harder to read because they look exactly like normal properly access but do something more complicated. Code is hard to read when you expect it to do one thing but it does another.

You can also read more opinions here: https://www.quora.com/Why-are-getters-and-setters-increasingly-considered-an-anti-pattern-in-programming-circles

cc-ebay:
Regarding: 
> I would also add, let's only check expander in places that its causing an issue. I don't think we need to check for expander to be defined in every single case, only in the one that's breaking., which is enter and escape keydown. The rest should be left as it is.

I disagree. Anything that can break will eventually break. If an object may not exist, its existence need to be checked. Refactoring of the code, updating dependencies, etc may change how things works and create a bug. You can easily avoid future bugs related to the existence of the `extender` by checking if it exists every time it is accessed.

In fact, if its existence was always checked before, you would not ever had bug #1771.

See also "Defensive programming".

agliga:
> Regarding:
> 
> > I would also add, let's only check expander in places that its causing an issue. I don't think we need to check for expander to be defined in every single case, only in the one that's breaking., which is enter and escape keydown. The rest should be left as it is.
> 
> I disagree. Anything that can break will eventually break. If an object may not exist, its existence needs to be checked. Refactoring of the code, updating dependencies, etc may change how things works and create a bug. You can easily avoid future bugs related to the existence of the `extender` by checking if it exists every time it is accessed.
> 
> In fact, if its existence was always checked before, you would not ever had bug #1771.
> 
> See also "Defensive programming".

So then why is expander not defined? I think we should figure that out first. It should always be set.

src/components/ebay-combobox/component.js
agliga: The thing is the previous call is not being cleared correctly because the html does not exist anymore. 
we have this pattern in all our other components. Not going to deviate from that. 
lets add that reset when we open the dropdown. Is that causing issues or not working?

agliga:
Also please follow the code conventions of this repo and not what some random article says. 
If you have suggestions we can take it offline and refactor if we need to. But we are not deviating from the way our code is currently setup.

cc-ebay:
> So then why is expander not defined? I think we should figure that out first. It should always be set.

In the specific here:
- `this.expander` is created in `_setupMakeup` that is called on `onCreate`, so there is a part of lifecycle in whick the component exist and `this.expander` is not defined.
-   In ` _cleanupMakeup()` you have `this.activeDescendant = null`

When you have `this.a.b` in dynamic language like Javascript checking the existence of `this.a` is always a good practice.  The `?.` operator was introduced because it is very common to check the existence of "parent objects".  When you write a piece of code is a good practice not only check for what is possible in the "context code" at the time the piece of code is written (it is also easy to miss some complex side effect), but for the future code that is changed because of maintenance.
 
What is the cost of checking ? Probably a micro second and less than ~1 byte is a gzipped JS.
What is the payoff of checking? Present and future proof robust code that will avoid bugs.

cc-ebay:
> Also please follow the code conventions of this repo and not what some random article says. If you have suggestions we can take it offline and refactor if we need to. But we are not deviating from the way our code is currently setup.

The `isExpanded()` was already present in the code and it was not a getter. I do not want to pejorate the readability of apiece of code that is already hard to understand.

I did not link something about the "getter and setter anti-pattern" so I do not have to write an article, not because I read it in that article but because you can read in detail why a getter here may not be a great idea. There is plenty of articles that tell you that it is not a good idea to use `getters` and `setters` when you can use one:

Screen Shot 2022-10-07 at 4 40 00 PM

agliga:
> > So then why is expander not defined? I think we should figure that out first. It should always be set.
> 
> In the specific here:
> 
> * `this.expander` is created in `_setupMakeup` that is **not** called on `onCreate`,  so there is a part of lifecycle in which the component exists and `this.expander` is not defined.
> * In ` _cleanupMakeup()` you have `this.activeDescendant = null`
> 
> When you have `this.a.b` in a dynamic language like Javascript checking the existence of `this.a` is always a good practice. The `?.` operator was introduced because it is very common to check the existence of "parent objects". When you write a piece of code, it is easy to miss some complex side effect that change a variable. But even more important, the code should be robust against future changes.
> 
> What is the cost of checking ? Probably a micro second and less than ~1 byte is a gzipped JS. What is the payoff of checking? Present and future proof robust code that will avoid bugs.

Actually I disagree. It creates a huge readability issue and complex if statements. If we have access to the `?` in js we can use that but it is not well supported yet from what I know.

Digging deeper, this console error is only caused when there are no visible items in the dropdown portion. Maybe we need to figure that out and address that issue.

agliga:
> > Also please follow the code conventions of this repo and not what some random article says. If you have suggestions we can take it offline and refactor if we need to. But we are not deviating from the way our code is currently setup.
> 
> The `isExpanded()` was **already present** in the code and it was **not** a **getter**. I do not want to pejorate the readability of a piece of code that is already hard to understand. Furthermore this is a breaking change since some code can accesses `isExpanded` as a function.
> 
> I did not link something about the "getter and setter anti-pattern" so I do not have to write an article, not because I read it in that link but because you can read in detail why a getter here may not be a great idea. There is plenty of articles that tell you that it is not a good idea to use `getters` and `setters` just because you can use one:
> 
> Screen Shot 2022-10-07 at 4 40 00 PM

We have been slowly changing these kind of methods to getters mainly for code readability. This is our decision as owners of this repo. 

That said, we can leave it as a method since it was like that initially.

src/components/ebay-combobox/component.js
```diff
@@ -15,11 +15,19 @@ export default {
...
+    },
+
+    expand() {
+        if (!this.isExpanded()) {
```
agliga, line 22: Actually, we might need an isCollapsed method since this can evaluate to true if expander is null.

ianmcburnie:
> a getter here may not be a great idea

https://github.com/makeup/makeup-js/blob/master/packages/makeup-expander/src/index.js#L225

Why is a getter here not a great idea? Serious question. The getter encapsulates/abstracts the specific ARIA attribute name on the DOM element. There is no access to a private variable. If the ARIA attribute changes under the hood, a boolean is still returned and the external code is protected. Sure, it could be written as `isExpanded()` (as it was originally) which might be a little less "deceptive" for those that believe getters should always access private variables.

So, is the getter here really causing any harm? Serious question.

I wouldn't disagree that getters and setters can be an anti-pattern, but they are not ALWAYS an anti-pattern, and do not always cause harm. Instead of routinely linking to quick google searches of "software best practices", which anybody could do and, quite frankly, can be considered condescending to some, why not take the time to say why YOU think they are an anti-pattern in the CONTEXT of the code we are working with. To me that would lead to a far more respectful and constructive dialog and better use of everybody's time.

cc-ebay:
> Instead of routinely linking to quick google searches of "software best practices", which anybody could do and, quite frankly, can be considered condescending to some, why not take the time to say why YOU think they are an anti-pattern in the CONTEXT of the code we are working with. To me that would lead to a far more respectful and constructive dialog and better use of everybody's time.

I actually gave my quick explanation at the [very beginning here](https://github.com/eBay/ebayui-core/pull/1785#discussion_r990534930):
> ... getters and setters ... make code harder to read because they look exactly like **normal properly access** but **do something more complicated**. Code is hard to read when you expect it to do one thing but it does another.

I can add that in real work situations, it is already hard trying to fix a bug looking at mutation of state of multiple objects involved. There is no reason to make harder to understand what really is going on by _camouflaging a function into a simple property access or assignment_. In C++ overriding math operators (`+`, `-`, etc) has been critizied for years for the same reasons and the recommendation is not to use this mechanism as a shortcut to call a function.

In that comment I tried to keep my response short and added one link as a way to learn more about `getters`/`setters`. When someone tell me something I do not completely grasp or disagree with, I normally google and skim at least 2-3 results.
It is easier to direct someone to read some articles instead of spending a few hours to write something to the extent of a "blog article" as an explanation. But here we are...

> Why is a getter here not a great idea?

 "If you can easily avoid using a `getters/setters`, avoid it" is both a simple corollary to the good philosophy "everything should be as simple as possible..." and it is also a way to prevent bug-proned code and unnecessary headaches.

 The only time it is _necessary_ to use `getters`/`setters` is to debug something, to build debugging tools or to inspect the access patterns of what is normally used as a normal property. For example if there is a bug related to a change in a property but it not clear what is changing that variable I can replace that property with a `setter/getter` with some logic and/or a `console.trace()` to find what is goign on.

 Specifically in the setter of [`expanded` in `makeup-js`](https://github.com/makeup/makeup-js/blob/master/packages/makeup-expander/src/index.js#L228) has so many collateral effects. I feel very uncomfortable with what look like a simple assignment to a property like `this.expander.expanded = true` really does. It changes a lot of internal states (`this.*`) and has a no so straightforward logic at the top that depends on the input of the expander, etc. **Nothing** in the `ebay-combobox` code or any other component that use it **tells you that there is a `setter` code that will run**. You need to be familiar with that code and _everytime_ you see it think "oh, we are calling the setter" that add cognitive overhead.

## Longer explanation of why `getters` and `setters` are always always harmful

### Broken assumptions

In both _writing code_ or _debugging a bug_, there are a few **assumptions** that our brain normally makes when we read an assignment (`obj.prop = 5`) and/or accessing a property (`aProp = obj.prop`). This is a not exaustive list of some of important assumptions that are **not guaranteed anymore** when we use `getters` and `setters`:

  1.  The only way `obj.prop = 5` or `obj.prop` can raise a runtime error is if `obj` is `undefined` or `null`. When `setters`/`getters` are involved there are multiple ways how an exception can be raised at runtime depending on the code. Now you will have to `try/catch` code that you normally do not expect needing it. More likely you are goign to be surprise with a runtime error.
  
  2.  You normally expect that a `obj.prop = 5;console.log(obj.prop)` will print `5` but it could print anything depending on what the code in the setters and getters do. They may introduce any kind of asymmetries between assignments and accessing the assigned value.

  3.  You normally expect that `obj.prop = 5` will only change one property (`prop`) in the state of the `obj` but a `setter` can change many things inside the object or any other variable in the scope of the `setter`/`getter`.

  4.  You normally expect that `obj.prop` will not change anything (especially the state of `obj`) but a `getter` can change everything it wants inside the object or any other variables in the scope of the `setter`/`getter`.

  5.  You normally expect that `obj.prop = 5` or `obj.prop` will take a very small amount of time but in fact it is not guaranteed that they will not take minutes or get in an infinite loop.

It is very useful to have the certainty that those assumptions are not violated.
In general I know when there is a **function call** easily recognizable by the `()` anything can happen and I am prepared. Since I cannot assume anymore that these general assumptions will be always statisfied, I will have to keep asking the question: "does this code do what it looks it does?". This adds a new level of complexity in reading code and make the **code more prone to bugs**.

### What I mean by getter and setter

I want to  be extra clear and make sure we are on that same page. In JS people sometimes use the term `getter` and `setter` for 3 different things:

1. By `getter` and `setter` as specified in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set - **This is what I specifically mean**.

2. `Object.defineProperty(obj, 'value', { get: ...})` may also be called `getter`. This is another form of a `getter` that behaves mostly similarly but sometime differently. It can also break the assumptions above so it is a good idea to **avoid this too**.

3. A property function that return an internal variable like this:

```
class MyClass {
   _myPrivateValue = 0,
   getValue = () => _myPrivateValue;
}
const obj = new MyClass();
console value = obj.getValue();
console.log(value)
```

`getValue` is also sometimes called a _"getter" function_. And this is a **GOOD practice** . Accessing directly `obj._myPrivateValue` would be a bad practice because it accesses the internal representation of the class.

### Advantages

#### What is the advantage of a `getter`?

1. A shortcut for not calling a function. In practice you do not have to write 2 characters (the 2 parenthesis `()`) to call the _getter function_.
2. Run some code when a property is accessed. It is good when you need to alter or inspect the beghavior of an existing app.
3. ??

#### What is the advantage of a `setter`?

1. Maybe writing `obj.value = newValue` instead of `obj.setValue(newValue)`?
2. Run some code when a property is assigned. It is good when you need to alter or inspect the behavior of an existing app.
3. ???

### Disadvantages

1. You CANNOT tell from reading a `obj.value = newValue` that you are actually calling a function that can do something more complex. In order to find out, you need to run the code step by step or you need to go search (often in other files) what that property really is. You would have to be very familiar with the code but if the code has a few getter and setter, it may be hard to remember what properties use `getters` and `setters`. Regardless **`getters` and `setters` add unnecessary cognitive overhead**. Is it not simpler to just use function calling (`f()`) syntax?

2. **A `getter` without a `setter` are a VERY bad idea** in JS because an assignment on that property will **silently** fail:

Screen Shot 2022-10-10 at 1 01 28 AM

You may think `obj.aValue = 7` would assign 7 but since there is no `setter` nothing is set. It still returns `7`  but `obj.aValue` is still `5`. And again, **no errors**.

3. `getters` and `setters` end up **creating confusing aliases**. Aliases add more unnecessary entropy/complexity. For example in this PR `this.isExpanded` would be _almost_ an alias of `this.expander.expanded` (more precisely of `this.expander && this.expander.expanded`). A user of the `ebay-combobox` may be expecting to be able to assign something to `.isExpanded` to collapse or expand the listbox. Currently since `isExpanded` does not have a `setter`, this would be just an example of (2) and an assignment would be **silently ignored**. So maybe we may want to create a `setter` for `.isExpanded`. But what are we going to do if `this.expander` is a falsy?
We have now 2 problems:
   a. `.isExpanded = true` does not always work. _What seems an alias is not really an exact alias_.
   b.  The `ebay-combobox` code can now change the `this.expander.expanded` in 2 ways: changing it directly or changing `.isExpanded`. Noting prevent us to use both in the code. If you need to find every time that the dropdown is expanded you need to _first realize that there is this "alias"_ (good luck) and then running a search for both strings. And of course you need to figure out how subtly different they may work when `this.expander = null`.

4. Nothing prevents a `getter` from mutating some internal state breaking assumption (4) above. Hopefully nobody is so evil to do that. But "if something can go wrong, it will". Now you may have some code that looks like only reading from an object but it is actually creating collateral effects in that object. Just another way things can get confusing.

5. Starting with few `getters` and `setters` will lead to more of them. That will lead to more cognitive overload. When you open up to a style that tollerate `getter` and `setter`, you end up pretty quickly to a point when you cannot trust a `obj.method` not to be something more complex. Good luck! Sooner than later will be debugging some code and being tricked into thinking it does something differnt than what it really does. You will be banging your head on the keyboard until you surrender to start over and checking each property access and assignment for a disguised function that violate one of the important assumptions above.

### Conclusions

I see no real advantage in application code in using `getters` annd `setters` and a lot of pretty bad disadvantages. So I think it is quite evident that the best policy is "if you can pretty easily avoid using a `getters/setters`, avoid it".

src/components/ebay-combobox/component.js
cc-ebay: > We have been slowly changing these kind of methods to getters mainly for code readability. This is our decision as owners of this repo.

What does "code readability" mean for the owner of the repo?

For me it is "easily understand what the code does".  And `setters` and `getters` make a lot hard to understand what the code really does. See https://github.com/eBay/ebayui-core/pull/1785#issuecomment-1273979641

agliga:
## In conclusion

We are the owners of this repo. We have decided that's the path we are going to go. We appreciate your input but as we discussed offline, we will stick with this. 
That's our choice. Since you are an external contributor, you should be respectful to our decisions and follow the guidelines of the current repo.

## Final notes
I'm tempted to close this PR because it is not following another guidelines we set for this repo, specifically: https://github.com/eBay/ebayui-core/blob/master/CONTRIBUTING.md#before-you-start 
(Pasted it here for convenience)
```
If you are creating a new component or significantly altering an existing one, please ensure that an issue has been created before you start any work, and that it has been assigned to you. The same goes for any big new additions or changes to the project structure, CI or documentation.

We ask this because:

we want to avoid cases where developers build something that does not align with our wants & needs
we want to be able to carefully plan our sprint and test cycles with minimal disruption
we want to avoid cases where two developers duplicate work
```
In fact, one of these issues was already assigned to our team member and he was planning on working on it, so that's a waste of resources

## What to do
I will give you one more chance, please update the code based on the review comments. This is counterproductive to both our time and yours.

src/components/ebay-combobox/component.js
cc-ebay: But that pattern is what creates the #1782 bugs. And it may always be the cause of others.

## A better pattern to guarantee clean up and resource availability

When you have an resource that need to be available during the lifecycle of a object you usually have this pattern:

```
  create() {
    this.obj = new Resource();
  }

   destroy() {
        this.obj.destroy();
   }
``` 

### Recreate
**FOR EACH each setup, you want ONE (and ONLY ONE) cleanup of the resource** so the chain look like: 
```
setup -> cleanup -> setup -> cleanup -> ... -> cleanup
```


If every once in a while you need to destroy the resource (cleaning its side effects) before a new one is created, having **one**  place where the destruction and (re)building are done together in order is the **simplest way** to guarantee that the resource will be **always** available and each setup will have one cleanup: 
```
  create() {
    this.obj = new Resource();
  }

  recreate() { // Keep them together without gaps!
      if (this.obj) this.obj.destroy();
      this.obj = new Resource()
   }
   
   destroy() {
        if (this.obj) this.obj.destroy();
   }
``` 

Or in other more general sudocode:

```
  create() {
    setup();
  }

  recreate() {
      cleanup();
      setup();
  }
   
  destroy() {
       cleanup();
  }
``` 

Since the actions are synchronous a `cleanup()` and a `setup()` run in `recreate()` **without gaps**.  In Javascript you are guaranteed that the resource reference will always be a valid object. It also impossible to overcall or undercall the `setup` or the `cleanup`.*
To make things even more DRY, you can have a `cleanupIfNecessaryAndSetup()` instead so there are only 2 functions. 

```
cleanupIfNecessaryAndSetup() {
   if (this.resource) cleanup();
   setup(); // This will set this.resource
}
...
{
  create() {
    cleanupIfNecessaryAndSetup();
  }

  recreate() {
      cleanupIfNecessaryAndSetup();
  }
   
  destroy() {
       cleanup();
  }
}
``` 

A **problem** can like occur instead if you break down the `recreate` in 2 methods. You are not guaranteed anymore that there is no gap (ex: callers are asynchronous) or `setup` and `cleanup` are not overcalled or undercalled.

This is what I did here in this PR moving the `this._cleanupMakeup()` inside the `_setupMakeup()`. So the `_setupMakeup` became the `cleanupIfNecessaryAndSetup()` above:
```
onMount() {
     cleanupIfNecessaryAndSetup()
     ...
}

onUpdate() {
     cleanupIfNecessaryAndSetup()
     ...
}

onDestroy() {
     cleanup()
     ...
}
```

These are the [Marko lifecycle](https://markojs.com/docs/class-components/#lifecycle-events) flows:
```
First render
emit('create') → emit('input') → emit('render') → emit('mount')

New input
emit('input') → emit('render') → emit('update')

Internal state change
emit('render') → emit('update')

Destroy
emit("destroy");
```

And each of the last actions in the chain (`mount`, `update`, `update` and `destroy`)  has the right call in it.

Now if you keep this structure the code is robust against changes and it is impossible to have a bug related to the call timing.

cc-ebay:
@agliga : 
1. Regarding your comment about the `onRender`, I am sure what you would like me to do because that is the fix of #1782, part of the fix of # the code work well as it is (see my last comemnt that explain what I did https://github.com/eBay/ebayui-core/pull/1785#discussion_r991790978 )

2. Regarding `this.expander && this.expander.expanded` you wrote: 
> Actually I disagree. It creates a huge readability issue and complex if statements. If we have access to the `?.` in js we can use that but it is not well supported yet from what I know.

I think you can actually just use `?.`.  I did not use it because the code was actually not using it and was exactly using [this.expander && !this.expander.expanded](https://github.com/eBay/ebayui-core/pull/1785/files#diff-3989921c7990e6038ef0b82eff0f698c230a1c327258db78bdf1c986e72614bcL74) in 2 places. A standard setup of `marko` with `webpack` or `lasso@4` will transpile it.  So using it is a quick change here.

But I think the code already improved. It is actually only used in one place instead of 2 like it was before. And frankly any JS developer that has been around in the last 4 years finds and write pretty often: `a && a.b && a.b.c` etc. It is not hard to understand what it does.

ianmcburnie:
> Specifically in the setter of expanded in makeup-js has so many collateral effects.

As a reminder, the whole *purpose* of makeupjs is that it is an imperative style library that touches the DOM and adds the appropriate ARIA roles, states and properties. There is no getting away from the fact that whether we call `expander.expanded = true` or `expander.expand()` a bunch of things are going to happen to the DOM (including the firing of events). That is the very purpose of this module.

As mentioned in another thread, we are aware that the makeupjs way of working is not a seamless match for Marko, and that is why a Marko version of Makeup is on the cards.

As far as getters/setters go, philosophically I'm with you (didn't need a novel for that) - but it's not like they are being grossly abused in this context. As maintainers of the code we decided we were happy with reducing the previous [3 methods](https://github.com/makeup/makeup-js/tree/master/packages/makeup-expander#methods) down to a getter/setter. I'd be just as happy with the previous 3 methods too. Both ways are fine here in this very simple context. It just doesn't seem worth squabbling over or writing a novel about. It's not like an entire, large library of code has been littered with gross misuse of the getter setter/pattern (which by the way, as software engineers we *are* familiar with).

Sure, it's not going to win any awards, but we find [this code](https://github.com/makeup/makeup-js/blob/master/packages/makeup-expander/src/index.js#L224) short and easy enough to follow. I'm sorry you find it difficult.

src/components/ebay-combobox/component.js
DylanPiercey: @cc-ebay without the full context of reading everything here:

The main issue is that although makeup follows a sane lifecycle, because it's an imperative API that touches the DOM in many places there is awkwardness here. Specifically, every time Marko renders the Makeup changes need to be _undone_ in order for Marko to properly complete it's update.

I believe it was mentioned in here somewhere that the ultimate goal is to recreate makeup or big portions of it as components to avoid this issue, but for now this is what we've got.

agliga: Just to add, spoke to Dylan as well

We need to keep the onRender cleanup because by the time we hit onUpdate, the html has already been removed and there's nothing to cleanup. And we don't want to call cleanup twice. 
I think there's more to this that's causing this bug. It looks like the previous cleanup might not be resetting correctly or such. 
By adding that reset as I recommended, should resolve this issue while allowing the cleanup to happen properly as before.

cc-ebay: Ok. I understand that this could be a better idea to split the `reload` in 2. Then we are not guaranteed anymore that the resources will be always available.

cc-ebay: I also found the root cause of https://github.com/eBay/ebayui-core/issues/1782:
 https://github.com/eBay/ebayui-core/blob/5f1e819603d7776beb03083d374971a3b7e12456/src/components/ebay-combobox/component.js#L235
 
 Call again the `_setupMakeup()` so it is called twice on load. That setup 2 `activeDescendant` without cleaning in the middle.
 
Since the `_setupMakeup()` here is setting up 3 resources (`activeDescendant`, `expander`, `_floatingLabel`), it would be good to guard the actual call of the setup function for each resources checking for existence. Ex:
 ```
 if (this.activeDescendant) this.activeDescendant = createLinear(..
 ...
 if (!this.expander) this.expander = new Expander(...
 ...
 ```
 
 The `_floatingLabel` could also move in its own function (ex:  `_setupFloatingLabel`) and instead of calling the whole  `_setupMakeup()` on the `load` event, can call itself.

src/components/ebay-combobox/component.js
cc-ebay: I realized that the last comment is in the wrong section.

cc-ebay: You are right that there is a problem here. 

Adding `isCollapsed` works to prevent accessing `this.expander` when it is `null`.

agliga (commented):
I think it's much better now. I think it's close to being able to merged. There are some code review/cleanup left to do.
Can you please target `10.1.0` branch with this?

I still think it can be simplified as a getter/setter so it matches makeup, but if you really don't want to do it, we can clean it up later.

src/components/ebay-combobox/component.js
```diff
@@ -62,37 +74,37 @@ export default {
...
         }
     },
 
     handleComboboxKeyDown(originalEvent) {
         eventUtils.handleUpDownArrowsKeydown(originalEvent, () => {
             originalEvent.preventDefault();
 
-            if (this.expander && !this.expander.expanded) {
+            if (!this.isExpanded()) {
                 this.activeDescendant.reset();
```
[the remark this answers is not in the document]
cc-ebay: It seems to work fine

agliga (commented):
Overall looks fine to merge codewise. I’m going to try it out locally and then I’ll approve

agliga (commented):
@cc-ebay 
So tried it out locally. Looks good.

I think there is a lint error. I ran `npm run build` locally and got this error:

```
File: ebayui-core/src/components/ebay-combobox/component.js
Line: 196 Unexpected trailing spaces found. [warning]
```

Can you make sure `npm run build` successfully passes. I can't run the checks on the PR since it's being run on your fork (but I'm fine merging it without the checks if it works locally, we can fix things later.)

src/components/ebay-combobox/component.js
```diff
@@ -189,52 +193,61 @@ export default {
...
+        } else if (document.readyState === 'complete') {
+            if (this.el) {
+                this._floatingLabel = new FloatingLabel(this.el);
+                this.handleFloatingLabelInit();
+            }
+        } else {
+            this.subscribeTo(window).once('load', () => {
+                this._setupFloatingLabel();
+            });
+        }
```
agliga, line 211: I see a test failure on browser tests.
This whole thing needs to be wrapped in `if (this.input.floatingLabel) {` otherwise it runs each time even if there is no floating label. 
We can either wrap this whole thing in the if statement or the function call in `_setupMakeup`

cc-ebay: Yep. I got lost in the refactoring. I put back in. Glad the tests caught it.

agliga:
@cc-ebay 
Still getting a lint error on like 196.
```
File: ebayui-core/src/components/ebay-combobox/component.js
Line: 196 Unexpected trailing spaces found. [warning]
```
Seems like there's an extra space after the brace `_setupFloatingLabel() { `
Once that's fixed I'll merge.

agliga (approved):

Issue #1764, opened 2022-09-20
tourtip: new guided module


agliga:
See https://github.com/eBay/skin/issues/1744

PR #1787, opened 2022-10-11
feat(tourtip): added footer

agliga:
## Description
Added footer to tourtip/base

## Context
* Added an index attribute to `@footer`. This will display the index values

## References
https://github.com/eBay/ebayui-core/issues/1764

## Screenshots
Screen Shot 2022-10-11 at 8 54 58 AM


ArtBlue (commented):
Code looks good. Just one thing - 

image

Can we allow for the setting of the value of `@footer` similar to `@heading` and `@content`? It would be especially helpful to allow people to quickly see what that variation looks like.

agliga:
> 

It should be available for the footer one
Screen Shot 2022-10-19 at 9 30 12 AM

I can add a default type to each of the args which would then allow them to be edited, but then users would have to type in `renderBody` and such to the object

ArtBlue:
Nice! DIdn't see it.

Issue #1772, opened 2022-09-29
ebay-filter: add support for "useAriaPressed" (default: true)


melquan:
# Feature Request
Add support for `useAriaPressed` for `ebay-filter`.  Default can be `true`.

## Description
We have a usecase where the `ebay-filter` is being used to open a menu so we do not need the `aria-pressed` set on the element.

PR #1791, opened 2022-10-14
feat(filter): added support to use aria-pressed

saiponnada:
## Description
Added a configuration `useAriaPressed` which defaults to `true`. This config can be set to `false` where using `aria-pressed` is inappropriate. 

## References
#1772 

## Screenshots

When `selected` and `useAriaPressed` set to `true`

image

When `selected` and `useAriaPressed` set to `false`

image

### DIFF
--- a/src/components/ebay-filter/index.marko
+++ b/src/components/ebay-filter/index.marko
@@ -2,7 +2,8 @@ import { processHtmlAttributes } from "../../common/html-attributes"
 
 static var ignoredAttributes = [
     "selected",
-    "a11ySelectedText"
+    "a11ySelectedText",
+    "useAriaPressed"
 ];
 
 $ var baseClass = input.href ? "filter-link" : "filter-button";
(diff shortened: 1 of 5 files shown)


src/components/ebay-filter/test/mock/index.js
```diff
@@ -2,12 +2,17 @@ import { createRenderBody } from '../../../../common/test-utils/shared';
...
 export const Basic = {
     renderBody: createRenderBody('text'),
+    useAriaPressed: true,
 };
 
 export const Selected = Object.assign({}, Basic, {
     selected: true,
 });
 
+export const withOutPressed = Object.assign({}, Selected, {
```
saiponnada, line 12: Is there a better way we can rename this?

src/components/ebay-filter/index.marko
```diff
@@ -2,20 +2,22 @@ import { processHtmlAttributes } from "../../common/html-attributes"
 
 static var ignoredAttributes = [
     "selected",
-    "a11ySelectedText"
+    "a11ySelectedText",
+    "useAriaPressed"
 ];
 
 $ var baseClass = input.href ? "filter-link" : "filter-button";
+$ var a11yAttributes = input.useAriaPressed ? { "aria-pressed": (!input.href && state.selected && "true") } : "";
```
agliga, line 10: We should make it like this:
`aria-pressed=(input.userAriaPressed !== false && !input.href && state.selected && "true")`

I'd also not put it as a variable and just pass it directly in the component.

agliga (commented):
We also need to change the `marko.json` to take `@useAriaPressed` so we can pass it as `use-aria-pressed`

agliga (approved):

ArtBlue (approved):

PR #1793, opened 2022-10-18
fix(filter-menu): disabled state and tests

agliga:
## Description
* Filter-menu disabled would work to grey out the item, but it was still clickable and would trigger the change event. Added a check to disable it
* Part of that, I went ahead and added tests to check for disabled states and that the events are not getting fired. This led me to also redo the tests using storybook mocks instead of mock/index.js

## References
This is part 2 of https://github.com/eBay/ebayui-core/pull/1760


ianmcburnie:
Only thing I notice is that `disabled` is also being rendered in the HTML:

Screen Shot 2022-10-20 at 8 57 13 PM

ArtBlue (approved):
The visual pieces look good.

agliga:
> Only thing I notice is that `disabled` is also being rendered in the HTML:
> 
> Screen Shot 2022-10-20 at 8 57 13 PM

Fixed in the latest push. Ill go ahead and merge after tests pass.

PR #1801, opened 2022-10-20
feat(date-picker): add date picker to ebayui-core

### LINKED ISSUE #1560: date-picker: move existing Terrapeak date picker under eBayUI project

LuLaValva:
- Resolves #1560

Added date picker to `ebayui-core`

## Prerequisites
- This must be tested in tandem with ebay/skin#1907
- `npm link` to skin on the `16.0.0` branch

## Description

- Full functionality for date picker has been moved to coreui
- Keyboard functionality is not yet implemented, it will likely be implemented into `makeup-js`. To be discussed offline.


src/components/components/ebay-date-picker-month/index.marko
```diff
@@ -0,0 +1,71 @@
...
+
+        <for|row| of=dateGrid>
+            <div role='row'>
+                <for|date| of=row>
+                    <if(date === undefined)>
+                        <div.date-picker__day--empty/>
+                    </if>
+                    <else>
+                        $ const timestamp = date.getTime();
+                        <div.date-picker__day
```
PrashantAshok, line 31: It would be good to move `day` specific logic to a child component?
`ebay-date-picker-day`

PrashantAshok:
Great work!. Are you trying to make sure the `ebay-date-picker` is drop in replacement (almost?) for `marko-date-picker`? It would be great if it is, would be easier for teams to adopt.

LuLaValva:
Thanks for the review @PrashantAshok, but I hope you didn't spend too much time on it 😬... we recently finished up changes on the `skin` side which ended up facilitating a big refactor here.

The primary goal of this component is to be a drop-in replacement for `marko-date-picker`, with updates to provide modern styles and better accessibility support. However, we may change a few of the attributes so it probably won't be _exactly_ one-to-one.

src/components/ebay-calendar/calendar.stories.js
```diff
@@ -0,0 +1,120 @@
...
+                defaultValue: {
+                    summary: 'undefined',
+                },
+            },
+        },
+    },
+};
+
+export const Standard = (args) => ({ input: args });
+Standard.parameters = {
```
agliga, line 114: There's an easier way to do this. 
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-button/button.stories.js#L220

src/components/ebay-calendar/calendar.stories.js
```diff
@@ -0,0 +1,120 @@
...
+            description:
+                'List of specific days that are disabled. Should be a list of date objects, but also accepts timestamps or ISO strings',
+            table: {
+                defaultValue: {
+                    summary: 'undefined',
+                },
+            },
+        },
+    },
+};
```
agliga, line 111: I don't see any events attached as args. We should attach them and document all events.

src/components/ebay-calendar/examples/default.marko
```diff
@@ -0,0 +1 @@
+<ebay-calendar ...input/>
```
agliga, line 1: All events should be emitted up one level from here so we can show them on storybook.

src/components/ebay-calendar/index.marko
```diff
@@ -0,0 +1,110 @@
...
+$ const monthDates = [...Array(numMonths)].map((_, i) => component.getMonthDate(state.offset + i))
+
+<div.calendar>
+    <if(input.navigable)>
+        <div.calendar__header>
+            <div.calendar__header--inner>
+                <ebay-icon-button
+                    transparent size="small"
+                    disabled=(state.disableBefore && component.getFirstVisibleISO() <= state.disableBefore)
+                    aria-label=`Show ${component.monthTitle(component.getMonthDate(state.offset - 1))}`
```
agliga, line 14: This needs to be configurable for localization. Maybe have it as `a11yPrevMonthText`

src/components/ebay-calendar/index.marko
```diff
@@ -0,0 +1,110 @@
...
+                    onClick('prevMonth')>
+                    <ebay-chevron-left-icon/>
+                </ebay-icon-button>
+                <for|monthDate| of=monthDates>
+                    <h3>${component.monthTitle(monthDate)}</h3>
+                </for>
+                <ebay-icon-button
+                    transparent size="small"
+                    disabled=(state.disableAfter && component.getLastVisibleISO() >= state.disableAfter)
+                    aria-label=`Show ${component.monthTitle(component.getMonthDate(state.offset + numMonths))}`
```
agliga, line 24: Same as above, have it as `a11yNextMonthText`

src/components/ebay-calendar/index.marko
```diff
@@ -0,0 +1,110 @@
+// @ts-check
+import {toISO} from "./component"
+
+$ const numMonths = input.numMonths || 1;
+$ const monthDates = [...Array(numMonths)].map((_, i) => component.getMonthDate(state.offset + i))
```
agliga, line 5: Can this only be 1 or 2? Or more?

src/components/ebay-calendar/index.marko
```diff
@@ -0,0 +1,110 @@
...
+                                    $ startDate = 1;
+                                </if>
+                                <else-if(endDate > daysInMonth)>
+                                    $ endDate = daysInMonth;
+                                </else-if>
+                                <for|day| from=startDate to=endDate>
+                                    $ const dayISO = toISO(new Date(year, month, day));
+                                    $ const isRangeStart = dayISO === state.rangeStart;
+                                    $ const isInRange = (state.rangeStart && state.rangeEnd
+                                                && dayISO >= state.rangeStart && dayISO <= state.rangeEnd);
```
agliga, line 63: This logic is a bit confusing and could possibly be very brittle. 
Having `&&` mixed with `>` like this is hard to understand. 
Could be better in a common method in the component where it has each of these statements line by line.

package.json
```diff
@@ -112,7 +112,7 @@
     "less-loader": "^11",
     "lint-staged": "^13.0.2",
     "lintspaces-cli": "^0.8.0",
-    "marko": "^5",
+    "marko": "^5.23.0",
```
agliga, line 115: Leave this as version 5, update only package.json. 

Is this upgrade needed?
Also do we need a peer depdency of marko or not necessarily?

src/components/ebay-calendar/index.marko
```diff
@@ -0,0 +1,110 @@
...
+                                        aria-selected=(
+                                            Array.isArray(input.selected)
+                                                ? input.selected.some((iso) => iso === dayISO)
+                                                : input.selected && input.selected === dayISO
+                                        ) && "true">
+                                        <${input.interactive ? "button" : "span"}
+                                            key=dayISO
+                                            class=input.interactive && disabled && "calendar__cell--disabled"
+                                            disabled=disabled
+                                            aria-label=(
```
agliga, line 82: Again, this needs to be thought out for localization.

src/components/ebay-calendar/component.js
```diff
@@ -0,0 +1,383 @@
...
+}
+
+/**
+ * @param {string | undefined} locale
+ */
+export function localeOverride(locale) {
+    if (locale) return locale;
+
+    return navigator.language;
+}
```
agliga, line 383: I'm tempted to have these in a helper.js file, which we can import from component and in the template. Just for organization.

src/components/ebay-calendar/component.js
```diff
@@ -0,0 +1,383 @@
...
+        }
+        this.emit('keydown', event);
+    }
+
+    /**
+     * @param {number} offset
+     */
+    getMonthDate(offset) {
+        const baseDate = fromISO(this.state.baseISO);
+        return new Date(baseDate.getFullYear(), baseDate.getMonth() + offset);
```
agliga, line 187: I see this kind of `new Date()` method being used a lot in similar ways. Maybe we can have a helper function which calculates what we need
Like
`generateDate(baseDate, {month: 1})` which increments month by 1,
`generateDate(baseDate, {day: 1})` which increments day by 1,
etc? Thoughts?

agliga (commented):
Initial reivew. Seems a good start. Couple of comments needed

I looked a bit at the datepicker input one, but some comments should apply for there too

Ideally I'd like to have a bit more comments on what each method does. It can be a bit hard to follow sometimes.

LuLaValva (commented):
Thanks for the quick review Andrew, much appreciated.

src/components/ebay-calendar/component.js
LuLaValva: Ok, I ended up just moving it into `helper.js` in `calendar`

src/components/ebay-calendar/calendar.stories.js
```diff
@@ -0,0 +1,150 @@
+import Readme from './README.md';
```
ArtBlue, line 1: This is throwing an error. I did an `npm i` before running. 
image

src/components/ebay-calendar/component.js
```diff
@@ -0,0 +1,316 @@
...
+        if (input.selected && input.range) {
+            // Determine range display (state.rangeStart-state.rangeEnd)
+            let iso1, iso2;
+            if (Array.isArray(input.selected)) {
+                [iso1, iso2] = input.selected;
+            } else if (this.state.focusISO) {
+                iso1 = input.selected;
+                iso2 = this.state.focusISO;
+            }
+            if (iso1 && iso2) {
```
ArtBlue, line 303: Some comments here to describe what is happening would be helpful.

src/components/ebay-calendar/component.js
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /**
+     * @param {DayISO} iso
+     */
+    setTabindexAndFocus(iso) {
+        this.state.tabindexISO = iso;
+        // After UI updates, focus on the new tabindex date
+        setTimeout(() => {
+            /** @type {HTMLButtonElement | undefined} */ (this.getEl(iso))?.focus();
```
ArtBlue, line 286: ❗ This appears to be called `onKeyDown`, which means every keydown (with conditionals met) we are traversing the DOM and selecting the element - very costly. We should cache the element on instantiation of the component and simply reference it where needed. This should be done for other elements as well.

ArtBlue (changes requested):
There are some issues to look at. Both Date Picker and Calendar stories throw an error for me. Once those are fixed, I can take another look.

src/components/ebay-calendar/component.js
LuLaValva: This is not looking for a single element... Each time a key is pressed and conditions are met, we need to focus on a _new_ day. Unless I'm misunderstanding, we would only benefit from caching here if the user clicked the left/right or up/down arrows repeatedly to continuously return to the same element.

src/components/ebay-calendar/calendar.stories.js
LuLaValva: Weird, it looks like moving functions out into a helper file didn't work. I moved everything back into `component.js`, it should be working now.

src/components/ebay-calendar/component.js
ArtBlue: Oh wow! Ok, but not great. Wasn't there a native keyboard solution for that we were considering during the markup determination phase?

src/components/ebay-calendar/component.js
LuLaValva: Sure, comments added

src/components/ebay-calendar/calendar.stories.js
```diff
@@ -0,0 +1,150 @@
...
+            description: 'Triggered when a date is selected',
+            table: {
+                category: 'Events',
+                defaultValue: {
+                    summary: '{ iso }',
+                },
+            },
+        },
+        onMonth: {
+            action: 'on-month',
```
agliga, line 121: I would make this more explicit
On-month-change or on-month-update etc

src/components/ebay-calendar/component.js
```diff
@@ -0,0 +1,401 @@
...
+    onDaySelect(day) {
+        this.emit('select', { iso: day });
+    }
+
+    /**
+     * @param {DayISO} day
+     */
+    onDayFocus(day) {
+        this.state.focusISO = this.state.tabindexISO = day;
+        this.calculateRangeDisplay();
```
agliga, line 124: Might consider adding events emitted from here. 
Might be used for tracking. I'm not sure so I'm fine if we dont.

src/components/ebay-calendar/component.js
agliga: I think that was going to be done later.

src/components/ebay-calendar/index.marko
agliga: No. We need to allow the users to pass in full text
Generally this text we have is not useful. Think of other languages, they might have the description done after the number or before. It might have to be "march show" or something. 
So there needs to be flexibility here. 
We can provide the basic template but allow it to be overriden. We should also make sure all the data they need to do so is emitted in the events that update it
See carousel also for this. To see how we update the current page etc.

src/components/ebay-calendar/index.marko
```diff
@@ -0,0 +1,110 @@
...
+
+$ const numMonths = input.numMonths || 1;
+$ const monthDates = [...Array(numMonths)].map((_, i) => component.getMonthDate(state.offset + i))
+
+<div.calendar>
+    <if(input.navigable)>
+        <div.calendar__header>
+            <div.calendar__header--inner>
+                <ebay-icon-button
+                    transparent size="small"
```
PrashantAshok, line 12: This should be ideally be in different lines. I misreading it as incorrect param. 
```
<ebay-icon-button
  transparent
  size="small"
  ......
/>
```

src/components/ebay-calendar/index.marko
PrashantAshok: +1 to what Andrew said. Ideally this can be extracted to a child component `ebay-calendar-month` or something similar.

src/components/ebay-date-textbox/component.js
```diff
@@ -0,0 +1,132 @@
...
+                      rangeStart: this.state.firstSelected,
+                      rangeEnd: this.state.secondSelected,
+                  }
+                : {
+                      selected: this.state.firstSelected,
+                  }
+        );
+    }
+
+    onCreate() {
```
PrashantAshok, line 107: minor: can the "initial" life cycle methods be not at the start of the file? Easier to read.

src/components/ebay-date-textbox/component.js
```diff
@@ -0,0 +1,132 @@
...
+        this.state = {
+            numMonths: 1,
+            firstSelected: undefined,
+            secondSelected: undefined,
+            popover: false,
+        };
+
+        this.calculateNumMonths();
+        document.addEventListener('resize', this.calculateNumMonths);
```
PrashantAshok, line 116: This should be defined inside `onMount` in case this component gets server rendered (shouldn't be the case). 
I think you should be able to use `this.subscribeTo` instead, you can avoid `removeListener` later.

src/components/ebay-calendar/index.marko
LuLaValva: Ok, I moved it into the component. Hopefully that's a little bit more clear.

src/components/ebay-calendar/calendar.stories.js
LuLaValva: I would like to change this, but I am unable to get storybook working with multi-word events. I don't this works for any of our components.

LuLaValva (commented):
Thanks Andrew, a few hanging threads are left but none that would introduce breaking changes in the future (except for maybe the `on-month` event name, more details in the comments)

LuLaValva:
Thanks for the review @PrashantAshok, hopefully we can have this published for your team very soon.

src/components/ebay-date-textbox/component.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    onMount() {
+        this.expander = new Expander(/** @type {HTMLElement} */ (this.el), {
+            hostSelector: '.ebay-date-textbox--main > .icon-btn',
+            contentSelector: '.date-textbox__popover',
+            expandOnClick: true,
+            autoCollapse: true,
+        });
+        this.subscribeTo(eventUtils.resizeUtil).on('resize', this.calculateNumMonths.bind(this));
```
agliga, line 50: There might be a better way to do this

Maybe we can use subscribe tag (see video)
In that case we only attach the event when the pane is opened. We shouldn't be recalculating when the pane is closed.

src/components/ebay-date-textbox/component.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * @param {{ iso: DayISO }} param0
+     */
+    onPopoverSelect({ iso }) {
+        if (this.input.range) {
+            if (this.state.firstSelected && this.state.secondSelected) {
+                this.state.firstSelected = iso;
+                this.state.secondSelected = undefined;
```
agliga, line 104: When we assign values we should use null not undefined. 
Undefined is only used when we dont assign anything.

src/components/ebay-date-textbox/component.js
```diff
@@ -0,0 +1,136 @@
...
+    closePopover() {
+        this.state.popover = false;
+    }
+
+    /**
+     * @param {{ iso: DayISO }} param0
+     */
+    onPopoverSelect({ iso }) {
+        if (this.input.range) {
+            if (this.state.firstSelected && this.state.secondSelected) {
```
agliga, line 102: This can be simplified. There is only 1 else if case where firstSected is not set to iso. Why not set it always and then just set it again in that other state. 
It seems like this can simply be:
Set first state
Check for that one off case and set iso to second state

src/components/ebay-date-textbox/date-textbox.stories.js
```diff
@@ -0,0 +1,127 @@
...
+                category: 'Events',
+                defaultValue: {
+                    summary: '{ selected } | { rangeStart, rangeEnd }',
+                },
+            },
+        },
+    },
+};
+
+export const Standard = (args) => ({ input: args });
```
agliga, line 120: Use Default and use that helper thing i have

src/components/ebay-date-textbox/date-textbox.stories.js
```diff
@@ -0,0 +1,127 @@
+import Readme from './README.md';
+import Component from './examples/default.marko';
+import code from './examples/default.marko?raw';
```
agliga, line 3: Id like this to be named
DefaultTemplateCode
And the template should be DefaultTemplate. That should be the new standard that we're using

src/components/ebay-date-textbox/index.marko
```diff
@@ -0,0 +1,49 @@
+// @ts-check
+
+<span.date-textbox
+    on-expander-expand("openPopover")
+    on-expander-collapse("closePopover")>
+    <if(input.range)>
+        <ebay-textbox
+            placeholder="YYYY-MM-DD"
```
agliga, line 8: This needs to be customizable.

src/components/ebay-date-textbox/index.marko
```diff
@@ -0,0 +1,49 @@
...
+            value=state.firstSelected
+            on-blur("handleInputChange", 0)>
+        </ebay-textbox>
+        <ebay-textbox
+            class="ebay-date-textbox--main"
+            placeholder="YYYY-MM-DD"
+            buttonAriaLabel="open calendar"
```
agliga, line 15: Localize text

src/components/ebay-date-textbox/index.marko
```diff
@@ -0,0 +1,49 @@
...
+    on-expander-collapse("closePopover")>
+    <if(input.range)>
+        <ebay-textbox
+            placeholder="YYYY-MM-DD"
+            value=state.firstSelected
+            on-blur("handleInputChange", 0)>
+        </ebay-textbox>
+        <ebay-textbox
+            class="ebay-date-textbox--main"
+            placeholder="YYYY-MM-DD"
```
agliga, line 14: Customizable

src/components/ebay-date-textbox/index.marko
```diff
@@ -0,0 +1,49 @@
+// @ts-check
+
+<span.date-textbox
+    on-expander-expand("openPopover")
+    on-expander-collapse("closePopover")>
+    <if(input.range)>
```
agliga, line 6: I think we only need the if statement to render one of the textboxes. They are so similar that we can probably render one by default and the second only when this is true.

agliga (commented):
Did another pass

Its close. 
One more thing i see a lot: best practice is to not assign undefined. Undefined is used only to check if a variable has a default assigned value or not. Null should be used when you want to assign to it. 

https://www.scaler.com/topics/javascript/null-and-undefined-in-javascript/

src/components/ebay-date-textbox/index.marko
LuLaValva: Ok, I changed some logic around to make this work

src/components/ebay-date-textbox/index.marko
LuLaValva: I made it customizable, but the logic might not be ideal. Let me know what you think.

src/components/ebay-date-textbox/index.marko
LuLaValva: See above

src/components/ebay-date-textbox/index.marko
LuLaValva: Good catch, thanks.

LuLaValva (commented):
Thanks Andrew! Comments should all be addressed now

src/components/ebay-calendar/calendar.stories.js
ArtBlue, line 1: The interactive version of `calendar` is not changing the visual date indicator to the date that was clicked on.

agliga:
> I'm taking one last look at this...
> 
> 
> 
> It looks like the static version of `calendar` (interactive being `false`) is emitting events. I don't think the static version should be emitting events.

I don't think it's a big deal if it emits events. There might be several reasons for people to want to listen to events on the static version (like having a static version and then clicking a date takes you to a new server side rendered page). 
They also don't have to listen to them. 
Overall I think we don't lose anything by doing so.

ArtBlue:
> don't think it's a big deal if it emits events. There might be several reasons for people to want to listen to events on the static version (like having a static version and then clicking a date takes you to a new server side rendered page).
> They also don't have to listen to them.
> Overall I think we don't lose anything by doing so.

Hmm...I have a few issues with this. What you've described is not a static use case. A truly static use case is a simple display, just a reference of days of a month to the days of the week. I'm not sure we should expect or even guide users to use a "static" version to accomplish the dynamic interaction you've described in your use case. 

Also, the specific use case you've described wouldn't pass `a11y` audits. The `<span>` elements cannot be traversed using a keyboard, they don't look clickable as they have the wrong cursor ("text selection"). We built the static version in Skin to be completely static; this is why these elements did not take the specific use case into consideration. 

Additionally, if an implementation such as the one you've described is to be supported, it should be done using actual anchor links to make the pages discoverable, indexable, and rankable by search engines. In that scenario, we wouldn't want to guide developers to adopt an inferior usage of the component in any case. That specific scenario would best be accomplished by allowing `<a>` to be used for the days instead of `<span>`. We can add this feature later, but I don't think we'd want to guide users to use the current one to achieve that use case.

Lastly, having an event handler just sitting there for all use cases, the use of which for a "static" implementation would NOT be expected to be there by developers, it's just a wasteful artifact. It's true they don't have to listen to them, but, as negligible as it may be for some use cases, there is memory loss taking place. This might be negligible in some circumstances, but not always. It's best to create technical scenarios that best fit the developer expectations and avoid making unexpected and "hidden" features.

ArtBlue (changes requested):
It looks like these components are still using the old/deprecated icons. All icons without a pixel unit postfix (deprecated, being deleted) should be changed to their respective matching new icons. Here's one: 
image

LuLaValva (commented):
Thanks for the review @ArtBlue, just one comment I'd like to hear your opinion about.

I agree that `<span>`s shouldn't have click handlers, they have been removed.

src/components/ebay-calendar/calendar.stories.js
LuLaValva: Ok, if that's the case then this is more of a `skin` issue. We should have a style for the `:active` selector. Maybe this can be added as a follow-up PR?

ArtBlue: We should probably use a `selected` class or something of the sort, but sure, that can be done later. It would probably need both a Skin piece and an ebay ui piece.

ArtBlue (changes requested):
It looks like after the icons PR was merged into Skin, there are some errors popping up now...

`````
Field 'browser' doesn't contain a valid alias configuration
.../node_modules/@ebay/skin/program-badge doesn't exist
`````

It's likely because program badges module was removed. We probably need to cut ties in ebay ui. 

Maybe check out Skin `16.0.0`, link to it in ebay ui, and fix those issues?

agliga:
> It looks like after the icons PR was merged into Skin, there are some errors popping up now...
> 
> ```
> Field 'browser' doesn't contain a valid alias configuration
> .../node_modules/@ebay/skin/program-badge doesn't exist
> ```
> 
> It's likely because program badges module was removed. We probably need to cut ties in ebay ui.
> 
> Maybe check out Skin `16.0.0`, link to it in ebay ui, and fix those issues?

Need to link skin on 16.0.0 branch to resolve that.

LuLaValva:
The program badge issues were present in `12.0.0` as well until a recent PR merge, which I have now pulled into this branch. I think this should be good to merge now.

ArtBlue (approved):
This looks good to me now. Once Andrew approves, it can be merged.

agliga (approved):

ianmcburnie:
@LuLaValva You may want to tackle this in a follow up PR, but before release we need to change all of the static span elements to use clipped text rather than `aria-disabled` and `aria-label`. This is because most screen readers ignore aria attributes that are on `p`, `div` and `span` (unless they have a role.... but no role seems appropriate for us here).

ianmcburnie:
Header buttons seem misaligned, possibly due to recent icon updates:

Screen Shot 2023-03-22 at 4 24 32 PM

ianmcburnie:
Seeing some unexpected behavior with ARROW KEY nav on interactive ebay-calendar. For example, pressing UP anywhere on top row takes me to 1st cell in row. Likewise on last row, DOWN key takes me to last cell in row. Seems like this is by design? Not a show-stopper but struck me as unexpected.

src/components/ebay-3d-viewer/style.less
ianmcburnie, line 1: Not sure why this file got changed. Seems unrelated.

ianmcburnie: But on the plus side, I'd completely forgotten about the `inset` property!

LuLaValva: Yeah, I didn't want to change it but stylelint was yelling at me... I figured it wouldn't be too awful to include _one_ out-of-scope change.

ianmcburnie (approved):
Approving for pre-release with the understanding this comment will be addressed before final release: https://github.com/eBay/ebayui-core/pull/1801#issuecomment-1480373978

Issue #1479, opened 2021-07-28
ebay-pagination: overflow updates (EVO)


ianmcburnie:
In v8.0.0 we took the new styles of evo from Skin without adding any new business logic.

We should next add the new logic for overflow positions and overflow button.

PR #1806, opened 2022-11-02
feat(pagination): added overflow menu

agliga:
__NOTE__ To get this to work, you need to remove the `overflow: hidden` property from the pagination. I will update skin directly with that change. That change was initially added because of an issue on ebayui causing too many items to be shown, which I fixed in this PR.

## Description
* Added `variant=overflow`. For now, it shows both overflow menus (on prev and next). 
* There was an issue where too many items were shown and not properly. I updated some of the calculation logic to include all items (and take into account the next/prev buttons, which were not)

## Context
* I had to update fake menu button (I also updated menu button), to support this new usecase.
* Added numerous tests

## References
#1479

## Screenshots
Screen Shot 2022-11-02 at 9 49 29 AM
page


src/components/ebay-icon-button/index.marko
```diff
@@ -28,7 +29,8 @@ $ {
     class=[
         input.class,
         'icon-btn',
-        isBadged && `icon-btn--badged`
+        isBadged && `icon-btn--badged`,
+        input.transparent && 'icon-btn--transparent'
```
ArtBlue, line 33: So after speaking with Randy about the buttons for the breadcrumbs overflow, he agreed that the button should have some indication of interactivity - I added a background color for that. We might want to verify with Randy that the same application should be used here. For the sake of consistency, we'd likely want to carry that over to this as well. 
image

ArtBlue: Now that I look at it, this comment might not have been added to the most proper place in the code though it might have ramifications for this too.

src/components/ebay-pagination/component.js
```diff
@@ -93,7 +116,18 @@ export default {
             }
         }
 
-        return { start, end, hideDots, dotsIndex };
+        if (hasLeadingDots) {
```
ArtBlue, line 119: This logic is a bit hard to follow. I think some comments would help.

ArtBlue (approved):
Looks good overall; just a couple of things...

src/components/ebay-pagination/component.js
agliga: The plan is to clean this part up in the tech debt.

agliga:
I am going to merge this. I need it for the breadcrumb change. I will revisit the comments before release, especially the transparent part.
The other part we will clean up in next tech debt.

ArtBlue:
> I am going to merge this. I need it for the breadcrumb change. I will revisit the comments before release, especially the transparent part. The other part we will clean up in next tech debt.

Was this intended for `master`?

agliga:
> > I am going to merge this. I need it for the breadcrumb change. I will revisit the comments before release, especially the transparent part. The other part we will clean up in next tech debt.
> 
> 
> 
> Was this intended for `master`?

It was not. I saw after i merged
I updated the feature branch with it. I was going to undo the master change but it left it in for now.

Issue #1792, opened 2022-10-17
ebay-menu-button: separator takes up a position


Castaliaa:
# Bug Report

## Description
I'm using `separator` in `ebay-menu-button` , but the `indexes` that `on-change` event emits seems strange, the separator itself takes up a position. For example, in this multi-select menu-button, when clicking item1, the index is 0, item3 is 3, and item6 is 7.

```
<ebay-menu-button type="checkbox" text="eBay Menu separator" on-change('onChange')>
    <@item>item 1</@item>
    <@item>item 2</@item>
    <@separator/>
    <@item>item 3</@item>
    <@item>item 4</@item>
    <@item>item 5</@item>
    <@separator/>
    <@item>item 6</@item>
</ebay-menu-button>

onChange({ checked }) {
    console.log('onChange', checked);
}
```
## Screenshots
![image](https://user-images.githubusercontent.com/7250941/196097460-c0e49d03-30e3-441e-8743-e2bcabfcd184.png)

PR #1807, opened 2022-11-03
fix(menu-button): remove index for seperator item

saiponnada:
## Description
Added a separator count and deducted the previous separators count from the item's index and passed that value to reflect the correct index if list item. 

## References
#1792

### DIFF
--- a/src/components/ebay-menu/index.marko
+++ b/src/components/ebay-menu/index.marko
@@ -24,6 +24,7 @@ $ var isRadio = input.type === "radio";
 $ var isCheckbox = input.type === "checkbox";
 $ var isNotCheckable = input.type !== "checkbox" && input.type !== "radio";
 $ var baseClass = input.classPrefix || "menu";
+$ var separatorCount = 0;
 <span
     ...processHtmlAttributes(input, ignoredAttributes)
     class=[
@@ -50,6 +51,7 @@ $ var baseClass = input.classPrefix || "menu";
                 var checked = component.isChecked(index);
             }
             <if (item.separator)>
+                $ separatorCount += 1;
                 <hr class=`${baseClass}__separator` role="separator" />
             </if>
             <else>
@@ -61,8 +63,8 @@ $ var baseClass = input.classPrefix || "menu";
                 aria-disabled=(item.disabled && "true")
                 href=(item.disabled ? null : item.href)
                 role=itemRole
-                onClick(!item.disabled && "handleItemClick", index)
-                onKeydown(!item.disabled && "handleItemKeydown", index)
+                onClick(!item.disabled && "handleItemClick", index - separatorCount)
+                onKeydown(!item.disabled && "handleItemKeydown", index - separatorCount)
                 onKeypress(!item.disabled && "handleItemKeypress")
                 key="item[]">
                 <span>


src/components/ebay-menu/index.marko
```diff
@@ -50,6 +51,9 @@ $ var baseClass = input.classPrefix || "menu";
                 var checked = component.isChecked(index);
             }
             <if (item.separator)>
+                $ {
+                    separatorCount += 1;
+                }
```
agliga, line 56: I think you can just do `$ separatorCount += 1;` in 1 line.

agliga (commented):
Looks great. 
I would add a quick browser test with separators. Don't worry about updating the tests to using the new syntax for now.
Use the code to render separators in the server tests, render one with separators and then click an item before and after the separator, and verify the index is correct. 
Separator render: https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu/test/test.server.js#L37-L38
Copy this test https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu/test/test.browser.js#L79-L99 (you might have to check for the index as well)

ArtBlue (commented):
There's something strange and inconsistent going on here.

1. The mouse-activated experience is not inline with the keyboard activated experience.
2. The keyboard activated experience seems to be skipping an item when using arrow keys.

https://user-images.githubusercontent.com/1675667/200351704-3d67593a-b162-4593-a3e1-e4778a4ee457.mp4

src/components/ebay-menu/index.marko
```diff
@@ -50,6 +51,9 @@ $ var baseClass = input.classPrefix || "menu";
...
                 <hr class=`${baseClass}__separator` role="separator" />
```
ArtBlue, line 57: Is any `js` dependent on or interacting with this `<hr>`? If not, could we not just do a `tabindex="-1"` and an aria hidden on this and fix this without having to fiddle with indexes?

agliga: I think the issue came from treating these as actual item elements. We loop through all the items and if they are a separator, we render it as so.
My guess is the keyboard part might be triggering from makeup.

ianmcburnie: Seems to be working fine on makeup side (https://makeup.github.io/makeup-js/makeup-menu-button/index.html). Is this Safari?

I am currently re-working all of the makeup keyboard stuff and should hopefully be easier to debug when done.

ArtBlue: This was Chrome

ArtBlue:
@saiponnada , also change the target from `master` to `15.2.0`.

src/components/ebay-menu/index.marko
agliga: > First I used the mouse to click into the menu, and then I used the keyboard to navigate the items. Then I collapsed the menu and then used the keyboard to gain focus on the menu and hit <ENTER> and then navigated the menu items with arrows once more.

So i would do one thing
Check if this issue is present on current production
If it is, lets file an issue. If not then we need to fix it
At least the first scenario seems correct, when we activate without keyboard it wont show focus on chrome but then it does if we press down and the second item is selected
My guess is the second usecase is just showing where you left off but doesnt show the initial focus.

ArtBlue: Ok. Yea, it does seem like in the second scenario, it continues from where it left off, which is not ideal. It should really start from the beginning again after the menu is expanded again.

ArtBlue: Yep, it looks to be a production issue. I created the bug issue: https://github.com/eBay/ebayui-core/issues/1809

src/components/ebay-menu/index.marko
saiponnada: Cool. Will do that.

agliga:
@saiponnada So lets just change the separator iterator for now. Lets get this merged and then we can work on tests next week.

saiponnada:
> @saiponnada So lets just change the separator iterator for now. Lets get this merged and then we can work on tests next week.
Agree @agliga . Will focus on code reviews for now.

src/components/ebay-menu/index.marko
saiponnada: Forgot to push this change. Let me do this.

agliga (approved):

PR #1812, opened 2022-11-09
feat(icon-button): add size attribute

LuLaValva:
## Description

Added the `size` attribute to `ebay-icon-button`.

## Context

- Styles do not change unless this is tested in tandem with ebay/skin#1930
- Right now `size` officially only supports "small" and "large", where `undefined` or any other string represents "default/normal".

### DIFF
--- a/src/components/ebay-icon-button/index.marko
+++ b/src/components/ebay-icon-button/index.marko
@@ -1,5 +1,7 @@
 import { processHtmlAttributes } from "../../common/html-attributes"
 
+static const VALID_SIZES = ['small', 'large']
+
 static function toJSON() {
     return {
         disabled: this.disabled
@@ -11,7 +13,8 @@ static var ignoredAttributes = [
     "badgeNumber",
     "badgeAriaLabel",
     "toJSON",
-    "transparent"
+    "transparent",
+    "size"
 ];
 
 $ {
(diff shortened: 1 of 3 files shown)


src/components/ebay-icon-button/index.marko
```diff
@@ -30,7 +30,9 @@ $ {
         input.class,
         'icon-btn',
         isBadged && `icon-btn--badged`,
-        input.transparent && 'icon-btn--transparent'
+        input.transparent && 'icon-btn--transparent',
+        input.size === 'small' && 'icon-btn--small',
+        input.size === 'large' && 'icon-btn--large'
```
agliga, line 35: We could do it like this:
```
$ const validSizes = ['small', 'large'];
```
And then:
```
validSizes.indexOf(input.size) > -1 && `icon-btn--${input.size}`
```
Just so we don't have too many of these statements, or if we need to add a new size its fine

agliga (commented):
Couple of issues:
1. set target to 10.2.0
2. add size to ignored attribute list
3. Should have `small` `regular` and `large`. Set regular as default (or just ignore it)
5. Set default value to normal
6. Need to add size as an enum to `marko-tag.json`

src/components/ebay-icon-button/index.marko
LuLaValva: Nice refactor, thanks. I've addressed all of your changes.

agliga (approved):

PR #1813, opened 2022-11-09
feat(fake-link): add standalone attribute

### LINKED ISSUE #1797: ebay-fake-link: add standalone variant
Original issue: https://github.com/eBay/skin/issues/1904
With this issue we might want to consider adding a `ebay-link` module since we have two variants of basic links now

LuLaValva:
- fixes #1797 

## Description

Add standalone attribute to `ebay-fake-link`, which simply adds the "standalone-link" class when true.

### DIFF
--- a/src/components/ebay-fake-link/fake-link.stories.js
+++ b/src/components/ebay-fake-link/fake-link.stories.js
@@ -32,12 +32,14 @@ export default {
                 },
             },
         },
-        standalone: {
-            description: '',
+        variant: {
+            description:
+                'Should only be standalone when it is clear contextually that this is a link, regardless of styles',
+            options: ['inline', 'standalone'],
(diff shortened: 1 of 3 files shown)


agliga (commented):
Couple of issues:
1. Target 10.2.0
2. I would probably make it as `variant="standlone"` (and have the other option as `inline`). Generally when we have a new look which might have more items in the future we set is as an enum rather than boolean
3. Add the new `standlone` or `variant` to ignored attributes list
4. Add an entry in marko.json
5. Add a description on where it's used like: "Standlone link is used when it is next to a button, but never inline with other text"

LuLaValva:
Thanks for the comments Andrew! Should be fixed now.

agliga (approved):

PR #1816, opened 2022-11-16
Charts

martin-ebay:
## Description
adding the 5 new chart components, area-chart, bar-chart, line-graph, trend-line and donut-chart (with donut-graph / chart legend sub components)


src/components/ebay-area-chart/component.js
```diff
@@ -0,0 +1,282 @@
...
+    patternFill(Highcharts);
+    ebayLegend(Highcharts);
+}
+const pointSize = 1.5;
+
+export default class {
+    onCreate() {
+        this.chartRef = null;
+        this.axisTicksLength = -1;
+        this.state = {
```
agliga, line 29: Do not assign something to state if we are not changing it
Use either `this` or set it as a getter or something

src/components/ebay-bar-chart/component.js
```diff
@@ -0,0 +1,292 @@
...
+                                }, 0);
+                            }
+                        },
+                    },
+                    className: this.input.stacked
+                        ? 'ebay-bar-chart-column-stacked'
+                        : 'ebay-bar-chart-column',
+                    stacking: this.input.stacked ? 'normal' : null,
+                    borderWidth: 1,
+                    borderColor: '#FFFFFF',
```
agliga, line 236: Use a token here

agliga (changes requested):
So couple of comments
Some of these apply to all chart components (like readme changes and doc changes)
Need to add more comments what things do

But regardless I thought we had discussed to remove donut chart for now and I see it here. 
We can't merge it until skin is merged. We need to have it split up

Please address these. We plan on releasing soon so if we want to get this in we need these changes.

src/common/charts/donut-chart.js
```diff
@@ -0,0 +1,169 @@
+const _polarToCartesian = (r, angle) => ({
```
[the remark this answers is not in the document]
ArtBlue: That was my impression too. I thought we were skipping donut for now.

src/components/ebay-bar-chart/component.js
```diff
@@ -0,0 +1,292 @@
...
+                    pointPadding: 0.15,
+                    states: {
+                        inactive: {
+                            opacity: 1,
+                        },
+                    },
+                    point: {
+                        events: {
+                            mouseOver: function () {
+                                const refPoint = this;
```
ArtBlue, line 248: This is a great example of what I was referring to in my previous comment about caching `this` to a meaningful const name. I would go bit further to explain what `refPoint` is, either make the const even more meaningful or add a comment.

src/components/ebay-bar-chart/component.js
```diff
@@ -0,0 +1,292 @@
...
+        let yLabelsItterator = 0;
+        const config = {
+            title: {
+                text: this.input.title,
+            },
+            chart: {
+                type: 'column',
+                backgroundColor: backgroundColor,
+                style: {
+                    fontFamily: '"Market Sans", Arial, sans-serif',
```
ArtBlue, line 70: This shouldn't be static. We should get it from some central global config.

martin-ebay:
let me see what happened cause this isn't matching my local
seems pushing to this branch from another didn't do as I expected
it's like it did a merge of them.

Issue #1820, opened 2022-12-02
3d-viewer: integrate with open source solution


agliga:
We need to change 3d-viewer to use an open source soltuion. Will use `google-model-viewer`

PR #1822, opened 2022-12-02
feat(3d-viewer): updated to use open source model-viewer

agliga:
## Description
* Added 3d-viewer to use `model-viewer`  open source project instead of restar. 
* Added scripts to generate CDN files for upload
* (also updated shaka player as part of this, since it was out of date)

This will be breaking for teams since they will need to migrate to the new 3d viewer and use `src` instead of `assetID`

## References
https://github.com/eBay/ebayui-core/issues/1820

## Screenshots
Screen Shot 2022-12-02 at 8 20 48 AM


src/components/ebay-3d-viewer/examples/02-with-options/template.marko
```diff
@@ -0,0 +1,16 @@
+class {}
+
+<ebay-3d-viewer
+    src='https://ir.ebaystatic.com/cr/v/c1/ebayui/3d/v1/image.glb'
```
ArtBlue, line 4: How come some are double-quoted while this one is single.

src/components/ebay-video/versions.json
```diff
@@ -1 +1 @@
-{ "//": "This is a generated file. Run script file to update", "shaka": "4.2.2" }
+{ "//": "This is a generated file. Run script file to update", "shaka": "4.3.0" }
```
ArtBlue, line 1: How come this video file got an update from the 3d viewer change?

ArtBlue (approved):
Looks great! Just a couple of questions.

src/components/ebay-video/versions.json
agliga: I mentioned that I updated it since it was out of date. I had not updated it earlier. 
The CDN generate script updates those version files so it tells the package which version to use. Since we have two files in CDN (shaka and media viewer), it got updated as well.

src/components/ebay-3d-viewer/examples/02-with-options/template.marko
agliga: It's because I am used to using prettier and it's not working locally. 😅

Issue #1804, opened 2022-10-26
listbox-button: match border & background color with other components like combobox or textbox


agliga:
See https://github.com/eBay/skin/issues/1842

Will need to add the option to support the change in skin

PR #1823, opened 2022-12-02
feat: add “form” variant to listbox-button

tigt:
## Description
Add plumbing so skin gets a styling hook for `<listbox-button variant="form">`.

Marked as Draft because I’m unfamiliar with matters of taste in ebayui-core — in particular, I dunno if I should test this in server, browser, or both?

## Context
I looked how the form variant was defined in `button`, and tried reapplying those codebits to `listbox-button`.

## References
Fixes https://github.com/eBay/ebayui-core/issues/1804

## Screenshots
As described by @agliga in the referenced issue, this adds an under-the-hood option that skin can use — as such, there are no visual differences.


agliga:
This looks great. 
Generally we begin with the skin portion because the class names might change. So we want to get that ironed out first before doing the component piece. 
The only change I would do is I would add a default value to variant, like `regular` or something. And then we should just do the check of: `class=[(input.variant === 'form' && 'listbox-button--form')]`

ArtBlue (changes requested):

ArtBlue (approved):
Nice!

ArtBlue:
Looks good to merge. Maybe just change the title of the PR?

Issue #1824, opened 2022-12-05
InvalidCharacterError on <ebay-filter-menu-button>


doahuang:
# Bug Report

## Description

when input `footerText` is passed as string, say, `this is footer text`, it will break
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-filter-menu-button/index.marko#L56
> InvalidCharacterError: “this is footer text” did not match the Name production: Unexpected syntax in top

PR #1825, opened 2022-12-06
fix(filter-menu): footer text passthrough

agliga:
## Description
* We were passing footer text as a render body element, but it has no render body. We want to pass it directly. Changed it to pass it as a string
* Added tests to catch this case.

## References
https://github.com/eBay/ebayui-core/issues/1824


ArtBlue (approved):
Looks good.

saiponnada (approved):
Looks good to me.

Issue #1826, opened 2022-12-06
ebay-menu-button: index not correctly matched


agliga:
> Hi, @agliga @saiponnada ,
Thanks for the fix. But I tested in v10.2.0, there is still some problem.
The indexes "change" emit is correct now, but in this multi-select, the check mark is not correct. For example, if I click item 4, there will be a check mark in item 3. 
Could you please take a look?

_Originally posted by @Castaliaa in https://github.com/eBay/ebayui-core/issues/1792#issuecomment-1336938509_

PR #1828, opened 2022-12-06
fix(menu-button): index not correctly matched

saiponnada:
## Description
To match indexes, passed filtered menu items excluding separators. Inserted separators at respective locations.

## References
#1826 #1792, #1829

## Screenshots

menu-button
image

fake-menu
image


agliga:
Goutham, there's a lot of methods in the menu utils that is using this.input.items. We should change it to use this.items. 
Otherwise it will give us the incorrect indexes when we have a separator still.

agliga:
I think there is still an issue with the separator. It's not rendering in the right place.
If you think about it it makes sense.

The user does:
```
item1
item2
separator
item3
item4
separator
item5
```

What we do is strip out the separators in one list, so that has
```
item1, item2, item3, item4, item5
```
And then we have another list with the separator indexes which has
```
false, false, true, false, false, true, false
```
If you look, the indexes on these two lists won't match. That means that we are placing the separators in the wrong position

If we want to fix this, we need to increment the index with how many previous separators were found.

ianmcburnie:
> Have we considered using pseudo elements for the separators?

Doesn't a separator represent something semantically though? If we are visually grouping things, doesn't that grouping represent information that we would lose with pseudo elements?

agliga:
We can use pseudo elements. I think it would be best to do it that way. 
However, we need to support this syntax till next major version, so we need to keep it in here anyways. 

Ideally we could do:
```
<ebay-menu-button>
   <@group>
      <@item>
      <@item>
   </@group>
   <@group>
       <@item>
      <@item>
  </@group>
</>
```

If that's what you meant.

src/components/ebay-menu/index.marko
```diff
@@ -25,6 +25,13 @@ $ var isCheckbox = input.type === "checkbox";
 $ var isNotCheckable = input.type !== "checkbox" && input.type !== "radio";
 $ var baseClass = input.classPrefix || "menu";
 $ var separatorCount = 0;
+$ var seperatorMap = (input.items || []).reduce((map, item, index) => {
+    if(item.separator) {
+        map[index - separatorCount] = true;
+        separatorCount++;
+    }
+    return map;
+}, {});
```
agliga, line 34: One more nit. Sorry, last one hopefully.
Let's put this in `menuUtils`. I think what might be easiest is to create a method on menuUtils, like `getSeparatorMap`. You can move the `separatorCount` there and such.
You can then call it here using `component.getSeparatorMap(input.items)`
This way it will be only in 1 place and we don't have it copied. (I see it's copied in the other PR for fake menu.

Otherwise, it looks great!

agliga (approved):

LuLaValva (approved):
This works for now and I haven't been able to find any places where it is broken, but there are a lot of places that have to be modified when the logic for separators is modified. Since we want to release soon I'd say push this through for now, but `ebay-menu` and `ebay-fake-menu` would definitely be good areas to clean up during tech debt month.

Issue #1829, opened 2022-12-06
ebay-fake-menu-button: separator takes up a position


saiponnada:
# Bug Report

## Description
Similar to issue with #1792, Separator takes up a position leading to incorrect indices across the items.

## Screenshots
When Item 3 is selected it should emit index of 2.

<img width="1144" alt="image" src="https://user-images.githubusercontent.com/6342519/206004814-fd480273-bb2e-4651-891c-378ff3ab66fe.png">

PR #1831, opened 2022-12-07
fix(fake-menu-button) separator takes up a position

saiponnada:
## Description
To match indexes, passed filtered menu items excluding separators. Inserted separators at respective locations.

## References
#1829 

## Screenshots
Menu items were selected from first item to last and we can see events fired with the right indexes.

image

### DIFF
--- a/src/components/ebay-fake-menu/component.js
+++ b/src/components/ebay-fake-menu/component.js
@@ -20,4 +20,8 @@ export default {
 
         this.emit(`${eventType}`, eventObj);
     },
+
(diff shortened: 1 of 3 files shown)


agliga:
@saiponnada 
Lets combine this PR with https://github.com/eBay/ebayui-core/pull/1828
We should use the same methods from there. Seems like menuUtils was updated and we should use that in this PR as well. 

If it's too much work to combine, lets first merge #1828 and then pull those changes into this PR and then get it to work.

PR #1833, opened 2022-12-15
highcharts charts

martin-ebay:
## Description
Initial build of all highchart based chart components plus the custom sparkline chart component


src/components/ebay-area-chart/README.md
```diff
@@ -0,0 +1,46 @@
...
+        label: '$1,255.00'
+    }]
+/>
+```
+
+## Description
+
+The area chart displays one to five series of data points as an interactive stacked area chart
+
+## Attributes
```
agliga, line 34: Can we update this to match `button`.
All attributes and how to use are in `storybook`
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-button/README.md

src/components/ebay-area-chart/area-chart.stories.js
```diff
@@ -0,0 +1,88 @@
+import { addRenderBodies } from '../../../.storybook/utils';
+import Readme from './README.md';
+import Component from './index.marko';
+
+function getPositiveTrend(days) {
```
agliga, line 5: This is good, but when users are looking through our docs, we want them to easily find how to add data. 
Maybe we should have this as a separate file? Or we can add a template with all the data hardcoded. (See button for examples on that https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-button/button.stories.js#L5-L6)

src/components/ebay-area-chart/component.js
```diff
@@ -0,0 +1,331 @@
...
+    // patternFill highcharts wrapper function enables rendering patterns instead of just solid colors
+    patternFill(Highcharts);
+    // add custom legend wrapper function
+    ebayLegend(Highcharts);
+}
+const pointSize = 1.5;
+
+export default class {
+    onCreate() {
+        this.chartRef = null;
```
agliga, line 31: Why is this needed?

src/components/ebay-area-chart/component.js
```diff
@@ -0,0 +1,331 @@
...
+        this.state = {
```
agliga, line 32: Do not use state unless we change it later. Either set it as a getter or use a common component method.

src/components/ebay-area-chart/component.js
```diff
@@ -0,0 +1,331 @@
...
+            containerId: `ebay-bar-chart-${this.id}`, // set unique ID for html container element that highcharts will use
+        };
+    }
+    onMount() {
```
agliga, line 36: Can we have a `setup` method, like we have for video, or other components (`_setupMakeup` etc)
`onMount` just calls those setup methods (we can have a `_setupEvents` and a `_setupCharts` one)

src/components/ebay-area-chart/component.js
```diff
@@ -0,0 +1,331 @@
...
+    // debounce used to help improve performance on mouse interactions
+    debounce(func, timeout = 100) {
+        let timer;
+        return (...args) => {
+            clearTimeout(timer);
+            timer = setTimeout(() => {
+                func.apply(this, args);
+            }, timeout);
+        };
+    }
```
agliga, line 255: Put this in utility

src/components/ebay-area-chart/component.js
```diff
@@ -0,0 +1,331 @@
...
+                        },
+                        false, // disable auto redraw
+                        false // disable auto animation
+                    );
+                }
+            });
+        });
+        this.chartRef.redraw(); // trigger redraw after all points have been updated
+    }
+}
```
agliga, line 332: Do we need any `teardown` methods?

src/components/ebay-bar-chart/component.js
```diff
@@ -0,0 +1,343 @@
...
+    legendHoverColor,
+    tooltipBackgroundColor,
+    tooltipShadows,
+    setSeriesColors,
+} from '../../common/charts/colors';
+
+import { ebayLegend } from '../../common/charts/legend';
+import { eBayColumns } from '../../common/charts/bar-chart';
+
+if (typeof Highcharts === 'object') {
```
agliga, line 21: If highcharts is not loaded in the page or if they load it later, then this won't be executed.
Maybe this should be put in an `onMount` and use `DOMContentedLoaded` or `load` event (see video `onMount`)

agliga (changes requested):
Added comments from the last PR.
Looking better overall.

src/components/ebay-area-chart/component.js
martin-ebay: so that I can store a reference to the chart once it is initialized.  Later in the code I often call the redraw method, or look at data elements from the base chartRef.

src/components/ebay-area-chart/component.js
martin-ebay: this state variable is used in the marko template, should I still move it?  I wanted easy access to the variable from the marko markup

src/components/ebay-area-chart/component.js
martin-ebay: not for redraw, redraw just tells highcharts to run it's basic draw method which Im suppressing above in the .update calls by the second and third parameters that are set to false.

src/components/ebay-bar-chart/component.js
martin-ebay: I have this conditional here so that it doesn't run on the server side.  I can see if I can move it into the onMount.  Not sure if it will cause any problems or not

src/components/ebay-area-chart/component.js
agliga: No, not for redraw. I'm pointing at the end of the file
Isn't there a method used to tear down any event listeners so we don't have memory leaks?

src/components/ebay-area-chart/component.js
agliga: state is only used when we want to rerender based on it's changes.
If we want to access something from a template, we can put it in a method on the `component`.
So like:
```
getContainerId() {
   return `ebay-bar-chart-${this.id}`;
}
```
And then in the template we call it using:
`<div id=component.getContainerId()>`

src/components/ebay-area-chart/component.js
agliga: The question is, why set it at null initially? It's not doing anything. There's no checks if it's null etc.

src/components/ebay-bar-chart/component.js
agliga: The issue isn't for server. It's on the client timing. This might not be called in time if `Highcharts` loads after this code. Some pages do optimizations where they load the code after page load to help site speed.

src/components/ebay-area-chart/component.js
martin-ebay: I thought it was nice to see a reference to a variable that is called later down, and the linter does let me create a variable with no assignment initially.  I'll just remove these lines then.  Was trying to make reading this file easier

src/components/ebay-area-chart/component.js
martin-ebay: I could add something

src/components/ebay-bar-chart/component.js
martin-ebay: So looking at the video player it is loading it's js manually from an ebaystatic url.  Are we planning to try and break highcharts out into a separate JS file to do this?  It tree shakes rather well when compiling the component.

agliga: So the thing is, highcharts is large. This could cause issues on loading the page if it's too big. I know with video this was a major concern on pages. 
We could change it like video to load it after page load.

src/components/ebay-line-graph/line-graph.stories.js
```diff
@@ -0,0 +1,230 @@
+import { addRenderBodies } from '../../../.storybook/utils';
+import Readme from './README.md';
+import Component from './index.marko';
+
+const sellingData = require('./selling-mock.json');
+
+const getLineGraphData = (days) => {
```
ArtBlue, line 7: Please add comments throughout this file.

src/components/ebay-spark-line/component.js
```diff
@@ -0,0 +1,37 @@
+export default {
+    onCreate(input) {
```
ArtBlue, line 2: A lot of what's taking place in here should be moved out into function(s) and be called from here to make things easier to read, reference, and understand.

ArtBlue (commented):
This is definitely better than the previous PR. A couple of things common to a lot of files here are (1) missing comments and (2) too many different things taking place inside single functions. 

Instead of this:
`````js
mount() {
  // 100 lines of code that does X
  // 100 lines of code that does Y
}
`````
we should be doing this:
`````js
mount() {
  doX();
  doY();
}

doX() {
// 100 lines of code that does X
}

doY() {
// 100 lines of code that does Y
}
`````

File: src/common/charts/shared.js [hunk 11]
```diff
@@ -0,0 +1,99 @@
+export const chartFontFamily = '"Market Sans", Arial, sans-serif';
```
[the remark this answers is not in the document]
martin-ebay: this gives a fun new parsing error ```101:7  error  Parsing error: This experimental syntax requires enabling the parser plugin: "exportDefaultFrom". ```. I'll see what I can do to make it easier to read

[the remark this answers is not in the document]
martin-ebay: I'll see what I can do to clean this up

src/components/ebay-area-chart/area-chart.stories.js
```diff
@@ -0,0 +1,88 @@
...
+            label: `$${v}`,
+        });
+    }
+
+    return {
+        data,
+    };
+}
+
+const multipleLineSeries = (num, days) => {
```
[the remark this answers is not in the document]
martin-ebay: I removed this with the last commit, this was just one of my random chart data generator helper functions.  I've since just made a hard coded data object.

src/common/charts/shared.js
```diff
@@ -0,0 +1,99 @@
...
+            break;
+        default:
+            colors = colorsMore;
+            series[0].lineColor = series[0].borderColor = chartSecondaryColor;
+            series[1].lineColor = series[1].borderColor = chartTertiaryStrokeColor;
+            series[2].lineColor = series[2].borderColor = chartPrimaryColor;
+            if (series.length > 3) {
+                series[3].lineColor = series[3].borderColor = chartQuaternaryStrokeColor;
+            }
+            if (series.length > 4) {
```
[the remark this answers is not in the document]
martin-ebay: This was done to prevent code duplication.  as I need the lines 87-90 to be executed for both and didn't think it was worth trying to break out into another function call

src/components/ebay-area-chart/component.js
```diff
@@ -0,0 +1,332 @@
...
+    legendInactiveColor,
+    legendHoverColor,
+    tooltipBackgroundColor,
+    tooltipShadows,
+    setSeriesColors,
+} from '../../common/charts/shared';
+
+import { ebayLegend } from '../../common/charts/legend';
+
+if (typeof Highcharts === 'object') {
```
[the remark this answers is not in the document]
martin-ebay: this has been removed now

src/components/ebay-area-chart/component.js
```diff
@@ -0,0 +1,332 @@
...
+                        {
+                            className: null, // nullify className if not active
+                            marker: {
+                                enabled: false, // disable marker if not active
+                            },
+                        },
+                        false, // disable auto redraw
+                        false // disable auto animation
+                    );
+                } else if (d.onTick && d.className === null) {
```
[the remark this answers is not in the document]
martin-ebay: it's similar but if onTick is true, and className !== null I don't want to run the update as there could be a lot of data points to cycle through and could cause a very slight performance hit.

src/components/ebay-area-chart/component.js
```diff
@@ -0,0 +1,332 @@
...
+                    d.update(
+                        {
+                            className: 'ebay-area-chart__marker--visible', // set classname
+                            onTick: d.onTick, // sets the onTick flag to keep track of the points enabled status for mouse events
+                            marker: {
+                                enabled: true, // set marker enabled
+                                radius: pointSize, // set the size of marker
+                                lineColor: backgroundColor, // set border color of hover markers
+                                lineWidth: 4, // set border width of hover markers
+                                fillColor: '#000000', // set fill color of markers
```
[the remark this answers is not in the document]
martin-ebay: oops yeah it should I totally missed moving that one out to a css var

src/components/ebay-area-chart/area-chart.stories.js
```diff
@@ -0,0 +1,913 @@
+import { tagToString } from '../../../.storybook/storybook-code-source';
+import { addRenderBodies } from '../../../.storybook/utils';
+import Readme from './README.md';
+import Component from './index.marko';
+
+const sampleSeriesData = [
```
agliga, line 6: Put this in an external file. Probably under `examples`. 
`examples/data.json`

src/components/ebay-bar-chart/component.js
```diff
@@ -0,0 +1,346 @@
...
+            },
+        };
+        this.chartRef = Highcharts.chart(this.getContainerId(), config);
+        this.chartRef.redraw();
+    }
+
+    getChartConfig() {
+        return {
+            type: 'column',
+            backgroundColor: backgroundColor, // set the chart background color
```
agliga, line 82: You can just do 
```
type: 'column',
backgroundColor,
style: {...
```
And we don't need these kind of comments . I think we can see that setting `backgroundColor` will set the chart background color :)

src/components/ebay-bar-chart/component.js
```diff
@@ -0,0 +1,346 @@
...
+        } else {
+            // if not stacked, set the top and bottom flag on each series so the single bar has rounded top and bottom corners
+            series.forEach((s) => {
+                s.top = true;
+                s.bottom = true;
+            });
+        }
+        const config = {
+            title: {
+                text: this.input.title, // set the title that will render above the chart
```
agliga, line 61: Can we set all the input's at the top of the function as variables

Like:
`const title = this.input.title`

And then we can use it later.
This way we can see at the very beginning what `inputs` are used and don't have to dig into the large config.

src/components/ebay-bar-chart/component.js
```diff
@@ -0,0 +1,346 @@
...
+            let s = `<b>${Highcharts.dateFormat('%b %e, %Y', this.x, false)}</b>`; // sets the displayed date at the top of the tooltip
+            if (component.input.stacked) {
+                // setup html for stacked tooltip
+                component.chartRef.series.forEach((serie) => {
+                    // cycle through each series
+                    serie.data.forEach((d) => {
+                        // cycle through each series data array to match x value with active hovered xAxis position
+                        if (d.x === this.x) {
+                            // when the x value matches the hovered xAxis position
+                            s += `<div style="display: flex; justify-content: space-between; width: 100%; align-items: flex-start;">${serie.name}<span style="margin-left: 16px">${d.label}</span></div>`;
```
agliga, line 176: Is there a better way to do this?
I mean I really don't want to stick in a string of HTML in javascript. This seems very backwards to me.

Can we have a sub template which we can just import here and then render it to string?

src/components/ebay-bar-chart/component.js
```diff
@@ -0,0 +1,346 @@
...
+    setSeriesColors,
+} from '../../common/charts/shared';
+
+import { ebayLegend } from '../../common/charts/legend';
+import { eBayColumns } from '../../common/charts/bar-chart';
+
+export default class {
+    onMount() {
+        this._initializeHighchartsExtensions();
+        this._setupCharts();
```
agliga, line 24: If this component rerenders, will highcharts still work? If I change input for example.

src/components/ebay-bar-chart/component.js
```diff
@@ -0,0 +1,346 @@
...
+                fontSize: '12px',
+            },
+            // this callback function is used to position the tooltip at the top of the stacked bars
+            positioner: this.input.stacked && this.tooltipPositioner,
+        };
+    }
+
+    legendItemClick() {
+        // returns a function so that can access input values
+        const component = this; // component reference used in item click functions that don't have the same scope
```
agliga, line 235: This isn't even needed. We can simply just add a `const stacked = this.input.stacked` at the top.
If we are worrying about change in inputs, we might need to revisit how to initialize highcharts.

agliga (commented):
I think it looks much better overall. Have some more comments and questions. 
The main one I want to resolve is the HTML in javascript. I really don't want to see any of that especially if we can use marko to render to string.

src/components/ebay-bar-chart/component.js
martin-ebay: not with highcharts. It doesn't actually render html either.

for reference https://api.highcharts.com/highcharts/tooltip.formatter

quote from the docs 

"A subset of HTML is supported. Unless useHTML is true, the HTML of the tooltip is parsed and converted to SVG, therefore this isn't a complete HTML renderer. The following HTML tags are supported: b, br, em, i, span, strong. Spans can be styled with a style attribute, but only text-related CSS, that is shared with SVG, is handled."

agliga: Marko has a renderToString. That should be better than this

```
Import subtemplate from 'subtemplate.marko'

subtemplate.renderToString(inputObject)
```

Should work

src/components/ebay-line-graph/style.less
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @media (prefers-color-scheme: dark) {
+        .highcharts-halo {
+            fill-opacity: 0.8; // this inverses the default 0.2 fill opacity in from light mode
+        }
+    }
+}
+
+// remove this before finished, just used as a sample
```
agliga, line 35: Please remove :)

src/components/ebay-spark-line/style.js
```diff
@@ -0,0 +1 @@
+require('./spark-line.less');
```
agliga, line 1: Why not just put `spark-line.less` contents in this file?

src/components/ebay-spark-line/test/test.server.js
```diff
@@ -0,0 +1,43 @@
...
+            y: 1,
+        },
+    ],
+};
+describe('spark-line', () => {
+    it('renders a blue line', async () => {
+        const { getByRole } = await render(template, mock);
+        const line = getByRole('graphics-symbol');
+        expect(line.classList.contains('ebay-spark-line__path-neutral'));
+    });
```
agliga, line 24: I'm happy to see tests.
Please look up `button` server tests. We have snapshot testing we use which is better than checking for class existence. 
We also use storybook as the basis of our mocks (see `button` again)

Some components are not changed, but we are going in that direction.

agliga (commented):
Overall I think it's almost there. Couple more nits I saw. 
I think after these changes and the template changes, I should be okay to merge this.

src/components/ebay-spark-line/style.js
martin-ebay: @agliga what would the syntax be for this. I don't think I've ever includes less directly in a js file like that.

agliga: Oh I see, it's a js file. 
Let me check on this.

agliga: So for now:
lets have a `style.less` with all the styles.
Lets have a `style.js` which imports `style.less`

martin-ebay: ok so leave it as is then

martin-ebay: oh wait I see rename the less file. change coming in a minute

src/components/ebay-spark-line/component.js
```diff
@@ -1,37 +1,56 @@
...
-            // set the y value minus the minimum y value times the normalize ratio + 10 padding and line width
-            path.push((p.y - yMin) * yScale + 10);
-        });
-        // set the viewBox x, y, width, height
-        const viewBox = `0 0 ${(endX - startX) * xScale + 12} ${(yMax - yMin) * yScale + 20}`;
-
-        this.state = {
-            path: path.join(' '), // join the path array into a string with spaces for use in the path d attribute
-            viewBox,
+        const yScale = 100 / (minMax.yMax - minMax.yMin);
```
ArtBlue, line 41: Nice cleanup!

ArtBlue (commented):
Looks good! Just one thing...

src/common/charts/shared.js
```diff
@@ -1,99 +1,99 @@
...
+                break;
+            default:
+                colors = colorsMore;
+                series[0].lineColor = series[0].borderColor = chartSecondaryColor;
+                series[1].lineColor = series[1].borderColor = chartTertiaryStrokeColor;
+                series[2].lineColor = series[2].borderColor = chartPrimaryColor;
+                if (series.length > 3) {
+                    series[3].lineColor = series[3].borderColor = chartQuaternaryStrokeColor;
+                }
+                if (series.length > 4) {
```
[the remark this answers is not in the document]
martin-ebay: this was done to prevent code duplication of these lines

colors = colorsMore;
                series[0].lineColor = series[0].borderColor = chartSecondaryColor;
                series[1].lineColor = series[1].borderColor = chartTertiaryStrokeColor;
                series[2].lineColor = series[2].borderColor = chartPrimaryColor;
                
                which are the same for both cases 3 and 4

src/common/charts/shared.js
```diff
@@ -0,0 +1,99 @@
...
+        pattern2,
+        chartQuinaryBackgroundColor,
+    ],
+    // function is used to set up the colors including lineColor(svg stroke) on each of the series objects
+    // based on the length of the series array
+    setSeriesColors = function (series) {
+        let colors;
+        switch (series.length) {
+            case 1:
+                colors = colorsSingle;
```
ArtBlue, line 72: @agliga , I just realized this function is just cherry-picking colors outside of itself. This is a bit brittle. We can keep it like this for now if we just want to get it in, but we should refactor this. Ideally, if we use a two-dimensional array here and pass the array into the function call, the function can then loop through and set the proper series colors without conditionals.

src/common/charts/shared.js
```diff
@@ -0,0 +1,99 @@
...
+    colorsSingle = [chartPrimaryColor],
+    colorsTwo = [chartSecondaryColor, chartPrimaryColor],
+    colorsThree = [chartSecondaryColor, chartPrimaryColor, pattern1],
+    colorsMore = [
+        chartSecondaryColor,
+        pattern1,
+        chartPrimaryColor,
+        pattern2,
+        chartQuinaryBackgroundColor,
+    ],
```
agliga, line 65: @martin-ebay can I get some clarity here, so we can refactor this part.
So if we have 1 series, we use primary as the main one, correct?
So in two series we are setting the secondaryColor as the first one and primary as secondary? Or is it going backwards?
In three series (This is where I get extremely confused), we have the first one as Secondary, second one as primary and the third as pattern1 (whatever pattern 1 is)... 
And then in more than 3, we have secondary, pattern1 (pattern1 moves to before primary?), primary, pattern2, and then quinary?

What is going on here?!

martin-ebay: so this all can be answered by looking at the figma 
https://www.figma.com/file/DVhjKXFZoTna19ffUkBk34/Evo-Data-Viz?node-id=5%3A27013&t=NsZjMIAAC8ri6lAn-0

if you look at the colors for charts portion of the figma you will see a table with the five "colors" with two of them being patterns.

Then if you look at the examples of the stacked bar charts the colors used are different depending on the number of series of data.

I'm going to post a couple images in slack to help point you at the right part of the figma

agliga: The more I think about it, the more this is off...
So if we have 1 series, the color is primary. 
If we have 2 series the first color is secondary, and then switches to primary.
If we have 3 series, then it goes primary, secondary, and then pattern1
If we have 4, then it goes to secondary, pattern1, primary, and then pattern2.

So the point of this was to make sure the order of lines look okay right? And this was called out as for accessibility so that the colors won't overlap or will be ordered correctly.... But what is to guarantee that series 1 will be the first one? 
What if series 1 has the largest amount of data on the y axis? Then the order would no longer look correct right and it would be on the top? I don't think you can guarantee that series 1 will always be the smallest dataset. It could overlap with others? 
Of course, on Figma everything looks nice and neat, but with real data, that no longer is the case. Things change, trends can overlap and then the charts will no longer look nice and neat like designers want them to. We need to figure out what the best way to overcome that would be, and I don't think by switching the series lines we would achieve that.

src/common/charts/shared.js
```diff
@@ -0,0 +1,79 @@
+export const chartFontFamily = '"Market Sans", Arial, sans-serif',
```
ArtBlue, line 1: Still not great to have this defined statically here instead of referencing something global, but we can change this in the future.

ArtBlue (approved):
Looks good for initial alpha release.

agliga (approved):

Issue #1837, opened 2023-01-04
feat(ebay-tourtip): Add support for heading elements


HenriqueLimas:
# Feature Request

## Description
Currently ebay-tourtip is rendering the headings as a `span` as per the base [tooltip](https://github.com/eBay/ebayui-core/blob/master/src/components/components/ebay-tooltip-overlay/index.marko#L38), but we should be able to allow a tourtip to render a heading element instead (ex. `h2`).

## Context
Skin uses heading elements for [tourtip](https://opensource.ebay.com/skin/#tourtip) so the same should be used on ebayui-core

## Screenshots

<img width="405" alt="Screen Shot 2023-01-04 at 10 18 16 AM" src="https://user-images.githubusercontent.com/2222191/210626726-78b8c0f8-45fb-4480-95d7-168f774d3d4a.png">

agliga:
We should add an `as=` attribute to `@heading` to override span
For now, we should leave span as the default so it doesn't break existing pages.

PR #1838, opened 2023-01-05
tips add support for heading elements

saiponnada:
## Description
Provided the ability to pass `as` attribute to heading tag to choose the right heading when needed. Default fallback is set to `span` to avoid breaking changes.

This change affects ebay-tourtip, ebay-infotip and ebay-tooltip.

## Context
#1837 

## Screenshots

image


ArtBlue (approved):
Looks good to me.

agliga (approved):

Issue #1835, opened 2023-01-03
data-viz: add line-charts, and other charts


agliga:
# Feature Request
With new designs from Data visualization, we should add these charts within our library. We should start with line chart

PR #1840, opened 2023-01-10
feat: added common cdn loader, and set highcharts to load from cdn

agliga:
fixes https://github.com/eBay/ebayui-core/issues/1835

## Description
Since highcharts is the third component to use CDN loading of scripts, I decided to take some time and refactor the CDN code to be more generic. (Also good as part of tech debt)

## Context
* Updated the generate CDN script to take a simple config object to generate the cdn files, and version.json file. There is only 1 version.json in the `common/cdn` directory
* Added a `common/cdn` file, which has a class that does all the cdn loading. 
* Added some tests for this

## Note
For high charts, the accessibility and pattern-fill modules required highcharts to be loaded. There was an issue if we would load all the scripts async, then there could be a race condition to cause the accessibility module to fail loading. I added a stagger option, so that it will not load another module, until the previous one is loaded. This could be improved to only wait for the high charts module, and then all others to load at the same time... however pattern-fill is used for other modules which we are not going to officially include. So I left it as is for now, and we can revisit when needed.

### DIFF
--- a/src/common/cdn/index.js
+++ b/src/common/cdn/index.js
@@ -10,7 +10,7 @@ const MAX_RETRIES = 3;
  * function setLoaded(boolean) : sets the is loaded state
  * function handleSuccess() : when cdn is succsssfully loaded
  * function handleError(err) : when the cdn loading fails
- * boolean stagger :  (optional) if all promoses should be staggered or executed in any order
+ * boolean stagger :  (optional) if all promises should be staggered or executed in any order
  */
 export class CDNLoader {
     constructor(self, { key, files, types, setLoading, handleSuccess, handleError, stagger }) {
@@ -56,7 +56,7 @@ export class CDNLoader {
         }
     }
 
-    loadCDN(immediate) {
+    loadCDN() {
         const _timeout =
             window.requestIdleCallback ||
             function (handler, arg) {
@@ -75,13 +75,8 @@ export class CDNLoader {
         this.setLoading(false);
 
         _cancel(this.loadDelay);
-        if (!immediate) {
-            this.setLoading(false);
-            this.loadDelay = _timeout(() => this._loadCDN(), { timeout: 100 });
-        } else {
-            this.setLoading(true);
-            this._loadCDN();
-        }
+        this.setLoading(true);
+        this.loadDelay = _timeout(() => this._loadCDN(), { timeout: 100 });
     }
 
     _catchError(err) {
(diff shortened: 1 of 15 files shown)


src/common/cdn/index.js
```diff
@@ -0,0 +1,108 @@
...
+            function (id) {
+                clearTimeout(id);
+            };
+
+        this.retryTimes = 0;
+        this.setLoading(false);
+
+        _cancel(this.loadDelay);
+        if (!immediate) {
+            this.setLoading(false);
```
saiponnada, line 79: Do we need this line? We already set it at L75. Also, I guess we need to set loading to true in L80 callback.

saiponnada (approved):
Pretty good refactor.  We are intentionally hiding all the charts except line chart for this release. correct? left minor comments.

src/common/cdn/index.js
ArtBlue: I'm guessing one of these needed to be `true` and the other sets it to the opposite based on the condition.

src/common/loader/index.js
```diff
@@ -57,23 +57,47 @@ function restArWorker(src) {
...
-                    } else if (typeList[key] === 'restar-worker') {
-                        return restArWorker(src, promiseKey);
-                    } else if (typeList[key] === 'model-viewer') {
-                        return srcLoad(src, promiseKey, 'module');
-                    }
-                }
-                return srcLoad(src, promiseKey);
-            })
-        );
+        if (stagger) {
```
ArtBlue, line 89: I like this idea of staggering. I think it could be optimized a bit more by identifying root flies that don't have dependencies and allow them to load asynchronously. That would speed things up more. Maybe that could be a future optimization?

ArtBlue (approved):

src/common/cdn/index.js
agliga: This is actually correct. If we trigger the immediate loading, we want to make sure the setLoading is set to false. This is if multiple components are on the page running concurrently.

src/common/loader/index.js
agliga: Yes, that's the plan. 
I think ideally we need to change loading to take an object which should have:
name - file to load from cdn
path - the path to the cdn
type - file type (css, script)
dependencies - array of all script that need to laod first. 

But this is for future.

Issue #1847, opened 2023-01-27
tech-debt: clean up storybook


agliga:
Plans for storybook, all components should have the following setup:

- All example directories should be flattened. It should be `example/{template name/type}.marko`
- All templates that are used in storybook need to emit all actions and take `...input` as spread.
- Stories should use templates only to show different variations that cannot be shown using args
- Add a helper to build these different varations
- Add templates that bridge gap of components.

agliga:
Closing this one. More than half components complete. Will open up another for the rest.

PR #1850, opened 2023-02-07
chore: tech debt on storybook stories 

agliga:
Please read first before reviewing

## Description
This is a large PR, the process for each should be the same. The idea should be the same for each component. 
I've changed `Standard` to `Default`. The issue with using default as the name was that it was a keyword. I got around this by making it capitalized. I think default is a better name than standard.
Basically all imports follow the pattern: `import NameTemplate from './examples/name.marko` where name of the type. We also import the code as `import NameTemplateCode from './examples/name/marko?raw`. I created a simple helper method to render this and the code in storybook (so we don't have to type all the definitions over and over). I started to change some code with the helper but wanted to get feedback for this.
For the code that has examples, I flattened it to simply be the `examples/template-name.marko`. 
The idea is for any code that has a simple interaction, we use storybook args directly. This is for cases like button or the base menu example. For anything more complex, we can render a template. Each template needs to have input spread and emit events. 
If this approach is good, then I will update all other components. 

## Context
- changed the following: eek, fake menu, infotip, inlinenotice, lightbox dialog, menu

## References
#1847


src/components/ebay-3d-viewer/3dViewer.stories.js
```diff
@@ -158,15 +151,15 @@ export default {
...
 
-export const Standard = Template.bind({});
-Standard.args = {
+export const Default = Template.bind({});
+Default.args = {
     src: 'https://ir.ebaystatic.com/cr/v/c1/ebayui/3d/v1/image.glb',
     a11yText: 'View these shoes for sale.',
 };
-Standard.parameters = {
+Default.parameters = {
```
ArtBlue, line 159: Yea, I do like `Default`; it's a lot clearer.

ArtBlue (commented):
The approach looks good to me.

ianmcburnie (approved):

Issue #1845, opened 2023-01-24
ebay-infotip: on-expand gets called twice when variant is set to modal


johnjhwang:
# Bug Report

## Description
When the infotip's variant is set to modal, the `on-expand` event calls the handler that's passed in twice, instead of once. Since this is just for the modal variant which uses drawer dialog, it may be due to some behavior related to drawer dialog and/or its animation. This is blocking the implementation of DSA click tracking as it's reliant on the on-expand event.

## Workaround
Not yet explored as on-expand is the easiest and cleanest way to implement the tracking. I can explore workarounds if the bug is not a simple fix.

## Screenshots
Clip demonstrating behavior:

https://user-images.githubusercontent.com/89167923/214404938-16c7403d-4172-459b-9233-58f2843f5869.mov

johnjhwang:
any update on this please?

PR #1851, opened 2023-02-09
fix(ebay-infotip): removed double events from infotip

agliga:
## Description
There was a problem with infotip calling expand twice. This is because the `on-click` method of the button was calling onExpand, but also the `drawer-dialog` `on-open` event was triggering the same `onExpand` method. I changed it so the `on-click` calls an open dialog method which just opens the dialog. The other event stays the same. This will make it so if it's programmatically called, it will still trigger properly. 

## References
https://github.com/eBay/ebayui-core/issues/1845

### DIFF
--- a/src/components/ebay-infotip/component.js
+++ b/src/components/ebay-infotip/component.js
@@ -11,6 +11,10 @@ export default {
         }
     },
 
+    handleOpenModal() {
+        this.setOpen(true);
+    },
+
     handleExpand() {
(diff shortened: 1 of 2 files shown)


LuLaValva (approved):
Good catch, seems to be working on my end

saiponnada (approved):
Looks good to me.

ianmcburnie (approved):

Issue #1856, opened 2023-02-24
tech-debt: clean up storybook (part 2)


agliga:
part 2 of https://github.com/eBay/ebayui-core/issues/1847

PR #1859, opened 2023-03-01
tech-debt: updated readme's and examples

agliga:
## Description
This is a large PR, I don't expect that each line is read and verified. 
* Updated all readmes to point to storybook. No readme's should have any more attributes or variables in them. This means that storybook is the source of truth
* Flattened all example directories. It should be `examples/{example-name}.marko` now.
* Updated menu and menu button tests to use storybook mock data

## References
https://github.com/eBay/ebayui-core/issues/1856


ArtBlue (approved):
Looks fantastic!

Issue #1863, opened 2023-03-10
fix: accessibility error on combobox


agliga:
When combobox is disabled, aria-expanded is not added because expander is skipped. Need to fix to add expander even when disabled.

PR #1864, opened 2023-03-10
fix(combobox): a11y errors when disabled 

saiponnada:
## Description
When the component is mounted in disabled state, makeup-expander is not associated leading to a11y error `Required ARIA attributes not present: aria-expanded, aria-controls`.

To fix this, added `aria-expanded` attribute set to false on mount. This attribute along with `aria-controls` updates to current state once expander attaches.

## Context
#1863

## References

Combobox pattern  tested using [mind pattern checklist](https://ebay.gitbook.io/mindpatterns/appendix/checklist#combobox)

## Screenshots

Before:
image

After:
image

### DIFF
--- a/src/components/ebay-combobox/index.marko
+++ b/src/components/ebay-combobox/index.marko
@@ -69,6 +69,7 @@ $ var id = input.id || component.getElId('input');
             aria-roledescription=roledescription
             aria-haspopup='listbox'
             autocomplete='off'
+            aria-expanded='false'
             onBlur('handleComboboxBlur')
             onClick('handleComboboxClick')
             onFocus('handleFocus')
--- a/src/components/ebay-combobox/test/test.browser.js
+++ b/src/components/ebay-combobox/test/test.browser.js
@@ -378,7 +378,7 @@ describe('given the combobox starts with zero options', () => {
         });
 
         it('then it should not be expanded', () => {
-            expect(component.getByRole('combobox')).not.has.attr('aria-expanded');
+            expect(component.getByRole('combobox')).has.attr('aria-expanded', 'false');
         });
     });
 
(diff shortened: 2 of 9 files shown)


src/components/ebay-combobox/test/test.browser.js
```diff
@@ -378,7 +378,7 @@ describe('given the combobox starts with zero options', () => {
         });
 
         it('then it should not be expanded', () => {
-            expect(component.getByRole('combobox')).not.has.attr('aria-expanded');
+            expect(component.getByRole('combobox')).has.attr('aria-expanded', 'false');
```
saiponnada, line 381: @ianmcburnie is it mandatory to not to have this attribute on load?

src/components/ebay-combobox/index.marko
```diff
@@ -69,6 +69,7 @@ $ var id = input.id || component.getElId('input');
             aria-roledescription=roledescription
             aria-haspopup='listbox'
             autocomplete='off'
+            aria-expanded='false'
```
agliga, line 72: If we render this with open as true or expanded, does it properly expand and set to true? I assume yes, but we should verify.

agliga (approved):
LGTM, just one usecase we should verify.

src/components/ebay-combobox/test/test.browser.js
ianmcburnie: I'd assumed absence of attribute on a combobox would be applied as aria-expanded=false. It definitely doesn't hurt to be explicit though and have it present in all cases, especially if that is what axe is looking for.

src/components/ebay-combobox/index.marko
saiponnada: I guess it should be. How can we test this? When I pass `&args=expanded:true` to [url](https://opensource.ebay.com/ebayui-core/?path=/story/form-input-ebay-combobox--floating-label&args=expanded:true), it doesn't load expanded on storybook.

agliga: We need just a test case. Although it probably shouldn't be an issue since it will open client side.

Issue #1866, opened 2023-03-15
fix: ebay-line-chart tooltip breaking for multiple series


figueiredo-lucas:
# Bug Report

## Description
There's a faulty variable on the code of the tooltip.marko which causes the rendering of the tooltip to break:

https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-line-chart/tooltip.marko#L8

## Workaround
To test, I've changed the variable name directly on node_modules. Basically changed `pnt` to `point` in that `line 8` show above.

## Screenshots
<img width="586" alt="image" src="https://user-images.githubusercontent.com/17763219/225396490-b0a35c32-97a6-4e70-bb32-70494ea2d5db.png">

PR #1868, opened 2023-03-15
fix: fixes typo on tooltip

figueiredo-lucas:
## Description
Fixes typo in variable that would prevent the component from rendering

## Context
There was a typo in tooltip.marko when having more than one series that would give several errors on log and prevent the tooltip to show up

## References
https://github.com/eBay/ebayui-core/issues/1866

## Screenshots
Can be found in the issue


LuLaValva (approved):
Great catch, thanks for the PR! Looks good on my end, but I'm not sure if we should do a patch release or put this out with `16.0.0` so we should wait for @agliga's approval before merging.

agliga:
Lets target 12.0.0 with this for now.

ianmcburnie (approved):

agliga (approved):

PR #1872, opened 2023-03-24
feat(calendar): maintain parity with skin's a11y updates

LuLaValva:
- more details on ebay/skin#2019
    - until merged, these changes won't work unless ebayui-core is locally linked with that skin on that branch

## Description
- Add configurable a11y text for "today" and "selected"
- Move a11y info to clipped span in static calendars
- Add classes and move attributes to satisfy new skin selectors

### DIFF
--- a/src/components/ebay-calendar/index.marko
+++ b/src/components/ebay-calendar/index.marko
@@ -1,20 +1,20 @@
 // @ts-check
-import {toISO} from "./component"
+import {localeOverride, toISO} from "./component"
 
 $ const numMonths = input.numMonths || 1;
 $ const monthDates = [...Array(numMonths)].map((_, i) => component.getMonthDate(state.offset + i));
 $ const getShowMonthText = input.getA11yShowMonthText ?? ((month) => `Show ${month}`);
 /**
-* @param {{selected: boolean, rangeStart: boolean, inRange: boolean, rangeEnd: boolean, today: boolean, disabled: boolean}} info
+* @param {{isSelected: boolean, isRangeStart: boolean, isInRange: boolean, isRangeEnd: boolean, isToday: boolean, isDisabled: boolean}} info
 */
 static function defaultBuildA11yCellText(info) {
     const parts = [""];
-    if (info.selected) parts.push("selected");
-    if (info.rangeStart) parts.push("start of range");
-    if (info.rangeEnd) parts.push("end of range");
-    if (!info.rangeStart && !info.rangeEnd && info.inRange) parts.push("in range");
-    if (info.today) parts.push("today");
-    if (info.disabled) parts.push("inactive");
+    if (info.isSelected) parts.push("selected");
+    if (info.isRangeStart) parts.push("start of range");
+    if (info.isRangeEnd) parts.push("end of range");
+    if (!info.isRangeStart && !info.isRangeEnd && info.isInRange) parts.push("in range");
+    if (info.isToday) parts.push("today");
+    if (info.isDisabled) parts.push("inactive");
     return parts.join(" - ");
 }
 $ const buildA11yCellText = input.buildA11yCellText ?? defaultBuildA11yCellText;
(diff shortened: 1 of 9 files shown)


src/components/ebay-calendar/index.marko
```diff
@@ -104,10 +103,12 @@ $ const rangeEndText = input.a11yRangeEndText ?? "end of range"
                                             </button>
                                         </if>
                                         <else>
-                                            <span
-                                                class=disabled && "calendar__cell--disabled"
-                                                aria-label=rangeAriaLabel>
-                                                ${day}
+                                            $ if (isToday) a11yText += ` - ${todayText}`;
```
agliga, line 106: How is this going to be localized? We also have to consider that in some other languages the day goes before the action. 
I think we should allow this to be fully changeable. We can give them the tools they need to do so however.

LuLaValva: I was thinking about this when I wrote the code, which is why I have hyphens between each label. So the longest label will be something like
```
10 - [selected] - [start of range] - [today]
```
where all of the text in `[]` brackets  is configurable. Is this not enough for localization?

src/components/ebay-date-textbox/date-textbox.stories.js
```diff
@@ -91,23 +91,16 @@ export default {
...
-            control: { type: 'text' },
-            description: 'Text to indicate a value that is inside the selected range',
-            table: { defaultValue: { summary: 'in range' } },
-        },
-        a11yRangeEndText: {
-            type: 'text',
-            control: { type: 'text' },
-            description: 'Text to indicate the end of a selected range',
-            table: { defaultValue: { summary: 'end of range' } },
+        buildA11yCellText: {
```
agliga, line 94: Can we have an example of this in storybook?

agliga (commented):
The only change id like is the changing of params. The rest we can do post release.

agliga (commented):
Also should probably document it in date textbox too. I looked if it was but i didnt see it.

LuLaValva (commented):
Thanks for the review Andrew, I'd like to follow up with additional documentation but these changes should be in soon so for now I think this will suffice

agliga (approved):

src/components/ebay-date-textbox/examples/localized.marko
```diff
@@ -0,0 +1,15 @@
+<ebay-date-textbox
+    disableBefore="2023-01-12"
+    disableAfter="2023-02-19"
+    range
+    locale="af"
+    getA11yShowMonthText=(monthName) => `Gaan na ${monthName}`
+    buildA11yCellText=({isSelected, isRangeStart, isInRange, isRangeEnd}) => {
+        let string = ""
```
ArtBlue, line 8: missing `;`

ArtBlue (approved):
Approved; just add the missing `;` Nice work!

src/components/ebay-calendar/index.marko
ianmcburnie: Yes to what Luke says. The idea with the hyphens is that this is not a grammatical construct/sentence. Instead we are mimicking ARIA states whereby we control the order in which they are read.

ianmcburnie:
One thing I noticed is `aria-selected` on the `<td>` elements for start and end of range. These ARIA semantics are redundant and will not be announced as user navigates from button to button. 

We should remove this and either:

* A) Add "selected" to clipped text of button
* B) Just let the "start of range" and "end of range" clipped text suffice

Screen Shot 2023-03-28 at 6 30 56 PM

LuLaValva:
> One thing I noticed is `aria-selected` on the `<td>` elements for start and end of range. These ARIA semantics are redundant and will not be announced as user navigates from button to button.

@ianmcburnie you may still have an older version of this branch cloned locally, I changed that behavior last week

> I also think aria-pressed on the start and end buttons might be a good idea. And on single select use case too.

Ok, I've added add `aria-pressed` to the selected buttons.

src/components/ebay-calendar/index.marko
```diff
@@ -64,37 +69,33 @@ $ const rangeEndText = input.a11yRangeEndText ?? "end of range"
...
-                                            : undefined
-                                        ))
-                                    <td
-                                        class={
+                                    $ const isDisabled = component.isDisabled(dayISO);
+                                    $ let a11yTexts = [""];
+                                    $ if (isSelected) a11yTexts.push(a11ySelectedText);
+                                    $ if (isRangeStart) a11yTexts.push(a11yRangeStartText);
+                                    $ if (!isRangeStart && !isRangeEnd && isInRange) a11yTexts.push(a11yInRangeText);
+                                    $ if (isRangeEnd) a11yTexts.push(a11yRangeEndText);
```
agliga, line 84: Can we use a block syntax for this?

```
$ {
let a11yTexts;
if etc
};
```

src/components/ebay-calendar/index.marko
```diff
@@ -64,37 +69,33 @@ $ const rangeEndText = input.a11yRangeEndText ?? "end of range"
...
-                                                : input.selected && input.selected === dayISO
-                                        ) && "true">
+                                        }>
                                         <if(input.interactive)>
                                             <button
                                                 key=dayISO
-                                                disabled=disabled
-                                                aria-label=rangeAriaLabel
+                                                disabled=isDisabled
+                                                aria-label=(a11yTexts.length > 1) && `${day}${a11yTexts.join(a11ySeparator)}`
```
agliga, line 96: Does this need a separator or something between day and a11yText's join?

LuLaValva: No separator is needed, because the array starts as a single empty string so effectively it begins with a separator

PR #1873, opened 2023-03-28
feat(calendar): add option for links in static calendar

LuLaValva:
- This should be reviewed and merged after #1872, as it includes changes from that branch

## Description

- A record of ISO dates to links can now be passed to the calendar component, which transforms date cells into links.


agliga (approved):
LGTM

agliga (changes requested):
Lets hold off till we figure out the right API for this.

PR #1874, opened 2023-03-28
fix: ebay-line-chart - Fix to onInput crashing because this.input is undefined

figueiredo-lucas:
## Description
Adds a fix to onInput to so it will have this.input whenever this._setupChart() gets called

## Context
Whenever ebay-line-chart gets rerendered, onInput gets called and throws an error because this.input is undefined

## Screenshots
image

### DIFF
--- a/src/components/ebay-line-chart/component.js
+++ b/src/components/ebay-line-chart/component.js
@@ -53,7 +53,8 @@ export default class {
             .mount();
     }
 
-    onInput() {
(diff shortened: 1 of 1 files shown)


LuLaValva (approved):
Thanks for the quick PR! Looks good to me.

agliga (approved):
Technically the right solution would move the setup in onUpdate. But this is fine for now.

Issue #1881, opened 2023-04-13
ebay-video: play button barely visible


darkwebdev:
# Bug Report

## Description
`Play` button has very low contrast

## Screenshots
Before Skin 16:
<img width="181" alt="image" src="https://user-images.githubusercontent.com/2737310/231731050-8549c779-ef68-4145-baa4-e592d45d659c.png">

After:
<img width="197" alt="image" src="https://user-images.githubusercontent.com/2737310/231731170-9ff80706-8b42-479a-855a-3f8f6b8b4dd5.png">

ArtBlue:
I don't think this is an ebay ui core bug. If anything, it's arguably an issue in Skin, but it's probably not that either. We simply need a more bulletproof icon from DS to work well in both use cases - dark background and light.

agliga:
The icon should be the play filled one. 
Although i see the icon show up as blue and not like in this image. It fits better

agliga:
Spoke to Ryan, it should be `play-64-colored`

darkwebdev:
Thanks @agliga !

PR #1882, opened 2023-04-13
fix: updated play icon to be play-64

agliga:
## Description
Updated play icon on video and 3d-player to be `play-64-colored-icon`

## References
https://github.com/eBay/ebayui-core/issues/1881


LuLaValva (approved):
Thanks Andrew, looks like an edge case was missed on the `ebay-video` component so I pushed up a commit with a fix. I think it should be good now so I'll approve, but I'd like to also see one from someone else since I contributed.

ianmcburnie (approved):

ArtBlue (approved):
Looks great!

Issue #1896, opened 2023-05-03
avatar: new module


agliga:
Component version of https://github.com/eBay/skin/issues/1948

PR #1897, opened 2023-05-03
feat(avatar): new module

agliga:
## Description
Added avatar module

## References
https://github.com/eBay/ebayui-core/issues/1896

## Screenshots
Screen Shot 2023-05-03 at 2 02 08 PM

### DIFF
--- a/src/components/ebay-avatar/examples/signedout.marko
+++ b/src/components/ebay-avatar/examples/signedout.marko
@@ -1,0 +1,1 @@
+<ebay-avatar a11y-text="Signed out" variant="signedout" ...input/>
--- a/src/components/ebay-avatar/examples/signedout.marko
+++ b/src/components/ebay-avatar/examples/signedout.marko
@@ -1,1 +1,1 @@
-<ebay-avatar a11y-text="Signed out" variant="signedout" ...input/>
+<ebay-avatar a11y-text="Signed out" ...input/>
--- a/src/components/ebay-avatar/index.marko
+++ b/src/components/ebay-avatar/index.marko
@@ -1,17 +1,17 @@
 import { processHtmlAttributes } from "../../common/html-attributes";
-import {getColorForText} from './util';
-
-static var ignoredAttributes = [
-    "a11yText",
-    "class",
-    "size",
-    "color",
-    "img",
-    "username"
-];
-
+import { getColorForText } from "./util";
+static {
+    var ignoredAttributes = [
+        "a11yText",
+        "class",
+        "size",
+        "color",
+        "img",
+        "username",
+    ];
+}
 $ var isSignedOut = !input.username;
-$ var color = getColorForText(input.username, input.color)
+$ var color = getColorForText(input.username, input.color);
 
 <div
     ...processHtmlAttributes(input, ignoredAttributes)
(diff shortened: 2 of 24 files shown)


src/components/ebay-avatar/examples/signedout.marko
```diff
@@ -0,0 +1 @@
+<ebay-avatar a11y-text="Signed out" variant="signedout" ...input/>
```
LuLaValva, line 1: Is this `variant="signedout"` a remnant from an earlier iteration? It looks like you aren't using `input.variant` at all, instead you just check if `input.username` exists.

src/components/ebay-avatar/index.marko
```diff
@@ -0,0 +1,39 @@
...
+    </if>
+    <else-if(input.img)>
+        <img ...input.img/>
+    </else-if>
+    <else-if(input.username)>
+
+        ${input.username.charAt(0).toUpperCase()}
+    </else-if>
+    <else>
+        <${input.renderBody}/>
```
LuLaValva, line 37: This branch will never be reached, because `isSignedOut` will be true every time `input.username` is falsy and vice versa.

src/components/ebay-avatar/index.marko
```diff
@@ -0,0 +1,39 @@
...
+        `avatar--${color}`,
+    ]>
+    <if(isSignedOut) >
+        <ebay-avatar-signed-out-icon/>
+    </if>
+    <else-if(input.img)>
+        <img ...input.img/>
+    </else-if>
+    <else-if(input.username)>
+
```
LuLaValva, line 33: empty line

LuLaValva (commented):
Overall looks great! Just some minor code style suggestions.

Also, do we have a standard in place for what the avatar should look like when an image exists but has not loaded yet? Right now it is just an empty colored circle, but it might be more visually appealing if we to showed a letter instead.

src/components/ebay-avatar/index.marko
```diff
@@ -0,0 +1,39 @@
...
+    "a11yText",
+    "class",
+    "size",
+    "color",
+    "img",
+    "username"
+];
+
+$ var isSignedOut = !input.username;
+$ var color = getColorForText(input.username, input.color)
```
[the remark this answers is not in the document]
agliga: There are some issues with using const. 
It also makes it easier to have all the logic in the function rather than outside.

agliga:
> Overall looks great! Just some minor code style suggestions.
> 
> Also, do we have a standard in place for what the avatar should look like when an image exists but has not loaded yet? Right now it is just an empty colored circle, but it might be more visually appealing if we to showed a letter instead.

We can probably do that. I think the tricky part is to layer it over the letter which we need CSS to do so. I think part of the problem is this image I have is really large, normally it won't be.

src/components/ebay-avatar/index.marko
```diff
@@ -0,0 +1,41 @@
...
+    ]
+>
+    <if(input.img)>
+        <img ...input.img>
+    </if>
+
+    <else-if(input.renderBody)>
+        <${input}/>
+    </else-if>
+    <else-if(!isSignedOut)>
```
LuLaValva, line 34: With this conditional structure refactor it might be simpler if you remove the `isSignedOut` variable entirely and just check `else-if(input.username)`

LuLaValva (approved):
Nice! Just one suggested refactor, but I don't mind if this is merged without it. Approving.

src/components/ebay-avatar/index.marko
agliga: Good catch, its easy to miss simple things like this.

Issue #1892, opened 2023-04-20
ebay-fake-menu-button: callbacks use wrong arguments


darkwebdev:
# Bug Report

## Description
Documentation reads that:
- ~`onSelect/onKeydown` callbacks should use `{ el, index, checked }` as argument~ - fixed, thanks!
- ~`onExpand/onCollapse/onMouseDown` shouldn't pass any arguments~ - fixed, thanks!

but in storybook I see:
- `on-select: {el: HTMLAnchorElement, originalEvent: MouseEvent, index: 1}`
- `on-collapse: {el: undefined, originalEvent: undefined, index: undefined}`
- `on-expand: {el: undefined, originalEvent: undefined, index: undefined}`
- `on-mousedown: {el: undefined, originalEvent: undefined, index: undefined}`

I've failed to trigger `onKeydown` event (should it be triggered on menu button keyDown or on menu item keyDown?)
One more mistake: storybook reads `onMousedown`: `Triggered on mouse down on menu item`, but in reality it's also triggered on menu button click.

darkwebdev:
I'm afraid the problem here is not with documentation, but with the code: when I click on the menu button most of the arguments are undefined:
<img width="1244" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/a47532c1-a7c3-4121-a80a-1a4d99c23029">

saiponnada:
@agliga  @darkwebdev 

Okay I have reviewed the code again(apologies for quick review & fix) and noticed that previous changes were pretty accurate and looks like we need to roll back documentation.  Mainly there are two types of events here...

DOM events
- `on-select` event is a component event that triggers a dom `select` event and works as expected.
-  `on-keydown` event is a dom event and works as expected. It might not be visible on storybook but using keyboard, try pressing enter or spacebar on the menu button and that triggers the event. @agliga If storybook defines an event is it mandatory to output every event in actions tab?

Hence below statement is not true.
> I've failed to trigger onKeydown event (should it be triggered on menu button keyDown or on menu item keyDown?)

Component events
- `onExpand/onCollapse/onMouseDown` shouldn't pass any arguments and docs need to be rolled back to previous state. Currently we just emit these events with undefined event object across the storybook.

```js
handleExpand() {
    this.emitComponentEvent({ eventType: "expand" });
},
emitComponentEvent({ eventType, el, originalEvent, index }) {
    const eventObj = {
            el,
            originalEvent,
            index,
        };
[...]

Issue #1891, opened 2023-04-20
ebay-fake-menu: callback arguments documented wrongly?


darkwebdev:
# Bug Report

## Description
Documentation reads that:
- `onSelect/onKeydown` callbacks should use `{ el, index, checked }` as argument

but in storybook I see:
<img width="612" alt="image" src="https://user-images.githubusercontent.com/2737310/233414070-e045d92d-f42d-422c-8db0-36f205010945.png">

darkwebdev:
thank you!

Issue #1889, opened 2023-04-20
ebay-checkbox: callback arguments are wrong


darkwebdev:
# Bug Report

## Description
Documentation reads that:
- `onChange` callback should use `{ originalEvent, value, checked }` as argument
- `onKeydown/onFocus` use `{ originalEvent, value }`

but in storybook I get additional arguments:
- `onChange/onKeydown/onFocus` all use `{ originalEvent, value, checked }, ComponentClass` - not sure what `ComponentClass` is, element or something else...

<img width="348" alt="image" src="https://user-images.githubusercontent.com/2737310/233403501-2e8f39a4-c435-4fce-902c-9f72677b0fd7.png">

Not sure if that's a bug or documentation mistake.

PR #1899, opened 2023-05-08
docs(checkbox, fake-menu, fake-menu-button): synced docs with code

saiponnada:
## Description

Updated the storybook documentation to sync with current code implementation.

## References

 #1889, #1892, #1891

### DIFF
--- a/src/components/ebay-checkbox/checkbox.stories.js
+++ b/src/components/ebay-checkbox/checkbox.stories.js
@@ -71,7 +71,7 @@ export default {
             table: {
                 category: "Events",
                 defaultValue: {
-                    summary: "{ originalEvent, value }",
+                    summary: "{ originalEvent, value, checked }",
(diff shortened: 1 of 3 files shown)


src/components/ebay-checkbox/checkbox.stories.js
```diff
@@ -71,7 +71,7 @@ export default {
             table: {
                 category: "Events",
                 defaultValue: {
-                    summary: "{ originalEvent, value }",
+                    summary: "{ originalEvent, value, checked }",
```
saiponnada, line 74: Not passing the component class as param here. Do we need to pass it? I don't see any value.

agliga: That's due to storybook implementation

agliga (approved):

PR #1900, opened 2023-05-09
fix(date-textbox): add guard against document for server side

LuLaValva:
## Description

Added a type guard against document for date textbox during server rendering.

## Context

The document object is not accessible from the server side.

### DIFF
--- a/src/components/ebay-date-textbox/component.js
+++ b/src/components/ebay-date-textbox/component.js
@@ -66,10 +66,12 @@ export default class extends Marko.Component {
     }
 
     calculateNumMonths() {
-        this.state.numMonths =
-            document.documentElement.clientWidth < MIN_WIDTH_FOR_DOUBLE_PANE
-                ? 1
-                : 2;
+        if (document !== undefined) {
+            this.state.numMonths =
+                document.documentElement.clientWidth < MIN_WIDTH_FOR_DOUBLE_PANE
+                    ? 1
+                    : 2;
+        }
     }
 
     /**
--- a/src/components/ebay-date-textbox/component.js
+++ b/src/components/ebay-date-textbox/component.js
@@ -66,7 +66,7 @@ export default class extends Marko.Component {
     }
 
     calculateNumMonths() {
-        if (document !== undefined) {
+        if (typeof document !== "undefined") {
             this.state.numMonths =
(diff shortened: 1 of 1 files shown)


agliga (approved):

src/components/ebay-date-textbox/component.js
```diff
@@ -66,10 +66,12 @@ export default class extends Marko.Component {
     }
 
     calculateNumMonths() {
-        this.state.numMonths =
-            document.documentElement.clientWidth < MIN_WIDTH_FOR_DOUBLE_PANE
-                ? 1
-                : 2;
+        if (document !== undefined) {
```
PrashantAshok, line 69: This check won't work, `document` would be undefined in server side. You might have to change this to
```
typeof document === 'undefined'
```

ArtBlue: Can't we just check whether `document` is falsy? And to keep conditionals shorter and the code easier to read, it's usually better to exit functions as quickly as possible...

`````js
if (!document) {
  return;
}
...
`````

Comment (agliga) [reply src/components/ebay-date-textbox/component.js:69]: Yes we need to check for typeof. This can throw a compile error if document is undefined.

**Review by agliga — CHANGES_REQUESTED** [review]
Need to change to check typeof

File: src/components/ebay-date-textbox/component.js (hunk 1 above)
Comment (LuLaValva) [reply src/components/ebay-date-textbox/component.js]: Thanks for the input everyone, I've updated it now

File: src/components/ebay-date-textbox/component.js [hunk 2]
```diff
@@ -66,10 +66,12 @@ export default class extends Marko.Component {
     }
 
     calculateNumMonths() {
-        this.state.numMonths =
-            document.documentElement.clientWidth < MIN_WIDTH_FOR_DOUBLE_PANE
-                ? 1
-                : 2;
+        if (typeof document !== "undefined") {
```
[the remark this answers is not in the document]
agliga: Agreed. However I'm fine with it as is for now.

agliga (approved):
Approving. We can change it to a guard if we like. Either way is fine by me.

agliga (approved):
Approving. We can change it to a guard if we like. Either way is fine by me.

agliga (approved):

Issue #1898, opened 2023-05-05
segmented-buttons: new module


agliga:
Original skin issue [#1976](https://github.com/eBay/skin/issues/1976)

PR #1902, opened 2023-05-10
feat(segmented-buttons): new component

agliga:
## Description

Component part for segmented button

## References
https://github.com/eBay/ebayui-core/issues/1898

## Screenshots
Screen Shot 2023-05-10 at 1 53 54 PM


LuLaValva (commented):
Looks awesome, the only problem I could find is that the storybook file and the directory disagree about the name of the component. Is it `segmented-button` or `segmented-buttons`?

saiponnada (changes requested):
Looks good to me.  Only suggestion that might need to be considered(not mandatory) is making sure these buttons are 44 x 44 dp touch target size so that users might not bump into one another.

As Luke mentioned, minor tweaks. Looks like the name need to be changed in few places 
 - Folder need to be renamed to ebay-seg**~~e~~**mented-button**s**
 - Server test describe needs to be updated as well - L11
 - Story book docs need be updated at L95

agliga:
> Looks good to me. Only suggestion that might need to be considered(not mandatory) is making sure these buttons are 44 x 44 dp touch target size so that users might not bump into one another.
> 
> As Luke mentioned, minor tweaks. Looks like the name need to be changed in few places
> 
> * Folder need to be renamed to ebay-seg**~e~**mented-button**s**
> * Server test describe needs to be updated as well - L11
> * Story book docs need be updated at L95

Okay, should be ready for rereview. 
The 44x44 part should be a skin change. I think we can revisit later. 
But renamed all tests and code.

LuLaValva (approved):
Thanks Andrew! Looks great

saiponnada:
yeah, looks like new rule mandates 24 x 24 dp and is not effective yet. so we can just keep it in backlog and make sure it corresponds to that before the rule becomes effective.

saiponnada (approved):
Looks good to me.

PR #1910, opened 2023-05-19
feat: convert to TypeScript

### LINKED ISSUE #1767: Support TypeScript
We possibly do this in two stages.

Details to follow...

LuLaValva:
- resolves #1767 

## Description

**BIG** pull request, all components have been updated to TypeScript. Some refactors were necessary, but I tried to keep all  functionality exactly the same.


src/components/ebay-carousel/examples/autoplay.marko
```diff
@@ -1,3 +1,16 @@
...
+    itemsPerSlide?: number;
+    gap?: number;
+    a11yPreviousText?: string;
+    a11yNextText?: string;
+    autoplay?: boolean;
+    paused?: boolean;
+    a11yPauseText?: string;
+    a11yPlayText?: string;
+}
+
```
agliga, line 13: This is probably not needed. Initially we were using another method in storybook to expose variables and this is a remnant of that. I would just simply spread input in the carousel in this file.

src/components/ebay-checkbox/index.marko
```diff
@@ -1,20 +1,29 @@
 import { processHtmlAttributes } from "../../common/html-attributes"
 
+export interface Input extends Omit<Marko.Input<'input'>, `on${string}` | 'type'> {
+    toJSON?: any;
```
agliga, line 4: The toJSON: this is internal and should not be exposed. It's used for optimization.

src/types.d.ts
```diff
@@ -0,0 +1,19 @@
+declare global {
+    namespace Marko {
+        interface HTMLAttributes {
+            /* makeup */
+            "on-expander-expand"?: (event: any) => void;
+            "on-expander-collapse"?: (event: any) => void;
+            "on-activeDescendantChange"?: (event: any) => void;
```
agliga, line 7: Kinda strange that this is all that's needed. Is that correct or just leftover?

agliga (commented):
So first pass, looks great! 

Couple of comments. Overall I'm fine with it going in, but I want to get some clarity on some things and get Dylans input. 

Things of note and questions:
1. I noticed there are no changes in package.json scripts. Do our prepublish scripts publish typescript files or javascript files? How to teams without typescript make it work? Will it automatically convert to JS when publishing with babel? 
2. Where is the right place to put input interface exports? I see some places its component, others its template (I assume thats because there is no component). I find that the template seems to be more of a place to hold input definitions (that's where they are commonly used), and that state variables should be declared in component. However I'm open to whatever. 
3. I mentioned it in one place, but I definitely think our examples should be almost free of typescript, or at least maybe a typescript version and a regular one. People tend to copy and paste this code in their app, so if it's not configured correctly, they will get errors (and we'll get slack messages). I think at least for the first pass, lets keep all examples as base marko as we can. (I'm fine having TS examples as well, as long as they are called out and not the default ones) Once people start adopting TS more, then we can switch to typescript in our examples. 
4. Our main thing when we started ebayui was to make sure it is as close to HTML syntax as you can get. We still want to maintain that ideology. I noticed all our examples are now changed to use camel case in templates. As a general rule, we want to follow the HTML dash case when we can. (I also commented on this in the code)
5. I noticed no `marko.json` files have been modified. Do we need those? It seems almost redundant to have the types declared in two places. Maybe we need to loop in Dylan with this.
6. The main thing: are there any breaking changes? Will teams be able to use ebayui _as is_ without any changes to their apps?

darkwebdev:
Hey guys, just an unsolicited advice: I would suggest to use TS in storybook and tests so that you can find (potential) problems in types earlier than the users who will try to use those types.

.nycrc
```diff
@@ -1,12 +1,12 @@
 {
-    "include": ["src/**/*.js"],
+    "include": ["src/**/*.js", "src/**/*.ts"],
```
montoya332, line 2: woud [jt]s also work

agliga (approved):
Approving for now. 
From what I see it looks fine. If we have any other changes we can do it in a followup

agliga (approved):

agliga (approved):

Issue #1917, opened 2023-06-07
combobox: storybook not working with list selection manual


saiponnada:
# Bug Report

## Description

In eBay UI storybook when list-selection value is populated with `manual` the combobox component is not reflected properly.

## Workaround

This is because of the typo in the storybook configuration `list-selection` needs to be changed to `listSelection`

Issue #1912, opened 2023-05-24
drawer-dialog: aria labelledby is not mapped to heading id


saiponnada:
# Bug Report

## Description

Dialog header id is not properly reflected on the dialog's `aria-labelledBy`

## Screenshots

<img width="1324" alt="image" src="https://github.com/eBay/ebayui-core/assets/6342519/f61c0eab-060e-4915-a499-17e3bc7cf850">

PR #1918, opened 2023-06-12
fix: storybook a11y bug fixes

saiponnada:
## Description

- Fixed an issue in storybook for combobox where list-selection manual is not working as expected.
- Fixed an issue in storybook for drawer dialog where aria-labelledby is not mapped to drawer dialog title.

## References

#1912, #1917

## Screenshots

**Combobox**
image

**Drawer dialog**
image

### DIFF
--- a/src/components/ebay-combobox/combobox.stories.js
+++ b/src/components/ebay-combobox/combobox.stories.js
@@ -48,7 +48,7 @@ export default {
             description:
                 "default is `none`; available values are `none` or `list`. For list, will automatically filter results while typing.",
         },
-        "list-selection": {
+        listSelection: {
             control: { type: "text" },
             description:
                 "default is `automatic`; available values are `automatic`, `manual`. If set to automatic will automatically fill in the input with the currently highlighted item when using the up/down keys.",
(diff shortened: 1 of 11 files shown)


src/components/ebay-combobox/combobox.stories.js
```diff
@@ -48,7 +48,7 @@ export default {
             description:
                 "default is `none`; available values are `none` or `list`. For list, will automatically filter results while typing.",
         },
-        "list-selection": {
+        listSelection: {
             control: { type: "text" },
```
ArtBlue, line 52: This is likely out of the scope for this issue, but this control type should probably be `type: "select"` since there are known values that can be set.

src/components/ebay-combobox/combobox.stories.js
```diff
@@ -48,7 +48,7 @@ export default {
             description:
                 "default is `none`; available values are `none` or `list`. For list, will automatically filter results while typing.",
         },
-        "list-selection": {
+        listSelection: {
```
ArtBlue, line 51: I'm not sure why changing the name actually makes this work since there are other options that have the same naming convention. Maybe some internals were not aligned with the name?

ArtBlue (approved):
Looks good overall. Just a couple of notes.

src/components/ebay-combobox/combobox.stories.js
saiponnada: Even I was surprised with that. I see at some places where this works just fine. may be @agliga can shed some light.

PR #1925, opened 2023-06-21
fix(calendar): stop displaying completely disabled months

### LINKED ISSUE #1924: calendar: initially load on the first month that has selectable dates
# Bug Report

 If I were to initialize the component with
"disableBefore": "2023-07-05",
"disableAfter": "2023-07-20"

with current date is 6/20. I see that the component is showing the month of June with all dates disabled, is there a way to have to it showing July at the start?

## eBayUI Version: x.x.x
"@ebay/ebayui-core@
  version "12.1.1"

## Description

Options proposed: 

I believe the intended behavior was to prevent users from navigating to months with only disabled dates, but if June should be shown then this needs to be rethought. The way it stands, I think there are two options:

- fix the issue as you initially proposed it, and continue preventing users from navigating to June
- keep the initial focus on the current month, and always allow for navigation up to and including the current month even if it is completely disabled

Ultimately we want lets work for our users.  Let’s go with option 1, which is more intuitive.  

In summary, please load the calendar on the month with selectable dates.  User would not be able to navigate to prior months with disabled dates.  

Thanks Luke!

## Workaround
n/a

## Screenshots
image

LuLaValva:
- Fixes #1924 

## Description

If `disableBefore` or `disabledAfter` were used to disable the entire current month, the calendar started by showing a month that was completely disabled. Now we move to the closest month in the valid range to the current month.

### DIFF
--- a/src/components/ebay-calendar/component.js
+++ b/src/components/ebay-calendar/component.js
@@ -106,7 +106,26 @@ export default class extends Marko.Component {
             input.disableList?.map(dateArgToISO) ?? []
         );
         if (this.isDisabled(this.state.tabindexISO)) {
-            this.state.tabindexISO = this.getFirstActiveISO(input);
+            const firstActive = this.getFirstActiveISO(input);
+            if (firstActive) {
+                this.state.tabindexISO = firstActive;
+            } else if (
+                this.state.disableBefore &&
+                this.state.tabindexISO < this.state.disableBefore
+            ) {
+                this.state.baseISO = this.state.disableBefore;
+                this.state.offset = 0;
+                this.state.tabindexISO =
+                    this.getFirstActiveISO(input) ?? this.state.disableBefore;
+            } else if (
+                this.state.disableAfter &&
+                this.state.tabindexISO > this.state.disableAfter
+            ) {
+                this.state.baseISO = this.state.disableAfter;
+                this.state.offset = 0;
+                this.state.tabindexISO =
+                    this.getFirstActiveISO(input) ?? this.state.disableAfter;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/components/ebay-calendar/component.js
```diff
@@ -106,7 +106,26 @@ export default class extends Marko.Component {
             input.disableList?.map(dateArgToISO) ?? []
         );
         if (this.isDisabled(this.state.tabindexISO)) {
-            this.state.tabindexISO = this.getFirstActiveISO(input);
+            const firstActive = this.getFirstActiveISO(input);
```
ArtBlue, line 109: Let's get some comments in here to help understand the long conditional blocks.

src/components/ebay-calendar/component.js
```diff
@@ -106,7 +106,26 @@ export default class extends Marko.Component {
...
+                this.state.tabindexISO = firstActive;
+            } else if (
+                this.state.disableBefore &&
+                this.state.tabindexISO < this.state.disableBefore
+            ) {
+                this.state.baseISO = this.state.disableBefore;
+                this.state.offset = 0;
+                this.state.tabindexISO =
+                    this.getFirstActiveISO(input) ?? this.state.disableBefore;
+            } else if (
```
ArtBlue, line 120: Is this `if/else if` exhaustive of all potential options? In other words, is there any other possibility outside of the conditions covered in the `if/else if`?

ArtBlue (changes requested):

src/components/ebay-calendar/component.js
LuLaValva: There actually is an additional undocumented edge case that I left unaddressed here, which we may have to include as an anti-pattern. Suppose we have the following component:
```marko
<ebay-calendar disable-before="2023-06-30" disable-list=["2023-06-30"] />
```
In this case, there is still one day in June which is within the possible range, but that day has been _manually disabled_. I believe it is very unlikely that this problem will ever come up, but if it does the solution is to expand `disable-before`:
```marko
<ebay-calendar disable-before="2023-07-01" />
```

ArtBlue: If that is the only possible alternative, we should address it in an `else` block. and comment what condition that block is covering.

LuLaValva: I think this should be left as a fall-through case where we stay in the current month even though all days are disabled, because if we start searching for a day we may cause infinite loops unless we are very careful. I still added an `else` with documentation, but did not execute any code in it.

ArtBlue: Ok, that's a fair compromise. The comment in the `else` is helpful.

ArtBlue (approved):

PR #1928, opened 2023-06-22
v12.3.0

ArtBlue:
Release PR for `12.3.0`


LuLaValva (approved):
Seems to be working on my end

ianmcburnie (commented):
One issue I noticed in Storybook is that if I select Toggle-Button's With Image story and then change layout from gallery to minimal, under the hood the html gets changed to `toggle-button--list-layout`.

PR #1929, opened 2023-06-26
feat(confirm-dialog): added destructive CTA button variant

### LINKED ISSUE #1919: confirm-dialog: add destructive button option for primary cta
# Feature Request

## Description

For the confirm dialog, add the ability for users to use the destructive button for the primary cta button instead of the typical primary button.

Note: the destructive modifier already exists in Skin. The optional flag would simply append the modifier class.

## Context

Some confirm dialogs are confirming a "destructive" action. As such, we need to support this use case.

ArtBlue:
Fixes #1919 

## Description

Added variant option to the confirm dialog main CTA button with the first supported variant of `destructive`.

## Context

I allowed flexibility to add other potential variants for better scalability.

## Screenshots

image

### DIFF
--- a/src/components/ebay-confirm-dialog/confirm-dialog.stories.js
+++ b/src/components/ebay-confirm-dialog/confirm-dialog.stories.js
@@ -40,7 +40,7 @@ export default {
             control: { type: "text" },
             description: "Text for reject button",
         },
-        variant: {
+        "confirm-cta-variant": {
             options: ["none", "destructive"],
             description: "The alternative main CTA button variant to use.",
             table: {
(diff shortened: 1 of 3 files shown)


LuLaValva:
Looks great, but should this be called something `confirm-cta-variant` instead? That would allow for us to eventually modify styles for the reject button as well, and any additional buttons that may be added.

ArtBlue:
> Looks great, but should this be called something like `confirm-cta-variant` instead? That would allow for us to eventually modify styles for the reject button as well, and any additional buttons that may be added.

Good call. Done.

LuLaValva (approved):
Ok, looks good. Thanks @ArtBlue

PR #1944, opened 2023-07-20
feat(toggle-button-group): add toggle button group

### LINKED ISSUE #1908: toggle button group - new component
ebayui portion of [#2060](https://github.com/eBay/skin/issues/2060)

LuLaValva:
- Resolves #1908 

> **Warning**
> Must be tested while linked with the `2060-toggle-button-group` branch in `skin`.

## Description

- Add toggle button group to sync with ebay/skin#2095
- I snuck in a few changes for `toggle-button` as well, hopefully they're not too invasive.

## Screenshots

image
image

### DIFF
--- a/src/components/ebay-toggle-button-group/examples/icons.marko
+++ b/src/components/ebay-toggle-button-group/examples/icons.marko
@@ -1,30 +1,42 @@
 <ebay-toggle-button-group columns=3 layoutType="list" ...input>
-    <@button title="Icon 1" subtitle="Car">
+    <@button>
+        Icon 1
+        <@subtitle>Car</@subtitle>
         <@icon>
             <ebay-car-24-icon/>
         </@icon>
     </@button>
-    <@button title="Icon 2" subtitle="ATV">
+    <@button>
+        Icon 2
+        <@subtitle>ATV</@subtitle>
         <@icon>
             <ebay-atv-24-icon/>
         </@icon>
     </@button>
-    <@button title="Icon 3" subtitle="Cart">
+    <@button>
+        Icon 3
+        <@subtitle>Cart</@subtitle>
         <@icon>
             <ebay-cart-24-icon/>
         </@icon>
     </@button>
-    <@button title="Icon 4" subtitle="Jet Ski">
+    <@button>
+        Icon 4
+        <@subtitle>Jet Ski</@subtitle>
         <@icon>
             <ebay-jet-ski-24-icon/>
         </@icon>
     </@button>
-    <@button title="Icon 5" subtitle="Motorcycle">
+    <@button>
+        Icon 5
+        <@subtitle>Motorcycle</@subtitle>
         <@icon>
             <ebay-motorcycle-24-icon/>
         </@icon>
     </@button>
-    <@button title="Icon 6" subtitle="Snowmobile">
+    <@button>
+        Icon 6
+        <@subtitle>Snowmobile</@subtitle>
         <@icon>
             <ebay-snowmobile-24-icon/>
         </@icon>
--- a/src/components/ebay-toggle-button-group/examples/icons.marko
+++ b/src/components/ebay-toggle-button-group/examples/icons.marko
@@ -1,0 +1,32 @@
+<ebay-toggle-button-group columns=3 layoutType="list" ...input>
+    <@toggle-button title="Icon 1" subtitle="Car">
+        <@icon>
+            <ebay-car-24-icon/>
+        </@icon>
+    </@toggle-button>
+    <@toggle-button title="Icon 2" subtitle="ATV">
+        <@icon>
+            <ebay-atv-24-icon/>
+        </@icon>
+    </@toggle-button>
+    <@toggle-button title="Icon 3" subtitle="Cart">
+        <@icon>
+            <ebay-cart-24-icon/>
+        </@icon>
+    </@toggle-button>
+    <@toggle-button title="Icon 4" subtitle="Jet Ski">
+        <@icon>
+            <ebay-jet-ski-24-icon/>
+        </@icon>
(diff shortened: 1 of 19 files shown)


src/components/ebay-toggle-button-group/index.marko
```diff
@@ -0,0 +1,19 @@
+import { processHtmlAttributes } from "../../common/html-attributes";
+static var ignoredAttributes = ["class", "columns"];
+
+<ul
+    class=["toggle-button-group", input.class]
+    data-columns=input.columns
+    ...processHtmlAttributes(input, ignoredAttributes)
+>
+    <for|{ layoutType = input.layoutType, ...buttonInput }, i| of=input.buttons>
```
agliga, line 9: Good! I assume this removes layoutType in buttonInput and then we don't need to pass ignoredArrtibutes anymore. 
Maybe we should do something in other places. I like this approach here (and I cant stand the whole ignoredAttributes object)

agliga (commented):
LGTM overall.
We should add some snapshot tests and some interaction tests. I can help you out with those if needed.

LuLaValva (commented):
Thanks for the review, all good changes. Quick note:

With the `pressed` implementation that you suggested, the values that are selected are currently reset when _any_ part of input changes (not just the values which are pressed). This is also the case for components that use `menuUtils` so I figured it would be better to stay consistent, but I could also store the last default and only reset when it is changed.

src/components/ebay-toggle-button-group/examples/icons.marko
```diff
@@ -0,0 +1,32 @@
+<ebay-toggle-button-group columns=3 layoutType="list" ...input>
+    <@toggle-button title="Icon 1" subtitle="Car">
```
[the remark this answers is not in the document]
agliga: We should change it there. 
For now lets keep title support so its not breaking. Just do a check if title is present render with that otherwise renderBody
And remove title from docs. 
Just one more question. Should subtitle be an attribute tag? Could there be anything in subtitle other than text?

src/components/ebay-toggle-button-group/toggle-button-group.stories.js
```diff
@@ -0,0 +1,123 @@
...
+    component,
+    parameters: {
+        docs: {
+            description: {
+                component: readme,
+            },
+        },
+    },
+    argTypes: {
+        radio: {
```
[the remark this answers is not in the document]
agliga: I agree (and i missed this part where radio is a boolean)
In general we want to have more explicit options in ebayui, even if we have only two variabts
Like variant=radio instead of radio=true

This is so if things change we have less breaking changes. 
But in this case we need variant since there are three.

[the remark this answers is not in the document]
LuLaValva: Ok, I called the third option `single` but I'm not super happy with that name

src/components/ebay-toggle-button-group/toggle-button-group.stories.js
```diff
@@ -0,0 +1,123 @@
...
+            type: "boolean",
+            control: { type: "boolean" },
+            description: "Whether the buttons act as radio buttons",
+        },
+        columns: {
+            type: "number",
+            control: { type: "number" },
+            description: "Suggested number of columns",
```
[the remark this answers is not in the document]
LuLaValva: Ok, I'll switch to keep it consistent with skin but I'm not super happy with "preferred minimum" either.

LuLaValva (commented):
Thanks for the comments! I'm going to keep the current "reset" behavior on input change instead of doing the comparison to keep it consistent with the rest of eBayUI, and we can point people to the "Controlled" example I added in the last commit.

src/components/ebay-toggle-button-group/toggle-button-group.stories.js
```diff
@@ -0,0 +1,125 @@
...
+            description: {
+                component: readme,
+            },
+        },
+    },
+    argTypes: {
+        variant: {
+            type: "string",
+            control: { type: "select" },
+            options: ["checkbox", "radio", "single"],
```
ArtBlue, line 30: @LuLaValva , I agree that's not a great name; too oversimplified. Are there any concerns over user confusion about these variant names as a whole? Is it not better to avoid confusion with native behavior and avoid "checkbox" and "radio?" Why not keep it simple and aligned with Skin - 

1. Multi-select
2. Single-select optional
3. Single-select required

?

I think these options provide the most clarity directly instead of going through detours of other patterns for analogous behaviors that users will need to make connections with.

LuLaValva: After discussing offline, we decided to go with `checkbox`, `radio`, and `radio-toggle`

ArtBlue (approved):

src/components/ebay-toggle-button-group/toggle-button-group.stories.js
```diff
@@ -0,0 +1,125 @@
...
+export const Default = Template.bind({});
+Default.args = {
+    columns: 3,
+    buttons: [
+        { title: "Button 1" },
+        { title: "Button 2" },
+        { title: "Button 3" },
+        { title: "Button 4" },
+        { title: "Button 5" },
+        { title: "Button 6" },
```
agliga, line 76: Lets also change these.

agliga (commented):
LGTM just one comment but should be quick. Ready to approve otherwise

src/components/ebay-toggle-button-group/toggle-button-group.stories.js
LuLaValva: Ok, updated. I also snuck in a little title update for one of the storybook pages.

PR #1948, opened 2023-07-26
fix(calendar): issue with time zones

LuLaValva:
## Description

JavaScript's native `Date` constructor does not always work as was expected across time zones, so `offsetISO` had a bug where in some cases it was one day behind. This meant that `offsetISO(date, 1)` sometimes returned the original date instead of the next day, which led to infinite loops in a few places.


agliga (approved):
That's great! Nice and easy. It makes sense since date will interpret the GMT timezone part.

PR #1951, opened 2023-08-01
fix: added button type "button" for ebay-drawer-dialog handle

### LINKED ISSUE #1953: ebay-drawer-dialog: fix handel not having type="submit" and thus submitting forms
# Bug Report

See more details in the PR: https://github.com/eBay/ebayui-core/pull/1951

ixam1:
## Description

Added `type="button"` to the ebay-drawer-dialog handle. Since by default without type any button inside a form will be treated as a "submit" button. This causes the handle to unintentionally submit a form wrapped around the dialog.

The close icon correctly already has `type="button"`:
https://github.com/eBay/ebayui-core/blob/84ba5f3ddb47105d4d3a0baba71941c181c6ef25/src/components/components/ebay-dialog-base/index.marko#L50-L69

## Context

I wrapped the drawer-dialog with a form, now the handle triggers the form submit, unless I specifically prevent it. What I cant prevent is that when in a form field enter is pressed, instead of submitting the form, the handle gets triggered.

Wrapping the dialog in a form is kind of necessary if you want to have the submit buttons in the footer, since the footer is only allowed as a direct child of the dialog component.

In any case, I think this change should be a no-brainer.

## Issue

fixes #1953

### DIFF
--- a/src/components/ebay-drawer-dialog/index.marko
+++ b/src/components/ebay-drawer-dialog/index.marko
@@ -23,6 +23,7 @@ $ var handleLabel = (!state.expanded ?
             <@top>
                 <button
                     aria-label=handleLabel
+                    type="button"
                     class=`drawer-dialog__handle`
                     onClick("handleExpand")
                     onTouchStart("handleTouchStart")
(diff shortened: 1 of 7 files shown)


agliga:
Thank you for your submission. It looks fine from my end.
Can you file an issue so we can track it
And can you target branch `12.5.0`?

ixam1:
> Thank you for your submission. It looks fine from my end. Can you file an issue so we can track it And can you target branch `12.5.0`?

Both done, hope the minimalist issue description is fine

agliga (approved):

PR #1957, opened 2023-08-04
fix(date-textbox): properly handle textbox changes based on index

### LINKED ISSUE #1952: ebay-date-textbox end range value is reset on blur
# Bug Report

## Description

After entering a start range and end range value and blurring the input, the start range value gets set to the end range value. At the same time the end range value is reset.

Expected behavior: start range value stays the same, end range value does not get reset.

## Workaround

Currently using the calendar view popover as the workaround.

## Screenshots

https://github.com/eBay/ebayui-core/assets/12465178/f4c8b84a-2946-44d6-a8ed-6cf283b7ade6

LuLaValva:
- Fixes #1952 

## Description

This ended up being based on an embarrassingly simple mistake 🙈. Quick one-line fix.

### DIFF
--- a/src/components/ebay-date-textbox/index.marko
+++ b/src/components/ebay-date-textbox/index.marko
@@ -27,7 +27,7 @@ $ {
         placeholder=mainPlaceholder
         buttonAriaLabel=openPopoverText
         value=input.range ? state.secondSelected : state.firstSelected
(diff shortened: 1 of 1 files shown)


LuLaValva:
I reran the CI last night, seemed to fix the issue.

Issue #1959, opened 2023-08-04
ebay-confirm-dialog: changing the confirm button variant is not working


sauermar:
# Bug Report

## Description
`<ebay-confirm-dialog>` component has `confirm-cta-variant` attribute for changing the confirm button variant.
This is the related code from the ebayui-core library:

```javascript
<ebay-button
            priority="primary"
            on-click("emit", "confirm")
            id=confirmId
            aria-describedby=mainId
            class=[
                "confirm-dialog__confirm",
                input['confirm-cta-variant'] && input['confirm-cta-variant'] !== 'none' && `btn--${input['confirm-cta-variant']}`
            ]>
            ${input.confirmText}
        </ebay-button>
```

The problem is in the line 
```
input['confirm-cta-variant'] && input['confirm-cta-variant'] !== 'none' && `btn--${input['confirm-cta-variant']}`.
```
Marko automatically transforms the input from: 
```
{
   confirm-cta-variant: "destructive",
}
```
into:
```
{
    confirmCtaVariant: 'destructive'
}
```

this means that anywhere we try to access the `input['confirm-cta-variant']` it results in `undefined` and thus the line will always result in `false`.

Please change the code to use either `input.confirmCtaVariant` or `input['confirmCtaVariant']` syntax everywhere.

It would also maybe make sense to revisit the idea behind giving the options `destructive` and `none` when `undefined` equals to `none`.

## Workaround

You can force the variant through styling the button in css.

PR #1963, opened 2023-08-09
fix(confirm-dialog): updated input of confirm to be proper marko syntax

agliga:
## Description
Changed the input for confirm dialog to be correct camel case. 
Also updated tests

## References
https://github.com/eBay/ebayui-core/issues/1959


LuLaValva (approved):
I like it, thanks for the light refactors in addition to the storybook changes! Makes a big difference.

ArtBlue (approved):

PR #1965, opened 2023-08-11
feat(character-count): create component

### LINKED ISSUE #1921: field: add helper text sections
# Feature Request

## Description

Add helper text sections in `field`.

## Context

This is the ebay ui companion to the Skin issue: https://github.com/eBay/skin/issues/2047

## Screenshots

image

LuLaValva:
- Resolves #1921

## Description

- Add `ebay-character-count` component (edit: renamed from `char-count`)

## Context

- After lengthy discussion offline with @agliga and @ianmcburnie we have decided to create a new character count component instead of adding it to textbox, as was done in #1956.
- This may require more ebay developer work than the alternative for accessibility, but it allows for more flexibility than the alternative _and_ it doesn't require that all textboxes become stateful.

## Screenshots
image
image
image


ianmcburnie:
Can we rename it to `ebay-character-count`. One very early call we made with skin and coreui is to avoid abbreviations in our naming where possible (I'm still not happy about button being btn in skin, ha ha)

LuLaValva:
> Can we rename it to `ebay-character-count`

Sure thing! Updated.

agliga (commented):
Can we also add some simple tests?
Lgtm otherwise

agliga (approved):

LuLaValva:
Looks good to merge @agliga, thanks for catching the `typeof window` check.

Issue #1962, opened 2023-08-09
ebay-video: remove index.marko references


agliga:
Should remove index.marko require. Should have it render in template and then clone the icon code into where it needs to go.

Issue #1960, opened 2023-08-08
ebay-video : expose ended event when shaka video ends


agunaganti:
# Bug Report

## Description

we were trying to add tracking whenever user tries to replay a ebay-video, to do so would like to check , if it is possible to emit ended event as soon as video ends.

**Expected:**
ebay-video should expose - ended event  by adding "ended" to this list: https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-video/component.js#L251-L254

**Actual:**
ended event is not exposed today

## Screenshots

Today Shaka player exposes ended event
<img width="434" alt="Screen Shot 2023-08-08 at 2 20 22 PM" src="https://github.com/eBay/ebayui-core/assets/110565292/3ee02103-5ec2-466f-9fc8-db2335cd7df2">

PR #1966, opened 2023-08-11
fix(video): added all shaka events and moved icons to be in template

agliga:
## Description

* As per marko team changed video tag to import icons from template (load the icons in the template and then clone them to use them in other places)
* Added all shaka events to video. 

## References
https://github.com/eBay/ebayui-core/issues/1960
https://github.com/eBay/ebayui-core/issues/1962

### DIFF
--- a/src/components/ebay-video/index.marko
+++ b/src/components/ebay-video/index.marko
@@ -66,4 +66,8 @@ static var ignoredAttributes = [
             <ebay-progress-spinner a11y-text=(input.a11yLoadText || "Loading")/>
         </else>
     </div>
+    <div style="display:none">
+        <ebay-play-64-colored-icon key="play-icon"/>
+        <ebay-flag-24-icon key="flag-icon"/>
+    </div>
 </div>
--- a/src/components/ebay-video/component.js
+++ b/src/components/ebay-video/component.js
@@ -1,6 +1,26 @@
 import { CDNLoader } from "../../common/cdn";
-import { getElements, playIcon } from "./elements";
+import { getElements } from "./elements";
 const DEFAULT_SPINNER_TIMEOUT = 2000;
+
+const eventList = [
+    "abort",
+    "canplay",
+    "canplaythrough",
+    "durationchange",
+    "emptied",
+    "encrypted",
+    "ended",
+    "error",
+    "loadstart",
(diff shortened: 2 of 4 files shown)


LuLaValva (commented):
Looks awesome! Just a few suggestions while you're touching events for this component:

- it looks like `onPause` is missing from storybook and also `onVolumeChange` should be switched to `"onVolume-change"` in `video.stories.js`
- Emitting `volume-change` instead of `volumechange` is a little inconsistent, maybe you could add `volumechange` to your new `eventList` and emit both for now?

agliga:
> Looks awesome! Just a few suggestions while you're touching events for this component:
> 
> * it looks like `onPause` is missing from storybook and also `onVolumeChange` should be switched to `"onVolume-change"` in `video.stories.js`
> * Emitting `volume-change` instead of `volumechange` is a little inconsistent, maybe you could add `volumechange` to your new `eventList` and emit both for now?

I'll add on pause. 
The others I wont change because we are using a custom event because we do some state changes on volume change. We can revisit next breaking but we need to leave it as is.

src/components/ebay-video/index.marko
```diff
@@ -66,4 +66,8 @@ static var ignoredAttributes = [
             <ebay-progress-spinner a11y-text=(input.a11yLoadText || "Loading")/>
         </else>
     </div>
+    <div style="display:none">
```
ArtBlue, line 69: Is this supposed to be inlined? Since it's hidden and I'm not seeing anything that would hook into it and change the display, what's the reason why it's there?

ArtBlue (commented):
Looks good. Just a question about the hidden element.

LuLaValva:
> I wont change because we are using a custom event because we do some state changes on volume change

Ok, that makes sense but can you change `onVolumeChange` to `"onVolume-change"` so the event is announced on storybook?

ArtBlue (approved):

agliga:
> > I wont change because we are using a custom event because we do some state changes on volume change
> 
> Ok, that makes sense but can you change `onVolumeChange` to `"onVolume-change"` so the event is announced on storybook?

Fixed this. I'll go ahead and merge. If there are any other issues I can do it in a followup.

Issue #1934, opened 2023-06-29
ebay-textbox: missing callbacks


darkwebdev:
# Bug Report

## Description
When user goes to storybook: https://opensource.ebay.com/ebayui-core/?path=/story/form-input-ebay-textbox--with-label
and edits the textbox field no actions are registered.

Expected:
<img width="525" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/452ae242-010e-4cff-a7af-e5024d4daef5">

Result:
<img width="793" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/f652cd45-8091-4d89-9ab1-e940923a1bc5">

Issue #1933, opened 2023-06-28
ebay-switch: missing callback


darkwebdev:
# Bug Report
When user clicks a switch in storybook no callback is registered.
https://opensource.ebay.com/ebayui-core/?path=/story/form-input-ebay-switch--with-label

## Description
Expected:
<img width="473" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/359c9983-25fd-4ffd-a898-2e72215e4777">

Result:
<img width="772" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/e24fc19b-036e-4389-8737-2b59120a7a7f">

Issue #1932, opened 2023-06-28
ebay-radio: missing callbacks


darkwebdev:
# Bug Report

## Description
In storybook when user clicks different radio buttons no action are registered.
https://opensource.ebay.com/ebayui-core/?path=/story/form-input-ebay-radio--group

Expected:
<img width="488" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/fb570822-9642-4c6d-86cc-16242514f85f">

Result:
<img width="791" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/65543291-0bdb-48f0-9f1b-b44e9f8f54cf">

Issue #1931, opened 2023-06-28
ebay-pagination: missing callbacks


darkwebdev:
# Bug Report

## Description
Maybe I'm doing something wrong, but when I click prev/next arrow button no callback is registered in storybook.
https://opensource.ebay.com/ebayui-core/?path=/story/navigation-disclosure-ebay-pagination--interactive

Expected:
<img width="466" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/d2f7f3e1-9aa2-46d6-8aea-b288f7f575b8">

Result:
<img width="796" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/02fc2e38-c403-4571-aef0-afd642655eed">

PR #1967, opened 2023-08-15
chore(storybook): exposed events

agliga:
## Description
* Added all events to storybook templates that had been missing them. 

## References

#1931 
#1932 
#1933 
#1934


ArtBlue (approved):
Looks great! Do we need tests to ensure ensure events are emitting properly?

agliga:
> Looks great! Do we need tests to ensure ensure events are emitting properly?

This is simply docs changes, so no tests needed.

PR #1972, opened 2023-08-29
feat(date-textbox): add collapseOnSelect

### LINKED ISSUE #1971: [collapseOnSelect] ebay-date-textbox popover should be dismissed on selecting the date.
# Feature Request
Provider option to collapse the calendar popover on selecting the date.

## Description

When a date is selected in the Date Picker, the Date Picker remains active. When I select a date, I expect for the Date Selector to go away so I don't have to find another way to dismiss that control. Can we add `collapseOnSelect` input property to this component to enable this behavior.

LuLaValva:
- Fixes #1971 

## Description

- Add `collapseOnSelect` option for date-textbox
- Defaults to `false` so this is not breaking, but maybe in the future it should default to `true` for single select and `false` for range


agliga (approved):
I know we discussed refactoring this, but lets leave it like this now now. We can revisit.

Issue #1930, opened 2023-06-28
ebay-page/section-notice: wrong callback arguments


darkwebdev:
# Bug Report

## Description
When user clicks `dismiss` component object is passed as an argument instead of original event.
https://opensource.ebay.com/ebayui-core/?path=/story/notices-tips-ebay-page-notice--with-dismiss
https://opensource.ebay.com/ebayui-core/?path=/story/notices-tips-ebay-section-notice--with-dismiss

Expected according to storybook:
<img width="404" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/484e6f84-a8a1-42c2-b2d6-3c8d6019235a">

Reality:
<img width="244" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/78c70486-682f-4f36-9ff7-73aaad2051de">

Issue #1890, opened 2023-04-20
ebay-drawer-dialog: callbacks use wrong arguments


darkwebdev:
# Bug Report

## Description
Documentation reads that:
- `onOpen/onClose/onExpand/onCollapse` callbacks should use `{ originalEvent }` as argument

but in storybook I get something else:
<img width="576" alt="image" src="https://user-images.githubusercontent.com/2737310/233406305-655b4507-8b17-418c-a0ca-d9b856cfd487.png">

saiponnada:
Not an issue. The events listed above use `{ originalEvent }` as argument

darkwebdev:
If you look at the labels of this issue you can notice an aspect called `documentation`. It's a hint that code and documentation have discrepancies. If you look at the screenshot you can see that there are two arguments for on-open callback: `0: ComponentClass and 1: ComponentClass`. Which of them is `OriginalEvent` and what is another one and why does it look different from other components (maybe there's a bug somewhere, hopefully only in documentation)?

If I'm spending my time on creating an issue please try to understand the reason behind it and don't just close it. If you have questions go ahead and ask me. I'm going to create more issues regarding similar problems with other components and I don't want to waste my time.

Thank you in advance.

saiponnada:
ebay-drawer-dialog will be **deprecated** and removed from the ebay UI core in next release.

ianmcburnie:
Hmm, not sure why this is closed. It doesn't sit well with me that we are keeping a known documentation discrepancy around. I can understand @darkwebdev 's frustration here. This should be a quick fix unless I'm missing something?

Issue #1887, opened 2023-04-19
ebay-confirm-dialog: onOpen arguments are wrong


darkwebdev:
# Bug Report

## Description
Readme says that `onOpen` callback should use `{ originalEvent }` as argument, but when I click/keypress in storybook I get this: `on-open: {name: "on-open", args: undefined}`

Not sure if that's a bug or readme mistake.

saiponnada:
Not an issue. The events listed above use `{ originalEvent }` as argument as per documentation.

darkwebdev:
I'm not saying that code is wrong, maybe storybook documentation is not updated. I don't see any properties of `OriginalEvent` here:

<img width="1005" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/10d5cedf-09c3-4763-91ea-c1c9a0a1c18a">

PR #1977, opened 2023-08-31
docs: storybook sync up with event arguments

saiponnada:
## Description

Updated storybook documentation to sync up with the component events. Both DOM and component events are being emitted and storybook displays these events in actions tab. During this process for some component events, marko class information is being emitted. We decided to leave that extra information as is. Other-than that, the callback arguments were appropriately corrected in the storybook documentation.

## References
Fixes following 
#1930
#1892 
#1887
#1890

## Screenshots

**Page/Section notice**
image
image

**Fake menu button**
image

**Confirm dialog**
image

**Drawer dialog**
image

### DIFF
--- a/src/components/ebay-fake-menu-button/component.js
+++ b/src/components/ebay-fake-menu-button/component.js
@@ -46,12 +46,9 @@ export default {
             originalEvent,
             index,
         };
-        eventObj = Object.entries(eventObj)
-            .filter(([, value]) => value !== undefined)
-            .reduce((obj, [key, value]) => {
-                obj[key] = value;
-                return obj;
-            }, {});
+        eventObj = Object.fromEntries(
+            Object.entries(eventObj).filter(([, value]) => value !== undefined)
+        );
         this.emit(`${eventType}`, eventObj);
     },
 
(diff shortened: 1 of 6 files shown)


src/components/ebay-fake-menu-button/component.js
```diff
@@ -41,11 +41,17 @@ export default {
         this.emitComponentEvent({ eventType: "mousedown", el, originalEvent });
     },
     emitComponentEvent({ eventType, el, originalEvent, index }) {
-        const eventObj = {
+        let eventObj = {
             el,
             originalEvent,
             index,
         };
+        eventObj = Object.entries(eventObj)
```
LuLaValva, line 49: If you'd like you can use `Object.fromEntries` here instead of `reduce`. I think it's a little cleaner but it is personal preference.

LuLaValva (approved):
Looks good to me! Just one super minor suggestion, but I don't mind if this is merged without changing

src/components/ebay-fake-menu-button/component.js
saiponnada: I like it. Looks cleaner. Rolled out the change.

PR #1979, opened 2023-09-01
fix(calendar): move utilities into separate file for Marko 4

### LINKED ISSUE #1978: Calendar: Marko 4 does not handle exports properly
# Bug Report

## Description

Calendar does not work in Marko 4 because it has multiple exports from its `component.js` file. This is a bug in Marko, not eBayUI, but we should fix it here for now until we find a fix on the Marko end.

LuLaValva:
- Fixes #1978 

## Description

- Move utility functions to separate file

### DIFF
--- a/src/components/ebay-calendar/component.js
+++ b/src/components/ebay-calendar/component.js
@@ -1,5 +1,14 @@
 // @ts-check
 
+import {
+    dateArgToISO,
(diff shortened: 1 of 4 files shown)


ArtBlue (approved):
LGTM

PR #1981, opened 2023-09-06
feat(chip): add chip component

### LINKED ISSUE #1973: ebay-chip: add new component
Original issue https://github.com/eBay/skin/issues/2108

LuLaValva:
- Resolves #1973 

> **Note**
> - Must be tested in tandem with ebay/skin#2141

## Description

- Add `ebay-chip` component to ebayui-core

## Screenshots

image
image
image

### DIFF
--- a/src/components/ebay-chip/chip.stories.js
+++ b/src/components/ebay-chip/chip.stories.js
@@ -1,0 +1,67 @@
+import { tagToString } from "../../../.storybook/storybook-code-source";
+import { addRenderBodies } from "../../../.storybook/utils";
+import Readme from "./README.md";
+import Component from "./index.marko";
+
+const Template = (args) => ({
+    input: addRenderBodies(args),
+});
+
+export default {
+    title: "building blocks/ebay-chip",
+    component: Component,
+    parameters: {
+        docs: {
+            description: {
+                component: Readme,
+            },
+        },
+    },
+
+    argTypes: {
+        renderBody: {
+            control: { type: "text" },
+            description: "Text to be displayed in the chip",
+        },
+        deleteButtonLabel: {
+            control: { type: "text" },
+            description:
+                "A11y text for the delete button, also determines if delete button is shown",
+        },
+        onDelete: {
+            action: "on-delete",
+            description: "Triggered when delete button is clicked",
+            table: {
+                category: "Events",
+                defaultValue: {
+                    summary: "{ el, checked, originalEvent }",
+                },
+            },
+        },
+    },
+};
+
+export const Standard = Template.bind({});
+Standard.args = {
+    renderBody: "chip text",
+};
+Standard.parameters = {
+    docs: {
+        source: {
+            code: tagToString("ebay-chip", Standard.args),
+        },
+    },
+};
+
+export const WithDelete = Template.bind({});
+WithDelete.args = {
+    renderBody: "chip text",
+    deleteButtonLabel: "Delete",
+};
+WithDelete.parameters = {
+    docs: {
+        source: {
+            code: tagToString("ebay-chip", Standard.args),
+        },
+    },
+};
--- a/src/components/ebay-chip/chip.stories.js
+++ b/src/components/ebay-chip/chip.stories.js
@@ -23,7 +23,7 @@ export default {
             control: { type: "text" },
             description: "Text to be displayed in the chip",
         },
-        deleteButtonLabel: {
+        a11yDeleteButton: {
             control: { type: "text" },
             description:
                 "A11y text for the delete button, also determines if delete button is shown",
@@ -41,14 +41,14 @@ export default {
     },
 };
 
-export const Standard = Template.bind({});
-Standard.args = {
+export const Default = Template.bind({});
+Default.args = {
     renderBody: "chip text",
 };
-Standard.parameters = {
+Default.parameters = {
     docs: {
         source: {
-            code: tagToString("ebay-chip", Standard.args),
+            code: tagToString("ebay-chip", Default.args),
         },
     },
 };
@@ -56,12 +56,12 @@ Standard.parameters = {
 export const WithDelete = Template.bind({});
 WithDelete.args = {
     renderBody: "chip text",
-    deleteButtonLabel: "Delete",
+    a11yDeleteButton: "Delete",
 };
 WithDelete.parameters = {
     docs: {
         source: {
-            code: tagToString("ebay-chip", Standard.args),
+            code: tagToString("ebay-chip", WithDelete.args),
         },
     },
 };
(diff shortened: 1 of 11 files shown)


src/components/ebay-chip/index.marko
```diff
@@ -0,0 +1,19 @@
+$ const textId = component.elId("text");
```
agliga, line 1: So if we don't allow the user to pass in a custom textid we should not do it this way but rather use `:scoped`
https://markojs.com/docs/class-components/#scoped

src/components/ebay-chip/index.marko
```diff
@@ -0,0 +1,19 @@
...
+
+<span class="chip">
```
agliga, line 3: We should allow spread for probably the root element.

agliga (changes requested):
Some comments to address before merging.

src/components/ebay-chip/index.marko
LuLaValva: This is nicer, thanks

src/components/ebay-chip/index.marko
LuLaValva: Added

src/components/ebay-chip/index.marko
```diff
@@ -0,0 +1,19 @@
...
+    <span class="chip__text" id=textId>
+        <${input.renderBody}/>
+    </span>
+    <if(input.deleteButtonLabel)>
+        <button
+            type="button"
+            class="chip__button"
+            type="button"
```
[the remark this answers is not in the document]
LuLaValva: Haha, good catch. I removed one of them.

LuLaValva (commented):
Thanks Andrew! Should be better now.

src/components/ebay-chip/chip.stories.js
```diff
@@ -0,0 +1,67 @@
...
+                category: "Events",
+                defaultValue: {
+                    summary: "{ el, checked, originalEvent }",
+                },
+            },
+        },
+    },
+};
+
+export const Standard = Template.bind({});
```
agliga, line 44: Lets call it `Default`. That's the name we want to call all out "Standard" templates now

src/components/ebay-chip/index.marko
```diff
@@ -0,0 +1,19 @@
+import { processHtmlAttributes } from "../../common/html-attributes";
+$ const { renderBody, deleteButtonLabel, ...htmlAttributes } = input;
+
+<span class="chip" ...processHtmlAttributes(htmlAttributes)>
+    <span class="chip__text" id:scoped="text">
+        <${input.renderBody}/>
```
agliga, line 6: Since we have `renderBody` defined at the top (so it doesnt get added in spread), why not use it here? At least that way we don't have any unused variables

src/components/ebay-chip/index.marko
```diff
@@ -0,0 +1,19 @@
...
+    </span>
+    <if(input.deleteButtonLabel)>
+        <button
+            type="button"
+            class="chip__button"
+            aria-label=input.deleteButtonLabel
```
agliga, line 12: We usually prefix aria-labels input variables with `a11y` something. So maybe call this `a11yDeleteButton`

agliga (commented):
Also should add examples in the `example` folder. We have a link to examples in readme but no examples. 
It should be fine after these changes.

src/components/ebay-chip/index.marko
LuLaValva: Thanks for the consistency

LuLaValva (commented):
Thanks @agliga, should be good to go now.

src/components/ebay-chip/chip.stories.js
```diff
@@ -0,0 +1,67 @@
...
+            description:
+                "A11y text for the delete button, also determines if delete button is shown",
+        },
+        onDelete: {
+            action: "on-delete",
+            description: "Triggered when delete button is clicked",
+            table: {
+                category: "Events",
+                defaultValue: {
+                    summary: "{ el, checked, originalEvent }",
```
ArtBlue, line 37: To what is `checked` referring? Should it be `deleted` instead?

src/components/ebay-chip/test/test.server.js
```diff
@@ -0,0 +1,16 @@
+import { use } from "chai";
+import { composeStories } from "@storybook/marko/dist/testing";
+import { snapshotHTML } from "../../../common/test-utils/snapshots";
+import * as stories from "../chip.stories"; // import all stories from the stories file
+const { Default, WithDelete } = composeStories(stories);
+const htmlSnap = snapshotHTML(__dirname);
+
+use(require("chai-dom"));
+
+it("renders default chip component", async () => {
```
ArtBlue, line 10: Can we rename these tests please?

1. The `default` chip is also static, so something like, `renders default static chip component`
2. The second one doesn't refer to chip; I think it should. Maybe something like, `renders interactive chip with close button when close label is provided`

ArtBlue (commented):
A couple of minor things...

src/components/ebay-chip/test/test.server.js
agliga: Also on this same line, we should group up all these tests in a `describe` and say: `ebay-chip` or something like that.
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-icon/test/test.server.js#L11
(I know that its not everywhere but I think we should start to add it in all places)

src/components/ebay-chip/index.marko
```diff
@@ -0,0 +1,19 @@
+import { processHtmlAttributes } from "../../common/html-attributes";
+$ const { renderBody, a11yDeleteButton, ...htmlAttributes } = input;
+
+<span class="chip" ...processHtmlAttributes(htmlAttributes)>
+    <span class="chip__text" id:scoped="text">
```
agliga, line 5: Just a nit, I would call the id `chip` or `title`. 
Text is pretty ambiguous. 
Its fine to leave as is though if there are no other changes.

agliga:
I think this PR is fine but need some minor nit cleanup.

src/components/ebay-chip/test/test.server.js
LuLaValva: Sounds good, I updated test names.

src/components/ebay-chip/chip.stories.js
LuLaValva: Good catch, this was actually completely wrong, must've been a copy-paste error. I fixed it.

LuLaValva (commented):
Thanks for the comments @agliga and @ArtBlue, should look cleaner now.

agliga (approved):

ArtBlue (approved):
Looks good.

Issue #1975, opened 2023-08-30
education-notice: add new variant


agliga:
Original issue:  https://github.com/eBay/skin/issues/1987
Need to add new type and option.

PR #1982, opened 2023-09-07
feat(section-notice): added education variant

saiponnada:
## Description

Added education variant to `ebay-section-notice`. By default section notice with status education shows lightbulb icon but has the ability to override to a custom icon if needed. This specific variant also has the ability to make the notice prominent by setting the prominent tag.

## Context

#1975 

## References

[Skin Issue](https://github.com/eBay/skin/issues/1987)
[Figma](https://www.figma.com/file/zEBdEhbonrBOGzZ0fXzWvM/eBay-Design-System?node-id=90732%3A97760&t=N2QRCppwp2G1BjRO-11)

## Screenshots
**Default**
image

**With Prominence**
image

**With custom Icon**
image

### DIFF
--- a/src/components/components/ebay-notice-base/index.marko
+++ b/src/components/components/ebay-notice-base/index.marko
@@ -28,7 +28,7 @@ $ var prefixClass = input.prefixClass;
 <${input.root || 'section'}
     aria-labelledby=!input.noA11yLabel && component.elId('status')
     aria-roledescription=input.a11yRoleDescription
-    class=[prefixClass, input.class, input.status === 'education' && input.prominent && `${prefixClass}--prominent`]
+    class=[prefixClass, input.class, input.status === 'education' && input.prominent && `${prefixClass}--education`]
     ...processHtmlAttributes(input, ignoredAttributes)
 >
     <if(input.icon !== 'none')>
--- a/src/components/components/ebay-notice-base/index.marko
+++ b/src/components/components/ebay-notice-base/index.marko
@@ -1,4 +1,5 @@
 import { processHtmlAttributes } from '../../../common/html-attributes';
+import LightBulbIcon from '<ebay-lightbulb-24-icon>';
 static {
     var ignoredAttributes = [
         'status',
@@ -17,6 +18,8 @@ static {
         'mainRoot',
         'noA11yLabel',
         'a11yDismissText',
+        'educationIcon',
+        'prominent',
     ];
 }
 $ var status = input.status;
@@ -25,7 +28,7 @@ $ var prefixClass = input.prefixClass;
 <${input.root || 'section'}
     aria-labelledby=!input.noA11yLabel && component.elId('status')
     aria-roledescription=input.a11yRoleDescription
-    class=[prefixClass, input.class]
+    class=[prefixClass, input.class, input.status === 'education' && input.prominent ? `${prefixClass}--prominent` : '' ]
     ...processHtmlAttributes(input, ignoredAttributes)
 >
     <if(input.icon !== 'none')>
@@ -50,6 +53,12 @@ $ var prefixClass = input.prefixClass;
                     a11y-text=input.a11yIconText || input.a11yText
                     class=input.iconClass
                 />
+            </else-if>
+            <else-if(status === 'education' && input.type === 'section')>
+                <${input.educationIcon || LightBulbIcon}
+                    a11y-variant='label'
+                    a11y-text=input.a11yIconText || input.a11yText
+                    class=input.iconClass/>
             </else-if>
         </>
     </if>
--- a/src/components/components/ebay-notice-base/index.marko
+++ b/src/components/components/ebay-notice-base/index.marko
@@ -28,7 +28,7 @@ $ var prefixClass = input.prefixClass;
 <${input.root || 'section'}
     aria-labelledby=!input.noA11yLabel && component.elId('status')
     aria-roledescription=input.a11yRoleDescription
-    class=[prefixClass, input.class, input.status === 'education' && input.prominent ? `${prefixClass}--prominent` : '' ]
+    class=[prefixClass, input.class, input.status === 'education' && input.prominent && `${prefixClass}--prominent`]
     ...processHtmlAttributes(input, ignoredAttributes)
 >
     <if(input.icon !== 'none')>
--- a/src/components/ebay-section-notice/examples/with-action.marko
+++ b/src/components/ebay-section-notice/examples/with-action.marko
@@ -3,7 +3,7 @@
         <div>Notice title</div>
     </@title>
     <@footer>
-        <div>H</div>
+        <ebay-fake-link>Footer</ebay-fake-link>
     </@footer>
     <p>
         <strong>Error:</strong> Please take another look at the following:
--- a/src/components/components/ebay-notice-base/test/mock/index.js
+++ b/src/components/components/ebay-notice-base/test/mock/index.js
@@ -32,3 +32,13 @@ export const titleFooterNotice = {
         renderBody: createRenderBody("footer"),
     },
 };
+
+export const educationSectionNotice = {
+    prefixClass: "section-notice",
+    status: "education",
+    a11yText: "education label",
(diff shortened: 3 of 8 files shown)


src/components/ebay-section-notice/examples/with-action.marko
```diff
@@ -3,7 +3,7 @@
         <div>Notice title</div>
     </@title>
     <@footer>
-        <div>H</div>
+        <ebay-fake-link>Footer</ebay-fake-link>
```
saiponnada, line 6: Refactored this for consistency.

src/components/components/ebay-notice-base/index.marko
```diff
@@ -25,7 +28,7 @@ $ var prefixClass = input.prefixClass;
 <${input.root || 'section'}
     aria-labelledby=!input.noA11yLabel && component.elId('status')
     aria-roledescription=input.a11yRoleDescription
-    class=[prefixClass, input.class]
+    class=[prefixClass, input.class, input.status === 'education' && input.prominent ? `${prefixClass}--prominent` : '' ]
```
LuLaValva, line 31: Marko automatically removes `null | undefined | false` values from the class list so you can simplify this a little bit:
```suggestion
    class=[prefixClass, input.class, input.status === 'education' && input.prominent && `${prefixClass}--prominent`]
```

LuLaValva: Also, now that I am looking at it again, do you _need_ to also check that `status === 'education'`? I know in `skin` we say it is "strictly for education notices only" but I don't think we need to enforce that in JS.

src/components/components/ebay-notice-base/index.marko
```diff
@@ -51,6 +54,12 @@ $ var prefixClass = input.prefixClass;
                     class=input.iconClass
                 />
             </else-if>
+            <else-if(status === 'education' && input.type === 'section')>
```
LuLaValva, line 57: If this is only available for section notices, should it be in `notice-base`? Maybe this should be moved to `ebay-section-notice` instead, or this is a hint that we need to do some refactoring somewhere

src/components/components/ebay-notice-base/index.marko
```diff
@@ -51,6 +54,12 @@ $ var prefixClass = input.prefixClass;
...
+                <${input.educationIcon || LightBulbIcon}
+                    a11y-variant='label'
```
LuLaValva, line 59: I don't think these attributes will ever reach the icon if it is passed on as an attribute tag as you have in `with-icon.marko`. We could potentially have people pass in the icon that was imported instead, like
```marko
import TheVaultIcon from '<ebay-the-ebay-vault-24-fit-icon>'
<ebay-section-notice status="education" icon=TheVaultIcon />
```
but I'm not sure if that is best practice for class API.

LuLaValva (commented):
Thanks Goutham, nice work! Just a few ideas for cleaning it up.

src/components/components/ebay-notice-base/index.marko
saiponnada: I guess we might need to as team tend to misuse all these attributes. Lets see others thoughts.

src/components/components/ebay-notice-base/index.marko
saiponnada: Definitely need a refactor here. We just want to release this for now and may be clean up in next major release.

ianmcburnie: I was also initially concerned that we might end up introducing too many conditionals. If it's starting to feel unwieldy. please consider a separate ebay-education-notice. If so, we might also need to split it out on the skin side too (for parity).

src/components/components/ebay-notice-base/index.marko
ArtBlue: I would say we should keep `status === 'education'` in the conditional as it removes DOM use cases that have no matching styles and reduces the likelihood of someone falsely expecting a "prominent" variation for other types of notices.

src/components/components/ebay-notice-base/index.marko
LuLaValva: This conditional and the other one for `input.type === 'education'` have me convinced that Ian is right and we should create a separate `ebay-education-notice`. I know it's not ideal and we'd be scrapping a lot of work that we already did, but if we don't make the switch now it will be a breaking change in the future.

saiponnada: As discussed, If we need to refactor this to separate component without notice-base, we are totally violating DRY here. If we want to create a dedicated story to improve DX, its pretty much same code with new story file and readme(as education is still a section notice under the hood). I am not sure doing either of the above yield anything other than extra work for us. Hence will leave this as is for now and we shall revisit when we are planning to make icon as a dedicated component.

LuLaValva (approved):
I still don't love this as a type of section notice instead of its own component, but if that isn't changing I like what you have now.

LuLaValva:
I do in theory, but since section notice leverages `notice-base` it actually has a surprisingly small amount of code that you'd have to duplicate. My perspective is weakly held, but I think as someone who is using eBayUI I think I may expect to use `<ebay-education-notice>` instead of `<ebay-section-notice type="education">` since to me they look different enough that they'd be separate components at first glance.

saiponnada:
> I do in theory, but since section notice leverages `notice-base` it actually has a surprisingly small amount of code that you'd have to duplicate. My perspective is weakly held, but I think as someone who is using eBayUI I think I may expect to use `<ebay-education-notice>` instead of `<ebay-section-notice type="education">` since to me they look different enough that they'd be separate components at first glance.

Fair. The way the current code is, I personally feel better to keep as is and we all can revisit this again as part of future tech debt if its justified.

LuLaValva:
Sure. I

> > I do in theory, but since section notice leverages `notice-base` it actually has a surprisingly small amount of code that you'd have to duplicate. My perspective is weakly held, but I think as someone who is using eBayUI I think I may expect to use `<ebay-education-notice>` instead of `<ebay-section-notice type="education">` since to me they look different enough that they'd be separate components at first glance.
> 
> Fair. The way the current code is, I personally feel better to keep as is and we all can revisit this again as part of future tech debt if its justified.

Sure. I approved as-is, we can maybe address this during the Marko 6 rewrite

ArtBlue:
> Sure. I
> 
> > > I do in theory, but since section notice leverages `notice-base` it actually has a surprisingly small amount of code that you'd have to duplicate. My perspective is weakly held, but I think as someone who is using eBayUI I think I may expect to use `<ebay-education-notice>` instead of `<ebay-section-notice type="education">` since to me they look different enough that they'd be separate components at first glance.
> > 
> > 
> > Fair. The way the current code is, I personally feel better to keep as is and we all can revisit this again as part of future tech debt if its justified.
> 
> Sure. I approved as-is, we can maybe address this during the Marko 6 rewrite

Yea, I don’t think it was even an option to redo this component at this point. The skin piece is mostly set and it is way too late to pivot anyway. It would be a breaking change so the earliest would be next major.

agliga:
> I do in theory, but since section notice leverages `notice-base` it actually has a surprisingly small amount of code that you'd have to duplicate. My perspective is weakly held, but I think as someone who is using eBayUI I think I may expect to use `<ebay-education-notice>` instead of `<ebay-section-notice type="education">` since to me they look different enough that they'd be separate components at first glance.

On paper, yeah it looks like a small amount.
But in practice it actually isn't. This is why we have notice base because we kept having to change things around in 3-4 different components if one of them changed. 90% of the code is the same among them. This is why we have reusable components and functions. Even the icon, we use the same 3 icons in all notices. The structure, the same etc. Just different classes and minor here and there's. 
So if we were to make a separate education notice (which we can, I don't mind if we do or not), I would still argue to use notice-base because that is what it is intended for. We could look at splitting out icon into it's own component piece (we should still keep that as a component because we would need to duplicate the logic for selecting what icon we want to show across 3 different components), but the rest should be the same.

ArtBlue (approved):
LGTM

Issue #1984, opened 2023-09-14
lightbox-dialog: wont close when parent state changes


agliga:
If the parent state changes (even if the open/close doesnt change), it causes the dialog to open again if it was in the midst of openning
To reproduce, use this code in one of our examples

```
<ebay-drawer-dialog
    a11y-close-text="Close drawer"
    open=state.open
    on-close("closeDrawer")
    on-expanded("emit", "expanded")
    on-collapsed("emit", "collapsed")
    on-open("emit", "open")
    ...input>
    <${input.renderBody}/>
</ebay-drawer-dialog>
<ebay-button on-click("openDrawer")>Open drawer</ebay-button>
class {
    onCreate() {
        this.state = { open: false, count: 0 };
    }
    onMount() {
        setInterval(() => {
            this.state.count++;
        }, 200)
    }
    openDrawer() {
        this.state.open = true;
    }
    closeDrawer() {
        this.state.open = false;
        this.emit('close', arguments);
    }
}
```

PR #1985, opened 2023-09-14
fix: added a check to not rerender dialog when closing/opening

agliga:
## Description
* Added a check if the dialog is animating. If so, then do not rerender the state.
* There was some issues with tests and timing, so I added a delay before the test

## References
#1984

### DIFF
--- a/src/components/ebay-drawer-dialog/index.marko
+++ b/src/components/ebay-drawer-dialog/index.marko
@@ -10,6 +10,7 @@ $ var handleLabel = (!state.expanded ?
     on-scroll('handleScroll')
     on-open("emit", "open")
     on-close("emit", "close")
+    on-toggled("emit", "toggled")
     class=[
         input.class,
         "drawer-dialog--mask-fade-slow"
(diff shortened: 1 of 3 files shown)


src/components/ebay-drawer-dialog/index.marko
```diff
@@ -10,6 +10,7 @@ $ var handleLabel = (!state.expanded ?
     on-scroll('handleScroll')
     on-open("emit", "open")
     on-close("emit", "close")
+    on-toggled("emit", "toggled")
```
LuLaValva, line 13: I think this may be a remnant from a previous attempt, but we should bring it back to the base component so it can be used for tracking and such. It also may be renamed to something like `on-close-click`, but that is up to you.

ArtBlue (approved):
LGTM

LuLaValva:
@agliga I think you should still include a `close-click` event for tracking (could be a separate PR, but it would fit here).

agliga:
> @agliga I think you should still include a `close-click` event for tracking (could be a separate PR, but it would fit here).

I think that's way outside the scope of this issue.

Issue #1987, opened 2023-09-21
breadcrumbs: missing a11y text for collapsed button


saiponnada:
# Bug Report

## eBayUI 

https://opensource.ebay.com/ebayui-core/?path=/story/navigation-disclosure-ebay-breadcrumbs--heading

## Description

The fake menu button doesn't have a11y text associated with it during collapsed state.

## Screenshots

![image](https://github.com/eBay/ebayui-core/assets/6342519/1419d125-c4a3-408c-9862-2768851a4955)

PR #1988, opened 2023-09-21
fix(breadcrumbs): added missing a11y text for collapsed button

saiponnada:
## Description

Fixed an accessibility issue where button doesn't have `aria-label` when collapsed.

## Context

#1987

## Screenshots

image


agliga (approved):
Also add an entry in the marko.json for autocomplete 
Otherwise looks good

PR #1996, opened 2023-10-04
feat(badge): add custom max value

### LINKED ISSUE #1995: Allow badges editable
In the current component, it shows 99+ but for Messages, we need to show 10+ only.
The dev team not able to make edit of it and we have to keep showing the 99+, is this a easy fix. Can we get this asap cause the badge itself is existing and all we need is to have one other component to show "10+". Thanks

LuLaValva:
- Resolves #1995 

## Description

Added new `max` attribute to `ebay-badge` which enables truncation to values besides 99

## Context

I considered naming this `truncateAfter` instead of `max`, but this seems simpler. Let me know if there is opposition to this.

## Screenshots

image

### DIFF
--- a/src/components/ebay-badge/badge.stories.js
+++ b/src/components/ebay-badge/badge.stories.js
@@ -20,6 +20,16 @@ export default {
             type: "number",
             description: "Used as the number to be placed in the badge",
         },
+        max: {
+            type: "number",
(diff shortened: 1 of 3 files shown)


LuLaValva:
Closed after decision that this should always be 99

PR #1999, opened 2023-10-10
docs(lightbox-dialog): add note

LuLaValva:
## Description

I think it would be nice to add a "Notes / FAQ" section to the README for components that have frequent questions asked about them. What do you all think of this idea?

## Context

I spent more time debugging a component than I should have, even though we have run into this problem before

### DIFF
--- a/src/components/ebay-lightbox-dialog/README.md
+++ b/src/components/ebay-lightbox-dialog/README.md
@@ -7,6 +7,10 @@
     </span>
 </h1>
 
+## Notes / FAQ
+
+-   `@header` is _required_, and styles will break if you do not include it
+
 ## Examples and Documentation
 
 -   [Storybook](https://ebay.github.io/ebayui-core/?path=/story/dialogs-ebay-lightbox-dialog)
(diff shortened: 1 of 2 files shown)


ianmcburnie:
Yes, we've had some discussion in the past on where this type of documentation should go (e.g storybook, standalone website. etc). It's definitely needed, so I agree let's just get it in the readme for each component.

ArtBlue:
Can't/Shouldn't we simply include the required flag in the descriptions with the notes there?  

Screenshot 2023-10-11 at 7 29 06 AM

agliga:
> Can't/Shouldn't we simply include the required flag in the descriptions with the notes there?
> 
> Screenshot 2023-10-11 at 7 29 06 AM
)

Agreed. We can do that by adding required: true
Screenshot 2023-10-11 at 9 34 32 AM
Screenshot 2023-10-11 at 9 34 04 AM

agliga:
I am in agreement of having a notes section. We do have it for other components as well, but not specifically in a separate section.

LuLaValva:
Good catch @ArtBlue, I added required to the attribute description. I still think it's a good idea to have a notes section for other non-attribute-related notes, so I kept it for now (no harm in redundancy, right?)

ArtBlue:
> Good catch @ArtBlue, I added required to the attribute description. I still think it's a good idea to have a notes section for other non-attribute-related notes, so I kept it for now (no harm in redundancy, right?)

Yep, sounds good.

agliga (approved):

Issue #2001, opened 2023-10-11
icons: add gradient support


agliga:
In skin we added support using defs. We need to make sure ebayui supports those defs in the script import

PR #2002, opened 2023-10-13
feat(icon): added gradient support

agliga:
## Description

In order to support new AI icons from skin, we need to add support for defs in our svg in ebayui

## Context
* This has a large changeset. In order to make it easier to look through, files that should be looked at: `component-browser.js`, `index.marko` in ebay-icons and `scripts/import-svg.js`
* I removed all the `xlink:href` from ebayui. Seems to have been lingering around (which is why we have so many files changed)
* This was a pretty complicated changeset. In order to get linearGradients to be working, we cannot use a base string. We have to build a linear gradient object and inject all the attributes that way. Part of the change required to use `createElementNS` and put an svg namespace in there. 
* Im open to other suggestions on how to build linear gradients without hardcoding some of the values, but I could not figure it out for now. For now it seems like these attributes are the only ones used. We can revisit to make it more generic

## References
https://github.com/eBay/ebayui-core/issues/2001


ArtBlue (commented):
Yea, the `xlink` changes seem arguably out of scope here; wish it was done in a follow-up PR to avoid the noise.

src/components/ebay-icon/component-browser.js
```diff
@@ -1,28 +1,71 @@
 let rootSvg;
+let svgDefs;
+
+function createLinearGradientFromString(data) {
+    // This is needed because if we add linear gradient as a string it will not be rendered.
+    const linearGradient = document.createElementNS(
+        "http://www.w3.org/2000/svg",
+        "linearGradient"
+    );
+    linearGradient.setAttribute("id", data.id);
```
ArtBlue, line 10: I think you should be able to loop through `data` and get the key/value pairs to use for these `setAtrribute` lines to avoid explicit attributes. That should make it more scalable.

src/components/ebay-icon/component-browser.js
```diff
@@ -1,28 +1,71 @@
...
         // If there were any symbols rendered then we move them to the svg above after rendering them.
         const defs = this.getEl("defs");
 
         if (defs) {
+            let defItem;
             if (this.input && this.input._themes) {
                 defs.innerHTML = this.input._themes();
             }
+            if (this.input && this.input._def) {
+                defItem = createLinearGradientFromString(this.input._def());
```
ArtBlue, line 58: Can we create a logical fork here to make it a bit more scalable right off the bat? Check inside `defs` with a `switch` statement and crate the linear gradient when `linearGradient` is inside? We're probably going to have more gradient types upcoming. It will be easier to manage this if the fork is already set up.

ArtBlue (commented):
Just a couple of notes; looks great, otherwise.

src/components/ebay-icon/component-browser.js
agliga: As I mentioned in the description
> For now it seems like these attributes are the only ones used. We can revisit to make it more generic
Yes, when we get more types we can add that case and make the attributes setting more generic. I think what we will end up doing is putting the attributes in it's own key and iterating through them, like:
```
{ 
   type: "linearGradient",
   attributes: { id: 'id', x1: 'x1" etc}
   children: { //etc }
}
```
This won't be a breaking change and will make it much more generic. So when we do get more types, Im happy to add that, but since this is a limited change we can leave it for now. The main rationale is I want to see what the other types we get are and then figure it out to make it generic then. That way we won't be guessing and we will see what we need to support fully.

src/components/ebay-icon/component-browser.js
agliga: See the other comment for an answer to this.

src/components/ebay-icon/component-browser.js
ArtBlue: Yea, I saw that in the PR. I think it'd be best to bake that in ahead of time, but if you want to iterate on this, I guess that's fine too.

src/components/ebay-icon/component-browser.js
ArtBlue: If we keep it this way, the next time design introduces a gradient with an attribute previously unused, such as `gradientUnits`, the script will ignore it silently. The resulting icon will not look right, and we'd wonder why. Then, we'd have to troubleshoot to figure out why it doesn't look right and I doubt we'd going to remember this particular nuance in the script or this thread, especially if someone else takes the ticket. Once we find the issue, then we'd have to modify the script to add support for that one attribute and then rerun the script. It seems like if this is changed to more of a passthrough instead of explicitly setting things based on the specific attributes currently used, we avoid all of that.

agliga: Thats the point. I want to make sure the icon is broken so we can know to modify the script.
I don't know exactly what design will do. When we get there we can update as needed. This will both help double check our code being robust and also implement it then rather than second guessing now.

ArtBlue: Hmm.. why leave room for the possibility to have the icon be broken when we can create a simple passthrough to avoid all of it ahead of time? This doesn't seem to be over-engineering to me. Ok, I guess I'll go ahead and approve the PR.

ArtBlue (approved):

agliga:
I'll update this with your comments
I did the change but it wasnt working last night
I'll push the updates if i can get it working.

agliga:
@ArtBlue can you please rereview? I updated based on your comments.

ArtBlue:
> @ArtBlue can you please rereview? I updated based on your comments.

Looks great! Thanks!

Issue #2004, opened 2023-10-20
storybook: update to storybook 7


agliga:
Also use vite if we can

PR #2010, opened 2023-10-26
chore: updated to storybook 7 and added vite support

agliga:
## Description
* Updated storybook to storybook 7
* Added vite support
* Fixed code to support this
* For fixing readmes I added a plugin to read readmes as raw files (This is because lasso doesnt work well with `Readme.md?raw`. If we get rid of lasso in our tests, we should remove this
* Updated dev deps

## References
https://github.com/eBay/ebayui-core/issues/2004


LuLaValva (commented):
This looks awesome! Storybook 7 has even more little improvements than I thought. Just a few things I found when looking through (one comment below because file is not included in PR):

- The `clipped` class in the default story of `character-count` isn't being applied. You didn't modify this file so I'm not sure if this is a problem with storybook or the component.

agliga:
> This looks awesome! Storybook 7 has even more little improvements than I thought. Just a few things I found when looking through (one comment below because file is not included in PR):
> 
> 
> 
> - The `clipped` class in the default story of `character-count` isn't being applied. You didn't modify this file so I'm not sure if this is a problem with storybook or the component.
> 
> - Story order fix included in #2011 

I'll look. My guess is global styles aren't being loaded

agliga:
> > This looks awesome! Storybook 7 has even more little improvements than I thought. Just a few things I found when looking through (one comment below because file is not included in PR):
> > 
> > * The `clipped` class in the default story of `character-count` isn't being applied. You didn't modify this file so I'm not sure if this is a problem with storybook or the component.
> > * Story order fix included in [fix(storybook): order in storybook 7 #2011](https://github.com/eBay/ebayui-core/pull/2011)
> 
> I'll look. My guess is global styles aren't being loaded

Okay, resolved this. I added `@ebay/skin/utility` as a depdency in storybook. Should be ready for rereview now.

LuLaValva (approved):
Looks good!

It's a little annoying that clicking on a component goes straight to the documentation instead of the first example, but that might just be because it's not what I'm used to. Also it seems to be what  storybook is pushing for as a standard so we should keep it.

agliga:
Also as part of this PR, I updated dev deps. We usually do it separately, but in this case there were so many changes needed I did it as a part of this PR.

PR #2011, opened 2023-10-27
fix(storybook): order in storybook 7

LuLaValva:
Experimenting with micro-PR as a suggestion for your PR


.storybook/storyOrder.js
```diff
@@ -1,20 +0,0 @@
-const README = ["Welcome", "Getting Started", "Contributing"];
```
agliga, line 1: This readme section is missing. Will it show up without it?

LuLaValva: Yeah, it looks like storybook 7 automatically puts docs first so there's no need to specify these anymore

LuLaValva: That said, keeping `Welcome` first is a good idea. I added a commit that brings these back

agliga (approved):

PR #2013, opened 2023-11-01
fix(calendar): timezone bug

LuLaValva:
## Description

Months were being shifted on the first day of the month in CET timezone, an additional UTC cast fixed the issue.

### DIFF
--- a/src/components/ebay-calendar/component.js
+++ b/src/components/ebay-calendar/component.js
@@ -293,7 +293,9 @@ export default class extends Marko.Component {
                 year: "numeric",
             }
         );
-        return formatter.format(date);
+        return formatter.format(
+            new Date(date.getUTCFullYear(), date.getUTCMonth())
+        );
     }
 
     /**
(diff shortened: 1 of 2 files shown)


agliga:
Is this blocking? Do we need a patch or can it wait.

LuLaValva:
> Do we need a patch or can it wait?

It depends how long we are waiting for, but I would like to get this in soon so I was thinking a patch was a good idea

agliga:
> > Do we need a patch or can it wait?
> 
> 
> 
> It depends how long we are waiting for, but I would like to get this in soon so I was thinking a patch was a good idea

I understand. I think it's good to get it out soon
But our criteria for is only if the team is blocked completely and there is no workaround. 
So verify if this is blocking their release or code. If not we can put it in 12.8.0

LuLaValva:
> > > Do we need a patch or can it wait?
> > 
> > 
> > 
> > It depends how long we are waiting for, but I would like to get this in soon so I was thinking a patch was a good idea
> 
> I understand. I think it's good to get it out soon
> But our criteria for is only if the team is blocked completely and there is no workaround. 
> So verify if this is blocking their release or code. If not we can put it in 12.8.0 

This is blocking, and there is no workaround. The team in question is looking to start demos next week, but this effects all teams that use date textbox.

PR #2019, opened 2023-11-09
feat(skeleton-loader): new component

saiponnada:
## Description

Added new skeleton loader component. 

Note:
Snapshots for the tests are not being generated.

## Context
#1976 

## Screenshots
image
image


src/components/ebay-skeleton-loader/index.marko
```diff
@@ -0,0 +1,48 @@
...
+    "multiline",
+    "count",
+    "columns",
+    "isVisible",
+];
+static var validTypes = ["avatar", "button", "text", "form", "image", "tile"];
+static var validSizes = ["large", "small"];
+$ var size = validSizes.includes(input.size) ? input.size : "default";
+$ var type = validTypes.includes(input.type) ? input.type : "button";
+$ var typeClass = type === "button" ? "skeleton--btn" : `skeleton--${type}`;
```
LuLaValva, line 16: I think this is a sign that we either need to change `type` to `btn` or the class name to `skeleton--button`

src/components/ebay-skeleton-loader/index.marko
```diff
@@ -0,0 +1,48 @@
...
+    "type",
+    "multiline",
+    "count",
+    "columns",
+    "isVisible",
+];
+static var validTypes = ["avatar", "button", "text", "form", "image", "tile"];
+static var validSizes = ["large", "small"];
+$ var size = validSizes.includes(input.size) ? input.size : "default";
+$ var type = validTypes.includes(input.type) ? input.type : "button";
```
LuLaValva, line 15: I'm not sure how I feel about defaulting to `"button"`, is that the most likely object to be making a skeleton for?

src/components/ebay-skeleton-loader/index.marko
```diff
@@ -0,0 +1,48 @@
...
+$ var typeClass = type === "button" ? "skeleton--btn" : `skeleton--${type}`;
+$ var skeletonClass = (
+    size !== "default"
+        ? `${typeClass}-${size}`
+        : type === "text"
+        ? `${typeClass}-small`
+        : typeClass
+);
+
+<if(typeof input.isVisible === "undefined" ? true : input.isVisible)>
```
LuLaValva, line 25: Using `if` here means that we need to use JavaScript to create and destroy DOM nodes, which means that layout shift is likely and also a lot of work is being done each time a skeleton is shown or hidden. I think it would probably be preferable to toggle `style="visibility:hidden"` on the div instead.

src/components/ebay-skeleton-loader/index.marko
```diff
@@ -0,0 +1,48 @@
...
+        ...processHtmlAttributes(input, ignoredAttributes)
+        class=[
+            type !== "tile" ? `skeleton ${skeletonClass}` : typeClass,
+            input.multiline && type === "text" && "skeleton--text-multiline",
+            input.class,
+        ]
+        role="img"
+        aria-label=input.a11yText || "Loading"
+    >
+        <if(type === "tile")>
```
LuLaValva, line 36: This is a lot of complex logic for the `tile` case, maybe it's a reason to move `ebay-skeleton-tile` into a separate component which has consistent logic (not sure about this comment, just a thought)

src/components/ebay-skeleton-loader/skeleton-loader.stories.js
```diff
@@ -0,0 +1,194 @@
...
+    parameters: {
+        docs: {
+            description: {
+                component: Readme,
+            },
+        },
+    },
+
+    argTypes: {
+        size: {
```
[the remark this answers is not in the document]
saiponnada: Was contemplating about that. I personally don't have any strong preference but felt unnecessary. Do we need to do that as we are listing all the stories?

src/components/ebay-skeleton-loader/index.marko
saiponnada: This was tricky, the whole skin docs refers button as btn. I wish we named btn as button. Not sure why we did it. To keep it consistent, I can change type to `btn` but I really like type to be spelled out as button.

src/components/ebay-skeleton-loader/skeleton-loader.stories.js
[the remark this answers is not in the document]
LuLaValva: Yes! You should include it every time, just so developers can see at a glance which attributes are being used.

src/components/ebay-skeleton-loader/index.marko
saiponnada: The idea was to restrict to one of these types from ebay UI side. And picked button as random. We can also not load a skeleton if it doesn't match any type.

LuLaValva: I would imagine that a skeleton without a type just adds `skeleton` to the class list, and nothing else

src/components/ebay-skeleton-loader/index.marko
saiponnada: The idea for using `isVisible` is to set the flag to false to remove skeleton from dom and show content below. This is useful technique for hiding sibling skeleton. Preferably teams use other techniques like if-else or await placeholder than this approach. If we use `visibility:hidden`, won't the empty layout still visible beside the regular content if not used content swap approach?

agliga: Yes we should keep it like this. Visibility hidden wont work
At the same time the tricky part of layout shift isnt hiding elements but ensuring that everything is the same size

src/components/ebay-skeleton-loader/index.marko
agliga: Agreed. Unless its broken then we shouldn't default to that

src/components/ebay-skeleton-loader/index.marko
agliga: I believe this was because of marko 3. (That or following bootstrap)
We can change it to button next major if needed. 
Lets keep it like this for now though.

src/components/ebay-skeleton-loader/skeleton-loader.stories.js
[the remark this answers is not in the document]
agliga: We need type especially for booleans or enums.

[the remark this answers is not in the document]
saiponnada: Done.

src/components/ebay-skeleton-loader/index.marko
agliga: I think the right way is to make this component callable from others. 
Lets leave it as is for now. In future we should have button or tile having a "loading" option which would swap it to a skeleton component.

src/components/ebay-skeleton-loader/index.marko
LuLaValva: Wait I still disagree with this at a fundamental level. Right now all `isVisible` does is control a top-level if statement, so really we should just remove this entirely and have teams use
```marko
<if(isVisible)>
  <ebay-skeleton-loader />
</if>
```

src/components/ebay-skeleton-loader/index.marko
saiponnada: Done. renamed this to `skeleton-button`.

src/components/ebay-skeleton-loader/index.marko
saiponnada: As discussed, now defaulted to form as we do not have standalone `skeleton` class.

src/components/ebay-skeleton-loader/index.marko
saiponnada: yeah, for now I removed the wrapper condition and `isVisible`. Instead let's use documentation to explain the right techniques.

src/components/ebay-skeleton/index.marko
```diff
@@ -0,0 +1,46 @@
...
+    class=[
+        typeClass,
+        skeletonClass,
+        input.multiline && type === "text" && "skeleton-text--multiline",
+        input.class,
+    ]
+    role="img"
+    aria-label=input.a11yText || "Loading"
+>
+    <if(type === "tile")>
```
LuLaValva, line 37: I don't know if I like having this as an `input` option, because composite components won't look the same for every use case. It would be nice as an example, but I don't think it should be baked into the component like this.

saiponnada: No longer a component. Moved to composite example.

src/components/ebay-skeleton/index.marko
```diff
@@ -0,0 +1,46 @@
...
+<div
+    ...processHtmlAttributes(input, ignoredAttributes)
+    class=[
+        typeClass,
+        skeletonClass,
+        input.multiline && type === "text" && "skeleton-text--multiline",
+        input.class,
+    ]
+    role="img"
+    aria-label=input.a11yText || "Loading"
```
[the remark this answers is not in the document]
saiponnada: Regarding point 1, IMO majority of eBay users use English and, adding a consistent a11y text makes it consistent experience rather than not passing at all. All our components provide similar consistent experience for non-English app teams to pass a localized value, so hopefully it makes easier to adopt. Point 2 is taken care of with current code refactor as we are always providing the wrapper and adding a11yText on the wrapper.

src/components/ebay-skeleton/examples/withContent.marko
```diff
@@ -0,0 +1,38 @@
...
+        }
+    }
+}
+
+<style>
+    .demo-container {
+        width: 88px;
+    }
+</style>
+<div class="demo-container">
```
agliga, line 29: Id rename this more specific. I believe this might be used by storybook or in other places. 
Maybe `skeleton-example-container`

src/components/ebay-skeleton/skeleton.stories.js
```diff
@@ -0,0 +1,220 @@
...
+    component: withContentTemplate,
+});
+withContent.args = {};
+withContent.parameters = {
+    docs: {
+        source: {
+            code: withContentCode,
+        },
+    },
+};
```
agliga, line 220: I have a shortcut way of doing all this. 
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-button/button.stories.js#L224-L227

The third argument is args if needed. 
https://github.com/eBay/ebayui-core/blob/336467022d6d755e714e48b90e48aefc4d9bca0d/.storybook/utils.js#L24

agliga (commented):
Overall LGTM.

src/components/ebay-skeleton/examples/withContent.marko
saiponnada: Done, renamed consistent to other story class names.

src/components/ebay-skeleton/skeleton.stories.js
saiponnada: Done.

src/components/ebay-skeleton/components/ebay-skeleton-avatar/index.marko
```diff
@@ -0,0 +1,15 @@
+import { processHtmlAttributes } from "../../../../common/html-attributes";
+import type { AttrClass } from "marko/tags-html";
+static var ignoredAttributes = ["as", "class"];
```
agliga, line 3: Lets use the new syntax for this
Its somethibg like
```
$ const { as, class: className, ...rest} = input
<div class=className processHtmlAttributes(...rest)>
```
Check with Luke for the right syntax.

agliga (commented):
Lgtm overall
One comment. Otherwise its good to go imo

LuLaValva (approved):
Awesome work, thanks for making all of the modifications

src/components/ebay-skeleton/components/ebay-skeleton-avatar/index.marko
LuLaValva: This will be done before my refactor, so I think it's ok to merge now and I'll do this change after a rebase.

agliga (approved):

Issue #2007, opened 2023-10-20
ebay-flag: new module


agliga:
see https://github.com/eBay/skin/issues/2184

PR #2022, opened 2023-11-14
feat(flags): imported flags from skin as components

agliga:
## Description
Imported flags into ebayui. Added as a new module.

## References
https://github.com/eBay/ebayui-core/issues/2007

## Screenshots

Screenshot 2023-11-14 at 2 33 39 PM


agliga:
CI will fail until we prepublish skin

LuLaValva (commented):
Should we run prettier on all of these `.js` and `.json` files, to keep them consistent with the rest of the source? This is potentially also something we can add to the import script

agliga:
> Should we run prettier on all of these `.js` and `.json` files, to keep them consistent with the rest of the source? This is potentially also something we can add to the import script

Every time we commit prettier is run on those files that were changed

saiponnada (approved):
LGTM

LuLaValva:
> > Should we run prettier on all of these `.js` and `.json` files, to keep them consistent with the rest of the source? This is potentially also something we can add to the import script
> 
> Every time we commit prettier is run on those files that were changed

The files in this PR are not formatted with prettier.

LuLaValva:
> > Should we run prettier on all of these `.js` and `.json` files, to keep them consistent with the rest of the source? This is potentially also something we can add to the import script
> 
> Every time we commit prettier is run on those files that were changed

Ah I see, it's just the other icons that aren't formatted. We should format them in a separate PR.

LuLaValva (approved):

PR #2031, opened 2023-11-27
Fix list numbers in contributing doc

darkwebdev:
what title says


LuLaValva (approved):
Markdown viewers should automatically increment for us in this situation so the numbers aren't necessarily required, but I see no reason why we shouldn't do this.

agliga (approved):

Issue #2032, opened 2023-11-30
ebay-carousel: auto-play slide has issues


nagarajan-g:
# Bug Report
In Ebay Home page banner we use ebay-carousel for Top banner. Click Paddle(arrow) to scroll left a few times, after some click (after 4 clicks) the banner space becomes empty, just white space and after you move the mouse again all the banner skips to last slide quickly.

## eBayUI Version: x.x.x
12.6.2

## Description

1. Goto [www.ebay.com](http://www.ebay.com/) 
2. After banner is loaded click on "Left" arrow 4 times ( depends on number of slides)/ After each click move the cursor out of arrow. Please refer the video.
3. After 4th click you can see the white space in banner and after the mouse move all the banners stacks continuously.

## Workaround

No workaround.

## Screenshots

https://github.com/eBay/ebayui-core/assets/150841755/3ac06564-b33a-4458-bce6-de6316007cb3

https://github.com/eBay/ebayui-core/assets/150841755/34b9c8a7-1740-4f25-8f0e-c5de7e2a8bf8

PR #2033, opened 2023-11-30
fix(carousel): added delay to interacting state if carousel is moving

agliga:
## Description
* Added a delay to change state if carousel is moving
* I looked to see if this would cause other issues, but its such a short interval that its hard to trigger any other problems. 

## References
https://github.com/eBay/ebayui-core/issues/2032


LuLaValva (approved):
Looks good!

ArtBlue (approved):
Nice!

PR #2037, opened 2023-12-14
feat(tooltip): swapped to use floatingui

agliga:
## Description

Added floating ui to tooltip. 
At the moment left it to use limited positions. 
There might be a bug with positions but I think its a skin issue. 
This depends on the typescript PR to be merged first.


LuLaValva (approved):
Looks good, thanks Andrew! The only problem I noticed was that popper didn't change when the size of the container it was labeling changed (i. e. via screen resize), but I think that may be a limitation of floatingui and not your implementation
image

src/components/components/ebay-tooltip-base/constants.ts
```diff
@@ -0,0 +1,14 @@
+export const pointerStyles:{[index: string]: string} = {
```
LuLaValva, line 1: If you want to be more precise with this type, you can say that the value is `Placement`

LuLaValva (approved):
Quick comment, but you don't have to change it. I think you're good to merge.

src/components/components/ebay-tooltip-base/constants.ts
agliga: Good call... Ill swap it.

PR #2043, opened 2024-01-03
fix(combobox): prevent changing expanded state on input change

LuLaValva:
## Description

When `expanded` is not included as an attribute (and therefore is resolved as `undefined`), every time `input` changes the combobox closes.

### DIFF
--- a/src/components/ebay-combobox/component.ts
+++ b/src/components/ebay-combobox/component.ts
@@ -55,7 +55,7 @@ export default class Combobox extends Marko.Component<Input, State> {
     declare lastValue: Input["value"];
     declare autocomplete: NonNullable<Input["autocomplete"]>;
     declare listSelection: NonNullable<Input["listSelection"]>;
-    declare expanded: boolean;
+    declare expanded?: boolean;
     declare expandedChange: boolean;
     declare _floatingLabel: any;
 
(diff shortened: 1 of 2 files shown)


agliga (changes requested):
Okay, I see the issue
https://github.com/eBay/ebayui-core/pull/2043/files#diff-7bc6354baabed4ea933f1a5dff0401d8ca820c71a9ecf8922193edd2e7f86892R239
This line needs to be `this.expanded = input.expanded` simply. 
It used to be https://github.com/eBay/ebayui-core/blob/8b860920f7e20bfa258692bf29d5c72290f32ddc/src/components/ebay-combobox/component.js#L180
And it was like that because all it's doing is saving the last expanded state. If the last expanded state was null then it wouldn't do anything. Only when the `input.expanded` is swapped then it changes. 
I got this working locally, all we have to do is have `this.expanded` take nullable/undefined too.

ArtBlue (approved):

agliga (approved):

LuLaValva (commented):
Thanks @agliga and @ArtBlue, I've addressed the comments. Prettier added commas to the `test.browser.js` file, I hope that's ok.

ArtBlue (approved):

PR #2046, opened 2024-01-04
fix(ts): handle global imports in dist

LuLaValva:
## Description

Various TS fixes


agliga (approved):
Reviewed offline with Luke

PR #2047, opened 2024-01-04
chore(storybook): add missing 'skin/utility' and fix type imports

HenriqueLimas:
## Description

- Add `skin/utility` class on storybook
- Add `import type` to `global.ts` imports

## Context

Without `skin/utility` the `clipped` class is not styled properly. 
WIthout `import type`, the following error is show on storybook locally:

image

## Screenshots

Without skin/utility

image

With skin/utility

Screen Shot 2024-01-04 at 11 06 46 AM

### DIFF
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -1,8 +1,7 @@
 import "./custom-styles.less";
 import "@ebay/skin/tokens";
-import "@ebay/skin/global";
+import "@ebay/skin/core";
 import "@ebay/skin/marketsans";
-import "@ebay/skin/utility";
 
 export const parameters = {
     layout: "centered",
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -2,6 +2,7 @@ import "./custom-styles.less";
 import "@ebay/skin/tokens";
 import "@ebay/skin/global";
 import "@ebay/skin/marketsans";
+import "@ebay/skin/utility";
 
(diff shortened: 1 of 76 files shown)


.storybook/preview.ts
```diff
@@ -2,6 +2,7 @@ import "./custom-styles.less";
 import "@ebay/skin/tokens";
 import "@ebay/skin/global";
 import "@ebay/skin/marketsans";
+import "@ebay/skin/utility";
```
agliga, line 5: Lets remove this, change `@ebay/skin/global` to `@ebay/skin/core`
Core imports both global and utility.

package.json
```diff
@@ -33,7 +33,7 @@
         "tb": "npm run test:browser",
         "tf": "npm run test:fps",
         "tr": "npm run coverage && npm run report && open .coverage/index.html",
-        "storybook": "storybook -p 6006",
+        "storybook": "storybook dev -p 6006",
```
agliga, line 36: Why was this changed? Just curious

agliga (commented):
LGTM, two comments (one nit and another question)

package.json
HenriqueLimas: It is part of storybook 7 upgrade. The same is used on `npm start` so maybe this script is not needed

agliga (approved):

LuLaValva (approved):

PR #2049, opened 2024-01-04
refactor(ts): use AttrString for text instead of string

LuLaValva:
## Description

- Use `AttrString` instead of `string` when it makes for input types, so that `null` and `false` are accepted in addition to `undefined`. This is espeically useful for i18n, with internationalization functions that might be typed to return falsy values instead of `string | undefined`.


agliga:
LGTM, can you file an issue for this. I think we can get this in 13.1.0 branch

agliga (approved):

ArtBlue (approved):
Nice!

PR #2050, opened 2024-01-05
feat(date-textbox): allow clear by setting value to null

### LINKED ISSUE #2036: Clear input value on date-textbox
# Bug Report
On `ebay-date-textbox` it is not possible to clear the selected values

## eBayUI Version: x.x.x
DS v1.2.0

## Description
Need a way to clear the values in the inputs.
I have been using this date-textbox component: https://opensource.ebay.com/ebayui-core/?path=/story/form-input-ebay-date-textbox--localized&args=value:1701410400000;rangeEnd:1702620000000;range:!true and even after setting `value = undefined` the inputs don't clear

## Workaround
@LuLaValva mentioned this solution: https://ebay-eng.slack.com/archives/C0QG1HKFY/p1701818569015489?thread_ts=1701354710.424089&cid=C0QG1HKFY

## Screenshot
image

LuLaValva:
- resolves #2036 

## Description

- Allow textbox to be cleared by setting `value` to `null`

## Screenshots

image

### DIFF
--- a/src/components/ebay-date-textbox/component.ts
+++ b/src/components/ebay-date-textbox/component.ts
@@ -66,7 +66,7 @@ class DateTextbox extends Marko.Component<Input, State> {
     }
 
     onInput(input: Input) {
-        if (input.value) {
+        if (input.value !== undefined) {
             this.state.firstSelected = dateArgToISO(input.value);
         }
         if (input.rangeEnd) {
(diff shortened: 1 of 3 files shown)


src/components/ebay-date-textbox/component.ts
```diff
@@ -66,7 +66,7 @@ class DateTextbox extends Marko.Component<Input, State> {
     }
 
     onInput(input: Input) {
-        if (input.value) {
+        if (input.value !== undefined) {
```
ArtBlue, line 69: Do we need to account for empty strings in this conditional?

LuLaValva: No, we shouldn't need to. An empty string would also work to clear the date textbox, but devs shouldn't be passing any strings that don't work as ISO strings.

ArtBlue (approved):

agliga:
This is a feature and should be targeting `13.1.0`

agliga (approved):

PR #2052, opened 2024-01-10
fix(date-textbox): pass 'range' to calendar

HenriqueLimas:
## Description

- Pass `range` also to `ebay-calendar` to proper render the range

## Context

`...calendarInput` doesn't have `range` included since it is used also on date-textbox, so it needs to be passed down

## Screenshots

Before:

Screen Shot 2024-01-09 at 5 39 17 PM

After:
Screen Shot 2024-01-09 at 5 39 36 PM


LuLaValva (approved):
Thanks @HenriqueLimas, good catch.

agliga (approved):

PR #2055, opened 2024-01-10
feat(date-textbox): allow server rendering w/ document === undefined

LuLaValva:
## Description

When date textbox is server-rendered, it produces an error because document is not defined.

### DIFF
--- a/src/components/ebay-date-textbox/component.ts
+++ b/src/components/ebay-date-textbox/component.ts
@@ -49,7 +49,9 @@ class DateTextbox extends Marko.Component<Input, State> {
             popover: false,
         };
 
-        this.calculateNumMonths();
+        if (typeof document !== "undefined") {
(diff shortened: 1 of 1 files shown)


src/components/ebay-date-textbox/component.ts
```diff
@@ -79,6 +79,7 @@ class DateTextbox extends Marko.Component<Input, State> {
 
     calculateNumMonths() {
         this.state.numMonths =
+            typeof document !== "undefined" &&
```
agliga, line 82: I think it's fine,  but generally we put these calls into the lifecycle methods. So lets move this check to `onCreate` and then only call calulateNumMonths if that's defined.

LuLaValva: Good call, will move.

agliga (approved):

PR #2056, opened 2024-01-11
fix(breadcrumb): use right arguments

PrashantAshok:
## Description

Fixes the arguments being emitted. This also brings in the fix made here https://github.com/eBay/ebayui-core/pull/2015

Reproducible here https://opensource.ebay.com/ebayui-core/?path=/docs/navigation-disclosure-ebay-breadcrumbs--documentation

## Screenshots

Screen Shot 2024-01-11 at 2 46 50 PM


agliga (approved):

LuLaValva (approved):
Nice catch, thanks!

Issue #2053, opened 2024-01-10
ebay-tooltip: throws error on manually calling expand


agliga:
If expand is manually called, it can throw an error before it is properly initialized. 
Need to change
```    
handleExpand() {
        this.emit("base-expand");
        this.updateTip();
}
```
to 
```
    handleExpand() {
        this.emit("base-expand");
        if (this.hostEl && this.overlayEl) {
            this.updateTip();
        }
    }
```

PR #2061, opened 2024-01-17
fix(tooltip-base): added a check in case tooltip is not initialized

agliga:
## Description
* Added check if host is defined before expanding tooltip

## References
https://github.com/eBay/ebayui-core/issues/2053

### DIFF
--- a/src/components/components/ebay-tooltip-base/component-browser.ts
+++ b/src/components/components/ebay-tooltip-base/component-browser.ts
@@ -40,7 +40,9 @@ class TooltipBase extends Marko.Component<Input> {
 
     handleExpand() {
         this.emit("base-expand");
-        this.updateTip();
+        if (this.hostEl && this.overlayEl) {
+            this.updateTip();
+        }
     }
 
     handleCollapse() {


ArtBlue (approved):
Looks good.

src/components/components/ebay-tooltip-base/component-browser.ts
```diff
@@ -40,7 +40,9 @@ class TooltipBase extends Marko.Component<Input> {
 
     handleExpand() {
         this.emit("base-expand");
-        this.updateTip();
+        if (this.hostEl && this.overlayEl) {
```
LuLaValva, line 43: Would it make sense to move this check into the `updateTip` function instead of adding it as a wrapper each time `updateTip` is called?

LuLaValva (approved):
Just one comment, but if you disagree you can go ahead and merge.

src/components/components/ebay-tooltip-base/component-browser.ts
agliga: We can refactor, but the reason is because it already is being checked on setupExpander
```
        if (this.hostEl && this.overlayEl) {
            this.updateTip();
            this.cleanup = autoUpdate(
                this.hostEl,
                this.overlayEl,
                this.update.bind(this),
            );
        }
```
And we dont want to call autoUpdate without hostEl and overlayEl being defined. 
So I don't want this check to be called multiple times.

PR #2062, opened 2024-01-18
feat(chips-combobox): add component

LuLaValva:
## Description

- Add `chips-combobox` component
- Some modifications needed to be made to `chips` and `combobox`, but none are breaking
- Right now the only behavior allows any string to be included as a chip regardless of whether it's an autofill option

## Screenshots

image


src/components/ebay-chips-combobox/index.marko
```diff
@@ -0,0 +1,42 @@
...
+                <li>
+                    <ebay-chip a11y-delete-button=a11yDeleteButton on-delete("handleDelete", i) disabled=disabled>
+                        ${option}
+                    </ebay-chip>
+                </li>
+            </for>
+        </ul>
+    </if>
+    <ebay-combobox
+        class=["chips-combobox__combobox", inputClass]
```
agliga, line 26: Generally we want to put the extra classes on the root. The reason is sometimes users might want to be able to use those classes as selectors.

tsconfig.json
```diff
@@ -9,8 +9,8 @@
     "compilerOptions": {
         "allowSyntheticDefaultImports": true,
         "incremental": true,
-        "lib": ["dom"],
-        "module": "commonjs",
+        "lib": ["dom", "ESNext"],
```
agliga, line 12: Lets push this change directly to master or to 13.1.0. Lets not have it part of this PR

agliga (commented):
LGTM overall I think. Couple of minor comments but I think its good to go

ArtBlue (commented):
Nice work! The code itself looks good. 

There appear to be `a11y` violations. Some, for example, the first two about `forms` may not be legitimate, but others likely are.

@ianmcburnie, @saiponnada , thoughts? 

Maybe we missed a couple of these in Skin?

image

agliga:
> Nice work! The code itself looks good. 
> 
> 
> 
> There appear to be `a11y` violations. Some, for example, the first two about `forms` may not be legitimate, but others likely are.
> 
> 
> 
> @ianmcburnie, @saiponnada , thoughts? 
> 
> 
> 
> Maybe we missed a couple of these in Skin?
> 
> 
> 
> image
> 
> 

For isolated view this should be fine
We might need an example which has a label.

ArtBlue (approved):
It looks like `combobox` has the same issues marked as "violations." If we determine, these are issues, we can fix them together at some later point.

The rest looks good.

agliga (approved):

Issue #2057, opened 2024-01-12
ebay-listbox-button: add error state


agliga:
Skin issue: https://github.com/eBay/skin/issues/2063

PR #2063, opened 2024-01-19
feat(listbox-button): added error state

agliga:
## Description
In order to support new listbox error, added a new error state
We can use `isInvalid` attribute but that was existing and I did not want to overload it.

## References
https://github.com/eBay/ebayui-core/issues/2057

## Screenshots

Screenshot 2024-01-18 at 9 53 51 AM


ArtBlue (approved):
Nice!

Issue #2066, opened 2024-01-22
Docs: alert-dialog, confirm-dialog adjustments


ArtBlue:
# Docs Issue
The `a11y` team found some minor things in the storybook can be confusing for devs implementing Alert Dialog or Confirm Dialog.

## Description
1. Since Alert Dialog is simply an alert and not a confirm, we need to change the button language from "Confirm" to "OK."
2. Since Confirm Dialog requires the user to confirm, the current button with the text, "Okay" should be changed to "Confirm."
3. Instead of the text for both dialogs having placeholder lorem ipsum text, we need to change the text to be descriptive of the dialog as well as make it much shorter for a clear guidance on the length of those dialog texts being no more than 2 sentences. The reason for the imposed brevity is to provide quick and clear information as well as avoid multi-line centering of text in the dialogs. For guidance, the Skin docs currently contain what should be done here.

### Ebay UI
<img width="671" alt="image" src="https://github.com/eBay/ebayui-core/assets/1675667/fa620054-a60a-45d0-9bf2-181d07b2d101">

<img width="687" alt="image" src="https://github.com/eBay/ebayui-core/assets/1675667/9754c356-c631-4fea-b057-2495fb086ecb">

### Skin
<img width="668" alt="image" src="https://github.com/eBay/ebayui-core/assets/1675667/5486ce92-3481-41d4-9e00-b4ed4f048d50">

<img width="676" alt="image" src="https://github.com/eBay/ebayui-core/assets/1675667/07191109-3b52-4840-9892-a5267ccb5587">

PR #2067, opened 2024-01-22
docs(alert-dialog,confirm-dialog): made the docs match expected usage

ArtBlue:
## Description

Issue: #2066

Fixes some minor things in storybook docs that can be confusing for devs implementing Alert Dialog or Confirm Dialog. Also, provides some guidance for usage.

1. Since Alert Dialog is simply an alert and not a confirm, we need to change the button language from "Confirm" to "OK."
2. Since Confirm Dialog requires the user to confirm, the current button with the text, "Okay" should be changed to "Confirm."
3. Instead of the text for both dialogs having placeholder lorem ipsum text, we need to change the text to be descriptive of the dialog as well as make it much shorter for a clear guidance on the length of those dialog texts being no more than 2 sentences. The reason for the imposed brevity is to provide quick and clear information as well as avoid multi-line centering of text in the dialogs. For guidance, the Skin docs currently contain what should be done here.

### Before
**Alert Dialog**

<kbd>image</kbd>

**Confirm Dialog**

<kbd>image</kbd>

### After
**Alert Dialog**

<kbd>image</kbd>

**Confirm Dialog**

<kbd>image</kbd>


agliga:
Can you rebase from 13.1.0?

agliga (changes requested):
LGTM. Ill approve after rebase

agliga (approved):

PR #2073, opened 2024-01-25
refactor(chips-combobox): use attr tags for options

LuLaValva:
## Description

- Refactor to allow _either_ a list input or a list of attr tags

### DIFF
--- a/src/components/ebay-chips-combobox/marko-tag.json
+++ b/src/components/ebay-chips-combobox/marko-tag.json
@@ -9,7 +9,7 @@
     "@fluid": "boolean",
     "@error": "boolean",
     "@list-selection": "string",
-    "@option <option>": {
+    "@option <option>[]": {
         "attribute-groups": ["html-attributes"],
         "@*": {
             "targetProperty": null,
--- a/src/components/ebay-chips-combobox/marko-tag.json
+++ b/src/components/ebay-chips-combobox/marko-tag.json
@@ -5,23 +5,11 @@
         "type": "expression"
     },
     "@html-attributes": "expression",
-    "@borderless": "boolean",
-    "@autocomplete": "string",
     "@expanded": "boolean",
-    "@list-selection": {
-        "enum": ["automatic", "manual"]
-    },
-    "@disabled": "#html-disabled",
-    "@form": "#html-form",
-    "@name": "#html-name",
-    "@placeholder": "#html-placeholder",
-    "@required": "#html-required",
-    "@value": "#html-value",
-    "@role": "never",
-    "@aria-autocomplete": "never",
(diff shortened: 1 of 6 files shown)


agliga (approved):
Verify this approach with dylan. I think there might be an issue if we are defining options twice in marko.json (one as array and one as attribute tags)

LuLaValva:
>  I think there might be an issue if we are defining options twice
We're actually not defining options twice with this approach. We're defining `options` which can be an array, and `option` which is a repeatable list of attr tags.

agliga:
> >  I think there might be an issue if we are defining options twice
> 
> 
> 
> We're actually not defining options twice with this approach. We're defining `options` which can be an array, and `option` which is a repeatable list of attr tags.

Ok. But lets rename it to something else. Option is a bad name since it is a singular and there are multiple options

src/components/ebay-chips-combobox/marko-tag.json
```diff
@@ -5,43 +5,17 @@
...
-    "@value": "#html-value",
-    "@role": "never",
-    "@aria-autocomplete": "never",
-    "@aria-haspopup": "never",
-    "@floating-label": "string",
-    "@actionable <actionable>": {
+    "@fluid": "boolean",
+    "@error": "boolean",
+    "@list-selection": "string",
+    "@option <option>": {
```
agliga, line 12: I think this needs to be `"@options <option>[]": {`

See combobox

LuLaValva: This isn't true for Marko 5 anymore since the new `iterable` changes that were included around the same time as TS was introduced, but it might cause problems in Marko 4. Good catch, I changed it.

src/components/ebay-chips-combobox/marko-tag.json
```diff
@@ -5,43 +5,17 @@
...
-    "@value": "#html-value",
-    "@role": "never",
-    "@aria-autocomplete": "never",
-    "@aria-haspopup": "never",
-    "@floating-label": "string",
-    "@actionable <actionable>": {
+    "@fluid": "boolean",
+    "@error": "boolean",
+    "@list-selection": "string",
+    "@option <option>[]": {
```
agliga, line 12: I think this needs to be 
`"@options <option>[]": {`
Then in the code we should reference `input.options`
See https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-combobox/marko-tag.json#L32
`@options` is the name of the input when you get it in code, where as the `<option>` part is the name of the attribute tag.

agliga (changes requested):

agliga (approved):

PR #2080, opened 2024-01-30
fix(highcharts): use dynamic import instead of CDN, with lasso polyfill

LuLaValva:
## Description

- Use dynamic import instead of CDN, with lasso polyfill

## Context

- Highcharts was not working on storybook, and some apps reported that it caused errors.

### DIFF
--- a/src/common/cdn/index.ts
+++ b/src/common/cdn/index.ts
@@ -72,6 +72,7 @@ export class CDNLoader {
         this.url = `https://ir.ebaystatic.com/cr/v/c1/ebayui/${this.key}/v${
             version || versions[this.key as keyof typeof versions]
         }`;
+        console.log(this.url);
         for (let i = 0; i < this.files.length; i++) {
             if (overrides && overrides[i]) {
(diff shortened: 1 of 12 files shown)


agliga:
Not sure why this is changed. This was working before SWU. 

We should not change it like this becasue this will block pages loading. We want to load it after the fact. 

I don't think this is the right fix.

LuLaValva:
This is not blocking to the render because it is a dynamic import, but I agree it may not be the right fix. We should discuss with @DylanPiercey before moving forward.

agliga:
> This is not blocking to the render because it is a dynamic import, but I agree it may not be the right fix. We should discuss with @DylanPiercey before moving forward.

So one more issue. There is an issue assigned to me this sprint. I had already started to look at it but this seems very inefficient since we're both doing the same thing
I would also add this has two different fixes, one for charts and one for tooltip. We need to have these in separate prs. 
Please reach out to me before implementing changes so we can discuss and not have throw away code.

LuLaValva:
Closing in favor of #2083

PR #2100, opened 2024-02-15
fix(segmented-button): move types and validSizes to marko file

LuLaValva:
## Description

- Move types and `validSizes` to the `.marko` file from `component.ts`

### DIFF
--- a/.changeset/afraid-mugs-search.md
+++ b/.changeset/afraid-mugs-search.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
+---
+
+fix segmented-buttons-import error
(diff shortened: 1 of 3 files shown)


agliga (commented):
Do we really need all these changes?I thought we just need to change the import
Also, can we verify with Dylan that it's OK to import the template from the component?

LuLaValva:
Since we moved `validSizes` into the `.marko` file it made sense to also move the TypeScript interfaces (especially for consistency with other components that do the same thing). Without changing the `validSizes` from `static` to `export`, this is also the only place that we can access its type now.

agliga (approved):

PR #2105, opened 2024-02-27
13.2.0

agliga:
Merge to master for release


package.json
```diff
@@ -72,23 +74,23 @@
         "@commitlint/cli": "^18",
         "@commitlint/config-conventional": "^18",
         "@ebay/browserslist-config": "^2.7.0",
-        "@ebay/skin": "~17.1.0",
+        "@ebay/skin": "~17.2.0-next.0",
```
ArtBlue, line 77: Should this be changed to the actual release or will changeset do that automatically during the release?

docs/assets/getting-started.stories-fKLBwFDx.js
ArtBlue, line 1: What's up with all these auto-generated docs `js` files?

ArtBlue (commented):
Looks good. Just a couple of questions and a possible Skin version change.

docs/assets/getting-started.stories-fKLBwFDx.js
agliga: Those are part of the build and needed for release. 
Usually those are part of the commit that we push up to master but they ended up being part of this prerelease commit

package.json
agliga: I'll swap it after skin release

ArtBlue (approved):

PR #2110, opened 2024-03-05
fix(tooltip-base): added option for no flip

agliga:
There is an edge case where flip is not working randomly. I added an option to disable it for now.


ArtBlue (approved):
Looks good.

PR #2118, opened 2024-03-11
fix(date-textbox): SSR issue & dates past year 9999

### LINKED ISSUE #2087: ebay-date-textbox allows number input
# Bug Report

## Description

When we input a number which can be parsed as a date, and it ends up being buggy because it's so far in the future. 
We should filter out cases like this
## Screenshots

Screenshot 2024-02-01 at 3 24 15 PM

### LINKED ISSUE #2099: date-textbox misalign
## Description

when the calendar is shown at the date-textbox, it shows the dates with a margin offset

## Screenshots

image

LuLaValva:
## Description

- Added `type` to an import to prevent errors in storybook (bonus change)
- Force client recalculation of `firstDayOfWeek` in case the browser doesn't support it but the server does (fixes #2099)
- Ensure that selected dates can be represented in `YYYY-MM-DD` format (fixes #2087)

### DIFF
--- a/src/components/ebay-calendar/date-utils.ts
+++ b/src/components/ebay-calendar/date-utils.ts
@@ -32,7 +32,8 @@ export function getWeekdayInfo(localeName: string) {
 export function dateArgToISO(arg: DateConstructor["arguments"]) {
     if (!arg) return undefined;
     if (/^\d\d\d\d-\d\d-\d\d$/g.test(arg)) return arg;
-    return toISO(new Date(arg));
+    const date = toISO(new Date(arg));
+    return /^\d\d\d\d-\d\d-\d\d$/g.test(date) ? date : undefined;
 }
 
 export function toISO(date: Date): DayISO {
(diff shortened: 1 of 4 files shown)


src/components/ebay-calendar/date-utils.ts
```diff
@@ -32,7 +32,8 @@ export function getWeekdayInfo(localeName: string) {
 export function dateArgToISO(arg: DateConstructor["arguments"]) {
     if (!arg) return undefined;
     if (/^\d\d\d\d-\d\d-\d\d$/g.test(arg)) return arg;
-    return toISO(new Date(arg));
+    const date = toISO(new Date(arg));
+    return /^\d\d\d\d-\d\d-\d\d$/g.test(date) ? date : undefined;
```
agliga, line 36: Is this always going to match all formats? What if there is only 1 day in the date?

LuLaValva: Yes, this is ISO format. If there is one day it will be printed with a leading zero

agliga (approved):

PR #2122, opened 2024-03-12
fix(signal): custom class passthrough

### LINKED ISSUE #2121: ebay-signal does not passthrough `class`
# Bug Report

## Description

`ebay-signal` does not passthrough `class` attribute, e.g. when used like
```marko
<ebay-signal status=level class="ui-hotness-signal">
```
This seems to be a miss in #2039 as it is missing the `class: inputClass` added in other components.

mkchang:
## Description
* Resolves #2121

## Context
* Adding use of `inputClass` as in other components.

### DIFF
--- a/.changeset/hungry-taxis-behave.md
+++ b/.changeset/hungry-taxis-behave.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
+---
+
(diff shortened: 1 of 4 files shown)


LuLaValva (approved):
Looks great, thanks!

agliga (approved):

Issue #2120, opened 2024-03-12
ebay-education-notice: new component


ArtBlue:
# Feature Request

New interactive component based on https://github.com/eBay/skin/issues/2271 .

PR #2125, opened 2024-03-14
feat(education-notice): added component

agliga:
## Description
Added education notice new component

## Context
For now, I used notice-base but we need to refactor it a bit next major. 

## References
#2120

## Screenshots
Screenshot 2024-03-13 at 7 58 12 PM
Screenshot 2024-03-13 at 7 58 18 PM
Screenshot 2024-03-13 at 7 58 22 PM
Screenshot 2024-03-13 at 7 58 25 PM
Screenshot 2024-03-13 at 7 58 30 PM

### DIFF
--- a/src/components/components/ebay-notice-base/index.marko
+++ b/src/components/components/ebay-notice-base/index.marko
@@ -71,6 +71,7 @@ $ const isEducation = prefixClass === 'education-notice';
                 <if(a11yDismissText)>
                         <ebay-icon-button
                             aria-label=`${a11yDismissText}`
+                            size="small"
                             class=['fake-link', `${prefixClass}__dismiss`]
                             onClick('emit', 'dismiss')
                         >
--- a/src/components/components/ebay-notice-base/index.marko
+++ b/src/components/components/ebay-notice-base/index.marko
@@ -69,14 +69,13 @@ $ const isEducation = prefixClass === 'education-notice';
                     <${title}/>
                 </>
                 <if(a11yDismissText)>
-                        <button
+                        <ebay-icon-button
                             aria-label=`${a11yDismissText}`
                             class=['fake-link', `${prefixClass}__dismiss`]
                             onClick('emit', 'dismiss')
-                            onKeydown('emit', 'dismiss')
                         >
-                            <ebay-close-16-icon.icon.icon--close-16/>
-                        </button>
+                            <ebay-close-16-icon/>
+                        </ebay-icon-button>
                 </if>
             </if>
         </>
--- a/src/components/components/ebay-notice-base/index.marko
+++ b/src/components/components/ebay-notice-base/index.marko
@@ -19,11 +19,12 @@ $ const {
     noA11yLabel,
     a11yDismissText,
     educationIcon,
-    prominent,
+    prominent, // deprecated, remove next major
     cta,
     ...htmlInput
 } = input;
 
+$ const isEducation = prefixClass === 'education-notice';
 <${root || 'section'}
     aria-labelledby=!noA11yLabel && component.elId('status')
     aria-roledescription=a11yRoleDescription
@@ -53,16 +54,35 @@ $ const {
                     class=iconClass
                 />
             </else-if>
-            <else-if(status === 'education' && type === 'section')>
+            // deprecated status === education. We should probably remove this whole icon on the next major version
+            <else-if(isEducation || (status === 'education' && type === 'section'))>
                 <${educationIcon || LightBulbIcon}
                     a11y-variant='label'
                     a11y-text=a11yIconText || a11yText
                     class=iconClass/>
             </else-if>
+            <if(isEducation && title)>
+                <${title.as || 'h2'}
+                    ...title
+                    class=[`${prefixClass}__title`, title.class]
+                >
+                    <${title}/>
+                </>
+                <if(a11yDismissText)>
+                        <button
+                            aria-label=`${a11yDismissText}`
+                            class=['fake-link', `${prefixClass}__dismiss`]
+                            onClick('emit', 'dismiss')
+                            onKeydown('emit', 'dismiss')
+                        >
+                            <ebay-close-16-icon.icon.icon--close-16/>
+                        </button>
+                </if>
+            </if>
         </>
     </if>
     <${mainRoot || 'div'} class=`${prefixClass}__main`>
-        <if(title)>
+        <if(!isEducation && title)>
             <${title.as || 'h2'}
                 ...title
                 class=[`${prefixClass}__title`, title.class]
@@ -76,10 +96,10 @@ $ const {
         <p class=`${prefixClass}__cta`><a ...cta><${cta}/></a></p>
     </if>
 
-    <if(footer && !a11yDismissText)>
+    <if(footer && (isEducation || !a11yDismissText))>
         <div class=`${prefixClass}__footer`><${footer}/></div>
     </if>
-    <if(!footer && a11yDismissText)>
+    <if(!isEducation && !footer && a11yDismissText)>
         <div class=`${prefixClass}__footer`>
             <button
                 aria-label=`${a11yDismissText}`
--- a/src/components/ebay-education-notice/README.md
+++ b/src/components/ebay-education-notice/README.md
@@ -1,0 +1,16 @@
+<h1 style='display: flex; justify-content: space-between; align-items: center;'>
+    <span>
+        ebay-education-notice
+    </span>
+    <span style='font-weight: normal; font-size: medium; margin-bottom: -15px;'>
+        DS v1.0.0
+    </span>
+</h1>
+
+The `<ebay-education-notice>` is a tag used to create a custom-designed notice element. The notice can be single or multi-line but each line should be wrapped inside a `<p>` tag.
+
+## Examples and Documentation
+
+-   [Storybook](https://ebay.github.io/ebayui-core/?path=/story/notices-tips-ebay-education-notice)
+-   [Storybook Docs](https://ebay.github.io/ebayui-core/?path=/docs/notices-tips-ebay-education-notice)
+-   [Code Examples](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-education-notice/examples)
--- a/src/components/ebay-education-notice/README.md
+++ b/src/components/ebay-education-notice/README.md
@@ -7,7 +7,7 @@
     </span>
 </h1>
 
-The `<ebay-education-notice>` is a tag used to create a custom-designed notice element. The notice can be single or multi-line but each line should be wrapped inside a `<p>` tag.
+The `<ebay-education-notice>` is a tag used to create an education notice element. The notice requires a `@title` element and should be wrapped inside a `<p>` tag.
 
 ## Examples and Documentation
 
--- a/src/components/ebay-education-notice/component.ts
+++ b/src/components/ebay-education-notice/component.ts
@@ -1,7 +1,7 @@
 import type { Input as NoticeBaseInput } from "../components/ebay-notice-base/component";
 import type { WithNormalizedProps } from "../../global";
 
-interface SectionNoticeInput
+interface EducationNoticeInput
     extends Omit<
         NoticeBaseInput,
         "role" | "prefixClass" | "mainRoot" | `on${string}`
@@ -12,13 +12,13 @@ interface SectionNoticeInput
     'icon-variant'?: 'none' | 'prominent';
 }
 
-export interface Input extends WithNormalizedProps<SectionNoticeInput> {}
+export interface Input extends WithNormalizedProps<EducationNoticeInput> {}
 
 interface State {
     dismissed: boolean;
 }
 
-class SectionNotice extends Marko.Component<Input, State> {
+class EducationNotice extends Marko.Component<Input, State> {
     onCreate() {
         this.state = { dismissed: false };
     }
@@ -33,4 +33,4 @@ class SectionNotice extends Marko.Component<Input, State> {
     }
 }
 
-export default SectionNotice;
+export default EducationNotice;
--- a/src/components/ebay-education-notice/component.ts
+++ b/src/components/ebay-education-notice/component.ts
@@ -1,0 +1,36 @@
+import type { Input as NoticeBaseInput } from "../components/ebay-notice-base/component";
+import type { WithNormalizedProps } from "../../global";
+
+interface SectionNoticeInput
+    extends Omit<
+        NoticeBaseInput,
+        "role" | "prefixClass" | "mainRoot" | `on${string}`
+    > {
+    dismissed?: boolean;
+    "on-dismiss"?: () => void;
+    variant?: 'none' | 'prominent';
+    'icon-variant'?: 'none' | 'prominent';
+}
+
+export interface Input extends WithNormalizedProps<SectionNoticeInput> {}
+
+interface State {
+    dismissed: boolean;
+}
+
+class SectionNotice extends Marko.Component<Input, State> {
+    onCreate() {
+        this.state = { dismissed: false };
+    }
+
+    onInput(input: Input) {
+        this.state = { dismissed: input.dismissed || false };
+    }
+
+    onDismiss() {
+        this.state.dismissed = true;
+        this.emit("dismiss");
+    }
+}
+
+export default SectionNotice;
--- a/src/components/ebay-education-notice/examples/prominent-icon.marko
+++ b/src/components/ebay-education-notice/examples/prominent-icon.marko
@@ -1,0 +1,13 @@
+import vaultIcon from '<ebay-the-ebay-vault-24-icon>';
+
+<ebay-education-notice ...input a11y-text="ebay vault" icon-variant="prominent" education-icon=vaultIcon>
+    <@title>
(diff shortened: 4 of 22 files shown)


src/components/ebay-education-notice/examples/prominent-icon.marko
```diff
@@ -0,0 +1,13 @@
+import vaultIcon from '<ebay-the-ebay-vault-24-icon>';
+
+<ebay-education-notice ...input a11y-text="ebay vault" icon-variant="prominent" education-icon=vaultIcon>
+    <@title>
+        <div>Notice title</div>
+    </@title>
+    <p>
```
ArtBlue, line 7: The markup in the examples doesn't seem right. It should result in this:

```html
<section class="education-notice" role="region" aria-label="Education" aria-roledescription="Notice">
    <div class="education-notice__header">
        <svg class="icon icon--the-ebay-vault-24" height="24" width="24" aria-label="Education">
            <use href="#icon-the-ebay-vault-24"></use>
        </svg>
        <h3 class="education-notice__title">The Vault</h3>
        <button aria-label="Close notice" class="fake-link education-notice__dismiss">
            <svg aria-hidden="true" class="icon icon--close-16" height="16" width="16">
                <use href="#icon-close-16"></use>
            </svg>
        </button>
    </div>
    <div class="education-notice__main">
        <p>You can now use the Vault to store items like this securely.</p>
    </div>
    <div class="education-notice__footer">
        <button class="fake-link">Learn more button</button>
    </div>
</section>
```

The `p` tags here seem like they are just floating here and this is much more flat than the Skin markup. 
image

It results in incorrect spacing as there should be a gap between the description and the CTA.
image

src/components/ebay-education-notice/examples/with-footer.marko
```diff
@@ -0,0 +1,15 @@
+import type { Input as EducationNoticeInput } from "<ebay-education-notice>";
+export type Input = EducationNoticeInput;
+
+<ebay-education-notice ...input>
+    <@title>
+        <div>Educaiton title</div>
+    </@title>
+    <@footer>
```
ArtBlue, line 8: The CTA actually creates the `footer` since it's the primary (and currently only) element there.

src/components/ebay-education-notice/component.ts
```diff
@@ -0,0 +1,36 @@
+import type { Input as NoticeBaseInput } from "../components/ebay-notice-base/component";
+import type { WithNormalizedProps } from "../../global";
+
+interface SectionNoticeInput
+    extends Omit<
+        NoticeBaseInput,
+        "role" | "prefixClass" | "mainRoot" | `on${string}`
+    > {
+    dismissed?: boolean;
```
ArtBlue, line 9: `title` should probably be made required here.

src/components/components/ebay-notice-base/index.marko
```diff
@@ -53,16 +54,35 @@ $ const {
...
             </else-if>
-            <else-if(status === 'education' && type === 'section')>
+            // deprecated status === education. We should probably remove this whole icon on the next major version
+            <else-if(isEducation || (status === 'education' && type === 'section'))>
                 <${educationIcon || LightBulbIcon}
                     a11y-variant='label'
                     a11y-text=a11yIconText || a11yText
                     class=iconClass/>
             </else-if>
+            <if(isEducation && title)>
```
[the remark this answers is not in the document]
ArtBlue: Here's the Figma for reference: https://www.figma.com/file/zEBdEhbonrBOGzZ0fXzWvM/branch/rSJeczAN0rQ0jA41TOSJrY/eBay-Design-System?type=design&node-id=90732-97760&mode=design&t=8JhV7bFCyecyl5sE-0

ArtBlue (changes requested):

src/components/components/ebay-notice-base/index.marko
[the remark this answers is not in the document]
agliga: We have that in order to not throw an error if its not set.

src/components/components/ebay-notice-base/index.marko
```diff
@@ -53,16 +54,35 @@ $ const {
                     class=iconClass
                 />
             </else-if>
-            <else-if(status === 'education' && type === 'section')>
+            // deprecated status === education. We should probably remove this whole icon on the next major version
```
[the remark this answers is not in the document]
agliga: Not sure what you are trying to say here, users can add whatever they want here.

src/components/ebay-education-notice/component.ts
```diff
@@ -0,0 +1,36 @@
+import type { Input as NoticeBaseInput } from "../components/ebay-notice-base/component";
+import type { WithNormalizedProps } from "../../global";
+
+interface SectionNoticeInput
```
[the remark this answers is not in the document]
agliga: Yes, will change.

src/components/ebay-education-notice/examples/prominent-icon.marko
agliga: This just shows there are multiple <p> elements
The CTA should be usually in the footer.

src/components/ebay-education-notice/examples/with-footer.marko
agliga: Not sure what this means.
Footer allows them to put what they want. We put a CTA there to show thats what it should have.

src/components/ebay-education-notice/component.ts
agliga: Title comes from notice base input. We use it from there.

agliga (commented):
I think part of this as I mentioned in the description is done to get education notice out. 
When we refactor our notices next major version (which we should have last major just did not have time), we will also refactor notice-base. But for now, keeping notice base like this because it is easier this way.

src/components/ebay-education-notice/README.md
```diff
@@ -0,0 +1,16 @@
+<h1 style='display: flex; justify-content: space-between; align-items: center;'>
+    <span>
+        ebay-education-notice
+    </span>
+    <span style='font-weight: normal; font-size: medium; margin-bottom: -15px;'>
+        DS v1.0.0
+    </span>
+</h1>
+
+The `<ebay-education-notice>` is a tag used to create a custom-designed notice element. The notice can be single or multi-line but each line should be wrapped inside a `<p>` tag.
```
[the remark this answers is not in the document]
agliga: I think part of this is what we have in storybook already. We have prominent in stroybook already, we don't need it in two places. 
We were getting to really longwinded readme's and we wanted to make them very brief. This info is usually found in storybook.

src/components/components/ebay-notice-base/index.marko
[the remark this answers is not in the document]
agliga: I already have it required in storybook
https://github.com/eBay/ebayui-core/pull/2125/files#diff-33cda131078de01ebef1030c9a7f2a001384f71e149ff9dd6ed28eb310609c96R83

src/components/components/ebay-notice-base/index.marko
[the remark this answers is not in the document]
ArtBlue: The footer should only have a CTA; users should not be able to do whatever they want. I simply included a reference to the markup to demonstrate how it should be structured. `education-notice` is a lot more explicit how the sections should work and what they should have.

src/components/ebay-education-notice/README.md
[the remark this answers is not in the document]
ArtBlue: I don't get it. This is where the users land when they "enter" into the component docs. What's wrong with having more complete info? I guess you can ignore my suggestion if you really want to, but we should avoid having inaccurate information.

According to the DS specs, this is simply not accurate and misleading: `The notice can be single or multi-line`

src/components/components/ebay-notice-base/index.marko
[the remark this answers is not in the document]
agliga: For now, will leave it as is. 
There could be multiple ways a person would add a CTA, like they might want to add their own custom component for it, or use fake-link etc. This also requires us to handle all those events on the CTA like focus, click, link etc. 
We have some places where we have it more explicit if its simple events, but for more complex interactions we allow users to pass in their component.

src/components/ebay-education-notice/examples/prominent-icon.marko
ArtBlue: There are no examples with links in the body in the Figma, nor multiple `<p>` tags. Those are not approved use cases for this component. As such, I did not build affordance for those use cases into Skin. In fact, not only does the Figma not have the use case for multiple `<p>` tags, but it actually includes a character limit of 120 characters for the description/detail, which is indicative of a short single paragraph of text.

Building things into ebay ui that do not have corresponding use cases supported in Skin is simply going to break the layout.

ianmcburnie:
> I think part of this as I mentioned in the description is done to get education notice out.

Is there a crazy rush/deliverable date? Typically we prefer not to rush things, especially if there are outstanding PR comments/concerns. Let's take the time to address those properly. We can always release this mid-SWU cycle for those that want to pick it up before the next SWU if we need.

src/components/components/ebay-notice-base/index.marko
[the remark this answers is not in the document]
ArtBlue: If we do that, how can we ensure the styling from DS remains intact? What if they use a custom component that looks completely different and unsupported by DS and Skin? We're ok with the way things would look? Can't we just have a link with an href or a button that emits the click and the dev does whatever they want with the click?

agliga:
> > I think part of this as I mentioned in the description is done to get education notice out.
> 
> Is there a crazy rush/deliverable date? Typically we prefer not to rush things, especially if there are outstanding PR comments/concerns. Let's take the time to address those properly. We can always release this mid-SWU cycle for those that want to pick it up before the next SWU if we need.
> 
> I also understand that we might have to pick up some tech debt that will be cleaned up in next major.

There is no rush, that was not what I meant.
I said that this is done this way to get education notice out so we dont have to spend a long time refactoring code and doing tech debt, which as I mentioned, we should, but just not now. 
We can redo notice base completely but it seemed A) outside the scope of this and B) will be more tech debt until we fix the skin markup as well. 
This was brought up before and Im just calling it out so we don't go on a tangent about that. That's all.

ianmcburnie:
> There is no rush, that was not what I meant.

Gotcha! Thanks for clarifying, Andrew.

src/components/ebay-education-notice/examples/prominent-icon.marko
agliga: I removed the second `<p>` in the examples.

That said, we don't have any places where we restrict the character count. We have done `line-clamp` to restrict the line numbers, but not the amount of text. We have this on buttons which have a max char count as well but don't restrict it.

src/components/ebay-education-notice/component.ts
ArtBlue: In `notice-base` it's optional, right? In `education-notice` it's required.

src/components/components/ebay-notice-base/index.marko
```diff
@@ -76,10 +96,10 @@ $ const {
         <p class=`${prefixClass}__cta`><a ...cta><${cta}/></a></p>
     </if>
 
-    <if(footer && !a11yDismissText)>
+    <if(footer && (isEducation || !a11yDismissText))>
```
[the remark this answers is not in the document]
ArtBlue: I guess if the use cases work, that's fine. It just seemed like an issue at first glance.

src/components/ebay-education-notice/examples/prominent-icon.marko
ianmcburnie: Hmm, not sure I agree with design here in terms of trying to forcefully restrict the character count. For web I would prefer things to be more _resilient_, and not have layout breakage even if a user does something they shouldn't do and puts in multiple paragraphs. I think we have some stories to cover these kinds of edge cases in skin (maybe page notice). On web lets encourage wrapping of text where we can easily accommodate it.

src/components/ebay-education-notice/component.ts
ianmcburnie: Figma is showing title as optional:

Screenshot 2024-03-14 at 1 01 36 PM

ArtBlue: That's not what I see. Maybe another change that was made recently after I was done with the work?

image

ArtBlue: I'm not sure where/how you got to that Figma file, but the one linked in the story still has the title as required. What you're seeing might be older.

src/components/ebay-education-notice/examples/prominent-icon.marko
ArtBlue: There might be some confusion here. I didn't actually build anything in Skin to honor the character limit. I wasn't intending ebay ui to honor it either. I only mentioned it here to paint the specific limited use of the component as intended by DS. It would be very odd to allow multiple paragraphs of detailed text here if the original intent was to have a very brief snippet of text: "_is indicative of a short single paragraph of text._"

src/components/ebay-education-notice/README.md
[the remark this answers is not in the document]
ArtBlue: This is still inaccurate: "The notice can be single or multi-line"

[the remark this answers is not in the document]
agliga:
```
The `<ebay-education-notice>` is a tag used to create an education notice element. The notice requires a `@title` element and should be wrapped inside a `<p>` tag.
```
This is what I have currently in code. https://github.com/eBay/ebayui-core/pull/2125/files#diff-279dbdd569cd0d80d5c06efe3bd5f3e71f7376df7921f3ebd2453ad370182addR10

src/components/ebay-education-notice/examples/prominent-icon.marko
agliga: I mean even if you look in the initial example, with multiple paragraphs, it looks fine and not broken. There isn't a lot of spacing between them, but it seems fine.

src/components/ebay-education-notice/examples/with-footer.marko
agliga: Resolving for now, please reopen if needed and add more comments.

src/components/components/ebay-notice-base/index.marko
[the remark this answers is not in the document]
agliga: We can, but we also need to allow them to change it to a link or button. If they want it to be a link, it should have href, otherwise we need to emit a click event. 
This is what the code looks like in both cases:
Current
```
<ebay-education-notice>
   <@footer><ebay-fake-link on-click("trigger")>Link</></@footer>
</>
``` 

Proposed
```
<ebay-education-notice on-cta-button-click('trigger')>
   <@cta as="button" class="fake-link">Link</@cta>
</> 
```

I don't like how they have to add fake-link class if they set it to button personally. I think its better if they pass in a component.

src/components/ebay-education-notice/component.ts
ArtBlue: Sounds good. That takes care of the actual PR concern, but I'd still like to know where that Figma screenshot came from. Having an obsolete Figma file out there is a bit concerning. @ianmcburnie ?

src/components/components/ebay-notice-base/index.marko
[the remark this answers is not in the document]
ArtBlue: I see. In that case, can we add the fact that the component requires either an `<a>` or `<ebay-fake-link>` to look as it was intended. We're currently not enforcing those elements and nowhere that I can are we proving guidance for what it needs.

src/components/ebay-education-notice/examples/with-footer.marko
ArtBlue: Ok, that makes sense based on our approach. I was thinking the structural elements (i.e. `@footer`) would be put together automatically through internal conditionals based on the options passed in instead of the devs explicitly including them. Given our approach, what you have makes sense.

ArtBlue (approved):

agliga:
Added icon button for close
There was a bug with tests and added tests.

src/components/components/ebay-notice-base/index.marko
```diff
@@ -53,16 +54,34 @@ $ const {
...
             </else-if>
+            <if(isEducation && title)>
+                <${title.as || 'h2'}
+                    ...title
+                    class=[`${prefixClass}__title`, title.class]
+                >
+                    <${title}/>
+                </>
+                <if(a11yDismissText)>
+                        <ebay-icon-button
```
ArtBlue, line 72: Since for now Skin isn't using an icon button (because of the display issue with icon button), this should probably be changed to a regular button.

ArtBlue (changes requested):
Looks good, just reverting the icon button...

src/components/components/ebay-notice-base/index.marko
agliga: Yes. Since skin is applying transparency at the CSS level as long as the icon-btn class is used

ArtBlue: Sunds good.

Issue #2126, opened 2024-03-15
ebay-listbox-button: `borderless` doesn't render properly


darkwebdev:
# Bug Report

## eBayUI Version: latest
https://opensource.ebay.com/ebayui-core/?path=/story/buttons-ebay-listbox-button--default&args=borderless:!true

## Description

When using `borderless` option button still shows border.

## Screenshots
How it looks like now:
<img width="917" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/eb964108-bdee-44d9-b556-2c56f8064ab7">

How it should be:
<img width="922" alt="image" src="https://github.com/eBay/ebayui-core/assets/2737310/4ed953f6-025b-49a8-ae0b-2fbfada88bf2">

## How to fix it
Don't use `btn-form` together with `btn-borderless`, they are multually-exclusive.

agliga:
Yes, they are not supposed to be used together
Is this requesting a docs change or did you want to use both together?

darkwebdev:
@agliga `borderless` in one of the `ebay-listbox-button` properties, but it doesn't work.

PR #2136, opened 2024-03-18
fix(listbox-button): removed btn--form when listbox is borderless

agliga:
## Description
Removed `btn--from` from listbox button when its borderless

## References
https://github.com/eBay/ebayui-core/issues/2126

## Screenshots
Screenshot 2024-03-18 at 1 18 58 PM

### DIFF
--- a/.changeset/sour-months-search.md
+++ b/.changeset/sour-months-search.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
+---
+
+listbox-button: removed btn--form whenn listbox is borderless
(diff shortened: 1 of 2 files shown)


.changeset/sour-months-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
```
ArtBlue, line 2: Since this is going into `13.3.0` it should be a `minor` release, right?

ArtBlue (commented):
Looks good, but the changeset might need an adjustment.

.changeset/sour-months-search.md
agliga: This change is a patch type change since it's a bugfix not a feature. 
So what will happen is the largest changeset will be picked when it's together with a bunch of other changes. 

So if we have a bunch of patches together with a minor change, the change will be minor because minor is the "largest" type changeset.

Generally I've been doing patch on bugfix changes, minor on features, and major on breaking changes.

ArtBlue (approved):

Issue #2116, opened 2024-03-08
chips-combobox: placeholder missing, icon size issue, delete labels not specific


ArtBlue:
# Bug Report

## Description

* The chips-combobox doesn't have the placeholder option, which is built into Skin.
* The delete icon should be `icon--close-12` instead of the `icon--close-16` used currently. 
* The delete icon doesn't have contextual relationship with the chip text. The delete button labels are "Remove" instead of the label being contextual - "Remove [chip_text]".

## Workaround

Users could technically add the placeholder themselves to be injected into `input`, but we should encourage its use by having an actual `a11y` option.

Users cannot work around the incorrect `svg` since it's markup. A user could override the sizing of the incorrect `svg` in `css` and force the correct sizes, but that would not be ideal.

PR #2137, opened 2024-03-18
fix(chips-combobox): changed a11y text, icon, and added placeholder docs

agliga:
## Description
* Fixed chips icon to be 12px
* Added placeholder in docs. I don't think we need `a11y` placeholder, but it should be documented now
* Added a11y text to include the option text (We might need to revisit this in case there are languages where `remove item` is gramatically incorrect, but for now, it's good enough)

## References
https://github.com/eBay/ebayui-core/issues/2116


LuLaValva (commented):
Changes look good! Would you mind also making it so that form submission is unconditionally blocked on `chips-combobox` on enter keypress? That was the other feedback from the accessibility review. It should be as simple as moving `preventDefault` out of its `if` statement in the keydown handler.

agliga:
> Changes look good! Would you mind also making it so that form submission is unconditionally blocked on `chips-combobox` on enter keypress? That was the other feedback from the accessibility review. It should be as simple as moving `preventDefault` out of its `if` statement in the keydown handler.

Done

LuLaValva (approved):
Thanks!

src/components/ebay-chips-combobox/chips-combobox.stories.ts
```diff
@@ -58,6 +58,11 @@ export default {
             description:
                 "The role description for accessibility. Default text is set and will be in english. Pass this to override for different locales",
         },
+        placeholder: {
```
ArtBlue, line 61: The placeholder was actually one of the most important aspects of this component as it promotes better usability and `a11y` and it was one of the sticking points we worked through with DS. To promote a more accessible and usable component use of the component, can we include a default placeholder? Maybe use "Add item" to be a bit generic?

src/components/ebay-chips-combobox/chips-combobox.stories.ts
ArtBlue, line 1: It would be great to have another story with an example of prefilled chips, but that's probably outside the scope of this issue.

ArtBlue (changes requested):
Looks great! Just one small change please...

src/components/ebay-chips-combobox/chips-combobox.stories.ts
agliga: Added it to the default example.

src/components/ebay-chips-combobox/chips-combobox.stories.ts
agliga: Added this. This is a good catch since we weren't even documenting on how to add a prefilled chip.

PR #2140, opened 2024-03-19
feat(ebay-donut-chart): implement donut chart with highcharts

mikehobi:
## Description

- Adds new module: `ebay-donut-chart`
- Adds new module: `ebay-chart-legend`

## Context

Create a reusable donut chart based on DS Data Viz component design

Github ticket: 

## References

- [Playbook](https://playbook.corp.ebay.com/PlaybookApp#/product-guidelines/patterns/data-visualization)
- [Figma](https://www.figma.com/design/zEBdEhbonrBOGzZ0fXzWvM/eBay-Design-System?m=auto&node-id=84110-89886&t=uiWPTvDZ2fyONLk3-1)


src/components/ebay-donut-chart/index.marko
```diff
@@ -0,0 +1,42 @@
...
+                ${metricValue}
+            </div>
+            <div class="ebay-donut-chart__metric-label">
+                ${metricLabel}
+            </div>
+        </div>
+        <div
+            id=component.getContainerId()
+            class="ebay-donut-chart__graph-container"
+            no-update
```
agliga, line 30: Why is there `no-update` here?

src/components/ebay-donut-chart/components/ebay-chart-legend/style.less
```diff
@@ -0,0 +1,72 @@
+.ebay-donut-chart__legend {
```
agliga, line 1: Do we need to add these styles in skin?

src/components/ebay-donut-chart/components/ebay-chart-legend/index.marko
```diff
@@ -0,0 +1,45 @@
...
+            on-mouseover("handleFocus", index)
+            on-mouseout("handleBlur")
+            on-click("handleFocus", index)
+            on-focus("handleFocus", index)
+            on-blur("handleBlur")
+        >
+            <div class="ebay-donut-chart__legend-label">
+                <div class=[
+                    "ebay-donut-chart__legend-symbol",
+                    `ebay-donut-chart__legend-symbol--${item.symbolClass}`,
```
agliga, line 36: Is this always defined?
If not, we might need to add a guard like
```item.symbolClass && `ebay-donut-chart__legend-symbol--${item.symbolClass}\`,```

File: src/components/ebay-donut-chart/components/ebay-chart-legend/index.marko [hunk 4]
```diff
@@ -0,0 +1,45 @@
...
+        this.emit("blur");
+    }
+}
+
+<div class="ebay-donut-chart__legend">
+    <for|item, index| of=input.items>
+        <div
+            class=[
+                "ebay-donut-chart__legend-item",
+                input.focusIndex !== -1 &&
```
agliga, line 23: Can we change this logic to be
`!(input.focusIndex === index) && "ebay-donut-chart__legend-item--inactive"`
I assume that index is not -1. If it can be, then maybe its fine as is

src/components/ebay-donut-chart/component.ts
```diff
@@ -0,0 +1,294 @@
...
+            chart,
+            colors,
+            title: {
+                text: undefined,
+            },
+            plotOptions,
+            series,
+            tooltip,
+            credits: {
+                enabled: false, // Sorry highcharts
```
agliga, line 145: Ha

agliga (commented):
Quick first look
Looks good. Lets try to get high charts uploaded to the CDN. Maybe we can setup some time to get that done.

src/components/ebay-donut-chart/style.less
```diff
@@ -0,0 +1,89 @@
...
+
+// Tooltip
+.tooltip-pointer {
+    position: absolute;
+    width: 0;
+    height: 0;
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
+    border-bottom: 5px solid var(--color-black);
+}
```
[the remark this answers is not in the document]
mikehobi: Ah, this is leftover from when I was overriding highcharts default styling (which has no pointer)

src/components/ebay-donut-chart/index.marko
mikehobi: I noticed the state update (for active portion hover) was causing the highcharts to re-render and destroying the chart instance.

I can work on another solution, as `no-update` would cause issues if there's a need for data to update dynamically

src/components/ebay-donut-chart/components/ebay-chart-legend/style.less
mikehobi: I can move them!

src/components/ebay-donut-chart/components/ebay-chart-legend/index.marko
mikehobi: Good call

src/components/ebay-donut-chart/components/ebay-chart-legend/index.marko
mikehobi: Yea, using `-1` is what I would consider an "old school" way of managing the null state of the focus index haha

So in this instance, none of these should have the inactive if there's no `active` index

I can update this logic to be a little better `input.focusIndex !== null &&`

ianmcburnie:
@mikehobi Is the small version accounted for (with different layout)? I don't see it when running storybook.

Screenshot 2024-04-03 at 12 34 28 PM

src/components/ebay-donut-chart/components/ebay-chart-legend/index.marko
```diff
@@ -0,0 +1,45 @@
...
+    handleFocus(id: number) {
+        this.emit("focus", id);
+    }
+
+    handleBlur() {
+        this.emit("blur");
+    }
+}
+
+<div class="ebay-donut-chart__legend">
```
ianmcburnie, line 18: We should maybe leverage an HTML description list here?

mikehobi:
> @mikehobi Is the small version accounted for (with different layout)? I don't see it when running storybook.
> 
> Screenshot 2024-04-03 at 12 34 28 PM

@ianmcburnie I first assumed these were two "layout" types, but confirmed with Randy that these are breakpoint sizes, so narrowing the screen should show the small layout

src/components/ebay-donut-chart/components/ebay-chart-legend/index.marko
mikehobi: I did update this to use ul/li

It gets a little bit tricky using `dl` `dt` `dd` when styling, I could be wrong, but I'm not sure it allows for other elements to be nested within `dl`

src/components/ebay-donut-chart/donut-tooltip.marko
```diff
@@ -0,0 +1,26 @@
...
+                    <br>
+                    <if(input.tooltip)>
+                        ${input.tooltip}
+                    </if>
+                    <else>${input.value}</else>
+                </div>
+            </div>
+        </div>
+    </div>
+</span>
```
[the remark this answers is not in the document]
ianmcburnie: Maybe just go with taking the BEM child element which should give you all the styles you need:

```
<div class="tooltip__overlay">
        <span class="tooltip__pointer tooltip__pointer--bottom"/>
        <div class="tooltip__mask">
            <div class="tooltip__cell">
                <div class="tooltip__content">
                    <b>${input.name}</b>
                    <br>
                    <if(input.tooltip)>
                        ${input.tooltip}
                    </if>
                    <else>${input.value}</else>
                </div>
            </div>
        </div>
    </div>
```

agliga (commented):
Screenshot 2024-08-20 at 11 09 52 AM
Seems like theres no padding on these? Is that expected?

Also can we switch this to use the CDN version so that it doesn't block the site when loading?

mikehobi:
> Screenshot 2024-08-20 at 11 09 52 AM Seems like theres no padding on these? Is that expected?
> Also can we switch this to use the CDN version so that it doesn't block the site when loading?

- I believe there shouldn't be padding, and that would be handled by the user/grid setup
- Should be loading from CDN:
image

- I am only importing the `type` from "highcharts"

agliga (commented):
Overall looks good codewise. 
will test it locally and merge it if it looks good.

agliga (approved):
Im approving for now
LGTM
Just fix the tests and we will try to get this in

LuLaValva (approved):

agliga (approved):

Issue #2016, opened 2023-11-08
ebay-expressive-loader: new component


cordeliadillon:
ebayui portion of https://github.com/eBay/skin/issues/2162

PR #2141, opened 2024-03-19
feat(progress-bar-expressive): new module

cordeliadillon:
## Description

New component for expressive progress bar (aka expressive loader)

## Context

Recently added the expressive progress bar to Skin, so this is the eBay UI Core implementation of it.

Note that this relies on https://github.com/eBay/skin/tree/17.3.0, which is not yet released.

## Question for reviewers
Would like to get your thoughts on handling long messages. Messages ideally only span one line, but with translations, it's possible some may overflow to two. Currently, we don't account for that; see first screenshot below. In practice, I think this component will most often be centered vertically in a larger container, so it's unlikely the text would get cut off. But for a standalone component, it seems like we should handle this case OOTB. Thoughts on adding another CSS class to Skin, e.g. `progress-bar-expressive--tall` that adds padding when one of the messages is above a certain character count?

_Current_
> Screenshot 2024-03-19 at 16 15 03

_Adding padding to container_
> Screenshot 2024-03-19 at 16 10 33

## References

#2016
https://github.com/eBay/skin/issues/2162

### Default with Messages
https://github.com/eBay/ebayui-core/assets/5630924/f302af8b-103a-4e62-b34f-faeb6771bb78

### Medium 
https://github.com/eBay/ebayui-core/assets/5630924/662aa83b-3994-46ce-8011-471804f4ef2b

### Reduced Motion
In reduced motion mode, messages display slightly longer than their specified duration

https://github.com/eBay/ebayui-core/assets/5630924/77f0177e-0175-4111-9179-762b6040e755

### Medium Reduced Motion
https://github.com/eBay/ebayui-core/assets/5630924/74acda8c-c2bc-49cc-a262-61b3a2604919


src/components/ebay-progress-bar-expressive/test/test.browser.js
```diff
@@ -0,0 +1,63 @@
...
+        });
+
+        it("then cycles through all messages", async () => {
+            const statusEl = component.getByRole("status");
+            const messages = WithMessages.args.messages;
+
+            // go through all messages once and loop back to first one
+            for (let i = 0; i <= messages.length; i++) {
+                const message = messages[i % messages.length];
+                await waitFor(
```
cordeliadillon, line 27: This is a long-running test (several seconds). Could use [SinonJS fake timers](https://sinonjs.org/releases/latest/fake-timers/) to simulate and speed up

src/components/ebay-progress-bar-expressive/component.ts
```diff
@@ -0,0 +1,154 @@
+import { AttrString } from "marko/tags-html";
+import type { WithNormalizedProps } from "../../global";
+
+export const messageDurationStandard = 1500;
+export const messageDurationReducedMotion = 2500;
```
cordeliadillon, line 5: Unused const - will remove

src/components/ebay-progress-bar-expressive/progress-bar-expressive.stories.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export const Localized = Template.bind({});
+Localized.args = {
+    a11yText: "Cargando...",
+    messages: [
+        { text: "Espera..." },
+        { text: "Estamos procesando tu pedido", duration: 2000 },
+        { text: "Sólo un momento más" },
+    ],
+};
```
agliga, line 164: One thing we have been trying to do is have a `localized.marko` template in examples, and importing that directly. 
I know we have this sprinkled everywhere, but going forward we want to do this method since its much more explicit when showing the code

See
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu-button/menu-button.stories.ts#L237-L241 and https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu-button/examples/icon-with-text.marko

src/components/ebay-progress-bar-expressive/component.ts
```diff
@@ -0,0 +1,152 @@
+import { AttrString } from "marko/tags-html";
+import type { WithNormalizedProps } from "../../global";
+
+export const messageDurationStandard = 1500;
+export const messageDurationReducedMotionMultiplier = 1.5;
+export const messageFadeInDuration = 833;
```
agliga, line 6: Do we want to allow users to modify these? Im fine leaving them as static variables though

src/components/ebay-progress-bar-expressive/component.ts
```diff
@@ -0,0 +1,152 @@
...
+
+class ProgressBarExpressive extends Marko.Component<Input, State> {
+    declare timeouts: { fadeIn?: NodeJS.Timeout; showMessage?: NodeJS.Timeout };
+    declare fadeInFirstMessage: boolean;
+
+    onCreate(input: Input) {
+        const messages = (input.messages as Message[]) || [];
+        const useReducedMotion =
+            typeof window !== "undefined" &&
+            window.matchMedia("(prefers-reduced-motion: reduce)").matches;
```
agliga, line 40: I believe we might need to move this outside of the class and in the global definitions. (after imports)
I think this is due to the create might not be called on the client if its initialized on the server. 

If we move it in the root, we can access it anywhere in the file which is great!

src/components/ebay-progress-bar-expressive/component.ts
```diff
@@ -0,0 +1,152 @@
...
+        this.timeouts = { fadeIn: undefined, showMessage: undefined };
+    }
+
+    onInput(input: Input) {
+        this.state.messages = (input.messages as Message[]) || [];
+        if (this.state.nextMessageIndex >= this.state.messages.length) {
+            this.state.nextMessageIndex = 0;
+        }
+
+        this.initializeMessageRotation();
```
agliga, line 61: Lets move this to 
`onUpdate` and `onMount`

We should have cleanups on `onRender` in addition to `onDestroy`

See https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-listbox-button/component.ts#L79-L95
(It might not be needed, but I think its good practice in case the parent rerenders or such)

agliga (commented):
LGTM overall, just a couple of comments

src/components/ebay-progress-bar-expressive/component.ts
cordeliadillon: I don't think we should. We have a way for them to override `messageDurationStandard` on a per-message basis with the `duration` attribute, but I don't see a need for them to modify these constants.

src/components/ebay-progress-bar-expressive/component.ts
```diff
@@ -0,0 +1,152 @@
...
+    declare fadeInFirstMessage: boolean;
+
+    onCreate(input: Input) {
+        const messages = (input.messages as Message[]) || [];
+        const useReducedMotion =
+            typeof window !== "undefined" &&
+            window.matchMedia("(prefers-reduced-motion: reduce)").matches;
+        // For medium sized text, display the first message immediately
+        this.fadeInFirstMessage = input.size !== "medium";
+        this.state = {
```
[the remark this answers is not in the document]
cordeliadillon: This makes sense! One of the reasons I was using state here was to cast and store the messages as `Message[]` type, rather than the `RepeatableAttrTag<Message>` type of `input.messages`. There are a few spots where I need to check the # of messages. Would you recommend casting it each time I do a length check - e.g. `(input.messages as Message[])?.length` - or storing message count as a state variable?

src/components/ebay-progress-bar-expressive/progress-bar-expressive.stories.ts
cordeliadillon: Ah yes, that is more explicit! I'll update, though it looks like I will also need to explicitly pass args so the Storybook control panel has the correct initial values. Are there plans to update the `buildExtensionTemplate` utility to automatically pull initial args from the example component?

src/components/ebay-progress-bar-expressive/progress-bar-expressive.stories.ts
```diff
@@ -0,0 +1,162 @@
...
+
+export const MediumSize = buildExtensionTemplate(
+    MediumTextTemplate,
+    MediumTextTemplateCode,
+);
+MediumSize.args = {
+    size: "medium",
+    messages: exampleMessages,
+};
```
agliga, line 162: I don't think we need this example. We have toggles for toggling medium/large. As long as we spread the input to the other stories we don't need to have this.

src/components/ebay-progress-bar-expressive/examples/default.marko
```diff
@@ -0,0 +1 @@
+<ebay-progress-bar-expressive/>
```
agliga, line 1: For all these, lets make sure we spread input.

Just change it to 
```
<ebay-progress-bar-expressive ...input/>
```
That way any changes on storybook will be reflected.

src/components/ebay-progress-bar-expressive/index.marko
```diff
@@ -0,0 +1,64 @@
...
+    ...htmlInput
+} = input;
+
+<div
+    ...processHtmlAttributes(htmlInput)
+    class=["progress-bar-expressive", inputClass]
+>
+    <if(messages && messages.length > 0)>
+        $ const currentMessage = messages[state.currentMessageIndex]?.text;
+        $ const nextMessage = messages[state.nextMessageIndex]?.text;
```
agliga, line 17: Can we maybe use render body here instead of text?

So have the user do
```
<ebay-progress-bar-expressive>
   <@message>My message</@message>
   <@message>My message 2</@message>
</ebay-progress-bar-expressive>
```

And then in the code we can do
```
$ const currentMessage = messages[state.currentMessageIndex]?.renderBody
 messages[state.nextMessageIndex]?.renderBody
<!-- Later in the code -->
<${currentMessage}>
```

This is done so we can be flexible in the future in case design changes or there are other requirements.

agliga (commented):
I think it's very close. After this pass it should be good to go.

src/components/ebay-progress-bar-expressive/examples/default.marko
cordeliadillon: Oh yes! Good catch. Thanks!

src/components/ebay-progress-bar-expressive/index.marko
cordeliadillon: This is something I was wondering about! `renderBody` is definitely more flexible than a `text` string, but I also worry it could be misused -- people throwing links and icons in there, etc. But we can draw that line in our design system docs and not necessarily in the code.

agliga: Yes that is an issue, but it does provide much more flexibility, especially if design changes to want to include other things in the messages. Im not saying that will happen, but we have seen that in the past. 
That said, I agree that if we document what can go in there in our design system docs or even in our documentation here (like storybook description for `@messages`), it should be good enough.

src/components/ebay-progress-bar-expressive/component.ts
cordeliadillon: Can you explain this a little more? We only need to re-initialize the message rotation when the messages input is set/changed, so it seems like `onInput` would be the right place for this. `onUpdate` and `onRender` get called frequently for this component, since CSS classes and text content change based on state.

src/components/ebay-progress-bar-expressive/index.marko
cordeliadillon: Updated to use renderBody in latest commit

src/components/ebay-progress-bar-expressive/marko-tag.json
```diff
@@ -0,0 +1,20 @@
...
+        "targetProperty": null,
+        "type": "expression"
+    },
+    "@html-attributes": "expression",
+    "@messages <message>[]": {
+        "@*": {
+            "targetProperty": null,
+            "type": "expression"
+        },
+        "@text": "string",
```
agliga, line 13: Lets remove this. 
Unfortunately for marko 4 apps, we need to keep this file in sync with our storybook attributes otherwise it won't work properly.

agliga (commented):
LGTM overall. Just need some minor changes in marko.json then it should be good to go

One additional question: do we need any events? Would emitting an event whenever a message changes be good to have? Or would there not be a reason to.
I think its probably fine to go as is without any events, but just wondering your thoughts.

cordeliadillon:
Thanks for your feedback so far, @agliga! Any thoughts on my question in the PR description about vertical padding for long text?

Re:
> One additional question: do we need any events? Would emitting an event whenever a message changes be good to have? Or would there not be a reason to. I think its probably fine to go as is without any events, but just wondering your thoughts.

I can't think of a scenario in which it would be useful for a developer to know exactly which message had just shown, since all the messages convey, in essence "something is happening, please wait," just in different words.

@ianmcburnie, @ArtBlue, have you gotten a chance to look at this PR?

src/components/ebay-progress-bar-expressive/progress-bar-expressive.stories.ts
cordeliadillon: Size doesn't just control the text size but also whether the message immediately appears - medium text messages appear immediately, while large text messages animates in after a short delay. Was thinking it would be useful to have a story to highlight this behavior. To your point, though, this behavior can also be seen when toggling the size in the Storybook controls. Just wondering if that might get missed unless people go digging for it.

agliga (approved):

agliga:
Im going to merge this, if anyone wants to look at it we can take the feedback at a later time and iterate.

PR #2146, opened 2024-03-27
fix(notice-base): stop dismissing on all keydowns

### LINKED ISSUE #2145: ebay-notice-base: dismiss button is considered clicked when tabbing through it
# Bug Report

## Description

When tabbing (pressing tab key) through the dismiss button on the `ebay-notice-base`

**Expected:** the dismiss button should not be clicked and the banner should remain
**Actual:** the dismiss button is clicked and the banner is dismissed

LuLaValva:
- Fixes #2145 

## Description

- Prevent notice base from emitting a dismiss event on keydown, instead only emit on click

### DIFF
--- a/.changeset/green-penguins-agree.md
+++ b/.changeset/green-penguins-agree.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
+---
+
+Notice base dismiss keydown fix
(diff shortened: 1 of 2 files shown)


ArtBlue (approved):
Looks good to me. Thanks for the quick work!

agliga (approved):
Ha i saw this when i did the education notice and wanted to change it.

agliga:
Lets target feature branch please

Issue #2006, opened 2023-10-20
ebay-phone-input: new component


agliga:
ebayui portion of https://github.com/eBay/skin/issues/1601

PR #2150, opened 2024-03-29
feat(phone-input): new component

agliga:
## Description
* New Phone input

## Context
* For this, I updated `import-svg` to do two new things
1. import svg will create a `ebay-flag` module, which will import a flag based on two letter country code. This file generate a large amount of conditional statements
2. import svg will also generate `src/common/countries/countries.ts`. This is using `libphonenumber-js`. This is important since we do not have to maintain this list of countries and phone prefix codes as well as the mask for each country. In case of udpates, this library will update the json file. At the same time, this will be done durring `import-svg` and we will extract 
the data we need so that we do not have to load this library on the application side at all. 
* ~For the client rendering: we are using `imask` to mask the input for phone numbers. This library is 14kb which can be a bit big. At the same time, we are downloading all the flags to the client side to show them in the dropdown. We might need to look at optimizing this since this will be pretty large, but there isn't really a way around this.~ We have our own custom mask implementation. 
* The mask will look up the current country code and apply it's mask. If the user types in too many keys, we will only mask the keys we know, and the rest will be appended after the mask portion (With a space)

## References
https://github.com/eBay/ebayui-core/issues/2006

## Screenshots
Screenshot 2024-03-29 at 11 18 25 AM
Screenshot 2024-03-29 at 11 19 21 AM


src/components/ebay-phone-input/examples/floating-label.marko
```diff
@@ -0,0 +1,12 @@
+<ebay-phone-input
+    floating-label="Date"
```
saiponnada, line 2: May be change this to `phone number`

saiponnada (changes requested):
This is coming along real nice, few things to note with my initial skim,

- Typeahead would be good to support. Not sure if we can leverage external library or use makeup module to support it.
- The way SR announces is different to the skin. In skin, we do not announce the country code with the list Box button (custom implementation), but we announce it with the textbox using `aria-describedby`.
- It would be nice to show one story with a visible label.
- The `aria-label` should not be on the span, but rather on the list box button element. Refer skin for example.

agliga:
> This is coming along real nice, few things to note with my initial skim,
> 
> * Typeahead would be good to support. Not sure if we can leverage external library or use makeup module to support it.
I think we can wait for v2. Also not sure what we would typeahead since we don't know the phone number.

> * The way SR announces is different to the skin. In skin, we do not announce the country code with the list Box button (custom implementation), but we announce it with the textbox using `aria-describedby`.
Okay, it already being announced with `aria-describeby` on the textbox as well if you look. 
However, isn't it a good idea to have the country code also announced on the listbox, since that is where it is changing. If I hear `usa 1`, then I know I have the right country code and I dont have to focus the next element to see that it's correct. 
I can remove it but it requires a lot of extra code to do so. Is it important to be removed?

> * It would be nice to show one story with a visible label.
We have toggles to toggle that. We don't need a story for every single variant. Just the major ones and examples that require major structure changes. 

> * The `aria-label` should not be on the span, but rather on the list box button element. Refer skin for example.
The listbox implementation is on the span because it is reusing the icon and inserting it into the button. Again this is done so there is minimal change to the listbox. 
Does this cause an issue? I assume it should be fine. If it doesn't announce correctly, I can look to change. 
We do have these minor differences in skin and ebayui, as long as it doesn't cause an issue we can leave it as is. 

Again, some of these are so that there is minimal change to listbox. I am trying to use listbox-button and textbox as is with minimal changes

src/components/ebay-phone-input/examples/floating-label.marko
agliga: I will do this.

src/components/ebay-phone-input/component.ts
```diff
@@ -0,0 +1,132 @@
...
+class PhoneInput extends Marko.Component<Input, State> {
+    mask: any;
+
+    onCreate() {
+        this.state = {
+            index: 0,
+        };
+    }
+
+    get selectedCountry(): CountryInterface {
```
LuLaValva, line 60: My opinion so it doesn't have to change, but I don't love getters like this because they hide the fact that you are doing work wherever they are called.

src/components/ebay-phone-input/component.ts
```diff
@@ -0,0 +1,132 @@
...
+        // Trigger change event after state is updated so that the right mask value will be emitted
+        setTimeout(() => {
+            this.forwardEvent("change", { originalEvent: e });
+        }, 0);
+    }
+
+    forwardEvent(eventName: string, event: PhoneInputEvent) {
+        const selectedCountry = this.selectedCountry;
+
+        setTimeout(() => {
```
LuLaValva, line 91: Why are doing this `setTimeout`?

src/common/mask/index.ts
```diff
@@ -0,0 +1,93 @@
+export default function (input:HTMLInputElement, initialMask:string) {
```
LuLaValva, line 1: The parts of this logic that handles events should probably move to `phone-input/component.ts`, while this file just works with the strings and _maybe_ with the DOM for cursor selection. Detaching events like this is harder to follow and also more brittle since you're escaping from Marko's event handling.

saiponnada:
> The way SR announces is different to the skin. In skin, we do not announce the country code with the list Box button (custom implementation), but we announce it with the textbox using aria-describedby.
It would be nice to show one story with a visible label.
The aria-label should not be on the span, but rather on the list box button element. Refer skin for example.

@agliga and I discussed and decided to address the above-mentioned issues.

src/components/ebay-phone-input/component.ts
```diff
@@ -0,0 +1,132 @@
...
+                originalEvent: event.originalEvent,
+                value: this.mask.value,
+                rawValue: stripNonDigits(this.mask.value || ""),
+                callingCode: selectedCountry.callingCode,
+                countryCode: selectedCountry.countryCode,
+            } satisfies PhoneInputEvent);
+        });
+    }
+
+    _initializeMask() {
```
LuLaValva, line 102: If you move event handling logic into this file instead of `mask` you'll be able to remove this function, `onMount`, and `onDestroy`. Up to you, but I think it's a cleaner solution

src/components/ebay-phone-input/index.marko
```diff
@@ -0,0 +1,75 @@
...
+    countryCode,
+    value,
+    a11yButtonText = "Country",
+    disabled,
+    renderBody,
+    countryNames,
+    floatingLabel,
+    readonly,
+    invalid,
+    ...phoneInput
```
LuLaValva, line 14: Not critical, but most other files call this `htmlInput`

LuLaValva (commented):
First pass, overall looks pretty good. No deal-breaking comments, if it comes to it this is ok to merge as-is

> Typeahead would be good to support.

For country name/code, I think this is pretty critical. Having to scroll through all countries to find yours without being able to type "United States" _or_ "+1" is frustrating for users.

src/components/ebay-phone-input/component.ts
agliga:
```
// Trigger change event after state is updated so that the right mask value will be emitted
 ```
The stack needs to be cleared so that the state is updated and then I pull the right value from the mask. 
This was broken with Imask, but might need to recheck for Michaels version.

File: src/components/ebay-phone-input/component.ts (hunk 5 above)
Comment (agliga) [reply src/components/ebay-phone-input/component.ts:102]: Not sure what you mean. 
Still need to have those events. We just need to call them directly instead of an external file.

File: src/common/mask/index.ts (hunk 4 above)
Comment (agliga) [reply src/common/mask/index.ts:1]: Can you explain what you mean? Not sure I follow.

File: src/components/ebay-phone-input/component.ts (hunk 5 above)
Comment (LuLaValva) [reply src/components/ebay-phone-input/component.ts:102]: Sure, but if you handle the events using Marko instead of attaching them directly with JavaScript you can be sure that they will be cleaned up properly and it is more obvious where they are coming form

File: src/components/ebay-phone-input/component.ts (hunk 3 above)
Comment (agliga) [reply src/components/ebay-phone-input/component.ts:91]: Yes this is still needed. I removed one of the timeouts, but the other one is needed.

What happens is the input doesn't get the masking applied from the new dropdown selection if setTimeout is removed. 
Try entering a phone number, and then change the value in the listbox
If we remove setTimoue the old mask is applied.

File: src/common/mask/index.ts (hunk 4 above)
Comment (LuLaValva) [reply src/common/mask/index.ts:1]: It's not the end of the world if you keep it like this, and if you end up extracting the mask logic to an external module like makeup this is the _only_ way to do it, but if you are only using logic inside Marko it typically makes more sense to use Marko's event handling system instead of `createEventListener`

File: src/components/ebay-phone-input/component.ts (hunk 5 above)
Comment (agliga) [reply src/components/ebay-phone-input/component.ts:102]: Resolving this for now, will use the other comment

**Review by saiponnada — APPROVED** [review]

**Comment by saiponnada** [comment]
The requested accessibility changes look good. Please check with @ianmcburnie on reading the country code twice even though visually it appears only once.

File: src/components/ebay-listbox-button/index.marko [hunk 7]
```diff
@@ -13,62 +12,89 @@ $ const {
...
+        aria-label=a11ySelectedIconText
         value=selectedText
         type="button"
         disabled=disabled
         name=buttonName
         aria-haspopup="listbox"
         aria-describedby=describedby
-        aria-labelledby=(labelId && `${prefixId} ${labelId}`)
-        aria-invalid=(invalid && "true")>
+        aria-labelledby=labelId && `${prefixId} ${labelId}`
```
ianmcburnie, line 76: Can't have aria-labelledby _and_ aria-label. It's one or the other.

ianmcburnie: Actually, just notice this is on listbox-button, not phone-input itself. So long as aria-label always takes precedence over aria-labelledby in listbox-button we should be good.

src/components/ebay-phone-input/component.ts
```diff
@@ -0,0 +1,132 @@
...
+    value?: string;
+    rawValue?: string;
+    callingCode?: string;
+    countryCode?: string;
+}
+
+interface PhoneInputInput extends Omit<MenuButtonInput, `on${string}`> {
+    disabled?: boolean;
+    "country-names"?: CountriesLanguageInterface;
+    "a11y-button-text"?: AttrString;
```
ianmcburnie, line 28: What are a11y-button-text and a11y-menu-text used for? I don't see them explained in storybook docs. We might need more descriptive names for them also. Is a11y-button-text the prefix for country code listbox button? e.g. "Country: "?

agliga: Done, fixed the documentation fort his.

src/components/ebay-phone-input/component.ts
```diff
@@ -0,0 +1,132 @@
+import type { WithNormalizedProps } from "../../global";
+import type { Input as MenuButtonInput } from "../ebay-menu-button/component";
```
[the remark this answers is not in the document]
agliga: Done

agliga:
> First pass, overall looks pretty good. No deal-breaking comments, if it comes to it this is ok to merge as-is
> 
> > Typeahead would be good to support.
> 
> For country name/code, I think this is pretty critical. Having to scroll through all countries to find yours without being able to type "United States" _or_ "+1" is frustrating for users.

Added typeahead support to listbox.

src/common/countries/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface CountryInterface {
+    countryCode: string;
+    callingCode: string;
+    mask: string;
+}
+
+export interface CountriesLanguageInterface {
+    [index: string]: string;
+}
```
LuLaValva, line 12: _Technically_ you can enforce that all countries are included with TypeScript
```suggestion
export type CountriesLanguageInterface = {
    [index in keyof typeof countriesEnglishJSON]: string;
}
```

LuLaValva (approved):
Looks great! Typeahead makes this much easier to use.

src/common/mask/index.ts
ArtBlue: Yea, I agree with Luke here. If ebay ui is to be a Marko component library, it would be great to utilize Marko's event handling instead of native `js`.

src/components/ebay-listbox-button/index.marko
```diff
@@ -13,62 +12,90 @@ $ const {
...
+            split !== "none" && `btn--split-${split}`,
+            borderless ? "btn--borderless" : "btn--form",
             truncate && "btn--truncated",
-            floatingLabel && "btn--floating-label"
+            floatingLabel && "btn--floating-label",
         ]
+        aria-label=a11ySelectedIconText && `${a11yIconPrefixText}: ${a11ySelectedIconText}`
         value=selectedText
         type="button"
         disabled=disabled
```
ArtBlue, line 73: Disabled is not disabling the look/feel of the flags. 
image

ArtBlue (commented):
Looks great! Just a couple of things...

src/components/ebay-listbox-button/index.marko
agliga: I don't see it like this
Did you link skin? Im using `17.4.0`

Screenshot 2024-04-18 at 8 00 17 AM
Screenshot 2024-04-18 at 8 00 57 AM

ArtBlue (approved):

Issue #2153, opened 2024-04-10
snackbar-dialog: update docs to include CTA


ArtBlue:
# Feature Request

## Description

~~Skin currently has `snackbar-dialog` `v2.0`, but Ebay UI does not. For this, implement `v2.0` in Ebay UI.~~
Need to update docs to show snackbar action

## Context

Skin module: https://opensource.ebay.com/skin/#snackbar-dialog~~

## Screenshots

<img width="520" alt="image" src="https://github.com/eBay/ebayui-core/assets/1675667/a86d44f6-88dc-4285-b391-8aeb133ba823">

PR #2154, opened 2024-04-10
docs(snackbar): updated docs to support @action

agliga:
## Description
* Added with action in docs
* Updated tests to use storybook mocks
* Cleaned up storybook docs for snackbar

## Context
* No changeset is needed, docs only change

## References
https://github.com/eBay/ebayui-core/issues/2153


ArtBlue (approved):
Looks great!!

Issue #2035, opened 2023-12-06
Float label support for date-textbox


fer-de-s-ebay:
# Feature Request
Float label support for date-textbox

## Description
The new ads dashboard design needs float labels for the `ebay-date-textbox` component, attached screenshot to show the use of `start` and `end` text as float labels in the component

## Context
New ads dashboard Figma design: https://www.figma.com/file/Lw6GFDvYTKRKyoSjLzmWdQ/Ads-Dashboard-2.0---April2024?node-id=6038%3A356632&mode=dev

## Screenshots

![image](https://github.com/eBay/ebayui-core/assets/153042539/1d078ff2-c82b-48cd-906d-f5f3315b78b1)

randybascue:
The title can float like other text fields

PR #2157, opened 2024-04-15
feat(date-textbox): added floating label support

agliga:
## Description
* Added floating label support for date-textbox (followed similar pattern to placeholder text)
* Added tests
* __Note__ After adding server tests, `navigator.language` was breaking on the server. Added a guard against that. 
* There is an open discussion about keeping the floating label always floated like we do with phone input. Regardless, if that is the way we are going to go, we will need a follow up after phone input is merged to add support for that. 

## References
https://github.com/eBay/ebayui-core/issues/2035

## Screenshots
Screenshot 2024-04-15 at 9 23 12 AM
Screenshot 2024-04-15 at 9 23 37 AM
Screenshot 2024-04-15 at 9 23 40 AM


ArtBlue (changes requested):
Everything looks good, except there's some weirdness with the option in the docs. It says, "If set then shows this text as the floating label," but entering a string doesn't work:

image

The docs are clear about setting the range of dates floating labels, but using the directions provided is still unclear about how to set it for one date.

agliga:
> Everything looks good, except there's some weirdness with the option in the docs. It says, "If set then shows this text as the floating label," but entering a string doesn't work:
> 
> image
> The docs are clear about setting the range of dates floating labels, but using the directions provided is still unclear about how to set it for one date.

This is due to storybook not fully supporting text/array. If you put it in quotes, it works fine.
Screenshot 2024-04-16 at 8 45 43 AM


Im not sure what more we can say on this. If you have any suggestions, let me know or commit it.
This is using the same docs as `inputPlaceholderText`. It shows that it can be an array or text.

src/components/ebay-date-textbox/date-textbox.stories.ts
```diff
@@ -46,6 +46,13 @@ export default {
                 },
             },
         },
+        floatingLabel: {
+            type: "text|array",
+            control: { type: "object" },
+            description:
+                'If set then shows this text as the floating label. If separate floating labels are required for a range display, use an array of two strings (i. e. `["Start", "End"]`)..',
```
ArtBlue, line 53: Regarding the docs option, something like this would be much clearer:
```suggestion
                'If set then shows this text as the floating label (i. e. `"Single floating label"`). If separate floating labels are required for a range display, use an array of two strings (i. e. `["Range start floating label", "Range end floating label"]`)..',
```

ArtBlue (changes requested):

src/components/ebay-date-textbox/date-textbox.stories.ts
ArtBlue: This helps avoid confusion. BTW, if we have something similar in other places, we may want to change those too. It's not clear that the string needs to be in quotes.

agliga:
I added it, although I don't think its that big of a problem since this is only affecting people entering information in the storybook. In the code when they actually use it, all strings need to be in quotes anyways.

ArtBlue:
> I added it, although I don't think its that big of a problem since this is only affecting people entering information in the storybook. In the code when they actually use it, all strings need to be in quotes anyways.

Yea, I realize it's just a storybooks docs thing and works well otherwise, but if the docs appear to break when a dev uses those docs in the most intuitive way they know (just enter text) with no other directions otherwise (need to wrap the text in quotes) to help them be able to use the docs, it's far less than ideal. 

I mean, part of our job is to help improve the DX, and clarity is a big part of that. Is it a big deal? Maybe. Maybe not. But does the change improve the DX? I would say that's categorically a 'yes.' When you consider that the change doesn't impact code in any way and simply provides clarity, I would say it would be wrong NOT to do it.

ArtBlue (approved):

PR #2160, opened 2024-04-22
feat: add localized formatting

LuLaValva:
Addresses: #2174 

## Description

Use `date-fns` to localize date format in date textbox instead of using ISO.

### Note
`date-fns` maintains _their own_ list of locales that likely doesn't exactly match the browser. This is likely something we will need to maintain ourselves, or provide/find a mapping

## Screenshots

image
image


agliga (changes requested):
Please update from master. Lets try to get this in a patch but lets fix conflicts first.

LuLaValva:
I'll update from master, but I think this should be part of at least a minor version since placeholder text will need to be updated in applications that are using the component.

LuLaValva:
This is actually nontrivial for teams, since the placeholder needs to switch from being unconditionally `YYYY-MM-DD` to a localized format like `MM/DD/YYYY` for each locale. Alternatively, the placeholder can be removed or replaced with something like `Date`

agliga:
> This is actually nontrivial for teams, since the placeholder needs to switch from being unconditionally `YYYY-MM-DD` to a localized format like `MM/DD/YYYY` for each locale. Alternatively, the placeholder can be removed or replaced with something like `Date`

So for now: lets not make this automatic. If they want to get this new functionality, have them pass a flag like `auto-localize=true`. 
Next major version we will remove that flag and have it turn on by default.

src/components/ebay-date-textbox/index.marko
```diff
@@ -1,8 +1,11 @@
+import { format, parseISO } from "date-fns";
+import { getDateFnsLocale, type DayISO } from "../ebay-calendar/date-utils";
 $ const {
     a11yOpenPopoverText = "open calendar",
     range,
-    inputPlaceholderText = "YYYY-MM-DD",
+    inputPlaceholderText = "MM/DD/YYYY",
```
agliga, line 6: We should probably keep the original default, otherwise it will be breaking if they don't pass in `localizeFormat`

agliga (approved):
Approving for now if that placeholder issue is fixed.

agliga:
Need to update snapshots in tests. @LuLaValva

agliga (approved):

agliga:
LGTM overall codewise
I see a couple of bugs.

In `localized` we should use `localizedFormat` by default. We should also update the placeholder there. That said, could we have the placeholder be automatically generated from the formatter?
Also theres a bad bug if I type in a bad date using dashes for example, it removes all the content. We should not select a date in that case but never remove the value of the textfield.

PR #2166, opened 2024-05-03
fix(date-textbox): ensure correct date range order

LuLaValva:
## Description

- Ensure that date range is always in the correct order, even when entered manually

### DIFF
--- a/.changeset/nine-ligers-shop.md
+++ b/.changeset/nine-ligers-shop.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
+---
+
+Order preservation fix
--- a/src/components/ebay-date-textbox/component.ts
+++ b/src/components/ebay-date-textbox/component.ts
@@ -2,6 +2,7 @@ import Expander from "makeup-expander";
 import { type DayISO, dateArgToISO } from "../ebay-calendar/date-utils";
 import type { WithNormalizedProps } from "../../global";
 import type { AttrString } from "marko/tags-html";
+import { TextboxEvent } from "../ebay-textbox/component-browser";
 
 const MIN_WIDTH_FOR_DOUBLE_PANE = 600;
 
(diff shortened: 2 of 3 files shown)


src/components/ebay-date-textbox/date-textbox.stories.ts
```diff
@@ -66,7 +66,8 @@ export default {
         locale: {
             type: "text",
             control: { type: "text" },
-            description: "Locale of the date picker",
+            description:
+                "Locale used for formatting dates and determining first day of the week. Defaults to `navigator.language`.",
```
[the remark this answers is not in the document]
LuLaValva: Not necessarily, but it's a documentation change so I don't think it really warrants its own PR.

.changeset/nine-ligers-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
+---
+
+Order preservation fix
```
agliga, line 5: Can we name this to be `date-textbox: order preservation fix`

src/components/ebay-date-textbox/component.ts
```diff
@@ -88,13 +89,23 @@ class DateTextbox extends Marko.Component<Input, State> {
                 : 2;
     }
 
-    handleInputChange(index: number, { value }: { value: string }) {
+    handleInputChange(index: number, { value }: TextboxEvent) {
         const valueDate = new Date(value);
         const iso = isNaN(valueDate.getTime()) ? null : dateArgToISO(valueDate);
         if (index === 0) {
-            this.state.firstSelected = iso;
+            if (this.state.secondSelected && iso > this.state.secondSelected) {
```
agliga, line 96: I think there can be a cleaner way to do this. 
Lets first fix the issue with the date being swapped when typing manually. And then I will revisit this.

agliga (changes requested):
I think there's a bad bug with this. 
If there's a date selected in both inputs and I type in a date after the second one then they swap. I think that creates a bad UX. 
My feeling in that case is that the second date should probably clear rather than swap. 

This case that we are trying to fix should only be applied when I use the calendar popover thingy.

agliga:
Blocking issue for now until we get more feedback on how to approach this.

PR #2167, opened 2024-05-03
feat(date-textbox): add disabled state

### LINKED ISSUE #2117: ebay-date-textbox and ebay-calendar missing disabled attribute
# Bug Report
Like any other form elements, ebay-calendar and ebay-date-texbox should support disabled feature. There is no way to stop user from interacting with calendar.

LuLaValva:
- Fixes #2117

## Description

- Add disabled state to `ebay-date-textbox`

## Screenshots

image
image

### DIFF
--- a/src/components/ebay-date-textbox/component.ts
+++ b/src/components/ebay-date-textbox/component.ts
@@ -10,6 +10,7 @@ interface DateTextboxInput {
     rangeEnd?: Date | number | string;
     locale?: string;
     range?: boolean;
+    disabled?: boolean;
     "disable-before"?: Date | number | string;
     "disable-after"?: Date | number | string;
     "disable-weekdays"?: number[];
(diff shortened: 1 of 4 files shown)


src/components/ebay-date-textbox/component.ts
```diff
@@ -100,7 +101,7 @@ class DateTextbox extends Marko.Component<Input, State> {
     }
 
     openPopover() {
-        this.state.popover = true;
+        this.state.popover = !this.input.disabled && true;
```
ArtBlue, line 104: Is this needed? When the input and button are both disabled, isn't this function unreachable since none of the necessary events are fireable?

ArtBlue (commented):
Looks good. Just one Q.

src/components/ebay-date-textbox/component.ts
agliga: If this is needed I would change this to an if statemenet rather than assignment. Its much more readable that way.
```
If (!this.input.disabled) {
   this.state.popover = truel;
}
```

LuLaValva: Good catch, this isn't needed. I removed it.

agliga (approved):

PR #2169, opened 2024-05-06
fix(flag): moved flag from icon depdencies to flag

agliga:
## Description
* We are loading the flag CSS in icon, this was before the background icon changes. Because of this the CSS is extra large when using icons. So removed flags from icons styles


ArtBlue (approved):
Looks good. Is a similar change required in Skin?

agliga:
> Looks good. Is a similar change required in Skin?

Not needed. This is only for imports

Issue #2151, opened 2024-04-01
ebay-confirm-dialog freezes after hitting esc from keyboard


sairam0903:
## Bug Report
Keyboard navigation is not working as expected when clicking Esc on the dialog.

## Version
@ebay/ebayui-core: ^13.1.3

## Steps To Reproduce
1.  Open the confirm dialog by clicking or by keyboard navigation to the button
2. Now hit "**Esc**" key on keyboard and the dialog closes as expected
3. Now try to click using mouse or keyboard navigate to the button and the component freezes

Please see the attached video for more details. 
https://github.com/eBay/ebayui-core/assets/8981107/cc52499b-0d78-4cc9-afeb-5c9478bad758

Image for reference
<img width="774" alt="Screenshot 2024-04-01 at 12 29 19 PM" src="https://github.com/eBay/ebayui-core/assets/8981107/06e5fcfb-5a69-4b8c-b1a5-00519ce0eb0d">

PR #2171, opened 2024-05-08
fix(confirm-dialog): added handler for escape to trigger reject

agliga:
## Description
* Added events for escape to trigger from confirm dialog

## References
https://github.com/eBay/ebayui-core/issues/2151


ArtBlue (approved):
LGTM

PR #2187, opened 2024-06-05
fix(date-textbox): ensure no breaking changes

LuLaValva:
## Description

- `date-utils` was moved during a minor version change, as it was thought to be an internal API, but users were importing from `"@ebay/ebayui-core/dist/components/ebay-calendar/date-utils"` so it is actually breaking.
- For this patch, APIs are duplicated until the next major version


agliga (approved):
We need to undo this in 13.0.0
Please add it to 13.0.0 breaking changes ticket to track

PR #2191, opened 2024-06-14
feat(date-textbox): localize date placeholders

### LINKED ISSUE #2189: ebay-date-textbox: not formatting date values correctly in input box for some locale input with country code
# Bug Report

eBayUI Version: 13.5.0

## Description

The displayed formatted date value in the input box is not correct based on the passed in locale passed when localized is enabled

`fr-FR`

image

See storybook
https://opensource.ebay.com/ebayui-core/?path=/story/form-input-ebay-date-textbox--default&args=localizeFormat:!true;locale:fr-FR

It works with just the language without the country code

`fr` (expected)

image

## Workaround

Use `fr` instead of `fr-FR`

LuLaValva:
- Fixes #2189 

## Description

- Pull date format data from CLDR
- Enforce localized date format
- Add automatic placeholders based on locale
- Various UX improvements
  - Automatically add localized d/m/y separators while typing
  - When 2-digit year is entered, default to 1900s or 2000s
- Remove deprecations in `date-textbox` and `calendar`

## Screenshots

image
image
image
image


.changeset/hot-maps-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
```
agliga, line 2: Should be major

scripts/generate-date-info.js
```diff
@@ -0,0 +1,176 @@
...
+    mon: 1,
+    tue: 2,
+    wed: 3,
+    thu: 4,
+    fri: 5,
+    sat: 6,
+};
+
+const modernLocales = JSON.parse(
+    fs.readFileSync("node_modules/cldr-core/availableLocales.json"),
```
agliga, line 28: Can we just do a json top level import?

agliga (commented):
Codewise looks good
Will verify it locally

.changeset/hot-maps-refuse.md
LuLaValva: I think this is actually because `14.0.0` does not match with `master`, I didn't add a changeset

scripts/generate-date-info.js
LuLaValva: Good idea, that's more clean for sure. Added.

agliga (approved):

Issue #2194, opened 2024-06-18
toggle-button-group: add equal heights to buttons and rework responsiveness


ArtBlue:
# Feature Request

## Description

Inherit breaking structure and attribute modifications from the respective Skin issue.

## Context

See Skin changes: https://github.com/eBay/skin/pull/2345

PR #2199, opened 2024-06-24
feat(toggle-button-group): removed columns and changed ul layout

agliga:
## Description
* Removed `columns`. 
* Changed structure to match skin
* Added tests

## References
https://github.com/eBay/ebayui-core/issues/2194


LuLaValva (approved):

src/components/ebay-toggle-button-group/index.marko
```diff
@@ -2,25 +2,24 @@ import { processHtmlAttributes } from "../../common/html-attributes";
...
 $ const {
     class: inputClass,
-    columns,
     layoutType: baseLayoutType,
     buttons = [],
     ...htmlInput
 } = input;
 
-<ul
+<div
```
ArtBlue, line 10: I think this is missing the breakpoint-specific attributes.

Also, the `a11y` attributes need to stay on the `<ul>`.

ArtBlue (changes requested):
Looks good overall, but some stuff needs fixing.

src/components/ebay-toggle-button-group/component.ts
```diff
@@ -13,6 +13,11 @@ interface ToggleButtonGroupInput
     extends Omit<Marko.Input<"span">, `on${string}`> {
     buttons?: Marko.RepeatableAttrTag<Omit<ToggleButtonInput, `on${string}`>>;
     variant?: "checkbox" | "radio" | "radio-toggle";
+    "a11y-text"?: string;
+    "a11y-label-id"?: string;
+    columnsXS?: number;
+    columnsSM?: number;
+    columnsMD?: number;
```
ArtBlue, line 20: We need one for `XL` too since Skin has it as well.

src/components/ebay-toggle-button-group/toggle-button-group.stories.ts
```diff
@@ -66,7 +86,10 @@ export default {
 
 export const Default = Template.bind({});
 Default.args = {
```
ArtBlue, line 88: The default should not have required column specifications. Those breakpoint-specific exceptions for devs are supposed to be overrides. 

<kbd>image</kbd>

src/components/ebay-toggle-button-group/toggle-button-group.stories.ts
```diff
@@ -85,41 +108,22 @@ Default.parameters = {
...
-});
-WithIcons.args = {};
-WithIcons.parameters = {
-    docs: {
-        source: {
-            code: withIconsCode,
-        },
-    },
-};
+export const WithIcons = buildExtensionTemplate(
```
ArtBlue, line 111: Even with the columns being null, we're still setting them in the markup. When those params are not included, those attributes should not even be there. 

<kbd>image</kbd>

ArtBlue (commented):
Nice! A couple of minor things. In general, we don't want to "advertise" the override exceptions too much because out of the box the responsive rules should be sufficient for most cases.

src/components/ebay-toggle-button-group/component.ts
agliga: Will add

src/components/ebay-toggle-button-group/toggle-button-group.stories.ts
agliga: We can remove them from examples. However, if teams use those examples to see what thigns should be like then they won't see them. 
They are not being set by default.

src/components/ebay-toggle-button-group/toggle-button-group.stories.ts
agliga: These are simply for examples. We can remove them from the default example.

src/components/ebay-toggle-button-group/examples/columns.marko
```diff
@@ -0,0 +1,26 @@
+<ebay-toggle-button-group columnsSM=3 columnsXS=2 columnsMD=6 columnsXL=8 ...input>
```
ArtBlue, line 1: Do we need `columns` and `controlled`? Aren't they the same thing?

Maybe we can combine them into a single one, titled "Exception Cases" or "Preferred Columns?"

ArtBlue, line 1: The default state of this example that has columns defined (see markup) doesn't match the options underneath that are all `null`.  That might be confusing to devs.

<kbd>image</kbd>

ArtBlue (commented):
Much better. A couple of last things.

src/components/ebay-toggle-button-group/examples/columns.marko
LuLaValva: Controlled is an important code example for devs who want to hold state in the parent instead of letting our component do it, but it may not be necessary to have in storybook

agliga: Will leave it. There's no other way for devs to know how to use columns

src/components/ebay-toggle-button-group/examples/controlled.marko
```diff
@@ -4,7 +4,7 @@ $ const titles = ["Option 1", "Option 2", "Option 3"];
     <ebay-button onClick("clearSelection")>
```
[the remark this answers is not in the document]
agliga: This was there earlier
Will leave it in. This is good for devs to ser

src/components/ebay-toggle-button-group/examples/columns.marko
agliga: Yes
This is fine. This is because we are setting it explicitly on the code
See show code button
We need an example with it set by default.

ArtBlue (approved):

PR #2210, opened 2024-07-08
feat(intl): override fr-CA date format

LuLaValva:
- Resolves https://github.com/eBay/ebayui-core/issues/2212

## Description

- CLDR uses ISO format for `fr-CA`, but eBay uses `dd/mm/yyyy`. This change is for consistency with eBay instead of the international standard.

## Screenshots

image
image


agliga (approved):

agliga:
Lets target `14.1.0`, good to merge though.

PR #2228, opened 2024-08-07
feat(toggle-button-group): added min columns option

ArtBlue:
## Description

Add support for the recently added data-columns-min in Skin for the default/minimum container number of buttons in a row.

## Context

This will allow users of ebay ui to have preferred columns of buttons for the smallest container.

## Links

The companion to the Skin update: https://github.com/eBay/skin/pull/2406


agliga (approved):
LGTM, just add a changeset

PR #2238, opened 2024-08-27
fix(pagination): change default number of pages shown

### LINKED ISSUE #2235: ebay-pagination: shorten default number of items
# Feature Request
Reduce the default width of ebay-pagination component.

## Description
When page renders slowly, and ebay-pagination has a large number of pages, the component renders wider than the page. When the page completes rendering, ebay-pagination returns to it's normal width. 

## Screenshots
pagination

LuLaValva:
- Resolves #2235 

## Description

- Pagination currently defaults to show _every_ page during initial render and then decreases after a page width is found. This seems reasonable at first glance, but it is problematic when there is a large number of pages and the component is rendered on the server. 

## Context & Screenshots

- See attached issue

### DIFF
--- a/.changeset/silly-grapes-rush.md
+++ b/.changeset/silly-grapes-rush.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
+---
+
+Change default number of pages for pagination
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ integration/template.marko
 yarn.lock
 .env
 _cdn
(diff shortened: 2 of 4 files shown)


agliga (approved):

agliga (changes requested):
Getting some test failures

LuLaValva:
Bonus change: Test failures are reported in a `__screenshots__` directory, which should never be committed, so I added to `.gitignore`

agliga (approved):

PR #2253, opened 2024-09-09
feat(tests): moved to vitest

agliga:
Swapped all tests to use vitest.


agliga:
Ready for review, tests are passing in CI (and only takes 3 mins!)

LuLaValva (approved):
:100:

Issue #2260, opened 2024-09-13
textbox: Prefix-text not working for Marko 4


ManvithaC:
# Bug Report

## Description

@prefix-text and @postfix-text is not showing up for textbox, as per documentation, with Marko 4
looks like the input is not being normalized for some reason (see prefix-text="{}" in below screenshot)
eBayUI should add prefix-text and postfix-text to marko-tag.json

## Workaround

Use @prefixText and @postfixText.

## Screenshots

![image](https://github.com/user-attachments/assets/0a2e625f-27d0-4c75-8ded-0506a26357eb)

sairam0903:
+1

Seeing the same issue in [ebay-phone-input](https://opensource.ebay.com/ebayui-core/?path=/docs/form-input-ebay-phone-input--documentation). Country code is not showing when passed one.

PR #2266, opened 2024-09-18
fix(listbox/textbox): updated marko-jsons

agliga:
## Description
Updated marko-jsons with missing attributes 

## References
https://github.com/eBay/ebayui-core/issues/2260


LuLaValva (approved):
These changes are good so we should merge, but looks like some unrelated test had a failure? I don't see how your changes could have caused a Highcharts test to fail.

Issue #2271, opened 2024-09-23
ebay-line-chart: Tooltip positioning is off when rendering in a dialog/modal


lorenzolevy:
# Bug Report

## Description

When rendering the `ebay-line-chart` component in a modal, the tooltip is absolutely positioned relative to the whole page since it is rendered `outside` of the context of the chart [code](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-line-chart/component.ts#L324).

After setting this property to false in a local build, I was able to render the tooltips as expected. By exposing this highcharts configuration to end users of `ebay-line-chart` we can allow for easy rendering in modals.

## Workaround

N/A

## Screenshots

![chart](https://github.com/user-attachments/assets/13ae8cca-0bd3-4106-ac9e-9ecc3f648d5a)

PR #2273, opened 2024-09-23
fix(line-chart): adds outside input to allow for renders in modal

lorenzolevy:
## Description

Adds `outside` input onto `ebay-line-chart`.
## Context

Prior to this PR, the `outside` highcharts configuration hardcoded to `true` would cause the tooltip to render relative to the whole page rather than to the modal window or chart render. Allowing clients to set this value allows proper rendering of the tooltip + chart in a modal window.
## References

https://github.com/eBay/ebayui-core/issues/2271

## Screenshots

image

### DIFF
--- a/src/components/ebay-line-chart/component.ts
+++ b/src/components/ebay-line-chart/component.ts
@@ -40,6 +40,7 @@ interface LineChartInput
     "cdn-highcharts-accessibility"?: string;
     version?: string;
     series: SeriesLineOptions | SeriesLineOptions[];
+    outside?: boolean;
     trend?: "positive" | "negative" | "neutral";
 }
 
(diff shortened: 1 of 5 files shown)


src/components/ebay-line-chart/component.ts
```diff
@@ -321,7 +322,7 @@ class LineChart extends Marko.Component<Input> {
             backgroundColor: tooltipBackgroundColor, // sets tooltip background color
             borderWidth: 0, // hide the default border stroke
             borderRadius: 10, // set the border radius of the tooltip
-            outside: true, // used to render the tooltip outside of the main SVG element
+            outside: component.input.outside || true, // used to render the tooltip outside of the main SVG element
```
LuLaValva, line 325: Will this not _always_ be true? Maybe it is meant to be `component.input.outside !== false`?

lorenzolevy: Oops, thanks!

LuLaValva: Sure, that's the same as `component.input.outside !== false`

src/components/ebay-line-chart/component.ts
```diff
@@ -40,6 +40,7 @@ interface LineChartInput
     "cdn-highcharts-accessibility"?: string;
     version?: string;
     series: SeriesLineOptions | SeriesLineOptions[];
+    outside?: boolean;
```
mikehobi, line 43: Since this "outside" property pertains to the tooltip configuration, I'm wondering if this should be renamed to something like `tooltipOutside` or `renderTooltipOutside`... I'm not quite sure what would make the most sense

lorenzolevy: I was struggling back and forth on that naming while working on this. I had `renderTooltipOutside` initially, and I think it makes sense to go back to that flag naming.

LuLaValva (approved):
Looks good!

agliga (approved):

Issue #2234, opened 2024-08-23
table: new component


agliga:
https://github.com/eBay/skin/issues/1900

agliga:
https://github.com/eBay/ebayui-core/pull/2291

PR #2282, opened 2024-10-01
feat(ebay-table): ebay-table base component with density attribute

bill-min:
## Description

https://github.com/eBay/ebayui-core/issues/2234

## Screenshots

Screenshot 2024-10-04 at 11 23 08 AM

### DIFF
--- a/src/components/ebay-table/component.ts
+++ b/src/components/ebay-table/component.ts
@@ -1,0 +1,23 @@
+import { WithNormalizedProps } from "../../global";
+
+// export type ColumnType = "normal" | "numeric" | "row-header" | "none";
+export type TableDensity = "compact" | "relaxed" | "none";
+export interface TableHeader extends Omit<Marko.Input<"th">, `on${string}`> {
+    columnType?: string; // Use ColumnType after marko fix the ts error for attr tags
+    renderBody: Marko.Body;
+}
+export interface TableCell
+    extends Omit<Marko.Input<"td"> & Marko.Input<"th">, `on${string}`> {
+    renderBody: Marko.Body;
+}
+export interface TableRow extends Omit<Marko.Input<"tr">, `on${string}`> {
+    cell: Marko.RepeatableAttrTag<TableCell> | Marko.AttrTag<TableCell>[];
+}
+export interface TableInput extends Omit<Marko.Input<"div">, `on${string}`> {
+    header: Marko.RepeatableAttrTag<TableHeader> | Marko.AttrTag<TableHeader>[];
+    row?: Marko.RepeatableAttrTag<TableRow> | Marko.AttrTag<TableRow>[];
+    density?: TableDensity;
+}
+export interface Input extends WithNormalizedProps<TableInput> {}
+
+export default class extends Marko.Component<Input> {}
--- a/src/components/ebay-table/index.marko
+++ b/src/components/ebay-table/index.marko
@@ -1,0 +1,77 @@
+import { processHtmlAttributes } from "../../common/html-attributes";
+static var validDensity = ["compact", "relaxed"];
+$ const {
+    class: divClass,
+    density,
+    header: headers,
+    row: rows,
+    ...divInput
+} = input;
+
+<div
+    ...processHtmlAttributes(divInput)
+    class=[
+        "table",
+        density &&
+            validDensity.includes(density) &&
+            `table--density-${density}`,
+        divClass,
+    ]
+    role="group"
+    tabindex="0"
+>
+    <table>
+        <thead>
+            <tr>
+                <for|header| of=headers>
+                    $ const {
+                        columnType = "normal",
+                        class: thClass,
+                        renderBody,
+                        ...thInput
+                    } = header;
+                    <th
+                        ...processHtmlAttributes(thInput)
+                        class=[
+                            "table__cell",
+                            columnType === "numeric" && "table__cell--numeric",
+                            thClass,
+                        ]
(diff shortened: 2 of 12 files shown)


src/components/ebay-table/marko-tag.json
```diff
@@ -0,0 +1,48 @@
...
+    "attribute-groups": ["html-attributes"],
+    "@*": {
+        "type": "expression",
+        "targetProperty": null
+    },
+    "@html-attributes": "expression",
+    "@density": {
+        "enum": ["compact", "relaxed"]
+    },
+    "@column <column>[]": {
```
bill-min, line 11: @agliga Discussed with @LuLaValva and using `column` instead of `columns` makes it more compatible with Marko5/6 where we don't use `marko-tag.json`. Let me know your thoughts.

src/components/ebay-table/component.ts
LuLaValva, line 1: Will there eventually be state here? If not, we should either move this information into `index.marko` or rename to `component-browser.ts` so it can remain as a server component.

src/components/ebay-table/components/ebay-table-cell-layout.marko
LuLaValva, line 1: I'm not sure how I feel about this component as a whole since it is so simple. Maybe it would be better to have an attribute for `@cell` in `ebay-table` that creates a wrapper?

src/components/ebay-table/component.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface TableColumn {
+    type?: string; // Use ColumnType after marko fix the ts error for attr tags
+    header: Marko.AttrTag<TableHeader> | string;
+}
+export interface TableHeader extends Omit<Marko.Input<"th">, `on${string}`> {
+    renderBody: Marko.Body;
+}
+export interface TableCell
+    extends Omit<Marko.Input<"td"> & Marko.Input<"th">, `on${string}`> {
+    renderBody: Marko.Body;
```
LuLaValva, line 14: Mentioned in my other comment— would it make any sense to add `type: "layout" | "multiline" | "thumbnail"` here and remove the helpers in `components/`?

src/components/ebay-table/component.ts
bill-min: From previous conversation, looks like we want provide row selection feature and sort feature where it manage state internally then populate events.

src/components/ebay-table/component.ts
bill-min: I tried, but just for these 3, we have many kind of combinations, say `mutliline` can be with or without `layout`, and thumbnail can be first element or second element of layout or without layout. We are going to have more cell components and I think eventually we will not be able to deliver everything through attr tags. @agliga what do you think?

bill-min: Discussed with @agliga and will do these in separate PR.

src/components/ebay-table/index.marko
```diff
@@ -0,0 +1,70 @@
...
+                        class: thClass,
+                        renderBody,
+                        ...thInput
+                    } = header;
+                    <th
+                        ...processHtmlAttributes(thInput)
+                        class=[
+                            "table__cell",
+                            columnType === "numeric" && "table__cell--numeric",
+                            ,
```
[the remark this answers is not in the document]
bill-min: oops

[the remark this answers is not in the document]
bill-min: oops

agliga (approved):
Looks like a great first pass. I like the small prs so we can iterate

src/components/ebay-table/index.marko
```diff
@@ -0,0 +1,77 @@
...
+            `table--density-${density}`,
+        divClass,
+    ]
+    role="group"
+    tabindex="0"
+>
+    <table>
+        <thead>
+            <tr>
+                <for|header| of=headers>
```
LuLaValva, line 26: I know you enforce with TS that there is at least one header, but not everybody is using that and this will error if they try to render a table without any `@header`. You should either do `headers || []` here or add a default `header: headers = []` in the destructure for `input`.

LuLaValva (approved):
Looks awesome, just one small comment but it can be addressed in future PRs

src/components/ebay-table/index.marko
bill-min: @agliga do we have example in ebayui that if not providing required attribute it will error out? I think error out is the proper approach, just need a clean way to emit the error where developer can easily be notified.

src/components/ebay-table/table.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+            table: {
+                category: "@row attribute tags",
+            },
+        },
+    },
+};
+
+export const Default = buildExtensionTemplate(defaultTemplate, defaultCode);
+export const Dense = buildExtensionTemplate(defaultTemplate, defaultCode, {
+    density: "dense",
```
ArtBlue, line 58: This doesn't look right and is not defaulting to the correct density. Should be `compact`.

ArtBlue (changes requested):
LGTM. One minor thing with the examples.

src/components/ebay-table/table.stories.ts
bill-min: good catch

bill-min: @ArtBlue pushed the change

ArtBlue (approved):
Nice work!

PR #2286, opened 2024-10-07
feat(ebay-table): selection mode

bill-min:
## Screenshots

https://github.com/user-attachments/assets/3e50fd8d-a53c-47f6-89e1-d21c5ae43e92


src/components/ebay-table/component.ts
```diff
@@ -11,13 +19,119 @@ export interface TableCell
...
+            }, {} as TableRowStateMapping),
+        };
+        this.emit("select", {
+            selected: this.state.selected,
+        });
+    }
+
+    rowSelect(name: TableRowName, { checked }: CheckboxEvent) {
+        this.state.selected[name] = checked;
+        this.setStateDirty(RowState.SELECTED);
```
LuLaValva, line 128: This only works because `RowState.SELECTED` is `"selected"`. You should
1. Change the previous line to `this.state[RowState.SELECTED][name] = checked`
2. Change this line to `this.setStateDirty("selected")`
3. Duplicate the array without mutating so you don't have to `setStateDirty`

LuLaValva (commented):
Looks great! Just a few minor comments

src/components/ebay-table/component.ts
```diff
@@ -11,13 +19,119 @@ export interface TableCell
...
+            disabled[name] = !!row.disabled;
+        });
+        return { selected, disabled };
+    }
+
+    getDisabledFromInput(input: Input): TableRowStateMapping {
+        const disabled: TableRowStateMapping = {};
+        this.repeatableAttrTagToArr(input.row).forEach((row) => {
+            if (row.disabled) {
+                disabled[row.name || ""] = true;
```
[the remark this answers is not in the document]
bill-min: sry, forgot to remove this function... this whole function is not being used at all.

src/components/ebay-table/component.ts
bill-min: I was doing option 1, but somehow typescript compiler allows `this.state.disabled`, thought it is an acceptable pattern. I'll change it back if not preferred.

src/components/ebay-table/component.ts
```diff
@@ -11,13 +19,112 @@ export interface TableCell
...
     density?: TableDensity;
+    selected?: TableRowStateMapping;
+    "a11y-select-all-text"?: string;
+    "a11y-select-row-text"?: string;
+    "on-select"?: (event: TableSelectEvent) => void;
 }
 export interface Input extends WithNormalizedProps<TableInput> {}
 
-export default class extends Marko.Component<Input> {}
+enum RowState {
```
LuLaValva, line 38: Is this being used at all?

agliga (commented):
It looks good, there's a couple of comments I have

My main concern is about the disabled state. Is this in the design spec? What is the usecase for disabled? 
The reason why this is important is because there is logic here to not check disabled rows, which is probably okay. But we need to know the actual usecase to see if this is valid.

We should definitely not build something that isn't in the spec. If its not there we need design to update and add it before building it because there could be other design implications in case it changes.

src/components/ebay-table/component.ts
```diff
@@ -11,13 +19,104 @@ export interface TableCell
...
+    SELECTED = "selected",
+    DISABLED = "disabled",
```
ArtBlue, line 40: Let's remove this. It's not covered in Skin. We can always add it in later if/when the use case is covered in DS and build in Skin.

ArtBlue (commented):
Looks like Andrew and Luke already covered everything else; just one comment from me.

src/components/ebay-table/table.stories.ts
```diff
@@ -1,8 +1,11 @@
+/// <reference types="vite/client" />
```
bill-min, line 1: @agliga lmk if you don't like this line, it removes error prompt for importing raw marko file.

src/components/ebay-table/component.ts
```diff
@@ -11,13 +15,103 @@ export interface TableCell
...
+        this.state.selected = [...(this.input.row || [])].reduce(
+            (acc, { name }, i) => {
+                acc[name || i] = selectionState !== "all-selected";
+                return acc;
+            },
+            {} as Record<TableColRowName, boolean>,
+        );
+        this.emit("select", {
+            selected: this.state.selected,
+            selectionState:
```
agliga, line 105: I would emit the same kind of data that we do for tri-state-checkbox
So maybe have the key be allSelected and the values be true, false, or indeterminate

src/components/ebay-table/examples/selection.marko
```diff
@@ -0,0 +1,47 @@
...
+        Watchers
+    </@header>
+    <@header column-type="numeric">
+        Protection
+    </@header>
+    <@header>Shipping</@header>
+    <@header>Delivery</@header>
+    <for|r, i| of=data>
+        $ const name = i || "named_row";
+        <@row name=name selected=!i>
```
agliga, line 26: Not sure what this is trying to show. 
I would have a more concrete example.

src/components/ebay-table/index.marko
```diff
@@ -23,6 +28,15 @@ $ const {
     <table>
         <thead>
             <tr>
+                <if(mode === "selection")>
+                    <th class="table__cell">
+                        <ebay-tri-state-checkbox
+                            aria-label=(a11ySelectAllText ?? "Select all rows")
+                            checked=component.getSelectionAttrTriState(input)
```
agliga, line 35: This should probably be a state variable which is set on input or when the checked state of the other checkboxes is set.

src/components/ebay-table/table.stories.ts
agliga: Lets remove it. There's another pr that resolves that i thought

agliga (commented):
Overall lgtm. After these changes i started im fine approving.

src/components/ebay-table/component.ts
bill-min: Its `true`, `false` and `mixed`. Will update to make it consistent with tri-state checkbox.

src/components/ebay-table/component.ts
```diff
@@ -1,7 +1,12 @@
...
 import { WithNormalizedProps } from "../../global";
+import { CheckboxEvent } from "../ebay-checkbox/component-browser";
 
-// export type ColumnType = "normal" | "numeric" | "row-header" | "none";
-export type TableDensity = "compact" | "relaxed" | "none";
+type TableColRowName = string | number;
+export type TableSelectEvent = {
+    selected: Record<TableColRowName, boolean>;
+    allSelected?: AttrTriState;
+};
```
agliga, line 9: Do we need this? Might as well just define this inline. Thats what we do in other places

bill-min: Do you mean `AttrTriState` or this `TableSelectEvent`? [AttrTriState](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-tri-state-checkbox/component.ts#L23) is the same as tri-state checkbox.

For this TableSelectEvent, I can do it inline. But I find it useful when ppl actually using typescript and need to handle the event.

```MARKO
import { TableSelectEvent } from "<ebay-table>";

class {
  handler(event: TableSelectEvent) {
  // handler
  }
}

<ebay-table on-select("handler") />
```

bill-min: Actually, for event, tri-state checkbox is using same? https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-tri-state-checkbox/component.ts#L4

bill-min: Sure I can remove it. (It won't be inferred tho inside class component)

bill-min: updated please check

agliga (approved):
LGTM

PR #2289, opened 2024-10-08
feat(file-preview-cards): add preview card/cards

### LINKED ISSUE #2207: DS Component — File Upload
The File Upload component is designed to spec and is ready for code development:

Resources

[Design File](https://www.figma.com/design/Eq6RVl1YlbiMBpFS87idc7/File-Upload-Pattern?node-id=1-16&t=kQEZYhkWVPSmIRpp-1)
[Project Brief](https://docs.google.com/document/d/1skcmlauKjowhWnN1FotsJPKWFqRYUc49mz-4ZFobCkY/edit)
Lead Designer: Randy Bascue

LuLaValva:
- Resolves #2207 

## Description

- Create `ebay-file-preview-card`
- Create `ebay-file-preview-card-group`
- Add examples to `ebay-file-input`

## Screenshots

image


agliga (commented):
It looks good overall once we get the skin changes in
But one thing i see that im wondering about is the spacing between the file input and file list. It seems too close and is slightly overlapping 
That should be fixed for this composite component. 

I'm fine merging this once Arthur approves.

LuLaValva:
Looks like some older commits got added to this log after the rebase, but files changed still looks accurate. The only bonus change in this PR is adding `typeof` to a few out-of-scope references to `Marko.NativeTags` (I discovered that this type was incorrect, seemed too simple to make a whole extra PR for)

agliga:
I think it looks good overall
However one issue is that the spacing is too close I believe as per design. 
There should be a 24px distance between them. We should have this documented or added in the example.

agliga (approved):
Approving for now. 
The spacing part will be fixed in skin. 
Good work!

ianmcburnie:
Looks like footerSubtitle is overflowing outside of preview card? Skin issue?

Screenshot 2024-10-28 at 2 12 15 PM

ianmcburnie:
Minor, but should be fixed. The delete buttons for preview cards do not have an aria-label. Check the cancel buttons too.

Screenshot 2024-10-28 at 3 37 45 PM

src/components/ebay-skeleton/components/ebay-skeleton-avatar/index.marko
```diff
@@ -3,7 +3,7 @@ import type { AttrClass } from "marko/tags-html";
 import type { WithNormalizedProps } from '../../../../global';
 
 static interface SkeletonAvatarInput extends Omit<Marko.Input<"div">, `on${string}`> {
-    as?: Marko.NativeTags;
+    as?: keyof Marko.NativeTags;
```
ianmcburnie, line 6: What happened to cause these out of scope changes?

src/components/ebay-file-preview-card/file-preview-card.stories.ts
```diff
@@ -0,0 +1,260 @@
...
+        },
+    },
+};
+
+export const Video = Template.bind({});
+Video.args = {
+    a11yCancelUploadText: "Cancel upload",
+    deleteText: "Delete",
+    file: {
+        name: "file-name.mov",
```
ianmcburnie, line 164: Is this story using our ebay-video-player? It probably should, if not.

ianmcburnie:
I tried uploading a video and I didn't get a preview.

Screenshot 2024-10-28 at 3 50 10 PM

ianmcburnie:
It might be worth adding some additional description to these 3 sections:

**Default**: In this story you can trigger the native file input picker only. Uploading files will have no effect.
**With Preview Cards**: In this story you can trigger the native file input picker. Uploading files will render each ebay-file-preview-card component in preview status.
**With Mock Uploads**: In this story you can trigger the native file input picker. Uploading files will render each ebay-file-preview-card component in uploading status.

Screenshot 2024-10-28 at 3 56 16 PM

src/components/ebay-file-preview-card/file-preview-card.stories.ts
agliga: It doesn't need to. 
The ebay-video is a bit overkill for this usecase
I also think this PR might need to be updated to not show the preview. We had discussed with design that it was going to be a static image only.

agliga:
> I tried uploading a video and I didn't get a preview.
> 
> Screenshot 2024-10-28 at 3 50 10 PM

Is this an actual video? We do detect on the actual file and not extension. 
I tried to upload one and it came out just fine!
Screenshot 2024-10-29 at 8 07 52 AM

agliga:
> Looks like footerSubtitle is overflowing outside of preview card?
> 
> Screenshot 2024-10-28 at 2 12 15 PM
> On Skin it looks correct:
> 
> Screenshot 2024-10-28 at 3 40 34 PM

This is actually correct. In these examples theres no width constraint. If you shrink your screen it should resize accordingly.
We can fix it to a certain width but we wanted to show these as the base building block examples. 
The card group ones should be fine.

agliga:
@ianmcburnie should have addressed most of your comments, if you can take another look.

src/components/ebay-skeleton/components/ebay-skeleton-avatar/index.marko
LuLaValva: I don't think it was a merge conflict, I added them in a recent commit because I didn't think this TS change warranted a whole PR. Mentioned [here](https://github.com/eBay/ebayui-core/pull/2289#issuecomment-2438374282).

src/components/ebay-file-preview-card/file-preview-card.stories.ts
ianmcburnie: Why doesn't it need to? Will our end users not use the ebay-video component here? 

Moot point if it's going to be a static image.

agliga: It will be a static image. 
Not playable

ianmcburnie: I see it's the HTML video tag, but just using the `poster` url only with no controls. This is good.

ianmcburnie: Looks like Skin has the `src` set, not `poster`. So we need to clarify which attr is correct (but could be addressed after release), i.e. do we get an image url or movie url back from the backend on upload?

Screenshot 2024-10-30 at 2 51 56 PM

PR #2291, opened 2024-10-15
feat(ebay-table): support column sorting

bill-min:
## Screenshots

https://github.com/user-attachments/assets/f57d8a50-bf51-4e0c-9103-5d2a9deb244c

### DIFF
--- a/src/components/ebay-table/component.ts
+++ b/src/components/ebay-table/component.ts
@@ -31,7 +31,7 @@ export interface TableInput extends Omit<Marko.Input<"div">, `on${string}`> {
         selected: Record<string, boolean>;
         allSelected?: AttrTriState;
     }) => void;
-    "on-sort"?: (event: { sorted: Record<string, boolean> }) => void;
+    "on-sort"?: (event: { sorted: Record<string, TableSort> }) => void;
 }
 export interface Input extends WithNormalizedProps<TableInput> {}
 
--- a/src/components/ebay-table/component.ts
+++ b/src/components/ebay-table/component.ts
@@ -2,7 +2,7 @@ import { AttrString, AttrTriState } from "marko/tags-html";
 import { WithNormalizedProps } from "../../global";
 import { CheckboxEvent } from "../ebay-checkbox/component-browser";
 
-type TableSort = "asc" | "desc" | "none";
+export type TableSort = "asc" | "desc" | "none";
 export interface TableHeader extends Omit<Marko.Input<"th">, `on${string}`> {
     columnType?: "normal" | "numeric" | "row-header" | "layout" | "icon-action";
     name?: string;
--- a/src/components/ebay-table/component.ts
+++ b/src/components/ebay-table/component.ts
@@ -1,10 +1,14 @@
-import { AttrTriState } from "marko/tags-html";
+import { AttrString, AttrTriState } from "marko/tags-html";
 import { WithNormalizedProps } from "../../global";
 import { CheckboxEvent } from "../ebay-checkbox/component-browser";
 
 type TableColRowName = string | number;
+type TableSort = "asc" | "desc" | "none";
 export interface TableHeader extends Omit<Marko.Input<"th">, `on${string}`> {
     columnType?: "normal" | "numeric" | "row-header" | "layout" | "icon-action";
+    name?: TableColRowName;
+    sort?: TableSort | boolean;
+    href?: AttrString;
     renderBody: Marko.Body;
 }
 export interface TableCell
@@ -28,11 +32,13 @@ export interface TableInput extends Omit<Marko.Input<"div">, `on${string}`> {
         selected: Record<TableColRowName, boolean>;
         allSelected?: AttrTriState;
     }) => void;
+    "on-sort"?: (event: { sorted: Record<TableColRowName, boolean> }) => void;
 }
 export interface Input extends WithNormalizedProps<TableInput> {}
 
 interface State {
     selected: Record<TableColRowName, boolean>;
+    sorted: Record<TableColRowName, TableSort | undefined>;
     allSelected: AttrTriState;
 }
 
@@ -40,6 +46,7 @@ export default class EbayTable extends Marko.Component<Input, State> {
     onCreate() {
         this.state = {
             selected: {},
+            sorted: {},
             allSelected: "false",
         };
     }
@@ -47,6 +54,7 @@ export default class EbayTable extends Marko.Component<Input, State> {
     onInput(input: Input) {
         this.state.selected = this.getSelectedRowStateFromInput(input);
         this.state.allSelected = this.getAllSelectedState(input);
+        this.state.sorted = this.getSortedColStateFromInput(input);
     }
 
     getSelectedRowStateFromInput(input: Input) {
@@ -58,6 +66,19 @@ export default class EbayTable extends Marko.Component<Input, State> {
             }
         }
         return selected;
+    }
+
+    getSortedColStateFromInput(input: Input) {
+        const sorted: Record<TableColRowName, TableSort> = {};
+        for (const [i, header] of Object.entries([...input.header])) {
+            const name = header.name || i;
+            if (header.sort === true) {
+                sorted[name] = "none";
+            } else if (header.sort) {
+                sorted[name] = header.sort;
+            }
+        }
+        return sorted;
     }
 
     getAllSelectedState(input: Input): AttrTriState {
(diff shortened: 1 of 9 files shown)


src/components/ebay-table/index.marko
```diff
@@ -56,8 +60,38 @@ $ const {
...
+                        )
                     >
-                        <${renderBody}/>
+                        $ let sortEleAttr = {};
+                        $ if (href) {
+                            sortEleAttr = { href };
+                        } else if (sortOrder) {
+                            sortEleAttr = { type: "button" };
+                        }
+                        <${href ? "a" : sortOrder ? "button" : null}
```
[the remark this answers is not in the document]
ArtBlue: As we discussed offline, most of the a11y legwork is complete simply by the way we've structured the markup. Let's first test what we have using a screen reader and see if there are gaps before making more changes.

src/components/ebay-table/component.ts
```diff
@@ -1,10 +1,14 @@
-import { AttrTriState } from "marko/tags-html";
+import { AttrString, AttrTriState } from "marko/tags-html";
 import { WithNormalizedProps } from "../../global";
 import { CheckboxEvent } from "../ebay-checkbox/component-browser";
 
 type TableColRowName = string | number;
```
agliga, line 5: I know this isn't in this PR, but can we pick a better name?
Is this the name of the row or columns?

Maybe just TableRowName or TableRowIdentifier?

src/components/ebay-table/component.ts
```diff
@@ -105,4 +126,21 @@ export default class EbayTable extends Marko.Component<Input, State> {
             selected: this.state.selected,
         });
     }
+
+    sortColumn(name: TableColRowName) {
+        const sort = this.state.sorted[name];
+        if (sort) {
```
agliga, line 132: Might be cleaner to have a map in this case
```
const sortTo = {
   "asc": "desc",
   "desc": "none",
   "none": "asc"
}

this.state.sorted[name] = sortTo[sort || "none"];
```
Or we can do switch and case here.

src/components/ebay-table/component.ts
```diff
@@ -60,6 +68,19 @@ export default class EbayTable extends Marko.Component<Input, State> {
         return selected;
     }
 
+    getSortedColStateFromInput(input: Input) {
```
agliga, line 71: So this seems to support multi column sort, what if a team wants one column sort and everything else resets when one column is sorted?

src/components/ebay-table/examples/sort.marko
```diff
@@ -0,0 +1,56 @@
+import data from "./data.json";
+
+<ebay-table on-sort("emit", "sort") ...input>
+    <@header name="sellerCol" column-type="row-header" sort=("asc" as const)>
+        Seller
+    </@header>
+    <@header name="itemCol" sort>
```
agliga, line 7: Why do we have a simple sort for each column? Seems somewhat unnecessary.
I think for each column that is sorted I can pass in sort as needed from the `on-sort` method (we should demonstrate this). I assume if sort is not passed then its not going to be sorted.

agliga (commented):
Couple of comments, I think its close though!

src/components/ebay-table/component.ts
bill-min: Maybe force it to be `string` and just not to add this type at all? `string | number` just makes it easy for index-as-name, but we can cast it to string as well.

src/components/ebay-table/component.ts
ArtBlue: +1 on this. In fact, DS has maintained a single sort is the default with multi-sort possible, but somewhat unlikely to be implemented. I think we should build for single-sort for now.

bill-min: Good question, we should emit single clicked column in `on-sort` event

```
{
  sorted: { col1: "none", col2: "asc", col3: "asc" }
  sortedColumn: {col3: "asc"}
}
```

So user need to store `sortedColumn` as state then do

```
<@header sort=(sortedColumn["col" + index] ?? true)>
  Col ${index}
</@header>
```

bill-min: or add sorting mode `single | multiple` and emit `sorted` obj differently. I think emitting event like above gives user more choice.

src/components/ebay-table/examples/sort.marko
bill-min: If `sort` is `undefined | false`, no sorting function for the column. I think we discussed this before, that whether we want `sortable` + `sortOrder` or just simple `sort` and we decided to use simple `sort`. Main functionality here is to distinguish columns that can be sorted or not.

src/components/ebay-table/component.ts
bill-min: I think we can do single column `sorted: {col3:"asc"}` in sorting event and user can still implement multiple sort if sort state properly. Will refine PR.

src/components/ebay-table/examples/sort.marko
agliga: I think in this example we should change this to pass in the sort value from the `on-sort`. This example here isn't a good example and can be misleading.

bill-min: Updated example please check again thanks

agliga (approved):

ianmcburnie:
Running into this issue when launching storybook.

Screenshot 2024-10-28 at 4 24 47 PM

Related?

```The request url "/Users/snipsnip/GitHub/eBay/skin/tokens/evo-core.css" is outside of Vite serving allow list.```

**Comment by ianmcburnie** [comment]
Clicking sort doesn't appear to be re-rendering the table for me:

Screenshot 2024-10-28 at 4 33 55 PM
Screenshot 2024-10-28 at 4 33 49 PM
Screenshot 2024-10-28 at 4 33 44 PM

**Comment by agliga** [comment]
> Clicking sort doesn't appear to be re-rendering the table for me:
> 
> 
> 
> Screenshot 2024-10-28 at 4 33 55 PM
> 
> Screenshot 2024-10-28 at 4 33 49 PM
> 
> Screenshot 2024-10-28 at 4 33 44 PM
> 
> 

This is something we discussed with bill
We only emit an event and let teams to do the sort on their end. We don't want to include a janky sorting logic that 99% of the time wont be used. 
We only swap the sort icon. 
I believe this is mentioned in the on-sort event docs.

**Comment by bill-min** [comment]
@ianmcburnie Also we don't have reference to input data, so theoretically we cannot sort rows. We can add explanatory code in examples but that would make example file pretty big.

File: src/components/ebay-table/examples/sort-client-side.marko [hunk 6]
```diff
@@ -0,0 +1,93 @@
...
+    <@header name="protectionCol" column-type="numeric">
+        Protection
+    </@header>
+    <@header name="shippingCol">
+        Shipping
+    </@header>
+    <@header name="deliveryCol">
+        Delivery
+    </@header>
+    <for|r| of=data.sort((a, b) => {
```
LuLaValva, line 54: This will re-sort every rerender, I don't think it's okay to recommend this behavior even though technically in this example the only cause of a rerender is the sorting parameters changing. Maybe you can store this sorted data in `sorted` instead of the parameters for sorting?

bill-min: I'll use state

LuLaValva (approved):

ianmcburnie:
So this is more related to Skin (@ArtBlue), but I'm still not convinced we should be using aria-pressed on the sort buttons for ascending and descending state:

> Toggle buttons require a full press-and-release cycle to change their value. Pressing and releasing it once changes the value to true. If it's pressed and released again, the value changes back to false.

We do not meet the criteria for this press and release cycle because we have more than two values. We could possibly use a value of mixed, but I'm not sure there's any real benefit in that.

Source: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed

agliga:
One issue Im seeing on the sort is that when it is toggled to none, it should reset the list to the initial state. 
We might need to clone the list in that case and then set it to the initial clone when sorting is none.

bill-min:
> One issue Im seeing on the sort is that when it is toggled to none, it should reset the list to the initial state. We might need to clone the list in that case and then set it to the initial clone when sorting is none.

Fixed, good catch.

agliga (approved):
LGTM, tried it out locally and its working fine!

PR #2296, opened 2024-10-17
fix(date-textbox, calendar): clear behavior with ranges

### LINKED ISSUE #2295: ebay-date-textbox: the clear function isn’t working with range: true on the dateTo field, even when passing a null value or an empty string.
# Bug Report

## Description

 the clear function isn’t working with range: true on the dateTo field, even when passing a null value or an empty string.

## Screenshots

image

LuLaValva:
- Fixes #2295 

## Description

- Allow clearing date textbox with `range`
- Updated example to work with the changes. This makes the example more complicated, it may even be better to keep the simpler version even though it doesn't work when you toggle range to true.
- This uncovered a bug in `ebay-calendar` with clearing of dates, which I fixed

### DIFF
--- a/src/components/ebay-date-textbox/examples/with-clear.marko
+++ b/src/components/ebay-date-textbox/examples/with-clear.marko
@@ -2,19 +2,28 @@ class {
     onCreate() {
         this.state = {
             value: null,
+            rangeEnd: undefined
         };
     }
 
-    handleChange({ selected }) {
-        this.state.value = selected;
+    handleChange({ selected, rangeStart, rangeEnd }) {
+        if (selected) {
+            this.state.value = selected;
+        } else {
+            this.state.value = rangeStart;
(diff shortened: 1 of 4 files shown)


src/components/ebay-date-textbox/examples/with-clear.marko
LuLaValva, line 1: This makes the example a little complicated, what do you think about leaving as-is and adding a comment instead that says this naive implementation won't work with `range`?

agliga: Hmm, what makes them complicated. Not sure I follow

LuLaValva: It just makes the example a little longer since it has to handle both cases. There will realistically never be an app that needs to handle both range _and_ single select in the same spot.

agliga (approved):

Issue #2225, opened 2024-07-30
Evo DS Component — List


namtheho:
The List component is designed to spec and is ready for code development:

Resources

[Design File](https://www.figma.com/design/Zog0TjORKHyZBlgcZTbNAY/Lists?node-id=1-16&t=PKEZ6uaWmFsU2ony-0)

[Project Brief](https://docs.google.com/document/d/1QGCJ75UCezbDXDyl5dLiADjcEyvZSKRXiJwP5t3Q0Ek/edit#heading=h.859l5crh4sv0)

Lead Designer: Randy Bascue

PR #2297, opened 2024-10-18
feat(list): new component

agliga:
## Description
* List ebayui component

## Context
* Used `as` to determine the list item type
* For the click event, I only emit it on button

## References
https://github.com/eBay/ebayui-core/issues/2225

## Screenshots
Screenshot 2024-10-18 at 10 47 27 AM
Screenshot 2024-10-18 at 10 47 31 AM

### DIFF
--- a/src/components/ebay-list/component-browser.ts
+++ b/src/components/ebay-list/component-browser.ts
@@ -1,12 +1,12 @@
 import type { WithNormalizedProps } from "../../global";
-export interface Item extends Omit<Marko.Input<"item">, `on${string}`> {
-    leading?: Marko.RepeatableAttrTag<any>
-    trailing?: Marko.RepeatableAttrTag<any>
-    "on-click"?: (event: {index: number}) => void;
+export interface ListItem extends Omit<Marko.Input<"item">, `on${string}`> {
+    leading?: Marko.AttrTag<Marko.Input<"div">>;
+    trailing?: Marko.AttrTag<Marko.Input<"div">>;
 }
 
 interface ListInput extends Omit<Marko.Input<"div">, `on${string}`> {
-    items?: Marko.RepeatableAttrTag<Item>;
+    "on-item-click"?: (event: {index: number}) => void;
+    items?: Marko.RepeatableAttrTag<ListItem>;
     separator?: boolean;
 }
 
--- a/src/components/ebay-list/component-browser.ts
+++ b/src/components/ebay-list/component-browser.ts
@@ -7,7 +7,7 @@ export interface ListItem extends Omit<Marko.Input<"item">, `on${string}`> {
 interface ListInput extends Omit<Marko.Input<"div">, `on${string}`> {
     items?: Marko.RepeatableAttrTag<ListItem>;
     separator?: boolean;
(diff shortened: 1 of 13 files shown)


src/components/ebay-list/component-browser.ts
```diff
@@ -0,0 +1,17 @@
+import type { WithNormalizedProps } from "../../global";
+export interface Item extends Omit<Marko.Input<"item">, `on${string}`> {
+    leading?: Marko.RepeatableAttrTag<any>
+    trailing?: Marko.RepeatableAttrTag<any>
+    "on-click"?: (event: {index: number}) => void;
```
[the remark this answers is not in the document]
LuLaValva: Also, it should be `"on-item-click"` instead

src/components/ebay-list/index.marko
```diff
@@ -0,0 +1,74 @@
...
+                <hr>
+            </if>
+            <else>
+
+                <li>
+                    <${as || "div"}
+                        class=["list__body", itemInputClass]
+                        style=itemStyle
+                        on-click(
+                            as === "button" && "emit",
```
LuLaValva, line 31: Why only emit this when it's a button? Seems like an arbitrary restriction.

agliga: This is because when its static or an anchor tag, I don't want to emit an event. 
We shouldnt emit a click event on an anchor tag and on a static div.

agliga:
Resolved Lukes comments

src/components/ebay-list/index.marko
LuLaValva: Why not? `<div on-click("whatever")/>` works, I don't think it makes sense to prevent people from doing that.

src/components/ebay-list/component-browser.ts
```diff
@@ -0,0 +1,17 @@
+import type { WithNormalizedProps } from "../../global";
+export interface ListItem extends Omit<Marko.Input<"item">, `on${string}`> {
+    leading?: Marko.AttrTag<Marko.Input<"div">>;
+    trailing?: Marko.AttrTag<Marko.Input<"div">>;
+}
+
+interface ListInput extends Omit<Marko.Input<"div">, `on${string}`> {
+    "on-item-click"?: (event: {index: number}) => void;
```
LuLaValva, line 8: Sorry for the nitpick, but please put this at the end. We have event handlers at the end as a convention for all other components.

src/components/ebay-list/index.marko
agliga: It does. But in this case we want to emit an event for button click. This is because there is no other way for users to know if something was clicked. 
I don't want to render a static list and emit a click event on each item. The same with an anchor tag. This is because a static list should not be intractable. Same with the switch one. I don't want to emit an event on the switch being toggled because a click event was triggered on the static item.

LuLaValva: Okay, I see where you're coming from. Maybe this can be revisited/refined when there are other events also being emitted.

LuLaValva (approved):

PR #2304, opened 2024-10-24
fix(ts): Update to latest

### LINKED ISSUE #2303: typescript: update to latest
# Bug Report

## Description

- Marko TypeScript for `@attr` tags was improved and simplified, but it broke some workarounds in eBayUI
  - `Marko.RepeatableAttrTag` needs to be moved to `Marko.AttrTag` (they are now the same type)
  - We can not rely on repeated tags being arrays, and instead treat them as iterables.

LuLaValva:
- Fixes #2303 

## Description

- Modernize TS `@attr` Tags

## Context

- Attr tags have been simplified on the Marko side, but it means that a lot of the workarounds in eBayUI are now unnecessary or broken


agliga:
CI is failing due to some TS errors

LuLaValva:
Oh actually we're waiting on one more simple TS patch, I had it updated locally but didn't realize it wasn't included in `1.2.1`. Should be up soon.

LuLaValva:
Okay, should be good now @agliga

agliga (approved):
Looks like there are some conflicts, but good to go

PR #2317, opened 2024-11-03
refactor: (read desc) fully migrate stories to typescript and fix type issues in examples

bill-min:
## Description

The essential change of this PR is to remove stories and examples files from `exclude` property in tsconfig.json, plus adding `types: ["vite/client", "mdx"]` since `vite` provides the loader for `.marko?raw` and we also have `.mdx` imports.

The type fix for stories is quite straightforward and using type `any` as temp fix is acceptable in this case since the story files are not externally exposed. Ideally, using `buildExtensionTemplate` utility to build stories can avoid most of `any` which is a recommended way to build new stories.

The type fix for examples is a bit challenging:
- Avoid using `any` type
- Some examples are out of date
- Some of the component does not have proper up-to-date types
  - One of the reason why a proper typed example would be beneficial here and it can help find out bugs

Below are some fixes I made to components

### `<ebay-section-title>`

- [small](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-section-title/examples/small.marko) attr is no longer supported hence removal
- [overflow](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-section-title/index.marko#L11) should be `Marko.AttrTag<Omit<Marko.Input<"div">, `on${string}`>>`

### `<ebay-menu>`

Attr [text](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu/examples/sprites.marko#L14) is not supported hence removal.

### `<@separator>`

This is not added into types which is part of transformer.

Added for below files:
- menu-utils.ts -> BaseMenuInput
- <ebay-fake-menu-button> component.ts

Maybe we should have a better refactoring here but can be a separate PR.

### `<ebay-fake-menu-button`

`label` attr type should be `Marko.AttrTag<{ renderBody?: Marko.Body }>` instead of `{ renderBody?: Marko.Body }`

### `<ebay-date-textbox>`

[localizeFormat](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-date-textbox/examples/localized.marko#L10) is not supported hence removal.

### `<ebay-carousel>`

[on-carousel-update](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-carousel/examples/continuous-image-treatment.marko#L27) is not supported hence removal.

### `Marko.Renderable` type

This would cause error when using `renderBody` of the attr after recent language-server update and `Marko.Renderable` need to be changed to `Marko.Renderable  | Marko.AttrTag<{ renderBody: Marko.Body }>`. @LuLaValva there might be better solutions. The error is about missing iterators in `Marko.Renderable` type definition.

> I only made changes to a few components where it uses `renderBody` in examples, in case Luke can fix it from language server/marko source


agliga (commented):
LGTM overall
Can we target `14.5.0` branch
I will approve when Luke take a look as well

bill-min:
> LGTM overall Can we target `14.5.0` branch I will approve when Luke take a look as well

Done, do we want changeset for this?

src/components/ebay-carousel/examples/discrete-controlled.marko
```diff
@@ -12,12 +12,13 @@ export interface State {
...
 
 class {
+    declare state: State;
     onCreate() {
         this.state = {
             index: 0
         }
     }
-    onMove({visibleIndexes}) {
+    onMove({visibleIndexes}: {visibleIndexes: number[]}) {
```
LuLaValva, line 21: If we _really_ want proper types here, we can do
```marko
import type { Input as CarouselInput } from "<ebay-carousel>";
```
and 
```suggestion
    onMove({visibleIndexes}: Parameters<NonNullable<CarouselInput["on-move"]>>[0]) {
```
or (I think)
```suggestion
    onMove: CarouselInput["on-move"] = ({visibleIndexes}) => {
```

LuLaValva: This can be applied all over the place in this PR, no need to do it though if you don't think it's necessary. Your method is still type safe since it will error where the function is used if the types don't match.

src/components/ebay-confirm-dialog/examples/default.marko
```diff
@@ -38,18 +39,21 @@ $ const {
...
         this.state = { open: input.open };
     }
     openDialog() {
         this.state.open = true;
     }
-    closeDialog(ev) {
+    closeDialog() {
         this.state.open = false;
-        this.emit('reject', ev);
+        this.emit('reject', ...arguments);
```
LuLaValva, line 53: I'm not sure if we should recommend that people do this, since `arguments` isn't well-defined. Also, you don't have `on-reject` or `onReject` typed in `Input` so people wouldn't be able to use this component in a project.

src/components/ebay-lightbox-dialog/examples/default.marko
```diff
@@ -20,9 +24,9 @@ class {
     openDialog() {
         this.state.open = true;
     }
-    closeDialog(e) {
+    closeDialog() {
         this.state.open = false;
-        this.emit("close", e);
+        this.emit("close", ...arguments);
```
LuLaValva, line 29: Same comment here, I'm not sold on recommending `arguments`

LuLaValva (commented):
This is _amazing_, thank you @bill-min! Looks like you uncovered a good number of holes in our types 😁.

Just a few comments, but otherwise this is an awesome change.

src/components/ebay-carousel/examples/discrete-controlled.marko
bill-min: Will try it out, I remember last time I tried it it didn't work.

src/components/ebay-confirm-dialog/examples/default.marko
bill-min: I thought this emit is just for storybook actions to work (not for demo purpose), so I simplified it to avoid type error. I can refine it not big deal, only like 4-5 places.

LuLaValva: Yeah I think it's worth refining. A lot of people copy-paste from these examples so they should use best practices.

src/components/ebay-carousel/examples/discrete-controlled.marko
bill-min: It works!

LuLaValva: That would be great, Marko 6 won't have this problem but it would be nice to have the better inference in older code. No need to complete the refactor, again your method is still type-safe

agliga (approved):

PR #2319, opened 2024-11-06
feat(carousel): respect reduced-motion during autoplay

### LINKED ISSUE #2300: ebay-carousel: does not respect prefers reduced motion
# Bug Report

## Description
Carousel should not auto play, when a user enables prefer reduced motion settings system level.

## Workaround

Use css media query to respect user preferences.
```
@media (prefers-reduced-motion) {
  /* styles to apply if a user's device settings are set to reduced motion */
}
```

saiponnada:
## Description

Fixes #2300

## Context

If user has reduced motion setting turned on at OS level, the carousel doesnot autoplay by respecting the `prefers-reduced-motion` media query.

Reused the existing `useReducedMotion` code from expressive loader(moved to common/dom) for better component sharing.

### DIFF
--- a/src/common/dom/index.ts
+++ b/src/common/dom/index.ts
@@ -14,3 +14,17 @@ export function getMaxWidth(el: HTMLElement) {
     el.style.width = width;
     return result;
 }
+
+/**
+ * Determines if the user has requested reduced motion in their system preferences.
+ *
+ * This function checks the `prefers-reduced-motion` media query to see if the user
+ * has indicated that they prefer reduced motion. This can be useful for improving
+ * accessibility by disabling animations or transitions for users who may be sensitive
+ * to motion.
+ *
+ * @returns {boolean} `true` if the user prefers reduced motion, `false` otherwise.
+ */
+export const useReducedMotion =
+    typeof window !== "undefined" &&
+    window.matchMedia("(prefers-reduced-motion: reduce)").matches;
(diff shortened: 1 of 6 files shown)


src/common/dom/index.ts
```diff
@@ -14,3 +14,17 @@ export function getMaxWidth(el: HTMLElement) {
...
+ * This function checks the `prefers-reduced-motion` media query to see if the user
+ * has indicated that they prefer reduced motion. This can be useful for improving
+ * accessibility by disabling animations or transitions for users who may be sensitive
+ * to motion.
+ *
+ * @returns {boolean} `true` if the user prefers reduced motion, `false` otherwise.
+ */
+export const useReducedMotion =
+    typeof window !== "undefined" &&
+    window.matchMedia("(prefers-reduced-motion: reduce)").matches;
```
ArtBlue, line 30: Would this match always work? Is there any possibility for there to be no whitespace in any one of our supported browsers? Other than this, it looks good.

ArtBlue (commented):
Looks good. Just one questions...

src/common/dom/index.ts
agliga: This is the same code used in progress bar expressive. 
I would only add that we should probably make that generic in a common library but that can be done in a followup.

agliga (approved):
LGTM

PR #2320, opened 2024-11-06
fix(combobox): added viewAllOptions control

### LINKED ISSUE #2284: Combobox: autocomplete list is not filtering with mouse input
# Bug Report

## Description

When textbox value changes, the autocomplete list should always update. Currently this is only happening when textbox value is updated via keyboard, and not when updated via mouse (e.g. after clicking on a listbox option). This creates confusing inconsistency in the UX.

Story here: https://opensource.ebay.com/ebayui-core/?path=/story/form-input-ebay-combobox--isolated

W3C reference example here: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/

## Workaround

None.

saiponnada:
## Description

Fixes #2284 

- Exposed `viewAllOptions` configuration to opt to desired behaviour. 
- By default, if not set it lists all the options and does not filter options with selected text. 

## Screenshots
**view all options set to `true`**

image

**view all options set to `false`**
image

### DIFF
--- a/src/components/ebay-combobox/component.ts
+++ b/src/components/ebay-combobox/component.ts
@@ -27,7 +27,7 @@ interface ComboboxInput extends Omit<Marko.Input<"input">, `on${string}`> {
     autocomplete?: "list" | "none";
     "list-selection"?: "manual" | "automatic";
     "floating-label"?: boolean;
-    viewAllOptions?: boolean;
+    "view-all-options"?: boolean;
     button?: Marko.Input<"button"> & Marko.AttrTag<{
         htmlAttributes?: Record<string, unknown>;
         renderBody?: Marko.Body;
--- a/src/components/ebay-combobox/component.ts
+++ b/src/components/ebay-combobox/component.ts
@@ -27,6 +27,7 @@ interface ComboboxInput extends Omit<Marko.Input<"input">, `on${string}`> {
     autocomplete?: "list" | "none";
     "list-selection"?: "manual" | "automatic";
     "floating-label"?: boolean;
+    "viewAllOptions"?: boolean;
     button?: Marko.Input<"button"> & Marko.AttrTag<{
         htmlAttributes?: Record<string, unknown>;
         renderBody?: Marko.Body;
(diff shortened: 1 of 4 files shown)


src/components/ebay-combobox/component.ts
```diff
@@ -117,14 +118,7 @@ export default class Combobox extends Marko.Component<Input, State> {
...
 
     handleExpand() {
-        if (this.state.viewAllOptions) {
-            this.setSelectedView();
-        } else {
-            this.state.viewAllOptions = true;
-            this.once("update", () => {
-                this.setSelectedView();
-            });
-        }
```
saiponnada, line 127: I refactored this logic. Not sure If I am missing something, let me know.

src/components/ebay-combobox/component.ts
```diff
@@ -233,7 +226,7 @@ export default class Combobox extends Marko.Component<Input, State> {
         this.lastValue = input.value;
         this.state = {
             currentValue: this.lastValue,
-            viewAllOptions: (this.state && this.state.viewAllOptions) || true,
+            viewAllOptions:  input.viewAllOptions ?? true,
```
saiponnada, line 229: Should the default value be `false` instead of true ?

LuLaValva: Actually, it looks like `state.viewAllOptions` is never changed? If it's not a mutable value we should just remove it from state and replace all instances of `state.viewAllOptions` with `(input.viewAllOptions ?? true)`

src/components/ebay-combobox/component.ts
agliga: This seems like this won't work in the default usecase.

src/components/ebay-combobox/component.ts
agliga: Im almost of the opinion that the logic should stay the same, only the input should determine if the viewAllOptions is set to true or false. 
I don't think we should be setting this to true/false onInput. 
I think when we set `viewAllOptions` in the code, we should check the input version of it and then set it based on that.

src/components/ebay-combobox/component.ts
saiponnada: The way I think of it now is, if `input.viewAllOptions` set to true It always shows all options whereas filtered when set to false. Curious why we want to do both?

LuLaValva (commented):
Looks good! Just one more tiny change

agliga (approved):

Issue #2305, opened 2024-10-25
cdn-loader: swap to use async loading instead of CDN


lorenzolevy:
# Bug Report

## Description

Currently, if `ebay-line-chart` is loaded asynchronously, the component will mount and display an empty screen while it retrieves `highcharts.js` and `accessibility.js`. After those requests have been retrieved and run (self invoking), the chart will end up rendering as expected. The result is an awkward UX where the chart shifts layout after the necessary scripts are loaded in.

In order to load and initialize the scripts before the chart has loaded in, we should consider loading these resources before the component mounts. This could possibly be achieved through `browser.json` async loading.

## Workaround

N/A

## Screenshots
<img width="420" alt="Screenshot 2024-10-25 at 3 04 59 PM" src="https://github.com/user-attachments/assets/931fcdcb-1c99-4176-9b42-7296ed312e16">

PR #2321, opened 2024-11-11
feat: added async loading support

agliga:
## Description
* In order to make it so we dont upload compiled files to the CDN I changed the loading portion of our modules to load modules asynchronously. 
* Changed model-viewer, highcharts, and shaka-player to load asynchronously. 
* The changes to package.json, and prettier are needed because we introduced a nested `node_modules` directory

## Context
* Part of the issue is we need to support lasso and webpack async. These load modules asynchronously very differently.
* We added a "fake" `node_module` directory where we load a module using `import()` if its vite or webpack, but if its lasso we load it using `lassoLoader`
* The reason why there are multiple modules is mainly for bundling purposes. These will all be indexed properly. 
* These also need to be tested in a prerelase to verify that this is working fine is all various types of apps. 

__note__ I had to disable a highcharts test because it was not finding the text correctly. 

## References
https://github.com/eBay/ebayui-core/issues/2305


mikehobi:
@agliga quick glance at it, looks like pattern-fill is not working properly

image

agliga:
> @agliga quick glance at it, looks like pattern-fill is not working properly
> 
> image

What would cause this problem?

mikehobi:
> > @agliga quick glance at it, looks like pattern-fill is not working properly
> > image
> 
> What would cause this problem?

I think accessibility and pattern fill need to be initialized?

Something like this might work:

```
        highChartsLoad()
            .then(([highcharts, accessibility, patternFill]: any) => {
                window.Highcharts = highcharts.default;

                accessibility.default(window.Highcharts);

                patternFill.default(window.Highcharts);

                this.handleSuccess();
            })
            .catch((e: Error) => {
                this.handleError(e);
            });
```

agliga:
Should be good to go. Fixed mikes issue
Also the license thing is probably due to highcharts being a depdency now. I will ignore this since this is documented properly.

Issue #2326, opened 2024-11-18
ebay-ccd: new component


agliga:
https://github.com/eBay/skin/issues/2493

PR #2327, opened 2024-11-18
feat(ccd): added new component

agliga:
## Description
* Added CCD component

## Context
* We need to toggle between one and two figures in this. Added some boolean types to allow toggling

## References
https://github.com/eBay/ebayui-core/issues/2326

### DIFF
--- a/src/components/ebay-ccd/index.marko
+++ b/src/components/ebay-ccd/index.marko
@@ -1,10 +1,9 @@
 import { processHtmlAttributes } from "../../common/html-attributes";
 import type { WithNormalizedProps } from "../../global";
-
 static interface CCDInput extends Omit<Marko.Input<"div">, `on${string}`> {
     max?: string;
     min?: string;
-    "charger-icon"?: string;
+    "charger-icon"?: "included" | "not-included";
     units?: string;
     "secondary-type"?: string;
     "secondary-text"?: string;
@@ -17,7 +16,7 @@ $ const {
     class: inputClass,
     style,
     a11yText,
-    chargerIcon = "none",
+    chargerIcon,
     units = "W",
     a11yUnits = "watts",
     max,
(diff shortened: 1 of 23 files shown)


agliga:
Test is failing because we havent linked up skin. Should pass once we have the correct skin version.

ArtBlue (approved):
LGTM!

src/components/ebay-ccd/index.marko
```diff
@@ -0,0 +1,61 @@
+import { processHtmlAttributes } from "../../common/html-attributes";
+import type { WithNormalizedProps } from "../../global";
+
+static interface CCDInput extends Omit<Marko.Input<"div">, `on${string}`> {
+    max?: string;
+    min?: string;
+    "charger-icon"?: string;
```
LuLaValva, line 7: This should be an enum, like `"none" | "included" | "notIncluded"`

LuLaValva: Also, I think we generally use kebab-case for attributes like this. Is `"notIncluded"` standard terminology or was it chosen arbitrarily?

src/components/ebay-ccd/index.marko
```diff
@@ -0,0 +1,61 @@
...
+    secondaryText = "USB PD",
+    ...htmlInput
+} = input;
+$ (input as any).toJSON = noop;
+
+$ const hasFastCharging = secondaryType === "usbpd";
+
+$ const noIcon = chargerIcon !== "included" && chargerIcon !== "notIncluded";
+$ const hasCharger = chargerIcon === "included";
+$ const chargerA11yText = noIcon ? "" : `Charger${hasCharger ? " " : "not "} included.`;
```
LuLaValva, line 35: This is missing a space
```suggestion
$ const chargerA11yText = noIcon ? "" : `Charger${hasCharger ? " " : " not "} included.`;
```

LuLaValva (commented):
Just a few minor comments

PR #2330, opened 2024-11-19
feat(filter-menu): add header with search

### LINKED ISSUE #2329: filter-menu: add search header
# Enhancement Request

## Description

Add search header option to the `filter-menu` inline with Skin (https://github.com/eBay/skin/issues/2451) and allow filtering of the filter options.

LuLaValva:
- Resolves #2329 

## Description

- Add search header

## Context

- @ianmcburnie and I decided offline that filtering should happen _outside_ of eBayUI, because:
  1. There are likely multiple types of custom filtering (i. e. case sensitive, beginning of term vs within term)
  2. Simple cases that _would_ likely be a good candidate for eBayUI to do the filtering should likely not even have a search filter to begin with
  3. Network requests are impossible to handle in eBayUI

## Screenshots

image

image

### DIFF
--- a/src/components/ebay-filter-menu/component.ts
+++ b/src/components/ebay-filter-menu/component.ts
@@ -34,16 +34,23 @@ interface FilterMenuInput
             "a11y-footer-text"?: AttrString;
         }
     >;
+    "search-header-placeholder-text"?: string;
+    "a11y-search-header-clear-text"?: string;
     "render-body"?: Marko.Body;
     "on-footer-click"?: (event: FilterMenuEvent) => void;
     "on-form-submit"?: (event: FilterMenuEvent) => void;
     "on-change"?: (event: FilterMenuEvent) => void;
     "on-keydown"?: (event: FilterMenuEvent) => void;
+    "on-search-change"?: (value: string) => void;
 }
 
 export interface Input extends WithNormalizedProps<FilterMenuInput> {}
 
-export default class extends MenuUtils<Input, MenuState> {
+export interface State extends MenuState {
+    searchTerm?: string;
+}
+
+export default class extends MenuUtils<Input, State> {
     declare _rovingTabIndex: ReturnType<typeof createLinear>;
     declare lastTabIndexPosition: number;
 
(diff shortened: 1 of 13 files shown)


src/components/ebay-filter-menu/component.ts
```diff
@@ -56,6 +63,16 @@ export default class extends MenuUtils<Input, MenuState> {
         this.lastTabIndexPosition = 0;
     }
 
+    handleSearch(e: InputEvent) {
+        this.state.searchTerm = (e.target as HTMLInputElement).value;
+        this.emit("search-change", this.state.searchTerm);
```
agliga, line 68: Do we want to add an option to filter the results? Like we have with combobox

agliga: We can also iterate on this if needed

src/components/ebay-filter-menu/examples/with-search.marko
ArtBlue, line 1: A couple of things here...

(1) I think there's a bug here with more than one filtering.

* user enters search term
* user selects an item
* user clears search

The initial selection is reset/removed. It should be retained.

If a user is trying to find various items and doing multiple filtering, the prior selections should be maintained until manually removed to allow for more than one filtering and selection.

(2) A slight UX gap.

For longer lists, once a selection is made and context switched to other parts of the list via scrolling or filtering, the user does not have an easy way to reference the list of their current selections.

ArtBlue (changes requested):
Looks good overall. One minor bug and a UX gap that we may not be able to do anything at the moment. I just wanted to document it.

src/components/ebay-filter-menu/component.ts
LuLaValva: I think it's overengineering to add our own filtering option, see the issue description for the results of a discussion with @ianmcburnie

src/components/ebay-filter-menu/examples/with-search.marko
LuLaValva: Good catch @ArtBlue on (1), I just committed a fix.

src/components/ebay-filter-menu/examples/with-search.marko
```diff
@@ -1,31 +1,47 @@
...
     onCreate() {
         this.state = {
-            searchTerm: "",
+            checkedCodes: {},
+            visibleCountries: countries,
         };
     }
 
     handleSearchChange(value: string) {
-        this.state.searchTerm = value.toLowerCase();
```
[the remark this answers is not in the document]
agliga: Could there be a usecase where the developer of this component wants to pass in and sync up the state of the search
I'm thinking of page navigations or such. 
Generally I think its a good idea to emit the value of the search and then update a parents state and pass that in as input.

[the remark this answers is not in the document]
LuLaValva: I can't think of a usecase, but I guess there's no harm in adding it. I just pushed a commit that includes this functionality.

src/components/ebay-filter-menu/examples/with-search.marko
ArtBlue: @LuLaValva , looks awesome!

ArtBlue (approved):
Looks great! Let's just create an issue for that feature we can iterate on in the future.

Issue #2313, opened 2024-11-01
ebay-table: various fixes


agliga:
* column-type: row-header does not sit well in here. Should move out, make it a boolean

WinkeeFace:
Blocked: eBayUI is using an outdated version of @marko/type-check, and the latest version needed to fix the type narrowing issue is incompatible. Awaiting resolution from the Marko team to upgrade and ensure compatibility.

PR #2355, opened 2024-12-24
fix(table): replaced column-type of "row-header" with boolean row-header

WinkeeFace:
## Description

Replaced `column-type` of `row-header` with a boolean `row-header` in the `ebay-table` component. Updated the relevant Marko and JSON files to reflect this change.

## Context

This change was made to simplify the handling of row headers and improve ARIA accessibility. The boolean `row-header` is more intuitive and aligns better with accessibility standards.

## References

- Issue: [#2313](https://github.com/eBay/ebayui-core/issues/2313)

## Screenshots

N/A


src/components/ebay-table/examples/default.marko
```diff
@@ -1,7 +1,7 @@
 import data from "./data.json";
 
 <ebay-table ...input>
-    <@header column-type="row-header">
+    <@header row-header=true>
```
LuLaValva, line 4:
```suggestion
    <@header row-header>
```

LuLaValva (commented):
Well done, thanks @WinkeeFace! Just a few notes:

1. Those tests are failing because you didn't include `rowHeader` in the `index.marko` destructure [here](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-table/index.marko#L4-L14).
2. Since Marko 5 / eBayUI normalizes kebab-case to camelCase so both are accepted, that needs to be accommodated for in TypeScript. To do this, you can either add duplicate `"row-header"?: boolean;` and `"column-type"?: ...` types in the interface (probably not recommended), or you can switch everything to kebab-case and use `WithNormalizedProps<>` like we do for `Input` (preferred).
3. `=true` is implied in Marko attributes (same as HTML), so there are a few examples that can be simplified (I showed one example in a code comment)

WinkeeFace:
> Well done, thanks @WinkeeFace! Just a few notes:
> 
> 1. Those tests are failing because you didn't include `rowHeader` in the `index.marko` destructure [here](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-table/index.marko#L4-L14).
> 2. Since Marko 5 / eBayUI normalizes kebab-case to camelCase so both are accepted, that needs to be accommodated for in TypeScript. To do this, you can either add duplicate `"row-header"?: boolean;` and `"column-type"?: ...` types in the interface (probably not recommended), or you can switch everything to kebab-case and use `WithNormalizedProps<>` like we do for `Input` (preferred).
> 3. `=true` is implied in Marko attributes (same as HTML), so there are a few examples that can be simplified (I showed one example in a code comment)

Thank you for the feedback! I'll make those corrections.

LuLaValva:
@WinkeeFace I appreciate the effort of going through and adding `as const`, but as I mentioned offline this will be fixed in the next version of `@marko/type-check`. Please undo these commits or remove the `as const` changes, we are hoping to have eBayUI updated today

agliga (changes requested):
Please remove the `.DS_Store` files
You can add them to gitignore (which they should be there already)

WinkeeFace:
### **Additional Update: Remove `aria-pressed` from `ebay-table`**

This PR has been updated to address an additional issue:

#### **Description**
The `aria-pressed` attribute was incorrectly applied in the `ebay-table` component. This update removes its usage to ensure compliance with ARIA standards and accessibility best practices.

#### **Changes Made**
- Removed the `aria-pressed` attribute from `ebay-table`.

#### **Checklist**
- [ ] `aria-pressed` removed from `ebay-table`.
- [ ] Verified no regressions in the `ebay-table` component.
- [ ] Updated tests to reflect the removal of `aria-pressed`.

PR #2361, opened 2025-01-03
fix(ebay-textbox): update input interface to include textarea-specifi…

### LINKED ISSUE #2356: ebay-textbox: typescript type check breaks for `step` and other `input` attributes
# Bug Report

## Description

Since `ebay-textbox` only extends `"textarea"`, attributes such as `step` will fail type check

## Workaround

Adding spread props w/ `any`

```
<ebay-textbox
    type="number"
    ...({ step: "0.01" } as any)
/>
```

## Screenshots

image

WinkeeFace:
## Description

Updated the `component-browser.ts` file to reference shared attributes from `tags-html.d.ts` and added the `Textbox` class implementation.

## Context

These changes were made to reduce redundancy and ensure a single source of truth for the attributes used in the `TextboxInput` interface. By referencing the shared attributes, we can maintain consistency and avoid potential errors. Additionally, the `Textbox` class was implemented to handle the floating label functionality and event forwarding.

## References

Fixes #2356 

## Screenshots

N/A

### DIFF
--- a/src/components/ebay-textbox/component-browser.ts
+++ b/src/components/ebay-textbox/component-browser.ts
@@ -1,13 +1,13 @@
 import FloatingLabel from "makeup-floating-label";
 import type { WithNormalizedProps } from "../../global";
-import type { AttrString } from "marko/tags-html";
+import type { AttrOnOff, AttrString, AttrStringOrNumber } from "marko/tags-html";
 
 export interface TextboxEvent {
     originalEvent: Event;
     value: string;
 }
 
-interface TextboxInput extends Omit<Marko.Input<"textarea">, `on${string}`> {
+interface TextboxInput extends Omit<Marko.Input<"input">, `on${string}`> {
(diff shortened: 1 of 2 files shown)


src/components/ebay-textbox/component-browser.ts
```diff
@@ -95,4 +100,4 @@ class Textbox extends Marko.Component<Input> {
     }
 }
 
-export default Textbox;
+export default Textbox;
```
LuLaValva, line 103: Looks like a newline was removed, can you add it back?

LuLaValva (commented):
Thanks @WinkeeFace, looks good! Just one quick fix.

Also, can you `npm run change`? Looks like you're missing a changeset, this should be a `patch`.

LuLaValva (approved):

agliga (approved):

PR #2364, opened 2025-01-07
fix(3d-viewer): install three peer dep

LuLaValva:
Looks like `three` is marked as a `peerDependency` of `@google/model-viewer`, which means that it will be included by NPM but not by yarn (as far as I understand). This adds it as a direct dependency so that yarn also works out of the box, without installing `three`.


agliga (approved):
Just verify this is working in yarn applications

PR #2370, opened 2025-01-09
fix(ebay-combobox): floating label type

MaiDeveloper:
## Description
The `floating-label` property in the combobox component is currently set as a boolean value, but it should be defined as a string instead.

### DIFF
--- a/.changeset/sharp-bees-grin.md
+++ b/.changeset/sharp-bees-grin.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ebayui-core": patch
+---
+
+TS fix for combobox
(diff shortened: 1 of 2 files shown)


LuLaValva (commented):
Awesome, thanks for the quick PR! Before merging, can you `npm run change` to add a changeset? Description can be super simple, something like "TS fix for combobox" is fine. This will kick off our release cycle. Added bonus is the CI will give you credit in the patch notes 😉

agliga:
We also need an issue to track this first.

LuLaValva (approved):

agliga:
So there are a couple of extra files. Can you maybe undo the commit and merge from `15.0.0`, not the `changesets/15.0.0` branch?

MaiDeveloper:
Done

agliga (approved):

PR #2376, opened 2025-01-10
chore: upgrade marko & mtc, fix type errors

LuLaValva:
## Description

- Upgrade `marko`, `@marko/type-check`, and `@marko/compiler` to latest
- The `type-check` upgrade exposed a few types that were wrong in eBayUI, this fixes each of them.

> [!NOTE]
> The `ebay-breadcrumbs` example that I changed includes a few attributes which only make sense in some ebay apps, so they are not valid in types. Teams who need to include extra attributes should override `Marko.HTML.Nav` in their project, but that shouldn't happen in eBayUI.


src/components/ebay-breadcrumbs/examples/page-custom-attrs.marko
```diff
@@ -1,6 +1,6 @@
 <ebay-breadcrumbs a11yHeadingText="Page navigation">
-    <@item href="http://www.ebay.com/" navsrc="{}" _sp="p2345.m909.l34">eBay</@item>
```
agliga, line 2: I think this is fine but we need to figure out how to pass in "anything" into some components

agliga (approved):

Issue #2349, opened 2024-12-19
dropdowns: swap to using floating ui


agliga:
Skin issue: https://github.com/eBay/skin/issues/2514

PR #2381, opened 2025-01-15
feat(dropdowns): added support for floating-ui

agliga:
## Description
* Added floating ui support for all dropdowns
* Also added it for calendar
* For combobox, need to retrigger the show attribute on makeup change in case the expander is expanded on rerender
* For chips combobox there's an issue where we need to attach the dropdown to the root element. For this, passed in a function which resolves to the root elemenet into combobox. This attribute is not supposed to be documented outside of our code.
* Reverse will be handled by the popper automatically

## References
https://github.com/eBay/ebayui-core/issues/2349

### DIFF
--- a/src/common/dropdown/index.ts
+++ b/src/common/dropdown/index.ts
@@ -1,0 +1,54 @@
+import {
+    autoUpdate,
+    flip,
+    computePosition,
+    shift,
+    offset,
+    type ReferenceElement
+} from "@floating-ui/dom";
+
+interface DropdownUtilOptions {
+    reverse?: boolean;
+    offset?: number
+}
+
+export class DropdownUtil {
+    declare host: ReferenceElement;
+    declare overlay: HTMLElement;
+    declare cleanupFn: any;
+    declare options: DropdownUtilOptions;
+
+    constructor(host: HTMLElement, overlay: HTMLElement, options?: DropdownUtilOptions) {
+        this.host = host as ReferenceElement;
+        this.overlay = overlay as HTMLElement;
+        this.options = options ?? {};
+    }
+
+    show() {
+        this.cleanupFn = autoUpdate(
+            this.host,
+            this.overlay,
+            this.update.bind(this),
+        );
+    }
+
+    update() {
+        computePosition(this.host, this.overlay, {
+            placement: this.options.reverse ? "bottom-end" : "bottom-start",
+            middleware: [offset(this.options.offset ?? 4), flip(), shift()],
+        }).then(({ x, y }) => {
+            Object.assign(this.overlay.style, {
+                left: `${x}px`,
+                top: `${y}px`,
+            });
+        });
+    }
+
+    cleanup() {
+        this.cleanupFn?.();
+    }
+
+    hide() {
+        if (this.cleanup) this.cleanup();
+    }
+}
--- a/src/components/ebay-chips-combobox/index.marko
+++ b/src/components/ebay-chips-combobox/index.marko
@@ -13,6 +13,7 @@ $ const {
 $ const options = [...option || []].map((o) => o.text)
 
 <span
+    key="root"
     class=[
         "chips-combobox",
         fluid && "chips-combobox--fluid",
@@ -43,6 +44,7 @@ $ const options = [...option || []].map((o) => o.text)
         onExpand("emit", "expand")
         onCollapse("emit", "collapse")
         disabled=disabled
+        dropdown-element=() => component.getEl("root")
         chevron-size="large"
         ...comboboxInput
         autocomplete="list"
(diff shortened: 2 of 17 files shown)


src/components/ebay-chips-combobox/index.marko
```diff
@@ -43,6 +44,7 @@ $ const options = [...option || []].map((o) => o.text)
         onExpand("emit", "expand")
         onCollapse("emit", "collapse")
         disabled=disabled
+        dropdown-element=() => component.getEl("root")
```
LuLaValva, line 47: I think it should be okay in this case, but passing a function like this is a sketchy pattern in Marko 5 because I believe it will cause problems if the function is called during SSR. Not sure if I can think of anything better, but it is something to note.

src/components/ebay-combobox/component.ts
```diff
@@ -35,6 +36,7 @@ interface ComboboxInput extends Omit<Marko.HTML.Input, `on${string}`> {
         }>;
     options?: Marko.AttrTag<ComboboxOption>;
     "chevron-size"?: "large";
+    "dropdown-element"?:  () => HTMLElement;
```
LuLaValva, line 39: Initially it was not clear to me what this was supposed to be. I think `"dropdown-source"` might make more sense as a name.

Also, please add to `.stories.ts`, I looked for a description and couldn't find one 😢

src/components/ebay-combobox/component.ts
```diff
@@ -342,6 +354,8 @@ export default class Combobox extends Marko.Component<Input, State> {
             this._floatingLabel.destroy();
             this._floatingLabel = null;
         }
+
+        // this.dropdownUtil?.cleanup();
```
LuLaValva, line 358: Is this intentional? Should either be removed or uncommented

LuLaValva (commented):
First pass, left a few comments

Is there any world where `DropdownUtil` can be merged with the makeup setup, or does that not make sense?

src/components/ebay-date-textbox/component.ts
ArtBlue, line 1: The popover looks a bit mispositioned now (slightly over the textbox) vs prod (right under textbox).
image
image

ArtBlue (commented):
Looks good. One minor thing.

src/components/ebay-chips-combobox/index.marko
agliga: I did speak with Dylan about this approach. So it seems the only way

src/components/ebay-combobox/component.ts
agliga: This is simply an internal api. No need to document it.

src/components/ebay-date-textbox/component.ts
agliga: Is this with the latest skin code linked? I thought this was fixed when using the latest skin

agliga:
> First pass, left a few comments
> 
> 
> 
> Is there any world where `DropdownUtil` can be merged with the makeup setup, or does that not make sense?

Could be a good idea
It would be good to add floating ui to expander but that's outside this.

src/components/ebay-combobox/component.ts
LuLaValva: Ah okay, if it's internal only can you add a JSDoc comment for it? Something like `/** For internal use only */`

That said, are there any use cases where teams have custom comboboxes and might need to use this on their own?

agliga: I believe this is only needed when combobox styles are overridden such as in chips combobox.

src/components/ebay-combobox/component.ts
agliga: Will remove.

agliga:
Pr should be ready to re-review

src/common/dropdown/index.ts
```diff
@@ -0,0 +1,54 @@
...
+        });
+    }
+
+    cleanup() {
+        this.cleanupFn?.();
+    }
+
+    hide() {
+        if (this.cleanup) this.cleanup();
+    }
```
LuLaValva, line 53: I don't understand why we have this `hide()` method in _addition_ to `cleanup()`, especially since to my knowledge that `if` statement will always be true. Why not just have `cleanup()`?

LuLaValva (approved):
Seems to be working well, this will simplify the lives of so many people using eBayUI. Just one comment on code I'm confused about, but I think it's good to merge either way

src/common/dropdown/index.ts
agliga: It won't be true
Hide is there mostly there in order to distinguish between hiding the popup (where we have to run cleanup) or to cleanup. 
There was a case where this was getting called multiple times and caused a console error (even though the popup was closed) I believe this was for combobox when the popup goes away if there is no match and then comes back.

PR #2398, opened 2025-01-21
refactor(ts): use `Marko.HTML` instead of `Marko.Input` when possible

LuLaValva:
## Description

- Swap `Marko.Input<"*">` for `Marko.HTML.*`. Behavior is identical as the former reduces to the latter, but there are a few benefits.
  1. I think `Marko.HTML.*` is more readable and easier to follow without knowing about the magic `Input` keyword
  2. If an HTML tag doesn't exist, this method will surface an error where the previous didn't. In fact, these changes surfaced an issue with the types for `List`, which were previously casted to `any`.
  
### Bonus Change

- Icons for examples in `ebay-menu` were the wrong size


agliga (approved):

agliga:
Need to update snapshots before merging

PR #2406, opened 2025-01-24
chore: removed deprecations

agliga:
Removed all deprecated code
* removed separator and transformers
* Removed dialog handle
* Removed stepper is current (was deprecated 2 years ago!)


LuLaValva (approved):

src/components/ebay-fake-menu-button/examples/separator.marko
```diff
@@ -1,7 +1,7 @@
 <ebay-fake-menu-button text="eBay Menu separator">
     <@item href="#">item 1</@item>
     <@item href="#">item 2</@item>
-    <@separator/>
+    <@item separator/>
```
HenriqueLimas, line 4: Why are we removing this? Isn't semantically better to have a different component? A separator is not an item that has events and it is completely different with different roles. I feel it is over using the `@item` tag/component

src/components/ebay-lightbox-dialog/index.marko
```diff
@@ -54,25 +41,6 @@ $ var handleLabel = state.expanded ? a11yMinimizeText : a11yMaximizeText;
...
-                aria-label=handleLabel
-                type="button"
-                class=`lightbox-dialog__handle`
-                onClick("handleExpand")
-                onTouchstart("handleTouchStart")
-                onTouchmove("handleTouchMove")
-                onTouchend("handleTouchEnd")
-                onTouchcancel("handleTouchEnd")
-            ></button>
-            <if(bannerImgSrc)>
```
HenriqueLimas, line 70: The `bannerImgSrc` shouldn't be removed I suppose no?

HenriqueLimas (changes requested):

src/components/ebay-fake-menu-button/examples/separator.marko
LuLaValva: It might be semantically better, but Marko has no way of telling which order the attribute tags are in so this required a compiler hook (which is not recommended). Essentially without the compiler hook it looks like this in React:
```jsx
<EbayFakeMenuButton item=[...] separator=[...] />
```

src/components/ebay-lightbox-dialog/index.marko
agliga: Good catch!

agliga:
Should be ready again!

HenriqueLimas (approved):

PR #2407, opened 2025-01-24
fix(build): move `.less` files to style blocks

LuLaValva:
## Description

The `.less` files weren't being included in the build `dist/`, so I moved styles into the `.marko` files.


agliga (approved):

agliga:
Lets just target 15.0.0 for now, we will release that version soon.

Issue #2165, opened 2024-05-01
Eliminate `marko-tag.json` transforms where possible


LuLaValva:
# Bug Report

## Description

- Known limitation of Marko: attribute aliases in `marko-tag.json` aren't applied when the tag is discovered as a dynamic tag
- To support dynamic tags: we should remove attribute aliases and other transforms whenever it is possible

## Workaround

- Don't use dynamic tags for eBayUI components

WinkeeFace:
[`"@items <item>[]" changed to "@item <item>[]"`](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu/marko-tag.json#L15) 
needs to change from plural : singular into singular : singular

WinkeeFace:
### **Workflow:**

1. **Update the `marko-tag.json` File**:
   - Change any plural attributes (like `items`) to singular (`item`).
   - Adjust configurations, such as `nested-tags`, to reflect this change.
   - Ensure that references to singular attributes match the component's expected structure.
   
2. **Adjust the Component's `component.ts` File**:
   - Modify the code to deconstruct `item` instead of `items`.
   - Update how the properties are spread and passed down to the child components.
   - Ensure backward compatibility or document breaking changes for users.

3. **Modify the `index.marko` File**:
   - Update references from `items` to `item`.
   - If a loop is required, use a deconstructed array like `...item` to handle individual properties.
   - Add default values, like `item = []`, for properties to ensure stability.

4. **Update Stories**:
   - Change references in the Storybook or other story files from `items` to `item`.
   - Ensure the examples reflect the singular naming convention.
   - Verify that all examples are functional after these changes.

5. **Testing**:
   - Run tests to ensure no breaking changes for users adhering to proper tags.
   - Validate that all singular attributes are correctly passed and handled.

---

### **Example Implementation for the Workflow:**

#### **`marko-tag.json` Before:**
```json
{
  "@tab <tabs>[]": {
    "attribute-groups": ["html-attributes"],
    "@*": {
      "targetProperty": null,
      "type": "expression"
    },
    "@html-attributes": "expression",
    "@href": "#html-href"
  },
  "@tabs": "Marko.AttrTag<Tab>",
}
```

#### **`marko-tag.json` After:**
```json
{
  "@tab <tab>[]": {
    "attribute-groups": ["html-attributes"],
    "@*": {
      "targetProperty": null,
      "type": "expression"
    },
    "@html-attributes": "expression",
[...]

PR #2415, opened 2025-01-30
feat(marko-json): changed all attribute tags to be singular

agliga:
## Description
* Changed all marko attribute tags to be singular
Original change done by @WinkeeFace, just mainly fixed tests and merge conflicts with this

## References
#2165

### DIFF
--- a/src/common/menu-utils/index.ts
+++ b/src/common/menu-utils/index.ts
@@ -26,7 +26,7 @@ export class MenuUtils<
     State extends MenuState,
 > extends Component<Input, State> {
     declare type?: string;
-    declare item: MenuItem[];
+    declare items: MenuItem[];
 
     isRadio() {
         return this.type === "radio";
@@ -34,10 +34,10 @@ export class MenuUtils<
 
     getCheckedValues() {
         if (this.isRadio()) {
-            const item = this.item[this.state.checkedIndex!] || {};
+            const item = this.items[this.state.checkedIndex!] || {};
             return [item.value];
         }
-        return this.item
+        return this.items
             .filter((item, index) => this.state.checkedItems![index])
             .map((item) => item.value);
     }
@@ -48,7 +48,7 @@ export class MenuUtils<
                 ? undefined
                 : [this.state.checkedIndex];
         }
-        return this.item
+        return this.items
             .map((item, i) => this.state.checkedItems![i] && i)
             .filter((item) => item !== false && item !== undefined) as number[];
     }
@@ -59,17 +59,17 @@ export class MenuUtils<
             from items to pass correct indexes to state
             Any other component that doesn't have separator should pass through
         */
-        this.item = [...(input.item || [])].filter((item) => !item.separator);
+        this.items = [...(input.item || [])].filter((item) => !item.separator);
         this.type = input.type;
         if (this.isRadio()) {
             return {
-                checkedIndex: (this.item || []).findIndex(
+                checkedIndex: (this.items || []).findIndex(
                     (item) => item.checked || false,
                 ),
             };
         }
         return {
-            checkedItems: (this.item || []).map(
+            checkedItems: (this.items || []).map(
                 (item) => item.checked || false,
             ),
         };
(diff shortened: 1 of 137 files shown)


src/common/menu-utils/index.ts
```diff
@@ -26,18 +26,18 @@ export class MenuUtils<
     State extends MenuState,
 > extends Component<Input, State> {
     declare type?: string;
-    declare items: MenuItem[];
+    declare item: MenuItem[];
```
LuLaValva, line 29: Technically this one can be plural, since it's internal. Maybe `items` or `itemList` is better

src/components/ebay-breadcrumbs/index.marko
```diff
@@ -6,7 +6,7 @@ $ const {
     a11yHeadingTag,
     a11yMenuButtonText,
     class: inputClass,
-    items: attrTagItems,
+    item: attrTagItems = [],
     ...htmlInput
 } = input;
```
LuLaValva, line 12: GitHub won't let me comment there, but the line below this doesn't need `|| []` anymore because of the default above

src/components/ebay-chips-combobox/index.marko
```diff
@@ -10,7 +10,7 @@ $ const {
     ...comboboxInput
 } = input as Omit<typeof input, `on${string}`>;
 
-$ const options = [...option || []].map((o) => o.text)
+$ const mappedOptions = [...options || []].map((o) => o.text)
```
LuLaValva, line 13: No need for `|| []` here

LuLaValva: Also, I know it's a bonus change but I don't like `mappedOptions` as a name. Looks like the only place it's used is the for below, I vote to remove this and replace the for with
```marko
<for|{ text: option }| of=options>
```

src/components/ebay-filter-menu-button/index.marko
```diff
@@ -12,7 +12,7 @@ $ const {
     formName,
     formAction,
     formMethod,
-    items,
+    item: items =[],
```
LuLaValva, line 15: Missing a space, maybe a prettier problem

src/components/ebay-listbox/component.ts
```diff
@@ -156,7 +156,7 @@ class Listbox extends Marko.Component<Input, State> {
 
         // This `as any` is here for while `options` is coerced into an array from `marko-tag.json`.
         // After we move to the full `iterator` we can switch to `if (input.options && !input.disabled)`
-        if ((input.options as any)?.length && !input.disabled) {
+        if ((input.option as any)?.length && !input.disabled) {
```
LuLaValva, line 159: I think this comment has been satisfied, we can do what it says and remove it

src/components/ebay-select/select.stories.ts
```diff
@@ -138,7 +138,7 @@ ExternalLabel.parameters = {
 };
 
 ExternalLabel.args = {
-    options: [
+    option: [ //This breaks if I set it to option, but it looks like everything's already swamped.
```
LuLaValva, line 141: Is this something you meant to fix before merging? I'm not sure I understand the comment

src/components/ebay-breadcrumbs/index.marko
```diff
@@ -6,7 +6,7 @@ $ const {
     a11yHeadingTag,
     a11yMenuButtonText,
     class: inputClass,
-    items: attrTagItems,
+    item: attrTagItems = [],
```
LuLaValva, line 9: This should just be `items`, since that's the language you used for all of the other pluralizations.
```suggestion
    item: items = [],
```

LuLaValva (commented):
Looks great @agliga! Just a few minor comments, and then I'll do one more pass

src/components/ebay-breadcrumbs/index.marko
agliga: Leaving this as is, we need to convert the attrTagItems to items which is a true array.

agliga:
Done, should be ready for re-review

LuLaValva (approved):
Looks good, thanks @agliga and @WinkeeFace!

PR #2433, opened 2025-02-14
feat(avatar): added image position option

agliga:
## Description
* Added image position option
* Options are `fit`, `contain` and `auto`. Auto detects aspect ratio and sets fit or contain.

## References
https://github.com/eBay/ebayui-core/issues/2434

## Screenshots
Screenshot 2025-02-14 at 9 24 33 AM

### DIFF
--- a/src/components/ebay-avatar/component.ts
+++ b/src/components/ebay-avatar/component.ts
@@ -1,11 +1,9 @@
 import type { WithNormalizedProps } from "../../global";
-import type {
-    AttrString,
-} from "marko/tags-html";
+import type { AttrString } from "marko/tags-html";
 
 type Size = 32 | 40 | 48 | 56 | 64 | 96 | 128;
 
-type ImagePlacement = "cover" | "fit" | "auto";
+type ImagePlacement = "cover" | "fit";
 
 interface State {
     imagePlacement: ImagePlacement;
@@ -13,18 +11,11 @@ interface State {
 
 interface AvatarInput extends Omit<Marko.HTML.Div, `on${string}`> {
     username?: string;
-    "image-placement"?: ImagePlacement;
     color?: string;
     "a11y-text"?: AttrString;
     size?: Size | `${Size}`;
     img?: Marko.AttrTag<Omit<Marko.HTML.Img, `on${string}`>>;
 }
-
-const coverApsectRatios = [
-    0.75, // 3:4
-    1.33, // 4:3
-    1, // 1:1
-];
 
 export interface Input extends WithNormalizedProps<AvatarInput> {}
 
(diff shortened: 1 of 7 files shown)


src/components/ebay-avatar/component.ts
```diff
@@ -0,0 +1,64 @@
...
+                    this.state.imagePlacement = "cover";
+                } else {
+                    this.state.imagePlacement = "fit";
+                }
+            };
+        }
+    }
+    onInput(input: Input) {
+        let imagePlacement = input.imagePlacement || "cover";
+        if (input.imagePlacement === "auto") {
```
LuLaValva, line 55: Won't this mean that every time input is changed, `imagePlacement` is set to `"cover"`? This would override the automatic sizing that happens in `onMount`

LuLaValva (commented):
Mostly looks good, but philosophical question— why do we allow users to use anything besides `"auto"`? Shouldn't avatar look the same everywhere? I understand it's more overhead on the client if an app team already had information about the aspect ratio of the image, but I'm not sure that's worth accounting for.

agliga:
> Mostly looks good, but philosophical question— why do we allow users to use anything besides `"auto"`? Shouldn't avatar look the same everywhere? I understand it's more overhead on the client if an app team already had information about the aspect ratio of the image, but I'm not sure that's worth accounting for.

I only set it like this because if you can detect the dimensions on the server you should set it there. 

Also i set cover as default so it doesnt break but we can change it to auto as default if needed.

LuLaValva:
> I only set it like this because if you can detect the dimensions on the server you should set it there.

I don't think we should add the option to set it from the backend. We don't have clear documentation on the specific aspect ratio where it switches from "fit" to "cover", and if it ever changes then a bunch of apps will be out of sync

agliga (commented):
From design 
It should be
3:4 aspect ratio or 4:3 aspect ration and square (1/1) to be center
All others fit. If there is an irregular aspect ratio it should also be fit. 
This isnt taking square into account. Or its allowing more than it should.

LuLaValva:
> All others fit. If there is an irregular aspect ratio it should also be fit.

I will talk to a designer, but I don't think that's right. My understanding is that everything between `4:3` and `3:4` should be `"fill"`, based on this part of playbook:

> Crops narrower than a 4:3 aspect ratio, such as 16:9 or more extreme formats, are adjusted to “fit.”

src/components/ebay-avatar/index.marko
```diff
@@ -32,13 +18,14 @@ $ const {
     aria-label=a11yText
     class=[
         "avatar",
+        state.imagePlacement === "fit" && "avatar--fit",
         inputClass,
         size && `avatar--${size}`,
-        `avatar--${getColorForText(username, color)}`,
+        color && !img && `avatar--${getColorForText(username, color)}`,
```
LuLaValva, line 24: @agliga doesn't checking for `color` here defeat the purpose of `getColorForText`? Looks like `color` is used as the default, so this is the same as `` !img && `avatar--${color}` ``

agliga: We removed the default because of the image scrim.

LuLaValva: Okay, in that case we don't need `getColorForText` because this is the same as `` color && !img && `avatar--${color}` ``

LuLaValva (approved):

Issue #2425, opened 2025-02-10
ebay-icon-button: TypeScript check error on attribute target


abiyasa:
# Bug Report

## Description

When we want to use `<ebay-icon-button>` as link button (e.g with property `href` set) with attribute `target="_blank"`, the TypeScript compilation will complain with **error TS2353**

```
<ebay-icon-button
  aria-label=(model.accessibilityText)
  href="https://www.ebay.de/fdbk/leave_feedback"
  target="_blank"
  ^^^^^^ Object literal may only specify known properties, and '"target"' does not exist in type 'Directives & Input'.
>
  <ebay-feedback-16-icon/>
</ebay-icon-button>
```

## Workaround

You can use spread object which will ignore type check so far:

```marko
$ const { model, class: inputClass } = input;
$ const linkProps = { 
  href: "https://www.ebay.de/fdbk/leave_feedback",
  target: "_blank" 
};
<ebay-icon-button
  aria-label=(model.accessibilityText)
  ...linkProps
>
  <ebay-feedback-16-icon/>
</ebay-icon-button>
```

## Additional Notes

I think the TypeScript error is due to `<ebay-icon-button>` Input type is extended from `<button>` which does not support attributes from `<a>` link: https://github.com/eBay/ebayui-core/blob/1c09d20314ca7dd9c0b8f96b7032dd2f40698892/src/components/ebay-icon-button/component-browser.ts#L6

## Screenshots

N/A

agliga:
@LuLaValva 
Might need to extend both anchor and button

Issue #2424, opened 2025-02-10
ebay-icon-button: No style at all when href property is set


abiyasa:
# Bug Report

## Description

When you set property `href` to `<ebay-icon-button>`, which will render a `<a>` instead of `<button>`, the icon button does not style properly.

## Screenshots

**Without href** from [Storybook](https://opensource.ebay.com/ebayui-core/?path=/story/buttons-ebay-icon-button--default)

![Image](https://github.com/user-attachments/assets/baeb5b1b-52af-4731-bc79-19e393cf4085)

**With href set**

![Image](https://github.com/user-attachments/assets/0118b4b8-571c-42d6-a99d-b57cf8a38631)

All the styling (bg color, radius, color, ...) are not visible.

## Quick investigation on the issue

It seems eBay Skin class for icon button is either `button.icon-btn` or `a.icon-link`: 
https://github.com/eBay/skin/blob/78b68b0309c7167ad81824db3bfa275c70a1ba9f/src/sass/icon-button/icon-button.scss#L13-L18

The `<ebay-icon-button href=link>` however renders as `a.icon-btn`: https://github.com/eBay/ebayui-core/blob/1c09d20314ca7dd9c0b8f96b7032dd2f40698892/src/components/ebay-icon-button/index.marko#L34-L44

## Workaround

User has to manually set class `icon-link` to `<ebay-icon-button>` when href prop is set

PR #2445, opened 2025-02-28
fix(icon-btn): added swap to icon-link for href and _target support

agliga:
## Description
* Added support for `_target` in button and icon-btn
* Added a swap of classes for link icon-btn when its a `href`

## References
#2424 
#2425


src/components/ebay-button/index.marko
```diff
@@ -12,6 +12,7 @@ export interface ButtonEvent<T extends Event> {
 
 export interface ButtonInput extends Omit<Marko.HTML.Button, `on${string}`> {
```
LuLaValva, line 13: If you change this to extend multiple types then no other overrides are required, and the type will be more precise
```suggestion
export interface ButtonInput extends Omit<Marko.HTML.Button, `on${string}` | "type">, Omit<Marko.HTML.A, `on${string}`> {
```

agliga: Done

LuLaValva (approved):
Looks good! Thanks for pushing back to find a simpler solution

Issue #2431, opened 2025-02-13
event-utils: swap to use key


hongjilee-1:
# Bug Report

## Description

When using `"@testing-library/user-event": "^14.6.1"` to initiate key press interactions, the api:
```
await user.keyboard('{Escape}');
```
results in the deprecated `e.charCode` being undefined and `e.keyCode` as 0.

Hence trying to confirm that the escape event was emitted on an `<ebay-button />` will not pass using userEvent.

Location: https://github.com/eBay/ebayui-core/blob/master/src/common/event-utils/index.ts#L6-L21

## Workaround

Instead of using `userEvent.keyboard`, use a custom KeyboardEvent with the keyCode defined:

```
const event = new KeyboardEvent("keydown", { keyCode: 27 });
(container.querySelector('button')).dispatchEvent(event);
```

## Screenshots

When using `userEvent.keyboard`

![Image](https://github.com/user-attachments/assets/5cd002cd-e9d4-48f6-95b3-eab003faeafb)

agliga:
I'm not sure i understand this issue
We take the escape event and emit a new event. The event we emit should have `originalEvent` object which should have the right character code
So i recommend to use that. 
We cannot emit the char code on the escape event because this is a custom event. This is how marko handles these events. 

Please provide more info on this otherwise I will close the issue. Thanks!

hongjilee-1:
`@testing-library/user-event` removed support for the deprecated `keyCode` and `charCode` and only defines `key` like in the screenshot.

Any events handled eventUtils within ebayui-core components won't emit properly when using userEvent's keyboard to trigger the interaction in unit tests. Perhaps this should not be a bug but an enhancement

agliga:
Okay i see a benefit to remove event utils and swap to use key since the support is pretty good now. https://caniuse.com/keyboardevent-key

PR #2446, opened 2025-02-28
fix(event-utils): updated to use key

agliga:
## Description
* Swapped deprecated event `charCode` and `keyCode` to use `key`

## Context
For tests, I updated `"(Space key)"` to be just `" "`
## References
https://github.com/eBay/ebayui-core/issues/2431

### DIFF
--- a/src/common/event-utils/index.ts
+++ b/src/common/event-utils/index.ts
@@ -1,13 +1,9 @@
-function checkModifiers(e: KeyboardEvent): boolean {
-    return e.ctrlKey || e.altKey || e.metaKey || e.shiftKey;
-}
-
 function handleKeydown(
     keyCodes: string[],
     e: KeyboardEvent,
     callback: () => any,
 ) {
-    const keyCode = e.key
+    const keyCode = e.key;
     if (keyCodes.indexOf(keyCode) !== -1) {
         callback();
     }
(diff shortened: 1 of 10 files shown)


src/common/event-utils/index.ts
```diff
@@ -1,67 +1,67 @@
...
-        40, // down arrow
-        91, // "meta" key (Mac "command" key)
+        "Tab",
+        "Enter",
+        "Shift",
+        "Escape",
+        "ArrowUp",
+        "ArrowDown",
+        "ArrowLeft",
+        "ArrowRight"
```
LuLaValva, line 64: Was it intentional to remove modifier keys here (ctrl, alt, caps lock, cmd)?

LuLaValva (commented):
Good change! Just one question and then I'll approve

src/common/event-utils/index.ts
agliga: I moved them into the function itself since the recommended way to check for modifiers is using e.shiftKey
See check modifiers
To add this function is actually not being used anywhere except here. I was considering to remove it and put the check directly inline

LuLaValva: This function is used in `ebay-combobox`, where `checkModifiers` is _not_ called. After your changes, `on-input-change` is emitted every time the user presses `ctrl`, `option`, `cmd` (previously that was not the case)

agliga: So as i mentioned i added a check in handleNotKeydown where this list is being passed. I added a check to see no modiferiers is pressed.

LuLaValva: Sure, but I still see events emitted where they were not before in `ebay-combobox` storybook

Issue #2419, opened 2025-02-03
Video Player: Update to match specs


soniamacias:
# Enhancement Request

## Description

Apply design system specs to current build.  

## Context

- Currently using default shaka player UI. DS would like to update video player to follow our specs. 
- Kicked off w/Andrew, James, Randy on 2/3
- Andrew to review specs and see what can be built 
- Andrew to bring demo into weekly ebayUI x DS meeting (Thursdays) 
- Will be prioritized for March 

## Design Specs

- https://playbook.ebay.com/design-system/components/video-player
- Default to TRUE (muted)
- Icons are custom and in our [library](https://playbook.ebay.com/foundations/iconography/icon-library) 
- Overflow to include: closed caption, reporter, copy link

PR #2452, opened 2025-03-06
feat(video): updated to match new specs

agliga:
# ⚠️  _Note_ please read before reviewing
I included the CSS here currently only for review. BEFORE I merge, I will push this to skin and remove the CSS file. It is easier to review it here and change it and then update skin rather than creating two PRs and keeping them in sync. There's a lot of styles which are specific to the configuration.

## Description
* Created as a draft PR to get early reviews
* Added current time and end time components (copied the shaka player components, but split them up into two). This allows us to the current time before the video progress and total time after
* Added a custom mute/unmute control. This is mainly for the icon, but might be extended to have volume popover (currently on hold)
* Shifted video time scrubber to be inline to controls. I looked at the best way to do it, and what shaka themes overrides did is to position this using margin and absolute positioning. I don't like this solution, but there doesn't seem to be a better way (I tried forking it but there were issues with updating the time, plus any changes we have to fix our forked code)
* Removed overflow. Moved report to be inline. 
* Moved CC to be inline (note, CC only shows when a video has CC, none of our examples have it, so it won't show) 
* Misc cleanup of old code like removing some unused state variables

## References
https://github.com/eBay/ebayui-core/issues/2419


src/components/ebay-video/component.ts
```diff
@@ -345,10 +396,12 @@ class Video extends Marko.Component<Input, State> {
         shakaLoad()
             .then((shaka: any) => {
                 this.shaka = shaka.default || shaka;
+                window.shaka = this.shaka; // Set global object for some components to access
 
                 this.handleSuccess();
             })
             .catch((e: Error) => {
+                console.log(e);
```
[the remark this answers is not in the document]
LuLaValva: I think leaving it makes sense, especially if we're sure this error only shows up when the video actually doesn't load properly

src/components/ebay-video/component.ts
```diff
@@ -58,6 +64,10 @@ interface VideoInput extends Omit<Marko.HTML.Video, `on${string}`> {
     clip?: any[];
     source: Marko.AttrTag<Marko.HTML.Source>;
     "report-text"?: AttrString;
+    "mute-text"?: AttrString;
+    "unmute-text"?: AttrString;
+    "fullscreen-text"?: AttrString;
+    "exit-fullscreen-text"?: AttrString;
     "spinner-timeout"?: number;
```
LuLaValva, line 71: Any reason these don't have an `a11y` prefix? I understand it would be a breaking change to also modify `report-text` to `a11y-report-text`, but maybe we can make an alias there?

src/components/ebay-video/component.ts
```diff
@@ -79,13 +89,15 @@ interface State {
...
-    showLoading: boolean;
 }
 
 class Video extends Marko.Component<Input, State> {
     declare video: HTMLVideoElement;
     declare root: HTMLElement;
     declare containerEl: HTMLElement;
+    declare spacer: HTMLElement;
+    declare rangeContainer: HTMLElement;
+    declare buttonPanel: HTMLElement;
```
LuLaValva, line 100: These are all only used in one function, why keep them as class members? Will they ever be shared between methods?

src/components/ebay-video/component.ts
```diff
@@ -111,6 +123,23 @@ class Video extends Marko.Component<Input, State> {
...
+            this.buttonPanel = this.el.querySelector<HTMLElement>(
+                ".shaka-controls-button-panel",
+            )!;
+            this.spacer = this.buttonPanel.querySelector(".shaka-spacer")!;
+            this.rangeContainer = this.el.querySelector<HTMLElement>(
+                ".shaka-range-container",
+            )!;
+        }
+        const buttonPanelRect = this.buttonPanel.getBoundingClientRect();
+        const spacerRect = this.spacer.getBoundingClientRect();
```
LuLaValva, line 137: Won't `this.buttonPanel` be undefined if the prior if statement failed? That would lead to a "document is undefined" error

src/components/ebay-video/video.stories.ts
```diff
@@ -213,7 +213,7 @@ export default {
```
LuLaValva, line 213: Storybook should have entries for all of the new text attributes

src/components/ebay-video/component.ts
```diff
@@ -58,6 +64,10 @@ interface VideoInput extends Omit<Marko.HTML.Video, `on${string}`> {
     clip?: any[];
     source: Marko.AttrTag<Marko.HTML.Source>;
     "report-text"?: AttrString;
```
LuLaValva, line 66: Can you include an alias for this?
```suggestion
    /** @deprecated use `a11y-report-text` instead */
    "report-text"?: AttrString;
    "a11y-report-text"?: AttrString;
```

LuLaValva (commented):
Looks good! Just one nit, and then I'll approve after you move the `.css` file to skin

saiponnada (commented):
Few things I noticed,

As part of a simple keyboard test 
- Current time is being focused which is unnecessary and after looking further looks like it is a button. What does this button do?
- I understand you had issues with positioning this seek bar. But the problem here is focus order is not aligning with visual order i.e, focus jumps from play button to timestamp and after full screen button it returns back to the track where a user has the ability to fast track or rewind

When I did a quick screen reader test, few things I noticed
- When I am on seek slider, it announces some gibberish number. What would be ideal is on what time stamp we are on and what is overall duration of the video.
- IMO it is always nice to CC button visible but disabled if no captions available

agliga:
> Few things I noticed,
> 
> As part of a simple keyboard test
> 
> * Current time is being focused which is unnecessary and after looking further looks like it is a button. What does this button do?
I disabled the button so its no longer in focus state
> * I understand you had issues with positioning this seek bar. But the problem here is focus order is not aligning with visual order i.e, focus jumps from play button to timestamp and after full screen button it returns back to the track where a user has the ability to fast track or rewind
That is unfortunately the recommended way to do it in shaka player. I don't like the change but there's no other way to do this. 
> 
> When I did a quick screen reader test, few things I noticed
> 
> * When I am on seek slider, it announces some gibberish number. What would be ideal is on what time stamp we are on and what is overall duration of the video.
So again, this is what shaka player is providing. It could be due to the video we are testing. 

> * IMO it is always nice to CC button visible but disabled if no captions available
Again, shaka player implementation. 


I think the conclusion is to feel free to file a bug on their repo. They seem pretty responsive on that. We had an issue that they fixed pretty quickly after we filed it.

agliga:
This should be ready for re-review, removed the CSS and pushed it to skin

saiponnada (approved):
Looks like we have few limitations with third party library in making this accessible. We need to eventually figure out a solution that can solve these issues in long term. For now I am okay with moving forward.

Issue #2328, opened 2024-11-19
table: add loading state


ArtBlue:
# Enhancement Request

## Description

Add `table` loading state to stay inline with Skin (https://github.com/eBay/skin/issues/2470).

The detailed guidance on `a11y` and `ux` should be built built in for this `table` loading. Refer to the the _**Table in Loading State**_ section of https://opensource.ebay.com/skin/component/table/

PR #2453, opened 2025-03-12
feat(table): added loading body state

agliga:
## Description
* Added table loading state. Used `body-state`. Was looking at using mode initially but that can remove `selected` state (can have load + selected`)
* Added `isLoading` variable to disable all interactive elements in template
* Added check in component to check all focusable elements (IE elements which are added by user) and disable/add tabindex -1 for them.
* For undo only remove tabindex and disabled attribute for items found in the list of disabled items
* Added tests for checking loading state

## References
https://github.com/eBay/ebayui-core/issues/2328

## Screenshots
Screenshot 2025-03-11 at 4 04 19 PM

### DIFF
--- a/src/components/ebay-table/component.ts
+++ b/src/components/ebay-table/component.ts
@@ -48,8 +48,9 @@ interface State {
 }
 
 export default class EbayTable extends Marko.Component<Input, State> {
-    declare disabledItems: Map<HTMLElement, boolean>;
+    declare disabledItems: Set<HTMLElement>;
     declare tbody: HTMLElement;
+    declare animationFrame: number;
 
     onCreate() {
         this.state = {
@@ -60,7 +61,7 @@ export default class EbayTable extends Marko.Component<Input, State> {
     }
 
     onMount() {
-        this.disabledItems = new Map();
+        this.disabledItems = new Set();
         this.tbody = this.getEl("tbody");
         this.setLoading();
     }
@@ -73,6 +74,16 @@ export default class EbayTable extends Marko.Component<Input, State> {
 
     onUpdate() {
         this.setLoading();
+    }
+
+    onRender() {
+        if (typeof window !== "undefined") {
+            cancelAnimationFrame(this.animationFrame);
+        }
+    }
+
+    onDestroy() {
+        cancelAnimationFrame(this.animationFrame);
     }
 
     getSelectedRowStateFromInput(input: Input) {
@@ -148,18 +159,16 @@ export default class EbayTable extends Marko.Component<Input, State> {
     setLoading() {
         if (this.input.bodyState === "loading") {
             if (this.tbody) {
-                requestAnimationFrame(() => {
+                this.animationFrame = requestAnimationFrame(() => {
                     focusables(this.tbody).forEach((focusable: HTMLElement) => {
-                        if (!focusable.getAttribute("disabled")) {
-                            focusable.setAttribute("disabled", "true");
-                            focusable.setAttribute("tabindex", "-1");
-                            this.disabledItems.set(focusable, true);
-                        }
+                        focusable.setAttribute("disabled", "true");
+                        focusable.setAttribute("tabindex", "-1");
+                        this.disabledItems.add(focusable);
                     });
                 });
             }
         } else {
-            for (const [focusable] of this.disabledItems) {
+            for (const [focusable] of this.disabledItems.entries()) {
                 focusable.removeAttribute("disabled");
                 focusable.removeAttribute("tabindex");
             }
--- a/src/components/ebay-table/component.ts
+++ b/src/components/ebay-table/component.ts
@@ -1,3 +1,4 @@
+import focusables from "makeup-focusables";
 import { AttrString, AttrTriState } from "marko/tags-html";
 import { WithNormalizedProps } from "../../global";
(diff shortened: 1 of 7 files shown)


ArtBlue (changes requested):
This looks to be missing a lot of the requirements for a fully disabled table in loading state. 

>Table in Loading State
When a table is fetching data for sorting, filtering, or pagination, a loading state should be displayed to indicate that the table is in the process of loading. This can be achieved by adding the table--loading-state class to the .table element.

>Additionally, in the loading state, the module should be inaccessible to the user. This can be achieved by setting tabindex="-1" to the module container, and adding the inert attribute to the table inside the module. Since inert is not fully supported in older browsers, it is recommended to either use a polyfill or use JavaScript to disable all interactive elements inside the table when in loading state. Elements that support disabled should be disabled. Anchors should have their href attributes removed.

>Here are some other steps to manually ensure the module is inaccessible:

>All interactive elements inside table should be forced to NOT be focusable.
The module container should be forced to be unscrollable.
The progress bar should be forced to gain focus.
All interactive elements inside the table should be disabled.
The loading state should be displayed until the data is fully loaded. In addition to the loading state being set on the table, a progress bar expressive should be added to relay the loading state both visually and for screen readers. The live examples that follow use aria-live="off" to prevent the status element from proactively announcing new content. This is only for demo purposes, so that this documentation page does not overwhelm the viewer with example status updates. The aria-live override should not be included in implementations of this pattern.

https://opensource.ebay.com/skin/component/table/

agliga:
> This looks to be missing a lot of the requirements for a fully disabled table in loading state. 
> 
> 
> 
> >Table in Loading State
> 
> When a table is fetching data for sorting, filtering, or pagination, a loading state should be displayed to indicate that the table is in the process of loading. This can be achieved by adding the table--loading-state class to the .table element.
> 
> 
> 
> >Additionally, in the loading state, the module should be inaccessible to the user. This can be achieved by setting tabindex="-1" to the module container, and adding the inert attribute to the table inside the module. Since inert is not fully supported in older browsers, it is recommended to either use a polyfill or use JavaScript to disable all interactive elements inside the table when in loading state. Elements that support disabled should be disabled. Anchors should have their href attributes removed.
> 
> 
> 
> >Here are some other steps to manually ensure the module is inaccessible:
> 
> 
> 
> >All interactive elements inside table should be forced to NOT be focusable.
> 
> The module container should be forced to be unscrollable.
> 
> The progress bar should be forced to gain focus.
> 
> All interactive elements inside the table should be disabled.
> 
> The loading state should be displayed until the data is fully loaded. In addition to the loading state being set on the table, a progress bar expressive should be added to relay the loading state both visually and for screen readers. The live examples that follow use aria-live="off" to prevent the status element from proactively announcing new content. This is only for demo purposes, so that this documentation page does not overwhelm the viewer with example status updates. The aria-live override should not be included in implementations of this pattern.
> 
> 
> 
> https://opensource.ebay.com/skin/component/table/

Can you be specific which requirements are missing?
The non interactive Elements are present (adding tabindex=-1) 
Not sure what else is needed.

src/components/ebay-table/component.ts
```diff
@@ -129,6 +145,28 @@ export default class EbayTable extends Marko.Component<Input, State> {
         });
     }
 
+    setLoading() {
+        if (this.input.bodyState === "loading") {
+            if (this.tbody) {
+                requestAnimationFrame(() => {
+                    focusables(this.tbody).forEach((focusable: HTMLElement) => {
+                        if (!focusable.getAttribute("disabled")) {
```
saiponnada, line 153: Focusables should not return elements that are disabled. This additional check is unnecessary IMO.  I would leave it up to you if you want to keep code as is or use makeup to fix when an interactive element not under makeup list is being used.

src/components/ebay-table/table.stories.ts
```diff
@@ -112,6 +117,27 @@ export default {
                 category: "@row attribute tags",
             },
         },
+        a11yLoadingText: {
+            description: "Text for progress bar expressive when table is in loading body state",
```
saiponnada, line 121: It would be helpful to say `Localized text`. Helps this needs to be localized

saiponnada (commented):
Overall LGTM. minor suggestions.

src/components/ebay-table/table.stories.ts
agliga: I believe that should be understood from this. I think for each `a11y` text override its assumed that it's for localization. No need to have it called out for each one.

agliga:
Should be ready for review. Again, if there are any gaps in what skin has, please comment what is missing because I did follow those guidelines on this.

ArtBlue:
Here's Skin:

<kbd>image</kbd>

A couple of things I noticed right away:

1. The `div.table` doesn't have `tabindex="-1"`
2. The `inert` is not set. 
3. The link sorting can't disable anchors since `<a>` is not something that be disabled. What we need for anchors is to be removed.
<kbd>image</kbd>
<kbd>image</kbd>
4. To avoid having to keep track of all the anchor URLs before removing them and then resetting them, I'd suggest we simply change the attributes from `href` to `data-href` during loading, and when loading completed, changing them back to `href`.

ArtBlue (approved):

saiponnada (approved):
LGTM! Thanks for taking time and making these changes.

PR #2454, opened 2025-03-17
feat(skeleton): added example for grouped skeletons

### LINKED ISSUE #2416: <ebay-skeleton>: add a11y support for grouped blocks
# Enhancement Request

## Description

Show an example where a11y text is defined at container level rather than each block saying "loading..."

## Context

https://pages.github.corp.ebay.com/BusinessComponents/BusinessComponentsWeb/?path=/story/@ebay-business-components/recs-placement_shimmer--default

## Screenshots

Image

saiponnada:
## Description

- Added skeleton example for group tiles
- Removed un-necessary punctuation in `aria-label` to be consistent with docs.  

## References

Fixes #2416

## Screenshots

image

### DIFF
--- a/src/components/ebay-skeleton/examples/grouped-tile.marko
+++ b/src/components/ebay-skeleton/examples/grouped-tile.marko
@@ -1,9 +1,10 @@
 <style>
     .group-demo-tile-container{
         display: flex;
+        flex-wrap: wrap;
     }
     .demo-tile-container {
-        margin-right: 20px;
+        margin: 10px;
         width: 220px;
     }
     .custom-image {
--- a/src/components/ebay-skeleton/examples/grouped-tile.marko
+++ b/src/components/ebay-skeleton/examples/grouped-tile.marko
@@ -12,7 +12,8 @@
     }
 </style>
 
-<ebay-skeleton class="group-demo-tile-container">
+<!-- Example with localized a11y-text. "Laden" translates to "Loading" in German  -->
+<ebay-skeleton class="group-demo-tile-container" a11y-text="Laden">
     <div class="demo-tile-container">
         <ebay-skeleton-image class="custom-image"/>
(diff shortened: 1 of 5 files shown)


src/components/ebay-skeleton/examples/grouped-tile.marko
```diff
@@ -0,0 +1,32 @@
...
+        margin-right: 20px;
+        width: 220px;
+    }
+    .custom-image {
+        height: 220px;
+        width: 220px;
+    }
+</style>
+
+<ebay-skeleton class="group-demo-tile-container">
```
agliga, line 15: I would add a specific `a11yText` to this just to show that its custom

src/components/ebay-skeleton/skeleton.stories.ts
```diff
@@ -83,7 +85,7 @@ export default {
 export const Default = Template.bind({});
 Default.args = {
     style: "width: 220px",
-    renderBody: `<div class="skeleton__textbox" />` as any,
+    renderBody: `<div role="img" aria-label="Loading" class="skeleton__textbox" />` as any,
```
agliga, line 88: Is this needed still?

agliga (commented):
LGTM overall, some minor comments

src/components/ebay-skeleton/examples/grouped-tile.marko
saiponnada: I don't think it is necessary, it is consistent with other examples. There is no custom a11y text because of a grouped variant. It would be still loading IMO.

src/components/ebay-skeleton/skeleton.stories.ts
saiponnada: Good catch, I did not see the wrapper being passed.

agliga (approved):

Issue #2473, opened 2025-04-03
ebay-item-tile/tile-group: new component


agliga:
Add item tile and item tile group
Also add layout component under the hood. 
Skin issue https://github.com/eBay/skin/pull/2603

PR #2474, opened 2025-04-03
feat: added item-tile, item-tile-group, and layout-grid

agliga:
## Description
* Added item-tile
* Added item-tile-group
* Added layout-grid 
* Modified file-preview-card to allow custom icon-button actions and href

## Context
* I added a requirement for marko 5 for these new components. Since marko 4 is deprecated and adding support for it would be extra tech debt, any new components will be marko 5 or greater.  (This is why we dont have a marko-tag.json)
* Formatter for marko files is broken currently. Even if we merge as is, when we do the final release, we run prettier on the whole project, so those will be fixed.
* Added a new `layout` section in storybook where item-tile and item-tile-group as well as layout-grid is located

## Structure
The structure I ended up going with (seemed the simpliest but also offering the most flexibility) was:

```
<ebay-item-tile on-action("emit", "action") href="https://ebay.com"
        file={
            name: "file-name.jpg",
            type: "image/jpeg",
            src: "https://ir.ebaystatic.com/cr/v/c01/skin/docs/tb-real-square-pic.jpg",
          }
          ...input
>
    <@action aria-label="Add to watchlist">
        <ebay-heart-16-icon/>
    </@action>
    <@supertitle>
      <ebay-signal status="time-sensitive">
        Time Sensitive
      </ebay-signal>
    </@supertitle>
      <@title>
          Apple iPhone 11 Pro Max
      </@title>
      <@subtitle>
          256GB Space Gray
      </@subtitle>
    <@description>
        <div class="price">
        $29.99
        <span class="clipped"> Was: </span>
        <s class="list-price"> $68.99 </s>
      </div>
    </@description>
</ebay-item-tile>
```
Happy to change it around as needed, but please offer suggestions if you do not like it (and rationale)

## References
https://github.com/eBay/ebayui-core/issues/2473

## Screenshots
Screenshot 2025-04-03 at 10 07 43 AM
<br/>
Screenshot 2025-04-03 at 10 11 24 AM
<br/>
Screenshot 2025-04-03 at 10 12 44 AM

### DIFF
--- a/src/components/ebay-item-tile-group/examples/default.marko
+++ b/src/components/ebay-item-tile-group/examples/default.marko
@@ -21,20 +21,16 @@ class {}
       <@subtitle>
           256GB Space Gray
       </@subtitle>
-    <@description>
-        <div class="price">
-        $29.99
-        <span class="clipped"> Was: </span>
-        <s class="list-price"> $68.99 </s>
-      </div>
-      <div class="buy-links">
-        <div>
-          <a href="https://ebay.com"> Buy it now </a>
-        </div>
-        <div>Free shipping</div>
-      </div>
-      <div class="sponsored">Sponsored</div>
-    </@description>
+      <@description class="price">
+          $29.99
+          <span class="clipped"> Was: </span>
+          <s class="list-price"> $68.99 </s>
+      </@description>
+      <@description>
+            <a href="https://ebay.com"> Buy it now </a>
+      </@description>
+      <@description>Free shipping</@description>
+      <@description>Sponsored</@description>
     </@item>
     <@item  href="https://ebay.com"
         file={
@@ -57,20 +53,48 @@ class {}
       <@subtitle>
           256GB Space Gray
       </@subtitle>
-    <@description>
-        <div class="price">
-        $29.99
-        <span class="clipped"> Was: </span>
-        <s class="list-price"> $68.99 </s>
-      </div>
-      <div class="buy-links">
-        <div>
-          <a href="https://ebay.com"> Buy it now </a>
-        </div>
-        <div>Free shipping</div>
-      </div>
-      <div class="sponsored">Sponsored</div>
-    </@description>
+      <@description class="price">
+          $29.99
+          <span class="clipped"> Was: </span>
+          <s class="list-price"> $68.99 </s>
+      </@description>
+      <@description>
+            <a href="https://ebay.com"> Buy it now </a>
+      </@description>
+      <@description>Free shipping</@description>
+      <@description>Sponsored</@description>
+    </@item>
+    <@item href="https://ebay.com"
+        file={
+            name: "file-name.jpg",
+            type: "image/jpeg",
+            src: "https://ir.ebaystatic.com/cr/v/c01/skin/docs/tb-real-square-pic.jpg",
+          }
+>
+    <@action aria-label="Add to watchlist">
+        <ebay-heart-16-icon/>
+    </@action>
+    <@supertitle>
+      <ebay-signal status="time-sensitive">
+        Time Sensitive
+      </ebay-signal>
+    </@supertitle>
+      <@title>
+          Apple iPhone 11 Pro Max
+      </@title>
+      <@subtitle>
+          256GB Space Gray
+      </@subtitle>
+      <@description class="price">
+          $29.99
+          <span class="clipped"> Was: </span>
+          <s class="list-price"> $68.99 </s>
+      </@description>
+      <@description>
+            <a href="https://ebay.com"> Buy it now </a>
+      </@description>
+      <@description>Free shipping</@description>
+      <@description>Sponsored</@description>
     </@item>
     <@item  href="https://ebay.com"
         file={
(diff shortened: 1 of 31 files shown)


src/components/ebay-item-tile-group/examples/default.marko
```diff
@@ -0,0 +1,219 @@
+class {}
+<ebay-item-tile-group on-action("emit", "action") ...input>
```
ArtBlue, line 2: `item-tile-group` needs the `item-tile-group--list-view` in order to make the display adjustments for that view.

<kbd>image</kbd>

src/components/ebay-item-tile-group/item-tile-group.stories.ts
```diff
@@ -0,0 +1,65 @@
...
+        },
+        item: {
+            name: "@item",
+            table: {
+                category: "@attribute tags",
+            },
+            description:
+                "Item to be rendered inside the group. Arguments used are the same as for item-tile. See item-tile for more information.",
+        },
+        columns: {
```
ArtBlue, line 30: Per my previous comment, we can remove this. We just need to impose the layout rules directly in the markup on our own side.

src/components/ebay-item-tile/examples/default.marko
```diff
@@ -0,0 +1,38 @@
...
+        Time Sensitive
+      </ebay-signal>
+    </@supertitle>
+      <@title>
+          Apple iPhone 11 Pro Max
+      </@title>
+      <@subtitle>
+          256GB Space Gray
+      </@subtitle>
+    <@description>
```
ArtBlue, line 24: The various subgroups in this section need to use `<p>` instead of <`div>` because Skin has built in relative line spacing DS wanted. The current spacing is incorrect.

<kbd>image</kbd>

It may be a good idea to actually use `<@attribute>` to do the `<p>` wrapping on our end so teams can't use something that will result in spacing issues. There's a possibility for any host of attributes. That will probably also help business objects components so more brand-specific attributes continue to retain that spacing as well.

src/components/ebay-item-tile-group/examples/default.marko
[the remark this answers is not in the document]
agliga: Hmm this seems like something skin should be taking care of on its end if its predetermined 
If its customizable then yeah we can default to these. But if these are fixed shouldn't these be direct styles in item-tile-group? In order to not duplicate the styles they can be referenced as mixins.

agliga: Not sure what this means? I added a list view option in storybook.

src/components/ebay-item-tile/examples/default.marko
agliga: Sounds good. Will adjust

agliga: One comment, we should adjust skin since some examples have `<div>`
Screenshot 2025-04-04 at 9 20 54 AM

src/components/ebay-item-tile-group/examples/default.marko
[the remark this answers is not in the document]
ArtBlue: So, we used `layout-grid` for this as we had planned. If we added more styling in `itel-tile-group`, we'd be duplicating the styles in `lyout-grid`. The default styling in `layout-grid` won't work, so we need the attributes for custom overrides. The plan all along has been to remove responsive layout styling from the components themselves and have `layout-grid` be the only thing that handles those. The more we duplicate the styling already present in `layout-grid` in other components, the less value we will see from it, and we'd needlessly increase the bundle size.

We simply need those breakpoint attributes set on `.layout-grid` to achieve that. One of the reasons why originally I had a separate distinct level in `item-tile-group` was to make this a bit clearer.

ArtBlue: The component is missing `item-tile-group--list-view` for the list view. If you look at the Skin example, it should be evident.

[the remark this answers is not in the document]
agliga: So I agree with that, but shouldn't those attributes just be set by `item-tile-group`? `layout-grid` handles all the layout part, but `item-tile-group` simply defines what happens at what breakpoint.

It could simply be:
```
.item-tile-group {
    --layout-grid-columns-min: 1;
    --layout-grid-columns-xs: 2;
    // etc
}
``` 
And then it should work automatically correct?

agliga: Okay, I see it there, but I left it out since there are no styles associated with it. Its only the `item-tile item-tile--list-view` applying styles. 
Do we need it if there are no styles?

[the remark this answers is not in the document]
ArtBlue: Oh, I see. Yea, we can do that too. I wish we'd done the Skin part like that. As it stands now though, if we went with this option, there will need to be sweeping changes in Skin and this PR will be blocked until that's done. I'm fine if that's what we do, but will take longer.

ArtBlue: Yea, you're right. I think I originally had view-specific styling, but realized it was redundant.

Yes, this isn't an issue. The issue I saw had to do with the breakpoint rules. Once that is fixed, this should resolve itself.

ArtBlue (approved):
Looks good.

ArtBlue:
> It looks to be mostly there. It seems to still be missing the `a11y` label. 

Ok, we spoke offline about this. Looks good. Approved the PR.

